PR #23, opened 2021-07-08
Fix Prompt

jtribe32:
This fix stops location change when a prompt is triggered but dismissed and then state changes.


tannerlinsley:
Nice Work!

Issue #35, opened 2021-10-18
Unresolved entry for es package when using Vite


rabah:
**Describe the bug**
I installed `react-location` in my Vitejs app and set up the router exactly as the examples show. When running the dev server, I got an error basically saying it couldn't resolve the entry for `react-location`

**To Reproduce**
1. Install react-location in a Vite React App
2. Setup the router and run the app
3. This should trigger the error saying it cannot resolve the entry.

**Expected behavior**
I checked the `package.json` of `react-location` and it basically says:
```json
"main": "lib/index.js",
"module": "es/index.js",
```

It's not resolving because there is no `es` directory with the entry files in the package when it is installed, when I changed the `es` to `lib` it worked fine. I'm going to post in the error that came up below.

**Screenshots**

**Package folders**
<img width="293" alt="react-location-packages@2x" src="https://user-images.githubusercontent.com/31989196/137709927-5f941aa3-4c73-420f-a878-4797378f531b.png">

**Additional context**
```bash
> node_modules/vite/dist/node/chunks/dep-be032392.js:30265:10: error: [plugin: vite:dep-scan] Failed to resolve entry for package "react-location". The package may have incorrect main/module/exports specified in its package.json: Failed to resolve entry for package "react-location". The package may have incorrect main/module/exports specified in its package.json.
    30265 │     throw new Error(`Failed to resolve entry for package "${id}". ` +
          ╵           ^
    at packageEntryFailure (/Users/rabahtahraoui/Code/HackPartners/world-app/node_modules/vite/dist/node/chunks/dep-be032392.js:30265:11)
    at resolvePackageEntry (/Users/rabahtahraoui/Code/HackPartners/world-app/node_modules/vite/dist/node/chunks/dep-be032392.js:30261:9)
    at tryNodeResolve (/Users/rabahtahraoui/Code/HackPartners/world-app/node_modules/vite/dist/node/chunks/dep-be032392.js:30036:11)
    at Context.resolveId (/Users/rabahtahraoui/Code/HackPartners/world-app/node_modules/vite/dist/node/chunks/dep-be032392.js:29896:28)
    at Object.resolveId (/Users/rabahtahraoui/Code/HackPartners/world-app/node_modules/vite/dist/node/chunks/dep-be032392.js:42356:55)
    at async resolve (/Users/rabahtahraoui/Code/HackPartners/world-app/node_modules/vite/dist/node/chunks/dep-be032392.js:42547:26)
[...]

PR #36, opened 2021-10-18
fix: include es directory in packages json

rabah:
This PR just simply includes the `es` directory in the `package.json` which should output that directory in the final package. This is a fix for issue #35


tannerlinsley:
Excellent!

Issue #101, opened 2021-11-21
TypeError: Cannot read properties of undefined (reading 'listeners')


Codpoe:
**Describe the bug**
There is an error when jumping the link:
```
TypeError
Cannot read properties of undefined (reading 'listeners')
notify
https://klup1.csb.app/node_modules/react-location/es/index.js:123:10
eval
https://klup1.csb.app/node_modules/react-location/es/index.js:124:14
ReactLocation.notify
https://klup1.csb.app/node_modules/react-location/es/index.js:123:20
```

**To Reproduce**
1. Go to https://react-location.tanstack.com/
2. Scroll down to the example
3. Click any link in the example

**Expected behavior**
expect to navigate correctly

**Additional context**
- react-location@3.2.7

tannerlinsley:
Fixed  with 5e7950a61f2752b1b729c3dea435ca18c0108831

tannerlinsley:
This example should be working now :)

PR #102, opened 2021-11-21
fix: notify binding

Codpoe:
#101


tannerlinsley:
Oh snap, I didn't see this fix until now. I just pushed my own fix, plus a few other changes for this. Thank you :)

PR #167, opened 2021-12-20
Feat: Unloader function for hooking into `cleanMatchCache`

gatorcse:
Adds an unloader function to the route configuration. This is useful when your loader function is triggering requests through a stateful request library that has its own caching mechanism (Relay, Apollo, etc.) and you want to clean their caches when eager prefetch's aren't needed.

Docs site should probably be updated with the new api.

### DIFF
--- a/packages/react-location/src/index.tsx
+++ b/packages/react-location/src/index.tsx
@@ -101,6 +101,9 @@ export type RouteLoaders<TGenerics> = {
   pendingElement?: SyncOrAsyncElement<TGenerics>
   // An asynchronous function responsible for preparing or fetching data for the route before it is rendered
   loader?: LoaderFn<TGenerics>
+  // An asynchronous function responsible for cleaning up when the match cache is cleared. This is useful when
+  // the loader function uses libraries with their own request caching, and they need to be cleaned up.
+  unloader?: UnloaderFn<TGenerics>
   // An integer of milliseconds representing how long data should be cached for the route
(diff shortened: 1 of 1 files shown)


packages/react-location/src/index.tsx
```diff
@@ -101,6 +101,9 @@ export type RouteLoaders<TGenerics> = {
   pendingElement?: SyncOrAsyncElement<TGenerics>
   // An asynchronous function responsible for preparing or fetching data for the route before it is rendered
   loader?: LoaderFn<TGenerics>
+  // An asynchronous function responsible for cleaning up when the match cache is cleared. This is useful when
+  // the loader function uses libraries with their own request caching, and they need to be cleaned up.
```
gatorcse, line 105: Would welcome feedback on a better doc message.

packages/react-location/src/index.tsx
```diff
@@ -689,6 +697,14 @@ export class RouterInstance<
...
 
       if (!match.maxAge || age > match.maxAge) {
+        if (match.route.unloader) {
+          if (match.status === 'resolved') {
+            match.route.unloader(match, match.ownData)
+          } else {
+            match.route.unloader(match)
+          }
+        }
+
```
gatorcse, line 707: I'm not the best with the lifecycle of promises in JS. This will fire-and-forget the unloader promise, without awaiting on it or anything. Is this ok?

PR #168, opened 2021-12-20
Enter pending state if pendingMs is set to 0 ms

### LINKED ISSUE #162: pendingElement is not being displayed
**Describe the bug**
Similar to #112 the `pendingElement` is not being displayed using the async route loader or an async route element.

**To Reproduce**
Define a route with an async `loader` or an async `element` and add a `pendingElement`.

**Expected behavior**
The `pendingElement` should be displayed. Seems to be related to `pendingMs`. If I add `pendingMs: 1` it works.

**Desktop (please complete the following information):**
- OS: Windows
- Browser: Chrome
- Version: 3.3.0

Ben52:
This would fix #162  which also caught me off guard. I'm not sure if you should even need to set `pendingMs` if you set `pendingElement`, but I think if you set `pendingMs` to 0, you'd want it to always enter the pending state immediately.

### DIFF
--- a/packages/react-location/src/index.tsx
+++ b/packages/react-location/src/index.tsx
@@ -842,7 +842,7 @@ export class RouteMatch<TGenerics extends PartialGenerics = DefaultGenerics> {
       clearTimeout(this.pendingTimeout)
     }
 
-    if (this.route.pendingMs) {
+    if (this.route.pendingMs !== undefined) {
       this.pendingTimeout = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


joshuabrokaw:
I agree with your idea. If you set `pendintElement` and don't set `pendingMs`, it should default to 0. Curious to see Tanner's thoughts.

msalahz:
@tannerlinsley: Can you please merge this PR?

PR #172, opened 2021-12-22
Fix search matching

### LINKED ISSUE #164: Search matching routes that only differ in their search function gives duplicate route error
**Describe the bug**
I'm exploring the capabilities of react-location and am trying to build a stepper using route search matching. I love the idea of being able to define a route for each step based off a search param. However, if doing something like this example below... I get an error saying it's a duplicate route. Which makes sense because not specifying a path defaults to `"*"`, but how is search matching supposed to work if the only thing changing between two routes is the search match?

```
type LocationGenerics = MakeGenerics<{
  Search: {
    step?: number;
  };
}>;

const routes: Route<LocationGenerics>[] = [
  {
    path: "/stepper",
    element: <StepperContainer />,
    children: [
      {
        search: (search) => {
          return search.step === 0;
        },
        element: <Step1 />,
      },
      {
        search: (search) => {
          return search.step === 1;
        },
        element: <Step2 />,
      },
    ],
  },
];
```

**To Reproduce**
use the above LocationGenerics and routes in a `<Router />`

**Expected behavior**
I'd expect that routes that don't produce the same match based off search matching are treated as separate routes and not as duplicates

**Additional context**
What's the point of search matching if you can't have routes whose only difference is their search match?

I realize this can be done, by leveraging `useSearch` within `<StepperContainer />` and showing the correct step based off the step param, but I was interested to see if it could be accomplished using search matching in the routes.

joshuabrokaw:
This is fixes #164 

Currently search match is broken, because two routes with the same path will throw an error as duplicates.

This change allows there to be multiple routes in `routesById`, provided that they have `search`. If it doesn't contain a search, a duplicate error will be thrown.

### DIFF
--- a/packages/react-location/src/index.tsx
+++ b/packages/react-location/src/index.tsx
@@ -562,7 +562,7 @@ export class RouterInstance<
   defaultPendingMinMs?: number
   caseSensitive?: boolean
 
-  routesById: Record<string, Route<TGenerics>> = {}
+  routesById: Record<string, Route<TGenerics>[]> = {}
 
   constructor({
     location,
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Hmmm, does adding a unique `id` to the route solve the issue?

tannerlinsley:
It will probably be better to open a new PR with the example/docs changes. Thanks!

PR #233, opened 2022-03-06
fix: make disabled prop work correctly on Link

### LINKED ISSUE #230: Setting disabled to true on <Link /> does not prevent navigation
**Describe the bug**
If you create a `Link` like `<Link to="/" disabled={true} />` I would expect the button to do nothing however when pressed currently it will still navigate to the route specified in the `to` field.

I am unsure if this is a bug or just an incorrect assumption so let me know if this is the intended design. I however am leaning towards this being a bug as react-location sets the `aria-disabled` DOM attribute to true when the disabled prop is true.

**To Reproduce**
[CodePen](https://codesandbox.io/s/react-location-link-disabled-bug-700dfi)

The example CodePen has two buttons to change the page.  Both buttons have the `disabled={true}` prop but if you click them you will observe the page change.

**Expected behavior**
The page to not change if the `disabled` prop is set to `true`.

oscartbeaumont:
This change will prevent a `<Link />` element from navigating if the disabled prop is true,

The route will still prefetch on hover when disabled which I could also disable if requested.

Fixes #230

### DIFF
--- a/packages/react-location/src/index.tsx
+++ b/packages/react-location/src/index.tsx
@@ -1389,6 +1389,7 @@ export const Link = function Link<
 
   // The click handler
   const handleClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
+    if (disabled) return
     if (onClick) onClick(e)
 
     if (


oscartbeaumont:
@tannerlinsley This change has been merged for a while and I would like to use it in my project. I can see it was included in [pre-release version v3.6.5](https://github.com/TanStack/react-location/releases/tag/v3.6.5), however I can't find a matching version on npm. Can any of the pre-releases be installed from npm? Also, do you have a timeframe for the next stable release?

Thanks for all the work you do in these amazing open-source projects!

tannerlinsley:
Try @tanstack/react-location

PR #287, opened 2022-06-14
feat: location add state attribute

bingling0084:
#286 
Location type add `state`. You can use `navigate` with state like this:
`const navigate = useNavigate();`
`navigate({ to: '/', state: { name: 'Lily' } });`

And read `state`:
`const location = useLocation();`
get state by `location.current.state`


tannerlinsley:
This will need to be reworked for TanStack Router

PR #296, opened 2022-07-15
fix: qss.ts toValue function

anokthus:
keep url query like "?code=001".


anokthus:
@tannerlinsley 
why don't you merge this commit?
a search value "0101" is not equal to number 101 when it's ID or PIN or name or tel or something code like that.
this function cause fatal bug.

tannerlinsley:
Will you please make the same PR for the `beta` branch as well?

duongdev:
Hey @tannerlinsley when will this PR be released?

PR #313, opened 2022-08-19
fix: memoize matches provider data

rstrand:
The `MatchesProvider` value changes on every render every time the provider is rendered. This causes any component using the `useMatch` or `useMatches` hook to re-render 3 times when navigating. Two of these renders occur before the routes update.

I am using the `useMatch` hook to access route loader data on my pages, and they currently re-render twice before unmounting because of a route change.

This PR fixes this by performing a shallow compare on the RouteMatch array and returning the old value if they are equal.

### DIFF
--- a/packages/react-location/src/index.tsx
+++ b/packages/react-location/src/index.tsx
@@ -482,10 +482,24 @@ export type MatchesProviderProps<TGenerics> = {
   children: React.ReactNode
 }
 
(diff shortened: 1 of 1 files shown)


yergom:
I would love to see this change accepted. We've been bitten by the double re-render before the route is updated several times.

tannerlinsley:
This will need to be reworked for TanStack Router

PR #327, opened 2022-09-22
Export contexts

firtina-lyfta:
Sometimes we need to use these for example for forwarding the context via https://github.com/pmndrs/drei#usecontextbridge


tannerlinsley:
This might need to be reworked for TanStack Router

firtina-lyfta:
Can you please ping me in the issue / PR if so? I had to fork the repo for my version and would like to resync once it's implemented.

PR #332, opened 2022-09-29
Change order of function to match order in (1) 'parseSearch' (2) 'stringifySearch'

tordans:
Using the same order in the code helps me understand it faster.


tannerlinsley:
Will you also double check these docs on the alpha branch for me?

PR #353, opened 2022-11-09
Remove unused import in first example

alii:
Import was confusing as I was expecting to see some data-fetching or example of using with axios in this quick example


alii:
That was quick 😆 thanks so much for your hard work, Tanner!

PR #395, opened 2022-11-23
WIP - Preact adapter

funkyOne:
**This is a work in progress.**

This is a draft version of Preact adapter. I used the react adapter as a base and replaced all React features with Preact equivalents. The goal was zero requirements for `preact/compat`.

The main changes are in this commit - https://github.com/TanStack/router/commit/fb0f7eeb0ce29dc3ae04502a25068af120d48f09

Things that currently still require `preact/compat`:
- Suspense/lazy
- forwardRef
- useSyncExternalStore

All those features do not really have Preact equivalents. They **are** the Preact equivalents, they just reside in `preact/compat` module. Replacing them would essentially mean adding new implementation and including it in the adapter module. And those who use `preact/compat` would pay for imports twice.

I think a good option would be if we could import those features from `preact/compat` cleanly i.e. without triggerring the global tweaks that come with `preact/compat` import .

This is how I was able do this:
```
import { Suspense, lazy as preactLazy } from 'preact/compat/src/suspense'
```
I could do the same with `forwardRef`, but not with `useSyncExternalStore` since it lives in the `index.js`. But anyway importing from src feels like hack. @marvinhagemeister @developit is there a chance these apis could be cleanly exported from `preact/compat`?


andrewiggins:
We (the Preact team) have been discussing making compat more modular/"piecemeal", seeing how helpful it can be to many people. Still working on it though. But perhaps in the meantime we could at least move `useSyncExternalStore` to its own file

tannerlinsley:
Or, if you happen to have any patterns that functionally supersede these apis, happy to use those as well. Also, The async suspense rendering/streaming can’t come soon enough.

flemmingmiguel:
I've done some work on this and due to some of the same things above it was not feasible for my own use to fully port it yet, the upcoming changes to preact streaming and the changes to the router will make for a re-implementation anyways

I'm testing creating the possible scenarios with compat then turn it off and see what's broken and fix accordingly

funkyOne:
> But perhaps in the meantime we could at least move `useSyncExternalStore` to its own file

Yes this would be great! In our particular case a type definition for `forwardRef` would also help. I could create a pr for this if that helps?

> We (the Preact team) have been discussing making compat more modular/"piecemeal", seeing how helpful it can be to many people. 

Sounds exactly the use case here. We want the compat features but don't want the global compat magic. Looking forward to these changes!

funkyOne:
> I've done some work on this and due to some of the same things above it was not feasible for my own use to fully port it yet, the upcoming changes to preact streaming and the changes to the router will make for a re-implementation anyways

Had the same thoughts originally - does it make sense to create (and maintain) a separate adapter if all works ok with `preact/compat`? But given one of the benefits people choose Preact for - its bundle size, I think if we could:
- shave off at least half of the `preact/compat` module size
- avoid `preact/compat` global magic

it will be enough to justify the separate adapter existence.

>the upcoming changes to preact streaming and the changes to the router will make for a re-implementation anyways

what changes, do you envisage, will require a re-implementation?

Saw your examples PR after creating mine, great work converting them all! Prehaps we could sync our PRs at some point?

tannerlinsley:
This needs to be updated. Any takers?

funkyOne:
Hi, sorry for the late response. What significant changes have occurred that require an update?

PR #410, opened 2022-12-01
Adding navigation component and including in kitchen-sink example

jameshiester:
* Added Navigation component to react-router
* Added Navigation component to kitchen-sink example using existing layouts pages.

### DIFF
--- a/examples/react/kitchen-sink/src/main.tsx
+++ b/examples/react/kitchen-sink/src/main.tsx
@@ -13,6 +13,7 @@ import {
   useSearch,
   RegisteredAllRouteInfo,
   RegisteredRouter,
+  Navigate,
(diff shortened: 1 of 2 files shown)


examples/react/kitchen-sink/src/main.tsx
```diff
@@ -180,7 +181,7 @@ const loginRoute = rootRoute.createRoute({
 })
 
 const layoutRoute = rootRoute.createRoute({
-  id: 'layout',
+  path: 'layout',
```
s-cork, line 184: doesn't this defeat the purpose of the layout route in the example?
I thought a layoutRoute didn't have a path by design.

https://tanstack.com/router/v1/docs/guide/layout-routes

Issue #415, opened 2022-12-04
[DevTools]: lazy loading TanstackRouterDevtools resolves to undefined


Moshyfawn:
### Describe the bug

Following the [Only importing and using Devtools in Development](https://tanstack.com/router/v1/docs/devtools#only-importing-and-using-devtools-in-development) section of [Tanstack DevTools docs](https://tanstack.com/router/v1/docs/devtools), lazy loading the `TanstackRouterDevtools` component yields `undefined` 

### Your Example Website or App

https://codesandbox.io/p/sandbox/modern-sound-tk8lsm?file=%2Fsrc%2FApp.tsx&selection=%5B%7B%22endColumn%22%3A40%2C%22endLineNumber%22%3A12%2C%22startColumn%22%3A40%2C%22startLineNumber%22%3A12%7D%5D&workspace=%257B%2522activeFileId%2522%253A%2522cl9zkf3e9000alqiphi669wk8%2522%252C%2522openFiles%2522%253A%255B%2522%252FREADME.md%2522%252C%2522%252Fsrc%252FApp.tsx%2522%252C%2522%252Fpackage.json%2522%252C%2522%252Fnode_modules%252F%2540tanstack%252Freact-router%252Fbuild%252Ftypes%252Findex.d.ts%2522%255D%252C%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522gitSidebarPanel%2522%253A%2522COMMIT%2522%252C%2522spaces%2522%253A%257B%2522clb9kq7em000u2e68b4cw2ugc%2522%253A%257B%2522key%2522%253A%2522clb9kq7em000u2e68b4cw2ugc%2522%252C%2522name%2522%253A%2522Default%2520Space%2522%252C%2522devtools%2522%253A%255B%257B%2522type%2522%253A%2522PREVIEW%2522%252C%2522taskId%2522%253A%2522dev%2522%252C%2522port%2522%253A5173%252C%2522key%2522%253A%2522clb9kq7em000v2e68gobobjm8%2522%252C%2522isMinimized%2522%253Afalse%252C%2522path%2522%253A%2522%252Fblog%2522%257D%255D%257D%257D%252C%2522currentSpace%2522%253A%2522clb9kq7em000u2e68b4cw2ugc%2522%257D

### Steps to Reproduce the Bug or Issue

[...]

PR #420, opened 2022-12-05
📝 docs(DevTools): add a working way to lazy-load DevTools

Moshyfawn:
As per @GyanendroKh's https://github.com/TanStack/router/issues/415#issuecomment-1337289252

Closes: #415


GyanendroKh:
Hi, it has already been fixed by #419

PR #424, opened 2022-12-07
fix: npm start command

ChanhyukPark-Tech:
Thank you for providing me with a good library!

The yarn start command seems to be used in duplicate.
Separated npm and yarn commands.


ChanhyukPark-Tech (commented):
trigger vercel deployment

PR #432, opened 2022-12-08
docs: Fix Core API 404 and Typos

ajcwebdev:
The current link to `https://tanstack.com/router/v1/docs/api/core` results in a 404 so I created a `core.md` file in `docs/api`. I had a few other questions as well about overall documentation contributions.

## How do I doc?

I was wondering if it's possible to build and run the docs site locally? I didn't see anyway to do so by looking at the scripts in the `package.json`.

I wanted to fix the weird anchor link behavior for the Solid/Svelte/Vue coming soon links which result in things like:

```
https://tanstack.com/router/v1/docs/vue-coming-soon
https://tanstack.com/router/v1/docs/#solid-coming-soon
https://tanstack.com/router/v1/docs/vue-coming-soon#svelte-coming-soon
```

My intuition is that the solution would something along the lines of uncommenting these:

```ts
// scripts/config.ts

export const examplesDirs = [
  'examples/react',
  'examples/solid',
  'examples/svelte',
  'examples/vue',
]
```

Fixing the `to` routes defined under the Solid, Svelte, and Vue tutorials in `config.json`.

```json
{
  "menu": [
    {
      "label": "Solid Examples",
      "children": [
        {
          "label": "Coming Soon!",
          "to": "examples/solid/basic"
        }
      ]
    },
    {
      "label": "Svelte Examples",
      "children": [
        {
          "label": "Coming Soon!",
          "to": "examples/svelte/basic"
        }
      ]
    },
    {
      "label": "Vue Examples",
      "children": [
        {
          "label": "Coming Soon!",
          "to": "examples/vue/basic"
        }
      ]
    }
  ]
}
```

And then create the following directories and markdown files with Todo messages in each:

```bash
mkdir -p examples/solid/basic
mkdir -p examples/svelte/basic
mkdir -p examples/vue/basic
echo '# Solid Basic Example\n\nComing soon!' > examples/solid/basic/README.md
echo '# Svelte Basic Example\n\nComing soon!' > examples/svelte/basic/README.md
echo '# Vue Basic Example\n\nComing soon!' > examples/vue/basic/README.md
```

Problem is, I have no idea how these markdown files are being built and packaged and I have no way of knowing whether this would break in some ways until I can test it somehow.

## Possible Additions to the Quick Start

Final thought I had was, I always personally find it useful when Quick Start guides include all or at least most of the steps you'd need to get from 0 to 1. In that spirit, I'd add something like this before the primary code block:

```bash
npm create vite@latest tanstack-router-react -- --template react-ts
cd tanstack-router-react
npm i @tanstack/react-router@beta
```

And then this after the code block:

```bash
npm run dev
```

### DIFF
--- a/docs/adapters/react-router.md
+++ b/docs/adapters/react-router.md
@@ -2,19 +2,13 @@
 title: React Router
 ---
 
-You can install TanStack Router with [NPM](https://npmjs.com),
-[Yarn](https://yarnpkg.com), or a good ol' `<script>` via
-[unpkg.com](https://unpkg.com).
+You can install TanStack Router with [NPM](https://npmjs.com), [Yarn](https://yarnpkg.com), or a good ol' `<script>` via [unpkg.com](https://unpkg.com).
 
 ### NPM
 
 ```sh
 npm install @tanstack/react-router@beta --save
-```
-
-or
-
-```sh
+# or
 yarn add @tanstack/react-router@beta
 ```
 
--- a/docs/api/core.md
+++ b/docs/api/core.md
@@ -1,0 +1,5 @@
+---
+title: Core
+---
+
+TODO
--- a/docs/config.json
+++ b/docs/config.json
@@ -181,11 +181,11 @@
           "to": "examples/react/with-react-query?file=src%2Fmain.tsx"
         },
         {
-          "label": "With TRPC",
+          "label": "With tRPC",
           "to": "examples/react/with-trpc?file=client%2Fmain.tsx"
         },
         {
-          "label": "With TRPC & React Query",
+          "label": "With tRPC & React Query",
           "to": "examples/react/with-trpc-react-query?file=client%2Fmain.tsx"
         }
       ]
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -25,7 +25,7 @@ TanStack Router is a UI router for building applications in React, Preact, Vue, 
 
 TanStack Router builds on concepts and patterns popularized by many other OSS projects, including:
 
-- [TRPC](https://trpc.io/)
+- [tRPC](https://trpc.io/)
 - [Remix](https://remix.run)
 - [Chicane](https://swan-io.github.io/chicane/)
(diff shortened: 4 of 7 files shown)


tannerlinsley:
The site requires a bunch of keys for accessing github, airtable and discord🫣

ajcwebdev:
Interesting, I'm assuming that setup helps you automate a lot of things? I'd be curious what kind of functionality you are afforded specifically and whether it would be possible to either:

- Shift to a more traditional docs site built with a static site generator and markdown but without introducing a huge migration burden or loosing functionality.
- Create a mock environment with keys contributors can be given to develop the site independently from the production site.

Since it's your workflow I have no idea how easy or hard either of those options will be and the exact trade offs they'll introduce so only you'll really know what the right call is. My 2 cents is I'm a big fan of doc sites that keep things as basic as possible so new contributors can make changes without a lot of friction or onboarding.

tannerlinsley:
Alright, I just pushed a few changes that should allow the normal features of the site to work without credentials locally. Mainly, the Airtable and Github APIs have been mocked to empty data where necessary. This means the sponsors will not be pulled and thus the sponsor bubble pack is empty.

Carry on!

tannerlinsley:
> Final thought I had was, I always personally find it useful when Quick Start guides include all or at least most of the steps you'd need to get from 0 to 1. In that spirit, I'd add something like this before the primary code block:

I would be fine with this. A great addition.

tannerlinsley:
As for testing, you would need to clone both the tanstack.com and the router repos down to your filesystem and ensure that the local linking is working between them. It should "just work", and show you the local markdown files you're editing in the remix dev environment of the site.

As far as fixing the coming soon links, thats exactly what would need to happen, except for the `scripts/config.ts` bit. That file can remain as is until the examples become "real".

ajcwebdev:
Awesome, appreciate you taking the time to make those changes! I'll test out that flow and report back if anything weird happens and then I'll add the rest of the discussed changes to this PR.

PR #460, opened 2022-12-20
Docs: Add API docs

hyperNURb:
I started adding everything under API, but I assume the long-term plan is to split it into sections:
- Core API
- Framework specific API i.e. React API

### DIFF
--- a/docs/api/react/Link.md
+++ b/docs/api/react/Link.md
@@ -1,0 +1,11 @@
+---
+id: Link
+title: Link
+---
+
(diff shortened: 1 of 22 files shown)


tannerlinsley:
Nice! Can you add the links to these pages in the docs/config.json file too?

hyperNURb:
@tannerlinsley Ok, I've updated the `docs/config.json` and I've moved everything that was related to React into its own folder.

PR #461, opened 2022-12-20
fix: restore router.history and usePrompt behavior

tombuntus:
- restores `router.history` in the public API or router-core
- restores usePrompt block / unblock behavior

### DIFF
--- a/packages/react-router/src/index.tsx
+++ b/packages/react-router/src/index.tsx
@@ -761,17 +761,16 @@ export function usePrompt(message: string, when: boolean | any): void {
   React.useEffect(() => {
     if (!when) return
 
-    // TODO: bring this back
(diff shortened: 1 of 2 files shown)


tannerlinsley:
My bad, it needs a bit more. This line needs to go away and just use the routers property https://github.com/TanStack/router/blob/3d284a6b626229909f44c78f424d5c181364ae23/packages/router-core/src/router.ts#L413 and also anything referencing that let variable will need to be updated as well.

thomasmarr:
@tannerlinsley I missed that! If I understood correctly, this should be good now.

tombuntus:
@tannerlinsley I missed that! If I've understood correctly, this should be good now.

PR #463, opened 2022-12-21
[Docs] Add-  API definitions and examples for 'createRouteConfig'

saurabhan:
This is a work in progress for Docs, I would like to add definitions and examples for the APIs and components.
something similar to react-router docs, so it's easier to understand what things need to be passed and what they do.

- added options reference for "createRouteConfig", will add more explanation for the same with examples.

Do let me know if this approach is okay.


tannerlinsley:
Looks good!

saurabhan:
Added examples and definitions for createRouteConfig options.

saurabhan:
Can we go ahead with merging this?

tannerlinsley:
Looks like some files that weren’t supposed to be included are there and causing conflicts. Can you remove those files changes?

tannerlinsley:
Looks like you deleted the lock file. Instead, you need to unstage/remove it from the commit.

saurabhan:
Apologies, I am not that versed with git, I tried unstaging and removing seems like that did not work, should I just close this PR and create new from the start?
Thanks

tannerlinsley:
Sometimes that is easier. Go for it!

PR #473, opened 2022-12-31
Referencing pnpm workspace packages incorrectly

yellowsink:
Inside of a pnpm workspace, other packages in the same workspace should be referenced in `package.json` via a `workspace:` version.

This PR adds fixes this in the repo and changes `workspace:*` to `workspace:^`, which will publish to the `^x.x.x` version number scheme most common in JS projects, as opposed to an exact `x.x.x` version

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,10 +28,10 @@
   },
   "spnpm": {
     "overrides": {
-      "@tanstack/router-core": "workspace:*",
-      "@tanstack/react-router": "workspace:*",
-      "@tanstack/router-cli": "workspace:*",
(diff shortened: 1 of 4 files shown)


ap-1:
This change has introduced an error that prevents `@tanstack/react-router@beta` from being installed with npm and pnpm (https://github.com/TanStack/router/issues/474). It seems as though neither npm nor pnpm support this use of `workspace:^`.

tannerlinsley:
Working on a fix right now

yellowsink:
When using workspace: versions, you must use pnpm publish, not npm publish - it will replace it with a normal version number.

PR #481, opened 2023-01-05
Remove extraneous console log statements

SSHari:
First of all, great work on this package. Really cool stuff!

I noticed some extraneous logs cluttering up the console while I was setting things up. Looks like this section was maybe missed during the migration to vitest in this PR: https://github.com/TanStack/router/commit/6d3c0686d086d98eeff80d98aa213ef5af4b0c2c#diff-d59f1a00dff1823e6f16ddf621c708682b1acc37566b17d047bbe4263a580058R337

I know things are moving fast, so feel free to ignore this if it's not worth the time or if you already have it on your radar.


SSHari:
@SeanCassiere That is correct. I can remove the import as well.

PR #493, opened 2023-01-15
fixed useSearch: returns latest data from store

frank-kolesik:


tannerlinsley:
I wish it were this simple, but search params must first pass through search validation. My pending local changes have a fix for this.

Issue #499, opened 2023-01-19
router.reload() adds /undefined to url


hornta:
### Describe the bug

-

### Your Example Website or App

https://stackblitz.com/edit/github-mzt8nn?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

1. Click on click me
2. /undefined is appended to the url

### Expected behavior

-

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

judicaelandria:
we can close this now too @hornta @tannerlinsley

PR #503, opened 2023-01-22
fix(reload): undefined from pathname

judicaelandria:


judicaelandria:
close https://github.com/TanStack/router/issues/499

PR #516, opened 2023-02-03
Solid router adapter

Tanner-Scadden:
Probably needs some revisions to make it cleaner. Copied some of the setup from solid-query to get everything compiling correctly. Packages were created from the react versions to keep 1:1 consistencies and will probably need to be optimized better.

- Created solid-router
- Created solid-store
- Created solid-loaders
- Created solid-actions

A few bugs/issues I've noticed:

- Suspense is needing a wrapper around it for it to resolve and render the child. Not sure if this is a solidjs bug or tanstack router. I have a parent wrapper around it for now but this is adding an unexpected div around children when wrapped in suspense. 
- When the solid-router library is being compiled it is throwing errors for the CSSProperties in the `ErrorComponent`. Added ts-ignore for now, will need to be resolved  too. 

TO DO:
- Solid suspense only works with createResource. Need to figure out how we want this to work with the router. 
- Go through how I have things built and make improvements for performance
- Figure out how to remove values from the stores. This is important for search params. 
- Figure out how to do SSR correctly and create an example

Included Examples: 
- basic
- kitchen-sink-multi-file
- with-trpc-solid-query


packages/solid-loaders/src/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+  )
+
+  return (
+    <loaderClientContext.Provider value={loaderClient}>
+      {coreProps.children}
+    </loaderClientContext.Provider>
+  )
+}
+
+type StrictLoaderInstance_<TLoaderInstance> =
```
tannerlinsley, line 62: Decided to switch this in the core so `data` is always defined, but then in the adapter, we now have `NullableLoaderInstance_` which references the core `NullableLoaderInstance`.

packages/solid-loaders/src/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+        loader: TLoader
+      }
+  ) & {
+    strict?: TStrict
+    track?: (loaderStore: LoaderStore<TData, TError>) => any
+    throwOnError?: boolean
+  } & VariablesOptions<TVariables>,
+): TStrict extends false
+  ? TResolvedLoaderInstance
+  : StrictLoaderInstance_<TResolvedLoaderInstance> {
```
tannerlinsley, line 105: This condition need to be swapped for the change to `NullableLoaderInstance` instead of strict

packages/solid-router/src/index.tsx
```diff
@@ -0,0 +1,712 @@
...
+      <div>
+        <pre>
+          {error.message ? (
+            <code
+              style={{
+                'font-size': '.7em',
+                border: '1px solid red',
+                'border-radius': '.25rem',
+                padding: '.5rem',
+                color: 'red',
```
tannerlinsley, line 702: This markup/style has changed on the react-router package

tannerlinsley (changes requested):

packages/solid-store/src/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  routerStore: Store<TState, TUpdater>,
+  selector: (state: NoInfer<TState>) => TSelected = (d) => d as any,
+  compareShallow?: boolean,
+): Accessor<TSelected> {
+  const [state, setState] = createSignal<TSelected>(
+    selector(routerStore.state),
+    {
+      equals: compareShallow ? false : shallow,
+    },
+  )
```
tannerlinsley, line 20: This will likely need to turn into a solid store and uses solid-store's reconcile to make it granularly reactive.

tannerlinsley (changes requested):

Tanner-Scadden:
Note on the store adapter. When I added reconcile in it wasn't rendering anything. Not sure why, but that would be a huge performance gain.

PR #523, opened 2023-02-10
Allow redirects in route functions and update error handling

tomqchristensen:
This PR amends route error handling behaviour as follows:

Throwing a `redirect` in any of:

- `beforeLoad`
- `validateSearch`
- `onLoad`
- `onBeforeLoadError`
- `onValidateSearchError`
- `onLoadError`
- `onError`

will cause the router to navigate immediately unless preloading. If preloading, the router will bail early but won't navigate.

If anything other than a `redirect` is thrown in `beforeLoad`, `validateSearch` or `onLoad` it will be handled by the corresponding handler (i.e. `onBeforeLoadError` `onValidateSearchError`, or `onLoadError`). If there is no corresponding handler, it will be handled by `onError`. If there is no `onError`, it will be handled by the `errorComponent`.

If a `beforeLoad` error handler returns, the original error is logged in the console. If the handler throws an error, the new error is logged in the console. In either case, no component is rendered (side-note: not sure if this is the desired behaviour, but this is what the router does currently).

If any other error handler returns, the original error is caught by the `errorComponent`. If the handler throws an error, the new error is caught by the `errorComponent`.

This diagram shows what happens in different scenarios (assuming no preloading):

```mermaid
%%{
  init: {
    'theme': 'base',
    'themeVariables': {
      'secondaryColor': 'white',
      'tertiaryColor': '#fff'
    }
  },
}%%
graph LR
    subgraph Legend
        direction LR
        e1(( )) --->|Throw error| e2(( ))
        er1(( )) --->|Throw error or return| er2(( ))
        r1(( )) --->|Throw redirect| r2(( ))
        h1(( )) --->|Happy path| h2(( ))
    end

    A[Load]
    B[Render \n component]
    C[Navigate]

    D(beforeLoad)
    E(onBeforeLoadError \n or onError)

    F(validateSearch)
    G(onValidateSearchError \n or onError)

    H(onLoad)
    I(onLoadError \n or onError)

    J[Render \n errorComponent]
    A --> D

    D --> C
    D --> E
    D --> F

    E --> C
    E ---> X[Uncaught error]

    F --> C
    F --> G
    F --> H

    G --> C
    G ---> J
  
    H --> C
    H --> I
    H --> B
  
    I --> C
    I ---> J

    B --> J

    style Legend color:black,padding:5px;

    linkStyle 0,6,11,16,20 stroke:red,stroke-width:3px;
    linkStyle 1,9,14,19 stroke:orange,stroke-width:3px;
    linkStyle 2,5,8,10,13,15,18 stroke:blue,stroke-width:3px;
    linkStyle 3,4,7,12,17 stroke:lightgreen,stroke-width:3px;
```

### DIFF
--- a/packages/router/src/routeMatch.ts
+++ b/packages/router/src/routeMatch.ts
@@ -164,7 +164,8 @@ export class RouteMatch<
         throw err
       }
 
-      this.route.options.onValidateSearchError?.(err)
+      const errorHandler = this.route.options.onValidateSearchError ?? this.route.options.onError
+      errorHandler?.(err)
       const error = new (Error as any)('Invalid search params found', {
         cause: err,
       })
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Love this. I think the only thing missing (even from my original implementation) is allowing `errorComponent` to handle the uncaught errors from onBeforeLoad. Thoughts?

tannerlinsley:
Wait. Scratch that. `onBeforeLoad` is meant to be a guard against rendering at all. So we might need to think how else to handle that.

tannerlinsley:
Okay, I think I got it. The "required" path is to throw a redirect in `onBeforeLoad`. Throwing any other error just doesn't make sense.  But even in that case, there could be some uncaught error in the mix. If that's the case, we should still show the nearest `errorComponent` to that `onBeforeLoad`.

So back to what I originally thought. We should forward `onBeforeLoad` uncaught errors to the same route's nearest `errorComponent`.

tannerlinsley:
We can do a new PR with further changes. @bholmesdev is going to try this out right now.

Issue #526, opened 2023-02-11
hash link add an redundant # to the first of hash value and not working


mnlbox:
### Describe the bug

I just tried to use Transtack React Router today but when I tried to use the hash link with the bellow code I got a redundant `#` character at the beginning of the hash value, this is my link code:
```
        <Link to="/" hash="test" activeOptions={{ includeHash: true }}>
          Test
        </Link>
```
It created something like this for me as result and it's not working when I click on the generated link:
```
http://localhost:5173/##test
```

### Your Example Website or App

https://stackblitz.com/edit/github-eeczki?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

1. Go to https://stackblitz.com/edit/github-eeczki?file=src%2Fmain.tsx
2. Click on the first two links below this text "Link component not working: "
3. You can see these links are **not working**
4. Click on the last two links below this text "Working Correctly with normal HTML: "
5. You can see these links are **working**

### Expected behavior

As a user, I expected to see a link with only one `#` character and in clickable mode but now it contains two `#` in the generated link like `http://localhost:5173/##test` and it's not working when I click on them.

### Screenshots or Videos

Look at the bellow short video:

https://user-images.githubusercontent.com/25859075/218264155-b61b64fd-4c3c-408d-a711-aea205fe67c0.mp4

### Platform

- OS: Linux
- Browser: Chrome, Firefox
- Version:
    "@tanstack/react-actions": "0.0.1-beta.81",
    "@tanstack/react-loaders": "0.0.1-beta.81",
    "@tanstack/react-router": "0.0.1-beta.82",

PR #527, opened 2023-02-13
fix: remove redundant hash from url

judicaelandria:


judicaelandria:
closes https://github.com/TanStack/router/issues/526

PR #568, opened 2023-03-24
Fix quickstart for bug reports

happycollision:
I don't know if this automatically fixes the link in the issue template, but I think it does.


tannerlinsley:
This is now out of date and probably not required.

PR #577, opened 2023-04-10
docs: Update search-params example code

jafin:
Fix example for warning `Type '{}' is not assignable to type 'string'`


tannerlinsley:
I'll try and get this fixed higher up to avoid the cast.

PR #593, opened 2023-04-27
ensure all layout routes are matched

xyng:
In `beta.84` only the deepest nested layout route is matched, see the reproduction:
https://stackblitz.com/github/xyng/ts-router-nesting-bug-repro?file=package.json

I've noticed that in route matching, the parent routes for pathless (layout) routes are not carried over.
This PR changes the behavior so nested layout routes are correctly added to the match list.

It could possibly be that the matches are not sorted correctly - I'm unsure if that is a problem or not.

Feedback for problems or possible improvements are very welcome - I'm happy to adjust things! :-)

### DIFF
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -596,7 +596,11 @@ export class Router<
           const children = route.children as undefined | Route[]
           if (!route.path && children?.length) {
             return findMatchInRoutes(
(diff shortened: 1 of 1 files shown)


tannerlinsley:
I reimplemented this logic internally to holistically tackle the issues. That code has already been released.

PR #594, opened 2023-05-03
feat: named imports

orangeswim:
Add named imports to lazy wrapper.


orangeswim:
I would add to the kitchen sink react example, but that is currently broken with useLoader for invoices.

PR #608, opened 2023-06-05
fix: link to react adapter

ahmadalfy:
The link to React Adapter is not working

image

This is the correct URL:

image


tannerlinsley:
Out of date.

PR #636, opened 2023-07-12
fix: stricter actions payload type

pawelblaszczyk5:
Hello,
Per our discussion on Discord I tried to fix action payload types to differentiate between functions with and without arguments. I'm not 100% sure about my solution (especially the payload part) but wanted to give it a try so feel free to point out errors 😃 It correctly infers both payload in submit fn and in submissions.

Example with parametrized fn:

image

Example with fn without parameters:
image


pawelblaszczyk5:
Hmm, I'm not sure about the failing actions, any ideas? It seems like some build error

PR #657, opened 2023-07-31
examples: fix invalid root route usage in multi file usage example

lamualfa:
# Original example 
In the original example, the page doesn't show anything and gives an error in the dev tools.

https://codesandbox.io/s/github/tanstack/router/tree/beta/examples/react/kitchen-sink-multi-file

image

image

# Resolved example

https://codesandbox.io/p/sandbox/unruffled-dan-p3w7jg


markdalgleish:
I think this is caused by a circular dependency. I've opened another PR to address this: https://github.com/TanStack/router/pull/663

lamualfa:
> I think this is caused by a circular dependency. I've opened another PR to address this: https://github.com/TanStack/router/pull/663

Thanks. I'll close this PR for more reliable solution.

PR #663, opened 2023-08-01
examples: fix routerContext circular dependency

markdalgleish:
Without this change I get a blank page with the error `Uncaught ReferenceError: Cannot access 'routerContext' before initialization at root.tsx`. Here's a StackBlitz of the latest commit showing the issue: https://stackblitz.com/github/tanstack/router/tree/5547d5f0f3f142ce0a64047d476bc0f59d11f15b/examples/react/kitchen-sink-multi-file

Turns out it's because there's a circular dependency between `router.tsx` and `routes/root.tsx`. To fix this I've extracted `routerCounter` into its own file.


tannerlinsley:
Mark! What's up! Thanks for catching this one. 😄

PR #671, opened 2023-08-09
[WIP]: Solid Adapter

Tanner-Scadden:
Basic example loading root page. 

### Examples
- [ ] Kitchen Sink (good to catch bugs)

### Code Organization
- [ ] Split out into it's own package

### Reactivity 
- [ ] Link (active not working properly)

### Packages
- [ ] Loaders (Ported but not tested)
- [ ] Actions (Ported but not tested)

### Route Methods
- [x] useMatch
- [x] useLoader
- [ ] useContext (Ported just not tested)
- [ ] useRouteContext (Ported just not tested)
- [ ] useSearch (Ported just not tested)
- [ ] useParams (Ported just not tested)


tannerlinsley:
This is looking great. I think it would be really helpful to remove the core router files and just have a `solid-router.ts` file or something that consumes `@tanstack/router-core`. That will ensure that we aren't changing any of the core routing behavior (and if we need to, we can handle that upstream)

examples/solid/basic/src/router/solid-router.tsx
```diff
@@ -0,0 +1,1100 @@
...
+    'class',
+    'onClick',
+    'onFocus',
+    'onMouseEnter',
+    'onMouseLeave',
+    'onTouchStart',
+  ])
+
+  const linkInfo = useRouterState({
+    select() {
```
Tanner-Scadden, line 457: Stale closure. Will need to figure out a way to do this better.

tannerlinsley:
I can definitely change this however we need.

Tanner-Scadden:
> I can definitely change this however we need.
> […](#)
> On Aug 9, 2023 at 10:23 PM -0400, Tanner Scadden ***@***.***>, wrote: @Tanner-Scadden commented on this pull request. In examples/solid/basic/src/router/solid-router.tsx: > + 'to', + 'preload', + 'preloadDelay', + 'replace', + // element props + 'style', + 'class', + 'onClick', + 'onFocus', + 'onMouseEnter', + 'onMouseLeave', + 'onTouchStart', + ]) + + const linkInfo = useRouterState({ + select() { Stale closure. Will need to figure out a way to do this better. — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you commented.Message ID: ***@***.***>

I'll probably need a way to subscribe to router properties changing instead of just state to get this to be reactive. If I can already do this let me know and I'll add a special store for it!

PR #683, opened 2023-08-20
[WIP]: Svelte Adapter

Theo-Steiner:
## Background

I started building an adapter for @tanstack/router for svelte by incrementally porting from @tanstack/react-router to svelte.

Incremental approach: implement the examples for react and use them as test cases for reaching feature parity with @tanstack/react-router

## Progress

- [x] create base example for svelte
- [x] port base functionality to svelte
- [ ] output a package with uncompiled svelte components (currently here, rollup for now compiles the svelte components instead of keeping them as is)
- [ ] test the base functionality in the base example, fix bugs...
- [ ] move on to non-base example and repeat the above steps

## Current Blockers

- Need to figure out how to output *uncompiled* svelte components + d.ts files, so users can use them from their svelte apps
- Need to figure out a better error boundary mechanism (for now using window.onerror)
- Need to figure out a replacement mechanism for Suspense (might become more important when it comes to loading)
- Need to figure out end2end typesafety


tannerlinsley:
So much has changed... Let's circle back around fresh.

PR #690, opened 2023-08-23
Add state to location nextOpts

hjbolide:
Fixed the issue where location state is ignored during navigation.


hjbolide:
Next will be having a proper test suite covering the navigations

PR #700, opened 2023-08-28
fix(types): remove typeof this in static method causing typescript errors

KurtGokhan:
This was a weird one. Although it doesn't cause Typescript build to fail, it was generating a code like this when built:

```ts
static __onInit: (route: typeof this$1) => void;
```

This is obviously a wrong Typescript code and it was causing errors in my project. Not sure if this is a common error for everyone. But I think not using `typeof this` here would be safer.

On another note, I see that `__onInit` method is used as some form of polymorphism. Imho, real polymorphism should be used and each framework should export their own `Route` class. (And the base Route class could be renamed to `BaseRoute` or `AbstractRoute` to avoid confusion maybe). Eventually someone may need to have two different frameworks in the same runtime and it wouldn't work because both frameworks are overwriting the `__onInit` method.

### DIFF
--- a/packages/router-core/src/route.ts
+++ b/packages/router-core/src/route.ts
@@ -842,7 +842,7 @@ export class Route<
     return this
   }
 
-  static __onInit = (route: typeof this) => {
+  static __onInit = (route: Route) => {
     // This is a dummy static method that should get
     // replaced by a framework specific implementation if necessary
   }


tannerlinsley:
I initial went down this "route" 😂, but was quickly barred by TS. Having multiple versions of Route going around from different frameworks that extend the base class should have worked, but for reasons I don't understand, the generic system started breaking down and constraints failing (TS not compiling) when I modeled it that way. This hack got me around most  of that. 🤷‍♂️

KurtGokhan:
Oh, I understand that. You start relying too much on the TS and start typing everything perfectly. Then it just runs into a wall. The solution is usually breaking the type chain by inserting `any` somewhere and kinda resetting it. I guess this reduces the depth of the types and TS can breathe for a bit in order to process other types. It also reduces the risk of having circular types.

I wonder how libraries like [hotscript](https://github.com/gvergnaud/hotscript) can handle so much types, or if they are just a gimmick.

Issue #707, opened 2023-08-31
Support for static prefix on dynamic paths


FugiTech:
### Describe the bug

Currently TanStack Router supports dynamic path segments in the form `/$var/`. However it is sometimes useful to have a static prefix (or suffix) that is used as part of the matching. For example `youtube.com/@ChannelName`.

This is a feature request for Router to support the form `/prefix$var/` (or some other way of having a static prefix mixed with a dynamic variable). Optionally it would be nice to also support suffixes, perhaps in the form `/prefix$var$suffix/`?

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-jwzga9?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

```typescript
new Route({
  path: '/prefix$var',
  component: function User({ useParams }) {
    const params = useParams();
    return <div className="p-2">Hello {params.var}!</div>;
  },
});
```

This will not be recognized as a dynamic route and `var` will not be available in `params`

### Expected behavior

`var` would be available in `params`

### Platform

N/A

tannerlinsley:
Now that things are pretty stable, I'd like to see this PR updated.

FugiTech:
I updated the PR to remove the merge conflicts but didn't do an in-depth look at what's changed in the last 4 months

JacobWennebro:
Any update on this?

schiller-manuel:
this was closed: https://github.com/TanStack/router/pull/709

would need to be redone

enheit:
Any updates?

FugiTech:
My understanding is that while this feature is something they are willing to support, it is not a high priority and nobody on the core team is working on it. (I am not a maintainer and have no real knowledge of their priorities)

I originally suggested it since it'd be useful in a project I was working on, but I have since dropped that project and am not actively using TanStack router, therefore I do not have time to implement this feature myself.

If somebody is looking for this feature, I recommend filing a PR implementing it yourself. The previously mentioned PR _may_ be helpful as a starting point, but I imagine a lot has changed since I made it.

schiller-manuel:
@FugiTech is absolutely correct. core team is not working on this feature but we are happy to accept PRs. if anyone is interested in implementing this, reach out to us e.g. via discord

CanRau:
Hey would love to chime in here as this is currently kinda a roadblocker for me to potentially migrate our platform to Start as we have /@username paths.

[...]

PR #709, opened 2023-09-01
Support static prefixes & suffixes in dynamic path segments

FugiTech:
This address #707 by adding support for routes of the form `prefix$var$suffix`. This is technically a breaking change in that you can no longer have `$` in the name of a dynamic path segment param. I'm unsure if that is something people do. If it is concerning I can adjust this to only support prefixes which would not be a breaking change.

I'm not familiar with the code base so I apologize if this is incomplete or broken. It seemed like there wouldn't be any changes needed to react-router as all the logic was in router-core. I wasn't sure if the cli needed changes.


tannerlinsley:
I fixed up some more of the conflicts. It looks like a few examples broke with this (infinite loops?). So we'll need to hold off on this.

tannerlinsley:
This would need to be updated with all of the new stuff, including file based route generation.

Issue #722, opened 2023-09-11
`useNavigate` does not use the registered router for typesafety


FugiTech:
### Describe the bug

The generated type for `useNavigate` and `Navigate` in Beta 185 is 
```typescript
export declare function useNavigate<TDefaultFrom extends RoutePaths<RegisteredRouter['routeTree']> = '/'>(defaultOpts?: {
    from?: TDefaultFrom;
}): <TFrom extends unknown = TDefaultFrom, TTo extends string = "">(opts?: NavigateOptions<any, TFrom, TTo> | undefined) => Promise<void>;
export declare function Navigate<TFrom extends RoutePaths<RegisteredRouter['routeTree']> = '/', TTo extends string = ''>(props: NavigateOptions<RegisteredRouter['routeTree'], TFrom, TTo>): null;
```

The `NavigateOptions<any` in `useNavigate` should be `NavigateOptions<RegisteredRouter['routeTree']` to match `Navigate`. Without it we get no typesafety for calls to `useNavigate`.

See https://discord.com/channels/719702312431386674/1007702008448426065/1150468110546128948 for prior discussion

### Your Example Website or App

N/A

### Steps to Reproduce the Bug or Issue

1. Define a router with the paths `/` and `/example`
2. call `const navigate = useNavigate()`
3. call `navigate({ to: '/invalid' })`

### Expected behavior

It is expected to get a type error as `/invalid` is not a registered route. Instead it is accepted.

### Platform

- OS: macOS
- Browser: N/A
- Version: N/A

PR #723, opened 2023-09-11
Fix types for useNavigate and useMatchRoute

FugiTech:
Addresses #722 and also fixes the `any` typing for `useMatchRoute` that I noticed while working on this.

The root of the problem seems to be that `rollup-plugin-dts` sees `RegisteredRouter` as `AnyRouter` while generating types, so any types using it get compiled down into `any` and `unknown`. The trick appears to be to define dynamic types that are set to `RegisteredRouter` by default, which tricks `dts` into leaving it alone.

For `useMatchRoute` because `const router = useRouter()` comes out typed as `AnyRouter` (since `useRouter` returns type `RegisteredRouter`), the `.matchRoute` call returns `any`. Therefore we have to explicitly type the return value to be the same as `Router.matchRoute`.

In `useMatchRoute` I also changed the type of `TFrom` to `RoutePaths<TRouteTree>` instead of `string` to match `Router.matchRoute`. Please let me know if this was incorrect.

Also for `useMatchRoute` I noticed that the parameter splitting of `opts` only extracts `pending` and `caseSensitive`, but `MatchRouteOptions` also includes `includeSearch` and `fuzzy`. I was unsure if this was a bug or not so I left it alone, but wanted to bring it up.

### DIFF
--- a/packages/react-router/src/react.tsx
+++ b/packages/react-router/src/react.tsx
@@ -907,13 +907,7 @@ export function useNavigate<
       TMaskFrom extends RoutePaths<TRouteTree> = '/',
       TMaskTo extends string = '',
     >(
-      opts?: NavigateOptions<
-        RegisteredRouter['routeTree'],
-        TFrom,
-        TTo,
-        TMaskFrom,
-        TMaskTo
-      >,
+      opts?: NavigateOptions<TRouteTree, TFrom, TTo, TMaskFrom, TMaskTo>,
     ) => {
       return router.navigate({ ...defaultOpts, ...(opts as any) })
     },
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Could you rebase your changes on top of the latest?

FugiTech:
Done! Interestingly seems like `TRouteTree extends AnyRoute = RegisteredRouter['routeTree']` has to be done at the top-most function and can't be done in a nested callback or it turns back into `any`

tannerlinsley:
Lots of conflicts here now. Might be worth restaging the changes.

tannerlinsley:
Ahhh yeah

PR #744, opened 2023-09-26
chore: fix broken examples for anybody using pnpm >= v8

marbemac:
Anybody on pnpm >= 8 will be unable to run most of the examples. This is because pnpm v8 changed the default behavior of `auto-install-peers` from `false` to `true`. This results in multiple versions of react getting installed and used (18.2.0, and 18.3.0-canary-587ae49d7-20230720), which leads to the error in the image below. It seems `use-sync-external-store@1.2.0` actually lists this canary react version as a peer dep...

Screenshot 2023-09-26 at 11 12 57 AM

---

Several options to resolve this:

**1. Explicitly configure pnpm `auto-install-peers=false`, so that pnpm v8 behavior doesn't change vs v7. **

This is the change in this PR atm, BUT unfortunately it doesn't really solve the problem. 

This is because if npm is not able to read any env variables used in the npmrc file it IGNORES the entire file! https://github.com/pnpm/pnpm/issues/5914. Thus if the end user doesn't have the `NPM_TOKEN` env variable set when they run `pnpm i`, they'll still end up with the default auto-install-peers=true.

This might be why maintainers don't end up with this problem (if ya'll are on pnpm v8 and happen to have the NPM_TOKEN) env variable set.

**2. Set engines.pnpm in the root package.json to force pnpm 7 or bail out with an error**

```json
{
  "engines": {
    "pnpm": "^7"
  },
}
```

**3. Set react and react-dom in the root package.json overrides to point to `18.3.0-canary-587ae49d7-20230720`, since a key dep (use-sync-external-store) lists this as a peer dep.**

This might be the simplest and would cover all users without requiring changes on their end to their pnpm version or env variables.

---

Up to maintainers... which approach do ya'll want to go with? (or is there one I haven't though of?). I'll update the PR to reflect the decision.

SIde note, looks like maybe `pnpm i` hasn't been run in a while? Running it resulted in lots of cleanup (including removing old examples, etc) in the lock file, hence the large diff.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,5 @@
 //registry.npmjs.org/:_authToken=${NPM_TOKEN}
+
+# Note, the settings in this file are ignored by pnpm if the NPM_TOKEN env variable is not set...
+# https://github.com/pnpm/pnpm/issues/5914
+auto-install-peers=false
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -195,7 +195,7 @@ importers:
         specifier: workspace:*
         version: link:../../../packages/react-start
       '@tanstack/router-cli':
-        specifier: 0.0.1-beta.185
+        specifier: 0.0.1-beta.192
         version: link:../../../packages/router-cli
       '@tanstack/router-core':
         specifier: workspace:*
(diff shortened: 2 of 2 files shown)


tannerlinsley:
Fixed already

PR #756, opened 2023-10-04
docs: Fix a number of inconsistencies wrt. `getParentRoute()`

brncsk:
`getParentRoute()` was called `getParentPath()` or `getRootRoute()` in a number of places. I can't be sure but hope these were just oversights and were not referring to something else.


tannerlinsley:
Fixed already

PR #763, opened 2023-10-10
fix: added `isPreloaded` flag to `RouteMatch`

schiller-manuel:
this allows to only clean a match based on `preloadMaxAge` if this match was really preloaded


tannerlinsley:
Added already

PR #768, opened 2023-10-12
fix: expose createHistory function to public API

KurtGokhan:
Discussion [here](https://github.com/TanStack/router/discussions/764)


tannerlinsley:
Fixed!

Issue #786, opened 2023-10-31
Browser history navigation broken


mer10z:
### Describe the bug

In 0.0.1-beta.198, browser navigating backwards and forwards changes the URL but does not update the page. It looks like this was introduced with the recent history changes. Beta 193 is the most recent version that is not broken.

### Your Example Website or App

tanstack-router-react-example-with-loaders-kitchen-sink-code-split

### Steps to Reproduce the Bug or Issue

1. Run tanstack-router-react-example-with-loaders-kitchen-sink-code-split locally (not sure how to navigate in stackblitz)
2. Go to a few different routes
3. Try to navigate back and forward
4. URL changes but page content does not

### Expected behavior

Navigating back or forward updates page with proper route content

### Platform

- OS: macOS
- Browser: Chrome
- Version: 118.0.5993.88

plbstl:
+1

plbstl:
I am thinking `history.update` is meant to be called in lines 289 and 293 of the history package.

https://github.com/TanStack/router/blob/3a32a8fc3a00834106ed2142c0acccc42ad67379/packages/history/src/index.ts#L287-L294

mer10z:
Looks like this is fixed in beta 204

PR #787, opened 2023-11-01
fix: update page when navigating backwards and forwards

plbstl:
see #786

### DIFF
--- a/packages/history/src/index.ts
+++ b/packages/history/src/index.ts
@@ -279,19 +279,21 @@ export function createBrowserHistory(opts?: {
     destroy: () => {
       window.history.pushState = pushState
       window.history.replaceState = replaceState
-      window.removeEventListener(pushStateEvent, history.update)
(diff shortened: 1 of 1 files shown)


tannerlinsley:
This is the right fix, but it needs to extract the callbacks in a way that they can be removed in the destroy method.

plbstl:
Okay, would that be something like this?

```ts
  const history = createHistory({
    // ...
    destroy: () => {
      window.history.pushState = pushState
      window.history.replaceState = replaceState
      window.removeEventListener(pushStateEvent, updateCurrentLocationAndHistory)
      window.removeEventListener(popStateEvent, updateCurrentLocationAndHistory)
    },
  })

  function updateCurrentLocationAndHistory() {
    currentLocation = parseLocation(getHref(), window.history.state)
    history.update()
  }

  window.addEventListener(pushStateEvent, updateCurrentLocationAndHistory)
  window.addEventListener(popStateEvent, updateCurrentLocationAndHistory)
```

PR #790, opened 2023-11-02
feat: pass window instance to createBrowserHistory

pserrer1:
Adds support for passing a Window instance to createBrowserHistory. This is especially usefull in case you need to route on parent windows (i.e. inside Micro-Frontends)


packages/history/src/index.ts
```diff
@@ -174,20 +174,22 @@ function assignKey(state: HistoryState) {
...
 export function createBrowserHistory(opts?: {
   getHref?: () => string
   createHref?: (path: string) => string
+  win: Window = window
 }): RouterHistory {
   const getHref =
     opts?.getHref ??
     (() =>
-      `${window.location.pathname}${window.location.search}${window.location.hash}`)
+      `${win.location.pathname}${win.location.search}${win.location.hash}`)
```
xeinebiu, line 188: Perhaps rename the `win` to `_window` or to be consistent, refactor it to a callback instead.

getWindow()?: () => Window

pserrer1: What would be the benefit of a callback? It would only be called once as we otherwise would need some sort of subscription to get informed about "possible" changes of the returned window instance.

I still updated the signature slightly, as there was an error inside

tannerlinsley:
Fixed

pserrer1:
@tannerlinsley How? I don't see any changes regarding the used window?

tannerlinsley:
Sorry, I thought I had committed them already. https://github.com/TanStack/router/blob/beta/packages/history/src/index.ts#L184

Issue #842, opened 2023-12-11
AbortController abort not handled properly for data loader


orangeswim:
### Describe the bug
 EDIT: Originally titled Strict Mode not handled properly. Changed because bug happens as well when out of strict mode. See my latest response.

See stackblitz link on changes from data example. 

Instructions on usage of abortController, https://tanstack.com/router/v1/docs/guide/data-loading#using-the-abort-signal
Per react documentation, React.StrictMode is supposed to make sure your components handle unmount and remount properly.

Visit the stackbllitz link, navigate to an individual post. Then hit refresh on that page (in the frame) so that StrictMode will mount the route twice. 

Actual behavior, route stays in aborted mode. In a different copy of the code I printed out via console.log, signal.aborted, which is a boolean. It shows it is aborted on both mounts. 

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-yexyqa?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

1. Go to stackblitz
2. Navigate to Posts then click a post so you are on a path like `https://qgnimrlxlgithub-bmvg--3001--d3416dfd.local.webcontainer.io/posts/1`
3. Refresh the frame inside stackblitz to trigger StrictMode
4. Component shows an error

### Expected behavior

Expected behavior, even if the route has an error, because we are unmounting, this error should be discarded. The newly mounted route (second time) should behave normally.

### Screenshots or Videos

![image](https://github.com/TanStack/router/assets/1499212/a52dcd6d-58fa-49b5-a6dc-918659424d39)

### Platform

- OS: [macOS 14.1.2 (23B2091)]
- Browser: [Chrome Version 119.0.6045.199 (Official Build) (arm64), Safari Version 17.1.2 (19616.2.9.11.12)]

### Additional context

Would have to turn off strict mode in development which is not ideal.

orangeswim:
It looks like a new abortController isn't created even if we try to revisit the route.


1. Change loader to have a 1 second delay.
1. Enter into aborted state for /posts/1
2. Routes entering  /posts/1 will always get an abortController that is already aborted.
3. Switch to /posts/2 and observe abortController is false. Switch back before page loads and /posts/1 is still aborted.
4. Switch to /posts/3 again and wait until it complets.
5. This time switching to /posts/1 provides a new abortController

Looking at more console logs, the match is stored in state.matches

orangeswim:
We can also enter the bad state with a different method. Remove strict mode
[...]

PR #843, opened 2023-12-12
Update router.ts

orangeswim:
Refresh AbortController if needed.

See issue #842 on AbortController issue


orangeswim:
Issues in checks above don't appear related to th PR. Need to do some digging on why it failed

tannerlinsley:
Technically, we should be creating a new abort controller for each attempted load, so I superseded this PR with that change. it should now be available.

PR #845, opened 2023-12-12
Update router.ts to fix AbortController

orangeswim:
Move changes to loadMatches

See #842 

handleErrorAndRedirect gets it own AbortController.
Then we refresh the AbortController on each route. In a new fresh case, each Route already has its own AbortController.


tannerlinsley:
Ah! You already did!

Issue #881, opened 2023-12-26
RouteApi routeId issues


V1RE:
### Describe the bug

The actual id for a route can differ from the RouteApi's constructor ids shown from TS when using routegen and "layouts"

### Your Example Website or App

https://stackblitz.com/~/github.com/V1RE/tanstack-router-route-api-issue

### Steps to Reproduce the Bug or Issue

The layout-a page shows issues when trying to find the route based on the ts autocompleted route id.

### Expected behavior

As a user I expect the autocompletions for the routeApi to resolve to the correct route id's

### Screenshots or Videos

<img width="512" alt="image" src="https://github.com/TanStack/router/assets/20016103/526f51b2-9cae-4b96-9f9e-c5ac84bb8e9b">

### Platform

Irrelevant

### Additional context

Could either be solved by making the routeApi constructor match the id transformation of the route constructor, or by changing the auto suggestions, but the latter option seems counter-intuitive

schiller-manuel:
This seems to be a problem of file based routes, rather than `RouteApi`. Same issue with e.g. `useNavigate(from: '...')` hook.

V1RE:
@schiller-manuel It seems to stem from the [logic in the `Route` constructor](https://github.com/TanStack/router/blob/912fae7e67f8a5b6a9b7445044393e3259a2a717/packages/react-router/src/route.ts#L558) and the way the route get's [matched based on the "from" property](https://github.com/TanStack/router/blob/912fae7e67f8a5b6a9b7445044393e3259a2a717/packages/react-router/src/Matches.tsx#L310) if I'm not mistaken

schiller-manuel:
I think the issue lies in the generated code:

```typescript
import { Route as rootRoute } from "./routes/__root"
import { Route as IndexRoute } from "./routes/index"
import { Route as LayoutRoute } from "./routes/_layout"
import { Route as PostsRoute } from "./routes/posts"
import { Route as PostsIndexRoute } from "./routes/posts.index"
import { Route as LayoutLayoutARoute } from "./routes/_layout/layout-a"
import { Route as LayoutLayoutBRoute } from "./routes/_layout/layout-b"
import { Route as PostsPostIdRoute } from "./routes/posts.$postId"
import { Route as PostsPostIdDeepRoute } from "./routes/posts_.$postId.deep"

declare module "@tanstack/react-router" {
  interface FileRoutesByPath {
    "/": {
      parentRoute: typeof rootRoute
    }
    "/_layout": {
[...]

PR #884, opened 2023-12-26
Update logic to retrieve matching routes

V1RE:
There's probably something I completely missed, but this seems to fix the issues I reported in #881.

Tried to modify the generator in the CLI to make the paths match the final IDs of the routes, but ran into issues since the underscores seem to get used in the TS magic and didn't feel comfortable changing those types.

Also made the route matching first try to match the routeId as given and fallback to the opts.from without a `/` as last char.

Also added an example of RouteApi being used in a lazy-loaded component.


schiller-manuel:
Please have a look at PR #886, it solves the underlying problem

PR #903, opened 2024-01-02
fix: return correct `next` in `replaceEqualDeep`

kahosan:
When the lengths of `next` and `prev` are the same, if there is a field in `next` that does not exist in `prev`, and its value is `undefined`, it will lead to `undefined === prev[key] // undefined`, causing an increase in `equalItems`. This returns incorrect results.

example:

```ts
replaceEqualDeep({
  a: 2,
}, {
  b: undefined
}) // { a: 2 }
```


schiller-manuel:
can you add some unit tests for this?

kahosan:
> can you please add some unit tests for this?

I have added some unit tests, please take a look

packages/react-router/tests/utils.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    const obj2 = { a: 1, b: { c: 3 } };
+    const result = replaceEqualDeep(obj1, obj2);
+    expect(result).toStrictEqual(obj2);
+  });
+
+  it('should properly handle non-existent keys', () => {
+    const obj1 = { a: 2, c: 123 }
+    const obj2 = { a: 2, b: undefined };
+    const result = replaceEqualDeep(obj1, obj2);
+    expect(result).toStrictEqual(obj2);
```
schiller-manuel, line 36: @tannerlinsley  is this expectation correct?

the function's comment says:

> This function returns `a` if `b` is deeply equal.
> If not, it will replace any deeply equal children of `b` with those of `a`.

What about children that exist in `b` but not in `a`?

tannerlinsley: If children exist in B, then a new object will be returned with anything else that hasn't changed from A, but with the new stuff from B:

So it's possible than an object would be a mix:

{
  foo: A,
  bar: B,
  sub: {
    foo: B
  }
}

packages/react-router/tests/utils.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  it('should handle nested objects correctly', () => {
+    const obj1 = { a: 1, b: { c: 2 } };
+    const obj2 = { a: 1, b: { c: 3 } };
+    const result = replaceEqualDeep(obj1, obj2);
+    expect(result).toStrictEqual(obj2);
+  });
+
+  it('should properly handle non-existent keys', () => {
```
schiller-manuel, line 32: this test also passes with your fix :(

kahosan: > this test also passes without your fix :(

i found that the original replaceEqualDeep function also won't add fields that exist in a but not in b to b:

```
const obj1 = { a: 2, c: 123 }
const obj2 = { a: 2 };

result = { a: 2 };
```

can you tell me the expected result?

schiller-manuel:
Can you please explain where you had react-router behave in a way that led you to discover this issue?

kahosan:
> Can you please explain where you had react-router behave in an incorrect way that led you to discover this issue?

when I use the `search` property of `navigate`, previous params are
```
const a = {
  page: 1,
  age: 2,
  order: 'asc'
}
```

the new search params are

```
const b = {
  page: 1,
  age: 2,
  filter: undefined
}
```

then I found that the new search params were not updated.

schiller-manuel:
Thanks for the explanation, I was able to reproduce this issue here
https://codesandbox.io/p/devbox/nostalgic-sky-qntm45?file=%2Fsrc%2Fmain.tsx%3A123%2C13

https://github.com/TanStack/router/assets/6340397/db99c05b-088c-43a1-980e-0061e6c5cfe0

I need to fully understand `replaceEqualDeep` to decide whether your fix is the right one or if something else is needed.

schiller-manuel:
The following fixes the issue and seems more generic, what do you think?

```diff
diff --git a/packages/react-router/src/utils.ts b/packages/react-router/src/utils.ts
index 90859f0f..faa06830 100644
--- a/packages/react-router/src/utils.ts
+++ b/packages/react-router/src/utils.ts
@@ -156,7 +156,6 @@ export function replaceEqualDeep<T>(prev: any, _next: T): T {
   const array = Array.isArray(prev) && Array.isArray(next)
 
   if (array || (isPlainObject(prev) && isPlainObject(next))) {
-    const prevSize = array ? prev.length : Object.keys(prev).length
     const nextItems = array ? next : Object.keys(next)
     const nextSize = nextItems.length
     const copy: any = array ? [] : {}
@@ -166,14 +165,13 @@ export function replaceEqualDeep<T>(prev: any, _next: T): T {
     for (let i = 0; i < nextSize; i++) {
       const key = array ? i : nextItems[i]
       copy[key] = replaceEqualDeep(prev[key], next[key])
-      if (copy[key] === prev[key]) {
+      if (copy[key] === next[key]) {
         equalItems++
       }
     }
 
-    return prevSize === nextSize && equalItems === prevSize ? prev : copy
+    return equalItems === nextSize ? next : copy
   }
-
   return next
 }
 ```

PR #942, opened 2024-01-07
fix: Link forgets path param

schiller-manuel:


StireArLankar:
Hello
I would like to ask what is the reasoning behind this commit? Because with these changes `from` property in `Link`, `useNavigate`, etc (which was present as `dest.from`) no longer is used for actual navigation and becomes useless?

schiller-manuel:
The `from` property is just an aid for typesafety.
How could it even be used at runtime? when you click a `<Link>`, you are navigating from your current location.
A good example are path params:
At compile time, the path param "placeholder" `$postId is used` `<Link from="/posts/$postId" ...>`, while at runtime it is an actual value.

StireArLankar:
Oh, i investigated a bit more and it seems like i was looking for non existent feature since around ~188 beta

I am using version 0.0.1-beta.83 and these links behave the same way ([repro sandbox](https://codesandbox.io/p/devbox/infallible-keller-98czg5?file=%2Fsrc%2Fmain.tsx%3A86%2C29))
```tsx
<Link
  from={namespaceRoute.fullPath}
  to="./test"
>
  To /Namespace/test
</Link>
<Link
  to="/$namespace/test"
  params={{ namespace: "namespace" }}
>
  To /Namespace/test
</Link>
```

somewhere around beta.188 and up until this commit they behave like [this](https://codesandbox.io/p/devbox/zen-thunder-g3cch7?file=%2Fsrc%2Fmain.tsx%3A89%2C27) - link with `from` prop has `undefined` instead of actual param value

after this commit ([repro sandbox](https://codesandbox.io/p/devbox/focused-haslett-n7c9ft?file=%2Fsrc%2Fmain.tsx%3A89%2C27)) first link resolves href on mount relative to current location - as you stated in your comment (so if page loads on `/namespace`, then link is `/namespace/test`)
but!
1. it doesnt consider itself as active when second link is
2. if component with link renders from scratch (for example i reload component with new key), then link resolves href one more time, and if current path is `/namespace/test`, then href for link is resolved relative to current location and it becames `/namespace/test/test`

I am not really sure if users expect such behaviour - i mean, it says in [docs](https://tanstack.com/router/v1/docs/guide/navigation#relative-links) `If you want to make a link that is relative to the current route`, not relative to current location

I think its reasonable to expect that when you pass `from` that `to` will be actually resolved from this `from` (pun intended) value (with actual params, just like in first sandbox) and not from current location (which is actually not really current location in my sandbox, but location during component mount)

Issue #940, opened 2024-01-05
LoaderContext type is not exported from @tanstack/react-router


aadamw:
### Describe the bug

According to this section of documentation about code-splitting: https://tanstack.com/router/v1/docs/guide/code-splitting#data-loader-splitting 
`LoaderContext` type should be exported from `@tanstack/react-router`, but in reality is not.

### Your Example Website or App

https://stackblitz.com/edit/github-vwuepr?file=src%2Fposts.route.tsx

### Steps to Reproduce the Bug or Issue

1. Open the reproduction link
2. Go to `/src/posts.route.tsx` file
3. Hover over LoaderContext import
4. You should see following error:
```
'"@tanstack/react-router"' has no exported member named 'LoaderContext'. Did you mean 'LoaderFnContext'?ts(2724
```

### Expected behavior

`LoaderContext` type is exported from @tanstack/react-router

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

PR #944, opened 2024-01-07
fix: export loader context interface from @tanstack/react-router

buzzracharyeah-rejan:


packages/react-router/src/router.ts
```diff
@@ -163,6 +163,10 @@ export interface DehydratedRouterState {
   dehydratedMatches: DehydratedRouteMatch[]
 }
 
+export interface LoaderContext extends LoaderFnContext{
+  parentMatchPromise: Promise<void> | undefined
```
schiller-manuel, line 167: I don't think this should be exposed as public API.
@tannerlinsley  which parts of `LoaderFnContext`should be exposed for codesplitting purpose?
see also #940

PR #956, opened 2024-01-08
Add useSearchState method

Nick-Lucas:
This adds a useSearchState method, which behaves similarly to useState but with search params. This is useful for situations when using a Link or navigate() directly is unergonomic or impossible.

Quick example of the proposed API:

```ts
  const [sort = 'asc', setSort] = postsRoute.useSearchState({
    key: 'sort',
  })
```

Here's a comparison from my original userland draft of the before and after for users, having access to this hook makes for much cleaner code in the situations it's needed:

image

### DIFF
--- a/examples/react/basic-default-search-params/src/main.tsx
+++ b/examples/react/basic-default-search-params/src/main.tsx
@@ -96,31 +96,53 @@ const postsRoute = new Route({
   path: 'posts',
   loader: () => fetchPosts(),
   component: PostsComponent,
+  validateSearch: z.object({
+    sort: z.enum(['asc', 'desc']).optional(),
+  }),
 })
 
 function PostsComponent() {
   const posts = postsRoute.useLoaderData()
 
+  const [sort = 'asc', setSort] = postsRoute.useSearchState({
+    key: 'sort',
+  })
+
   return (
     <div className="p-2 flex gap-2">
       <div className="list-disc bg-gray-800/70 rounded-lg divide-y divide-green-500/30">
-        {posts.map((post, index) => {
-          return (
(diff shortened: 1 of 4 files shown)


examples/react/basic-default-search-params/src/main.tsx
```diff
@@ -96,31 +96,54 @@ const postsRoute = new Route({
...
+        <select
+        className='bg-gray-800/70 rounded-lg w-full'
+          value={sort}
+          onChange={(e) => setSort(e.target.value as any)}
+        >
+          <option value="asc">Asc</option>
+          <option value="desc">Desc</option>
+        </select>
+
+        {/* TODO: probably better to drive the loader with search but need to research how */}
```
Nick-Lucas, line 123: This is probably worth doing as it would be a little more correct to how the example is written. Need to check the docs!

tannerlinsley:
I *am* interested in some abstraction APIs around search, but we're currently exploring this pattern along with many others at Nozzle right now. I don't feel confident enough in both our APIs or this one (it's pretty good as well!) to commit them to the core. I would recommend maintaining this in userland and report back in a few months how well it holds up! We will need to address this eventually though.

Nick-Lucas:
No problem!

@tannerlinsley there is some prior art here which you might be interested in: https://github.com/pbeshai/use-query-params/tree/master/packages/use-query-params

I've been using that library for a few years, and it's pretty proven for me, plus being so similar to useState there's some evidence it's a good pattern.

The great thing is the new router hoists out some of the cruftier bits like encoder/decoder objects for types, so it can be way simpler here!

PR #971, opened 2024-01-11
feat: type safe useSetSearch

gronxb:
## Summary
Inspired by React Router's [useSearchParams](https://reactrouter.com/en/main/hooks/use-search-params), 

The `useNavigate` method is useful for navigating between pages. However, when performing filter-related functions on the current page, it's often necessary to set the state within the page itself. In such cases, a combined setter for search parameters and state would be more beneficial.

https://github.com/TanStack/router/assets/41789633/52751bd9-4d30-4a43-b9d5-62b095f1d433

image

## Simple Usage
```tsx
const countRoute = new Route({

  validateSearch: (search) =>
    search as {
      count: number;
      id: string;
    },
  path: "count",
  component: PostComponent,
});

export function PostComponent() {
  const { count } = countRoute.useSearch();
  const setSearch = countRoute.useSetSearch();
  // replace false
  // const setSearch = countRoute.useSetSearch({ replace: false }});

  return (
    <div>
      Count {count}
      <button
        onClick={() => {
          setSearch({
            count: count + 1,
          });
        }}
      >
        setSearch
      </button>{" "}
    </div>
  );
}
```


schiller-manuel:
why wouldn't you use `useNavigate` to update the search params without navigating away?

gronxb:
@schiller-manuel 
A scenario where you only want to control the serach state on the current page, without page navigation.

schiller-manuel:
I use `useNavigate` to update the state of a filtered, paginated list like this:

```tsx
const navigate = useNavigate({from: '/foo'});
navigate({search: (prev) => ({...prev, page: page + 1 });
```

PR #984, opened 2024-01-12
examples: add a basic vinxi example.

MAST1999:


nksaraf:
If we want to add server functions via use server, you should use @vinxi/server-functions which is an ootb way to add use server support in any Vinxi app

MAST1999:
> If we want to add server functions via use server, you should use @vinxi/server-functions which is an ootb way to add use server support in any Vinxi app

Got it, I'll look into it tonight.

BTW, what does ootb stand for?

nksaraf:
https://github.com/nksaraf/vinxi/blob/main/examples/solid/ssr/basic/app.config.js

Ootb -Out of the box

MAST1999:
Added `@vinxi/server-functions` as Nikhil suggested.

tannerlinsley:
So, I spent some time with @nksaraf and adapted this example into an SSR w/ streaming example. I just committed that example into the repo as `vinxi-basic-ssr-streaming`. 

If you can, please take a look at some of the code and, for the pieces that make sense, adapt it to this one. I think the most important distinction to make here is that we likely need a few examples for vinxi:

- vinxi-basic (client rendering + server functions)
- vinxi-basic-ssr (server rendering + server functions)
- vinxi-basic-ssr-streaming (server rendering + streaming + server function. This is the one I just committed)


Not sure where you want to land with this example (either basic or basic-ssr), so let me know what you're planning.

MAST1999:
> So, I spent some time with @nksaraf and adapted this example into an SSR w/ streaming example. I just committed that example into the repo as `vinxi-basic-ssr-streaming`.
> 
> If you can, please take a look at some of the code and, for the pieces that make sense, adapt it to this one. I think the most important distinction to make here is that we likely need a few examples for vinxi:
> 
>     * vinxi-basic (client rendering + server functions)
> 
>     * vinxi-basic-ssr (server rendering + server functions)
> 
>     * vinxi-basic-ssr-streaming (server rendering + streaming + server function. This is the one I just committed)
> 
> 
> Not sure where you want to land with this example (either basic or basic-ssr), so let me know what you're planning.

I can try to implement both client/server rendering + server functions examples.

I'll try to write it up over the weekend.

tannerlinsley:
It seems like what you have here is basically already the `client only + server functions` example, right?

MAST1999:
So the one I had also had SSR.

I added a client only with server functions as well.

So now we have these:
- Client only with server function -> vinxi-basic
- SSR with server function -> vinxi-basic-ssr
- SSR with streaming -> vinxi-ssr-streaming

I also updated the package names.

PR #990, opened 2024-01-14
fix: `useMatchRoute` was not respecting options `fuzzy` and `includeSearch`

schiller-manuel:


morzel85:
Related question about matching parent route: https://github.com/TanStack/router/discussions/977 (answered thanks to this fix).

PR #1000, opened 2024-01-15
fix `pendingComponent` not shown when `pendingMs` is 0

nklhtv:
`pendingConponent` is not shown when `pendingMs` is 0.
Is that a bug or intended behavior?
I am using `Number.EPSILON` in the meantime.


schiller-manuel:
Can you please provide a reproducer on e.g. codesandbox?

PR #1002, opened 2024-01-16
readdir with withFileTypes: true

Liooo:
https://nodejs.org/api/fs.html#fspromisesreaddirpath-options

Roughly speaking, should be faster in most cases.


tannerlinsley:
Awesome!

PR #1059, opened 2024-01-22
feat: Add data transformer API

iDarkLightning:
Add a data transformer API similar to [tRPC](https://trpc.io/docs/server/data-transformers). This allows the user to use a custom serializer such as [superjson](https://github.com/blitz-js/superjson) to ensure that dates and other JS objects can be serialized across the server-client boundary.

I can add the new API to the API reference documentation/add a section to the guide if the API can be finalized.


tannerlinsley:
Rebase and handle any upstream updates/confilcts and we'll get this merged. Really great work!

iDarkLightning:
Done, should be ready for merge

docs/framework/react/guide/ssr.md
```diff
@@ -258,3 +258,19 @@ transformedStream.pipe(res)
 ```
 
 With `renderToPipeableStream` and `transformStreamWithRouter`, TanStack Router is now configured to stream data to the client as it is rendered on the server!
+
+## Data Transformers
+
+When using SSR, data passed between the server and the client must be serialized before it is sent accross network-boundaries. By default, Tanstack Router will serialize data using the default `JSON.parse` and `JSON.stringify` implementations. This, however, can lead to incorrect type-definitions when using objects such as `Date`/`Map`/`Set` etc. The Data Transformer API allows the usage of a custom serializer that can allow us to transparently use these data types when communicating across the network.
```
schiller-manuel, line 264: Is `incorrect type-definitions` really the issue here? isn't it a runtime issue?

packages/react-router/src/router.ts
```diff
@@ -1663,10 +1675,10 @@ export class Router<
           typeof getData === 'function' ? await (getData as any)() : getData
         return `<script id='${id}' suppressHydrationWarning>window["__TSR_DEHYDRATED__${escapeJSON(
           strKey,
-        )}"] = ${JSON.stringify(data)}
+        )}"] = ${JSON.stringify(this.options.transformer.stringify(data))}
           ;(() => {
             var el = document.getElementById('${id}')
-            el.parentElement.removeChild(el)
+            el.parentElement.removeChild(el)
```
schiller-manuel, line 1681: can this whitespace be removed?

docs/framework/react/guide/ssr.md
iDarkLightning: Yes, the types yielded by `useLoaderData` will not match the runtime types, so I think both are fair?

Issue #1058, opened 2024-01-22
routeFilePrefix is being included in path


shreddish:
### Describe the bug

Reopening a similar [bug](https://github.com/TanStack/router/issues/750) that was closed and included a stackblitz link. When using the routeFilePrefix my routeTree.gen is including the `~` in the path names.

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-xlqig6?file=src%2Froutes%2F~posts.tsx

### Steps to Reproduce the Bug or Issue

Set up route generation using the tsr.config.json and specify a routeFilePrefix

```
{
  "routeFilePrefix": "~",
  "routesDirectory": "./src/routes",
  "generatedRouteTree": "./src/routeTree.gen.ts"
}

```

Create a file folder hierarchy similar to docs eg

```
~__root.tsx
~posts.index.tsx
~posts.tsx
```

Observe that the `~` becomes a part of the generated path names in the codegen 

### Expected behavior

As a user I would expect the routeFilePrefix  in the tsr.config (`~`) to be ignored in the route path from the codegen 

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

PR #1061, opened 2024-01-22
Strip out the routeFilePrefix in the routePath when gathering routes

shreddish:
Fixes [ #1058](https://github.com/TanStack/router/issues/1058)

### DIFF
--- a/packages/router-generator/src/generator.ts
+++ b/packages/router-generator/src/generator.ts
@@ -63,6 +63,11 @@ async function getRouteNodes(config: Config) {
           const filePathNoExt = removeExt(filePath)
           let routePath =
             cleanPath(`/${filePathNoExt.split('.').join('/')}`) || ''
+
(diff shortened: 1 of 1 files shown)


packages/router-generator/src/generator.ts
```diff
@@ -280,8 +283,7 @@ export async function generator(config: Config) {
   const routeConfigChildrenText = await buildRouteConfig(routeTree)
 
   const sortedRouteNodes = multiSortBy(routeNodes, [
-    (d) =>
-      d.routePath?.includes(`/${routePathIdPrefix + rootPathId}`) ? -1 : 1,
+    (d) => (d.routePath?.includes(`/${rootPathId}`) ? -1 : 1),
```
shreddish, line 286: Same as other check - since we've stripped out the prefix in the getRouteNodes function we don't have to include the routeFilePrefix when filtering the rootPath

PR #1099, opened 2024-01-29
fix: access data-status from linkProps which got from useLinkProps no…

iamchanii:
Currently (at least 1.14.6), `<Link />` component get `isActive` from component's `props` not `linkProps`. it should access `linkProps` which returned from `useLinkProps` for pass `isActive` to props children function as correctly.


mgcrea:
Encountering this as well, `<Link>{({ isActive }) => ()}</Link>` is currently broken.

PR #1113, opened 2024-01-30
fix: correct path interpolation for splat/catch-all routes

### LINKED ISSUE #1070: Navigating to splat/catch-all routes always return the data of the first match
### Describe the bug

The first time a splat/catch-all route is matched, the data is cached and the page is rendered correctly, but subsequent matches after that always return the first match.

https://stackblitz.com/edit/tanstack-router-rlxprr?file=src%2Froutes%2F%24.tsx

1. Click on "Splat Route" to navigate to the splat route.
2. Click on "/page" to navigate to the subroute.
3. Nothing happen unless the page is refreshed.

### Expected behavior

Clicking on "/page" should render the new match

### Screenshots or Videos

_No response_

### Platform

- OS: [Windows]
- Browser: [Chrome]
- Version: [120]

### Additional context

From what I can see in the Devtools, Splat/catch-all routes uses the route's path as the key for the cache while dynamic routes uses the captured section of the URL pathname.

For exemple, a dynamic route `services/$serviceId` with the param service-a will use the pathname `/services/service-a` as the key but the splat route `/services/$` will always use `/services/$`.

Maybe it could mean that the data returned from the cache is always the instance of the first match for a splat route.

GuillaumeLaroucheUQAT:
fixes #1070

### DIFF
--- a/packages/react-router/src/path.ts
+++ b/packages/react-router/src/path.ts
@@ -225,6 +225,7 @@ export function matchByPath(
             )
             // TODO: Deprecate *
             params['*'] = _splat
+            params['$'] = _splat
             params['_splat'] = _splat
             return true
           }


GuillaumeLaroucheUQAT:
[https://github.com/TanStack/router/commit/5521f9044e2dc98c8003965ba6d51e242bf524c3](https://github.com/TanStack/router/commit/5521f9044e2dc98c8003965ba6d51e242bf524c3)

Issue #966, opened 2024-01-10
SSR: throw redirect in beforeLoader does nothing


universse:
### Describe the bug

Throw redirect in beforeLoader does not work for SSR example.

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-u6w1y3?file=src%2Froutes%2F_auth.profile.tsx,src%2Froutes%2F_auth.tsx

### Steps to Reproduce the Bug or Issue

1. Add _auth.tsx with `beforeLoader` and _auth.profile.tsx.
2. Click on Profile link, redirect is thrown and index route is loaded.
3. Enter `/profile` route via address bar. User is not redirected to '/' route.

* Need to open StackBlitz browser URL in a new tab since the URL in iframe does not get updated.

### Expected behavior

User is redirected to '/' route.

### Platform

NA

tannerlinsley:
We're tracking this internally with some key features around SSR. Thanks!

PR #1125, opened 2024-01-31
examples: add ssrx-basic-ssr-streaming example

marbemac:
An example with [SSRx](https://github.com/marbemac/ssrx), a library I maintain.

Everything seems to be working well including deferred data and lazy loading. The only thing that I was not able to get working in the example was server side redirects - might be an issue in the router but I haven't had a chance to dig deeper yet.


iDarkLightning:
Yeah, the redirect issue was detailed in #966

schiller-manuel:
too much changed in the meantime, closing this

PR #1149, opened 2024-02-05
fix: routeTree.gen.ts compatible with ESLint and Prettier

recallwei:
- The Generated file should be compatible with ESLint and Prettier(Sometimes should do `eslint fix` or `prettier write`).
- These files are generated by Lib that we won't modify any more.
- This solution is inspired by `unplugin-auto-import`.

image


tannerlinsley:
And this still allows the types referenced outside of the file to work?

recallwei:
> And this still allows the types referenced outside of the file to work?

Yes, only disable the check(ESLint, Prettier, Type) for this generated file.

PR #1160, opened 2024-02-07
Option to add files extensions

vafanassieff:
Hey !

Thanks for the great work with all the TanStack packages.

I'm trying to use `@tanstack/router-vite-plugin` in a project where I'm using `"moduleResolution": "NodeNext",` and i'm unable to use the `bundler` module resolution in my monorepo, I have packages shared between my backend and my frontend code.

Best.

### DIFF
--- a/packages/router-generator/src/generator.ts
+++ b/packages/router-generator/src/generator.ts
@@ -365,6 +365,12 @@ export async function generator(config: Config) {
 
   const virtualRouteNodes = sortedRouteNodes.filter((d) => d.isVirtual)
 
+  const rootPathIdExtension = config.addExtensions && !config.disableTypes
+    ? '.tsx'
+    : config.addExtensions && config.disableTypes
+      ? '.jsx'
+      : ''
+
   const routeImports = [
     '/* prettier-ignore-start */',
     '/* eslint-disable */',
@@ -379,10 +385,11 @@ export async function generator(config: Config) {
       `import { Route as rootRoute } from './${replaceBackslash(
         path.relative(
           path.dirname(config.generatedRouteTree),
-          path.resolve(config.routesDirectory, routePathIdPrefix + rootPathId),
-        ) + config.addExtensions
-          ? '.tsx'
-          : '',
+          path.resolve(
+            config.routesDirectory,
+            `${routePathIdPrefix}${rootPathId}${rootPathIdExtension}`,
+          ),
+        ),
       )}'`,
       ...sortedRouteNodes
         .filter((d) => !d.isVirtual)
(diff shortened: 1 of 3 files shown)


packages/router-generator/src/generator.ts
```diff
@@ -380,20 +380,27 @@ export async function generator(config: Config) {
...
           return `import { Route as ${
             node.variableName
           }Import } from './${replaceBackslash(
-            removeExt(
-              path.relative(
-                path.dirname(config.generatedRouteTree),
-                path.resolve(config.routesDirectory, node.filePath),
-              ),
-            ),
+            config.addExtensions
```
schiller-manuel, line 393: to avoid code duplication, I would modify a function like this:

```tsx
export function removeExt(d: string, enabled:boolean = true) {
  return enabled ? d.substring(0, d.lastIndexOf('.')) || d : d
}
```
and then call removeExt with `config.addExtensions`

you could also create a new wrapping function around `removeExt` to separate those concerns

packages/router-generator/src/generator.ts
```diff
@@ -380,20 +380,27 @@ export async function generator(config: Config) {
         path.relative(
           path.dirname(config.generatedRouteTree),
           path.resolve(config.routesDirectory, routePathIdPrefix + rootPathId),
-        ),
+        ) + config.addExtensions
+          ? '.tsx'
```
schiller-manuel, line 384: what about js only mode?

packages/router-generator/src/generator.ts
vafanassieff: Thanks for pointing this, i pushed another version.

packages/router-generator/src/generator.ts
```diff
@@ -365,6 +365,12 @@ export async function generator(config: Config) {
 
   const virtualRouteNodes = sortedRouteNodes.filter((d) => d.isVirtual)
 
+  const rootPathIdExtension = config.addExtensions && !config.disableTypes
+    ? '.tsx'
+    : config.addExtensions && config.disableTypes
+      ? '.jsx'
```
schiller-manuel, line 371: why do we have to statically assign the file extension? why don't you just use the actual file extension of the file being imported? this would make this much more reliable

schiller-manuel: You need to find out the actual file extension at generation time.
AFAIK the file could be named:
- __root.tsx
- __root.jsx
- __root.ts
- __root.js

schiller-manuel:
sorry for the lengthy review, but there is one last thing I ask you do add:
Please document this feature.
See https://tanstack.com/router/v1/docs/framework/react/guide/file-based-routing#configuration

vafanassieff:
> sorry for the lengthy review, but there is one last thing I ask you do add: Please document this configuration option. See https://tanstack.com/router/v1/docs/framework/react/guide/file-based-routing#configuration

Thanks for the good review and hint :)

PR #1165, opened 2024-02-09
docs: fixed typo in quick start code.

atifmehmood53:
Fixed typo `!`.


SeanCassiere:
It's not a typo, since its return would infer the element possibly not being present on the page and the `!` forces *typescript* into believing that the element is defined.

SeanCassiere (commented):
Not a typo.

PR #1202, opened 2024-02-20
perf: improve typescript performance for Link and useNavigate absolut…

### LINKED ISSUE #1091: `useNavigate` causes slow TS performance
### Describe the bug

I have a fairly small project (it will grow), right now there are about ~20 pages, but I am already facing some performance issues after including the router package, mainly in the IntelliSense and auto-complete. After some investigations, I have narrowed down part of the issue to be that for every component I use `navigate` from `useNavigate`, I get a bad compilation time for that file. 

I have identified this using `--generateTrace` flag for `tsc`, it points to every single one of these files to the particular line of when the `navigate(...)` method is being used. 

 

```
npx @typescript/analyze-trace trace
Hot Spots
├─ Check file /users/me/dev/project/applications/web/dashboard/src/pages/policiesaddedit/policiesaddedit.tsx (1050ms)
├─ Check file /users/me/dev/project/packages/web/ui/src/components/sidebar/sidebar.tsx (1001ms)
│  └─ Check deferred node from (line 70, char 9) to (line 82, char 16) (979ms)
│     └─ Check expression in /users/me/dev/project/applications/web/dashboard/src/router/index.tsx from (line 84, char 24) to (line 84, char 30) (651ms)
│        └─ Check expression from (line 71, char 16) to (line 77, char 3) (651ms)
│           └─ Check expression from (line 71, char 27) to (line 77, char 2) (649ms)
│              └─ Check expression from (line 72, char 5) to (line 72, char 14) (642ms)
│                 └─ Check expression from (line 32, char 19) to (line 64, char 3) (626ms)
│                    └─ Check expression from (line 32, char 41) to (line 64, char 2) (626ms)
│                       └─ Check expression from (line 34, char 5) to (line 63, char 7) (444ms)
│                          └─ Check expression from (line 34, char 36) to (line 63, char 6) (435ms)
│                             └─ Check expression from (line 35, char 9) to (line 62, char 11) (435ms)
│                                └─ Check expression from (line 35, char 37) to (line 62, char 10) (430ms)
├─ Check file /users/me/dev/project/applications/web/dashboard/src/pages/signup/index.tsx (814ms)
│  └─ Check deferred node from (line 54, char 9) to (line 223, char 16) (735ms)
│     └─ Check expression from (line 55, char 13) to (line 58, char 44) (732ms)
│        └─ Check expression from (line 57, char 22) to (line 57, char 46) (731ms)
│           └─ Check expression from (line 57, char 23) to (line 57, char 45) (731ms)
│              └─ Compare types 137362 and 137370 (731ms)
│                 └─ Check expression from (line 39, char 20) to (line 48, char 19) (731ms)
│                    └─ Check expression from (line 39, char 20) to (line 45, char 23) (730ms)
│                       └─ Check expression from (line 39, char 20) to (line 44, char 96) (730ms)
│                          └─ Check expression from (line 44, char 23) to (line 44, char 95) (730ms)
│                             └─ Check expression from (line 44, char 29) to (line 44, char 95) (730ms)
├─ Check file /users/me/dev/project/applications/web/dashboard/src/pages/login/index.tsx (795ms)
│  └─ Check deferred node from (line 57, char 9) to (line 124, char 16) (725ms)
│     └─ Check expression from (line 58, char 13) to (line 61, char 44) (719ms)
│        └─ Check expression from (line 60, char 22) to (line 60, char 46) (718ms)
│           └─ Check expression from (line 60, char 23) to (line 60, char 45) (718ms)
│              └─ Compare types 134742 and 134800 (717ms)
│                 └─ Check expression from (line 34, char 20) to (line 51, char 14) (717ms)
│                    └─ Check expression from (line 38, char 17) to (line 50, char 18) (716ms)
│                       └─ Check expression from (line 45, char 32) to (line 45, char 113) (716ms)
├─ Check file /users/me/dev/project/applications/web/dashboard/src/pages/apikeysedit/index.tsx (703ms)
│  └─ Check deferred node from (line 35, char 9) to (line 65, char 10) (553ms)
│     └─ Check expression from (line 41, char 20) to (line 64, char 15) (553ms)
│        └─ Check expression from (line 58, char 21) to (line 64, char 14) (537ms)
│           └─ Check expression from (line 59, char 24) to (line 63, char 19) (537ms)
├─ Check file /users/me/dev/project/applications/web/dashboard/src/pages/accountedit/index.tsx (615ms)
│  └─ Check deferred node from (line 33, char 32) to (line 43, char 6) (516ms)
│     └─ Check expression from (line 36, char 13) to (line 36, char 62) (515ms)
│        └─ Check expression from (line 36, char 19) to (line 36, char 62) (515ms)
└─ Check file /users/me/dev/project/applications/web/dashboard/src/pages/resolversinstancesaddedit/resolversinstancesaddedit.tsx (513ms)
   └─ Check deferred node from (line 55, char 9) to (line 72, char 10) (377ms)
      └─ Check expression from (line 58, char 17) to (line 61, char 40) (373ms)
         └─ Check expression from (line 58, char 17) to (line 61, char 25) (373ms)
            └─ Check expression from (line 58, char 17) to (line 61, char 19) (373ms)
```

I am using the `navigate` like (example from SignUp):

```ts
export default function SignUpPage() {
    const navigate = useNavigate({ from: '/sign-up' });

    const onSubmit = useCallback(
        async (data: SignUpSchemaType) => {
            // auth code
                .then(() => navigate({ to: '/verify-account', search: { email: data.email } }))
        },
        [context.auth],
    );
```

From the example, on the line with `navigate` I get the hotspot from the report. 

Is there anything I can do to help out the compiler for faster resolution? Maybe if there are some recommendations, it would be helpful to add on the docs site.

/

1. Import via `useNavigate`
2. Use method `navigate`
3. Slow TS compile

### Expected behavior

Fast TS compile

### Screenshots or Videos

_No response_

### Platform

- OS: MacOS 2019, Intel i9, 16GB RAM
- Version: 1.14.0

### Additional context

I have may other instnaces where navigation is made via `Link` and those instances doesn't seem to cause slow compilation. Maybe because they are JSX?

chorobin:
I've been looking into improving typescript performance for Link and useNavigate. 

After tracing I noticed it was related to relative pathing for Link. SplitPaths was being defaulted to the result of splitting all paths. This is very slow when you have a large number of routes. I've inlined SplitPaths because when TFrom isn't specified, you should not need to use it as we're dealing with absolute paths. This improved performance significantly.

useNavigate was even slower, it looks like Typescript was doing some eager checking of the route tree. I've solved this the following - useNavigate was defaulting TFrom and TDefaultFrom to all paths but it was always a union of string | RoutePaths. Defaulting this to string and not constraining it to all paths limits the possibility distributing over such a huge union and auto complete can be kept intact by using TDefaultFrom | RoutePaths<TRouteTree> | (string & {})

I'm wondering if TFrom should be defaulted to '/' as mentioned in the docs

With testing I was able to get like Link down from around 12s to around 2s with around 600 routes.

Thinking and exploring further performance improvements. It will harder to approach improving performance of relative paths but it would be good to improve this too. But apart from that there is a slightly faster way of creating the ParseRoute union

fixes #1091

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -74,46 +74,53 @@ export type RelativeToPathAutoComplete<
   AllPaths extends string,
   TFrom extends string,
   TTo extends string,
-  SplitPaths extends string[] = Split<AllPaths, false>,
 > = TTo extends `..${infer _}`
-  ? SplitPaths extends [
-      ...Split<ResolveRelativePath<TFrom, TTo>, false>,
-      ...infer TToRest,
-    ]
-    ? `${CleanPath<
-        Join<
-          [
-            ...Split<TTo, false>,
-            ...(
-              | TToRest
-              | (Split<
-                  ResolveRelativePath<TFrom, TTo>,
-                  false
-                >['length'] extends 1
-                  ? never
-                  : ['../'])
-            ),
-          ]
-        >
-      >}`
+  ? Split<AllPaths, false> extends infer SplitPaths
+    ? SplitPaths extends [
+        ...Split<ResolveRelativePath<TFrom, TTo>, false>,
+        ...infer TToRest,
+      ]
+      ? `${CleanPath<
+          Join<
+            [
+              ...Split<TTo, false>,
+              ...(
+                | TToRest
+                | (Split<
+                    ResolveRelativePath<TFrom, TTo>,
+                    false
+                  >['length'] extends 1
+                    ? never
+                    : ['../'])
+              ),
+            ]
+          >
+        >}`
+      : never
     : never
   : TTo extends `./${infer RestTTo}`
-    ? SplitPaths extends [
-        ...Split<TFrom, false>,
-        ...Split<RestTTo, false>,
-        ...infer RestPath,
-      ]
-      ? `${TTo}${Join<RestPath>}`
+    ? Split<AllPaths, false> extends infer SplitPaths
+      ? SplitPaths extends [
+          ...Split<TFrom, false>,
+          ...Split<RestTTo, false>,
+          ...infer RestPath,
+        ]
+        ? `${TTo}${Join<RestPath>}`
+        : never
       : never
     :
-        | (TFrom extends `/`
-            ? never
-            : SplitPaths extends [...Split<TFrom, false>, ...infer RestPath]
-              ? Join<RestPath> extends { length: 0 }
-                ? never
-                : './'
-              : never)
-        | (TFrom extends `/` ? never : '../')
+        | (string extends TFrom
+            ? '/'
+            : TFrom extends `/`
+              ? never
+              : Split<AllPaths, false> extends infer SplitPaths
+                ? SplitPaths extends [...Split<TFrom, false>, ...infer RestPath]
+                  ? Join<RestPath> extends { length: 0 }
+                    ? never
+                    : './'
+                  : never
+                : never)
+        | (string extends TFrom ? '../' : TFrom extends `/` ? never : '../')
         | AllPaths
 
 export type NavigateOptions<
--- a/packages/react-router/src/useNavigate.tsx
+++ b/packages/react-router/src/useNavigate.tsx
@@ -5,12 +5,26 @@ import { LinkOptions, NavigateOptions } from './link'
 import { AnyRoute } from './route'
 import { RoutePaths } from './routeInfo'
 import { RegisteredRouter } from './router'
-import { StringLiteral } from './utils'
+
+export type UseNavigateResult<TDefaultFrom extends string> = <
+  TRouteTree extends AnyRoute = RegisteredRouter['routeTree'],
+  TFrom extends RoutePaths<TRouteTree> | string = TDefaultFrom,
+  TTo extends string = '',
+  TMaskFrom extends RoutePaths<TRouteTree> | string = TFrom,
+  TMaskTo extends string = '',
+>({
+  from,
+  ...rest
+}: NavigateOptions<TRouteTree, TFrom, TTo, TMaskFrom, TMaskTo>) => Promise<void>
 
 export function useNavigate<
-  TRouteTree extends AnyRoute = RegisteredRouter['routeTree'],
-  TDefaultFrom extends RoutePaths<TRouteTree> | string = RoutePaths<TRouteTree>,
->(_defaultOpts?: { from?: StringLiteral<TDefaultFrom> }) {
+  TDefaultFrom extends string = string,
+>(_defaultOpts?: {
+  from?:
+    | TDefaultFrom
+    | RoutePaths<RegisteredRouter['routeTree']>
+    | (string & {})
+}) {
   const { navigate } = useRouter()
 
   const matchPathname = useMatch({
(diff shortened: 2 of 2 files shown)


packages/react-router/src/link.tsx
```diff
@@ -74,46 +74,53 @@ export type RelativeToPathAutoComplete<
...
-                >['length'] extends 1
-                  ? never
-                  : ['../'])
-            ),
-          ]
-        >
-      >}`
-    : never
-  : TTo extends `./${infer RestTTo}`
+  ? Split<AllPaths, false> extends infer SplitPaths
```
chorobin, line 78: Inline like this should still distribute in the conditional case I think

packages/react-router/src/link.tsx
```diff
@@ -74,46 +74,53 @@ export type RelativeToPathAutoComplete<
...
     :
-        | (TFrom extends `/`
-            ? never
-            : SplitPaths extends [...Split<TFrom, false>, ...infer RestPath]
-              ? Join<RestPath> extends { length: 0 }
-                ? never
-                : './'
-              : never)
-        | (TFrom extends `/` ? never : '../')
+        | (string extends TFrom
```
chorobin, line 112: string is the current default for Link so I had to check for this. But I don't know if it should be '/'

packages/react-router/src/useNavigate.tsx
```diff
@@ -5,36 +5,41 @@ import { LinkOptions, NavigateOptions } from './link'
...
+>({
+  from,
+  ...rest
+}: NavigateOptions<TRouteTree, TFrom, TTo, TMaskFrom, TMaskTo>) => Promise<void>
+
+export function useNavigate<
+  TDefaultFrom extends string = string,
+>(_defaultOpts?: {
+  from?:
+    | TDefaultFrom
```
chorobin, line 24: This still provides auto complete without defaulting TFrom to a huge union

schiller-manuel: what exactly is the relevant change here?
would `TDefaultFrom extends RoutePaths<TRouteTree> | string = string,` result in the same improvement?

just trying to understand this better

chorobin: I tried this. I think it works for performance but it breaks autocomplete. I noticed you get autocomplete with:

`TDefaultFrom extends RoutePaths<TRouteTree> | string = RoutePaths<TRouteTree>`. Which might be why it was defaulting to the union in the first place.

`TDefaultFrom | RoutePaths<RouteTree> | (string & {})` provides loose autocomplete. But hopefully this gets fixed with just `TDefaultFrom | RoutePaths<RouteTree> | string` soon 

https://www.totaltypescript.com/typescript-5-3#automatic-loose-autocomplete-on-strings

This method keeps the TDefaultFrom looser but still provides autocomplete which I think is what you want here because the constraint was `RoutePaths<TRouteTree> | string` which is already very loose due to the `| string`

Anyway. The point is to remove the default to a large union but to keep autocomplete. I think defaulting this to a large union could easily sneak in performance issues as you just need to do one `extends` and you're dealing with it being distributed somewhere else and it could be hard to track this down.

chorobin: I will also explain a separate issue which is connected to this but it was why `useNavigate` seemed so slow. `useNavigate` captured `TRouteTree` as a type parameter and this was used in `navigate` which is a function you return. It looks like Typescript was checking `TRouteTree` again when `navigate` was called. If I let `navigate` capture `TRouteTree` instead, the extra checking didn't happen. With a large tree this checking was taking seconds for me. I'm not sure why but this might be to do with generic constraints on `TRouteTree` and variance. But `TRouteTree` will never change so this checking was only slowing things down. Especially if the `TRouteTree` is large

schiller-manuel:
thanks for this!

Issue #1231, opened 2024-02-28
docs: tracking missing reference documentation


SeanCassiere:
### Reference page missing documentation

Using, this to track the missing [reference page](https://tanstack.com/router/latest/docs/framework/react/api/router) documentation.

This is what's missing, and I'll cross (`x`) them out once I've written the documentation for them.
```[tasklist]
### Missing pages
- [x] createRootRouteWithContext
- [x] createLazyRoute
- [x] createLazyFileRoute
- [x] useLoaderDeps
- [x] useRouterState
- [x] useRouter
- [x] useLinkProps
- [x] useRouteContext
- [x] useSearch
```

PR #1232, opened 2024-02-28
docs: normalizing the hierarchy content in the reference pages

SeanCassiere:
* Touching up the hierarchical structure of the reference pages, which requires adding a bit of content.
* Adding hyperlinks between reference pages for the more common types.
* Adding more *examples* on the reference pages.

Nothing really new is being added in this PR.

```[tasklist]
### Reference sections I've completed
- [x] Functions
- [x] Components
- [x] Hooks
- [x] Types
- [x] Deprecated 
```


docs/framework/react/api/router/ParseRouteChildrenType.md
```diff
@@ -4,3 +4,9 @@ title: ParseRouteChildren type
 ---
 
 This type recursively parses a route's children and grandchildren into a single union of all possible routes.
+
+```tsx
+export type ParseRouteChildren = {
```
schiller-manuel, line 9: this type does not exist anymore.
even more proof we should not document each internal type as they might... change or vanish

SeanCassiere: Shall we keep a log of what doesn't exist anymore (and what shouldn't be exposed) and remove those in a following PR?
I'm currently keeping track of what's missing in #1231 to be filled in after this PR, let's track the stuff to be hidden from the user in an issue as well and fasttrack it before the new pages. Since it's just a matter of *deletes*, it ideally should take much time.

schiller-manuel: yeah sure let's remove it in a separate PR. I just noticed it during review.

schiller-manuel (approved):
LGTM

PR #1248, opened 2024-03-01
fix: ssr improvements

tannerlinsley:
- Reworks not-found errors and handling to better support SSR, including the ability to configure a default not-found component from the router
- Reworks redirect errors and handling to work for SSR
- Reworks error rendering to work for SSR
- Allows loaders to return never (for throwing redirects)
- Improves error handling and status code management for SSR
- Fixes a few edge cases where hovering or preloading a link would navigate to it

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,5 +4,4 @@
 **/build
 **/coverage
 **/dist
-**/docs
 pnpm-lock.yaml
--- a/docs/framework/react/guide/not-found-errors.md
+++ b/docs/framework/react/guide/not-found-errors.md
@@ -203,13 +203,15 @@ export const Route = createFileRoute('/_layout/a')({
 
 ### Manually targeting the root route
 
-You can also target the root route by passing `route: '__root__'` to the `notFound` function.
+You can also target the root route by passing the exported `rootRouteId` variable to the `notFound` function's `route` property:
 
 ```tsx
+import { rootRouteId } from '@tanstack/react-router'
+
 export const Route = createFileRoute('/posts/$postId')({
   loader: async ({ params: { postId } }) => {
     const post = await getPost(postId)
-    if (!post) throw notFound({ route: '__root__' })
+    if (!post) throw notFound({ route: rootRouteId })
     return { post }
   },
 })
--- a/docs/framework/react/guide/not-found-errors.md
+++ b/docs/framework/react/guide/not-found-errors.md
@@ -21,7 +21,7 @@ There are 2 uses for not-found errors in TanStack Router:
     - Attempting to access `/posts/1` when the post with ID 1 does not exist
     - Attempting to access `/docs/path/to/document` when the document does not exist
 
-Under the hood, both of these cases is implemented using the same `notFound` function and `notFoundComponent` API.
+Under the hood, both of these cases are implemented using the same `notFound` function and `notFoundComponent` API.
 
 ## The `notFoundMode` option
 
@@ -53,7 +53,7 @@ If provided the path of `/posts/1/edit`, the following component structure will 
 
 - `<Root>`
   - `<Posts>`
-    - `<postRoute.notFoundComponent>`
+    - `<Posts.notFoundComponent>`
 
 The `notFoundComponent` of the `posts` route will be rendered because it is the **nearest suitable parent route with children (and therefore an outlet) and a `notFoundComponent` configured**.
 
--- a/docs/framework/react/guide/not-found-errors.md
+++ b/docs/framework/react/guide/not-found-errors.md
@@ -4,11 +4,145 @@ title: Not Found Errors
 
 > ⚠️ This page covers the newer `notFound` function and `notFoundComponent` API for handling not found errors. The `NotFoundRoute` route is deprecated and will be removed in a future release. See [Migrating from `NotFoundRoute`](#migrating-from-notfoundroute) for more information.
 
-Not-found errors are a special class of errors that may be thrown in loader methods and components to signal that a resource cannot be found. TanStack Router has a special API for handling these errors, similar to Next.js' own not-found API.
+## Overview
 
-Beyond being able to display a not-found, TanStack Router also lets you specify where a not-found error gets handled. This allows you to handle not-found errors in a way that preserves layouts.
+There are 2 uses for not-found errors in TanStack Router:
 
-## `notFound` and `notFoundComponent`
+- **Non-matching route paths**: When a path does not match any known route matching pattern **OR** when it partially matches a route, but with extra path segments
+  - The **router** will automatically throw a not-found error when a path does not match any known route matching pattern
+  - If the router's `notFoundMode` is set to `fuzzy`, the nearest parent route with a `notFoundComponent` will handle the error. If the router's `notFoundMode` is set to `root`, the root route will handle the error.
+  - Examples:
+    - Attempting to access `/users` when there is no `/users` route
+    - Attempting to access `/posts/1/edit` when the route tree only handles `/posts/$postId`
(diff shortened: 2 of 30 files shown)


docs/framework/react/guide/not-found-errors.md
```diff
@@ -78,9 +201,19 @@ export const Route = createFileRoute('/_layout/a')({
 })
 ```
 
-### "Global" Not Found Errors
+### Manually targeting the root route
+
+You can also target the root route by passing `route: '__root__'` to the `notFound` function.
```
schiller-manuel, line 206: maybe it would be better to keep the `__root__` id as an implementation detail, and add an abstraction?

packages/react-router/src/Matches.tsx
```diff
@@ -165,21 +166,25 @@ export function Match({ matchId }: { matchId: string }) {
           getResetKey={() => router.state.resolvedLocation.state?.key!}
           errorComponent={routeErrorComponent}
           onCatch={(error) => {
+            console.log('Tanner')
```
schiller-manuel, line 169: remove

packages/react-router/src/Matches.tsx
```diff
@@ -165,21 +166,25 @@ export function Match({ matchId }: { matchId: string }) {
...
             // Forward not found errors (we don't want to show the error component for these)
             if (isNotFound(error)) throw error
             warning(false, `Error in route match: ${matchId}`)
-            console.error(error)
+            console.error('Caught!!!', error)
```
schiller-manuel, line 173: remove

docs/framework/react/guide/not-found-errors.md
tannerlinsley: Technically it's exported as `rootRouteId`

packages/react-router/src/Matches.tsx
```diff
@@ -212,24 +217,58 @@ function MatchInner({
...
+      return (
+        <RouteErrorComponent
+          error={match.error}
+          info={{
+            componentStack: '',
+          }}
+        />
+      )
+
+      // const SuspendedError = () => {
```
schiller-manuel, line 262: remove commented out code?

docs/framework/react/guide/not-found-errors.md
schiller-manuel: then at least use this instead of the raw string in the docs?

packages/react-router/src/RouterProvider.tsx
```diff
@@ -137,21 +137,21 @@ function Transitioner() {
...
       }
     })
 
-    const nextLocation = router.buildLocation({
-      search: true,
-      params: true,
-      hash: true,
-      state: true,
-    })
+    // const nextLocation = router.buildLocation({
```
schiller-manuel, line 145: remove commented out code? or enable it again?

packages/react-router/src/route.ts
```diff
@@ -220,14 +227,14 @@ export type UpdatableRouteOptions<
   onLeave?: (match: AnyRouteMatch) => void
   meta?: (ctx: {
     params: TAllParams
-    loaderData: TLoaderData
+    loaderData: TLoaderData extends never ? undefined : TLoaderData
```
schiller-manuel, line 230: when would `TLoaderData` be `never`?
Shouldn't this be already caught by the `TLoaderDataReturn` pattern?

packages/react-router/src/route.ts
```diff
@@ -220,14 +227,14 @@ export type UpdatableRouteOptions<
...
   }) =>
     | JSX.IntrinsicElements['meta'][]
     | Promise<JSX.IntrinsicElements['meta'][]>
   links?: () => JSX.IntrinsicElements['link'][]
   scripts?: () => JSX.IntrinsicElements['script'][]
   headers?: (ctx: {
-    loaderData: TLoaderData
+    loaderData: TLoaderData extends never ? undefined : TLoaderData
```
schiller-manuel, line 237: same as above

packages/react-router/src/router.ts
```diff
@@ -737,7 +764,14 @@ export class Router<
       // Waste not, want not. If we already have a match for this route,
       // reuse it. This is important for layout routes, which might stick
       // around between navigation actions that only change leaf routes.
-      const existingMatch = getRouteMatch(this.state, matchId)
+      let existingMatch = getRouteMatch(this.state, matchId)
+      // if (
```
schiller-manuel, line 768: remove commented out code?

packages/react-router/src/router.ts
```diff
@@ -70,7 +69,12 @@ import { NotFoundError, isNotFound } from './not-found'
 import { ResolveRelativePath, ToOptions } from './link'
 import { NoInfer } from '@tanstack/react-store'
 import warning from 'tiny-warning'
-import { DeferredPromiseState } from '.'
+import {
+  DeferredPromiseState,
+  NavigateOptions,
+  NotFoundRouteComponent,
+  notFound,
+} from '.'
```
schiller-manuel, line 77: I would rather import from the files that export those types, e.g. `./defer` etc

docs/framework/react/guide/not-found-errors.md
```diff
@@ -4,11 +4,145 @@ title: Not Found Errors
...
+      </div>
+    )
+  },
+  notFoundComponent: () => {
+    return <p>This setting page doesn't exist!</p>
+  },
+})
+```
+
+Or configuring a `notFoundComponent` for a `/posts/$postId` route to handle posts that don't exist:
```
schiller-manuel, line 99: doesn't this contract the above?

> The `notFoundComponent` of the `posts` route will be rendered because it is the **nearest suitable parent route with children (and therefore an outlet) and a `notFoundComponent` configured**.

packages/react-router/src/not-found.tsx
```diff
@@ -4,7 +4,7 @@ import { useRouterState } from './useRouterState'
 import { RegisteredRouter, RouteIds } from '.'
 
 export type NotFoundError = {
-  global?: boolean
+  _global?: boolean
```
SeanCassiere, line 7: Is there a reason for this api surface to change?
This'll be a breaking change for anyone currently calling `notFound({ global: true })` yes?

tannerlinsley: Hmmm yeah. But I thought I left in a backwards compatibility layer…

.prettierignore
```diff
@@ -4,5 +4,4 @@
 **/build
 **/coverage
 **/dist
-**/docs
```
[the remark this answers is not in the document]
tannerlinsley: Ah. I needed to remove that so the docs would get formatted. Is it not auto fixing?

packages/react-router/src/not-found.tsx
SeanCassiere: Not sure if that'd help this though, as this is a user input type.
Since the type `NotFoundError` is the same user input used when throwing a `notFound()`, changing it makes it so now you'd have to `throw notFound({ _global:true })` instead of just `throw notFound({ global: true })`.

tannerlinsley: I’ll fix this or at the very least support both officially with a deprecation notice.

.prettierignore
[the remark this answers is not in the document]
schiller-manuel: no, you need to execute

```shell
pnpm prettier:write 
```

docs/framework/react/guide/not-found-errors.md
tannerlinsley: Yes it does. I’ll fix that.

PR #1256, opened 2024-03-02
feat: disableLogging

itsyoboieltr:
Proposal to have a property on the config object to optionally disable logging. This would be relevant for my use-case, because I am experimenting with creating my own-stack on top of TanStack Router, and the logging of the route generation is messing with my logging.

- Adds `disableLogging` property to the config
- Adds `logging` function to utils
- Adds docs to document it


schiller-manuel:
I'd rather create a wrapper logging function that is called with the message to be logged and the config flag, and then replace every raw `console.log` it decides whether it should log or not.

itsyoboieltr:
I hope this is the implementation you had in mind @schiller-manuel. Looking forward to your feedback :)

schiller-manuel:
looking good!

PR #1257, opened 2024-03-03
Update file-based-routing.md

ianzone:


schiller-manuel:
why this change? path can either be represented by filename or directory structure

ianzone:
it's a bit confusing that the sentence emphasizes using `.route.tsx` as a suffix but gives a directory structure example.

schiller-manuel:
then the word "suffix" is the problem.
want to change this to maybe just "filename"?

ianzone:
added two examples

Issue #1240, opened 2024-03-01
routesDirectory treated differently in the vite-plugin and the router-generator


Kinqdos:
### Describe the bug

I noticed that vite-plugin and the router generator use a different root when handling the `routesDirectory` from the router config. 

Consider my vite root directory in `/src/renderer`. When I specify `./src/renderer/src/routes` as my `routesDirectory` the vite-plugin is joining the path together with `vite.root`, when handling a hot update. This results in `/src/renderer/src/renderer/src/routes`.

On the other hand the router-generator is joining the path with the projects root directory. This results in `/src/renderer/src/routes`, which is correct. The problem is now, that either the vite plugin is watching the wrong directory or the generator can't find the right files.

### Your Example Website or App

https://stackblitz.com/edit/github-k8oxxb

### Steps to Reproduce the Bug or Issue

1. Create a vite project with the vite config file at the project root.
2. Move the index.html and the src directory to ./ui
3. Specify `root` as `./ui` in the vite config
4. Setup TanstackRouter with the vite plugin and `routesDirectory: './ui'`

### Expected behavior

I expect that the routes gets processed properly and the vite plugin is detecting changes. The outcome is, that the routes gets generated at vite startup, but not when I do changes to the files.

### Platform

Not platform specific

### Additional context

See `packages/router-vite-plugin/src/index.ts` at [line 23](https://github.com/TanStack/router/blob/aa3da3192507fe87e30f8eed50f9b6ccf6474694/packages/router-vite-plugin/src/index.ts#L23)
and `packages/router-generator/src/generator.ts` at [line 136](https://github.com/TanStack/router/blob/aa3da3192507fe87e30f8eed50f9b6ccf6474694/packages/router-generator/src/generator.ts#L136)

linusrush:
Ran into the same problem while using electron-vite. 
The vite config is in project root, but ui root is in "src/renderer".
 
The routesDirectory from router config is handled differently in vite-plugin and router-generator.

Kinqdos:
Any updates?

schiller-manuel:
can you please provide an complete example project where this issue is happening?

Kinqdos:
https://stackblitz.com/edit/github-k8oxxb

You will notice, that the routes get generated on vite startup, but no changes to the files are recognized.
If you change the `routesDirectory` to `.src/routes` changes will be recognized, but the generator fails, cause it looks at the wrong path for the files.

dmitrymaslov99:
Any updates?

[...]

PR #1259, opened 2024-03-04
Fixed root path to line up with router-generator

carl-sauter:
Fixes issue #1240


schiller-manuel:
I am not sure `process.cwd()` is a good idea. What happens if vite is started from another directory?
I think (but I probably not understand the problem fully) there must be another solution that does not rely on `cwd`.

Kinqdos:
Yeah I see your point. Its the same behavior as the router cli. It searches for routes (and the config file) from the directory you started the cli / vite. This is already the behavior of the route generator. The problem is that the plugin wich handles the hmr updates always takes the vite root directory as "starting" point. (vite root is the directory with the index.html) 
Another solution would be passing the vite root to the route generator instead of passing a relative directory.

schiller-manuel:
"edit the router generator " in which way?

Kinqdos:
Because the generator function exported by the router-generator only takes the userConfig as argument. So you have to add the root directory as argument

schiller-manuel:
i see no problem with this additional change, can you update the PR?

Kinqdos:
Yes I can do! My concern is, that other packages wich also depend on the `generator`function, will break with the changes to the arguments.
P.S: The `getConfig` function from the router-generator also relies on cwd as fallback:
```typescript
export async function getConfig(
  inlineConfig: Partial<Config> = {},
  configDirectory?: string,
): Promise<Config> {
  if (configDirectory === undefined) {
    configDirectory = process.cwd()
  }
...
```

Kinqdos:
Any thoughts? Or should I start update the pr?

schiller-manuel:
can't you add the root directory as optional argument, if it is not set, fall back to `cwd`?

Kinqdos:
Just thinking why to change the behavior here? The router-generator relies on cwd why can't the vite plugin? With this changes we are introducing a different behavior between the vite plugin and the cli. The vite plugin is looking for config in the vite route directory, but if you switch over to using the cli nothing will work. The cli is looking for the config in the cwd and not in the vite root. So you have to move your config from the vite root to the cwd and update the paths in the config.

Is this a better solution than keeping the cwd solution in the vite-plugin?

schiller-manuel:
ok, let's merge this, and wait for complaints :)

PR #1272, opened 2024-03-07
fix: incorrect route generation for nested layout routes

### LINKED ISSUE #1150: Layout in folder doesn't work
### Describe the bug

Currently this works
/_layout/login.tsx -> /login
auth._layout.login.tsx -> /auth/login

Its just that this doesn't work
/auth/_layout/login.tsx -> /auth/layout/login

As you can see on the reproduction link, the _layout in the /auth/ folder is totally ignored by the router, ignoring my layout.

https://stackblitz.com/~/edit/github-odjbcq-wjavnw?file=src%2Froutes%2F__root.tsx

Go on /auth/login, the layout isn't applied.

### Expected behavior

Apply the layout on /auth/_layout pages

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: [Edge]
- Version 114.0.1823.82 (Version officielle) (arm64)

### Additional context

_No response_

SeanCassiere:
I fully acknowledge that this looks jank... there's no escaping this fact.
Most of it stems from having to short-circuit the `handleNode` function.

Fixes #1150

### DIFF
--- a/packages/router-generator/src/generator.ts
+++ b/packages/router-generator/src/generator.ts
@@ -16,8 +16,11 @@ export type RouteNode = {
   routePath?: string
   cleanedPath?: string
   path?: string
-  isNonPath?: boolean
-  isNonLayout?: boolean
+  isNonPath?: boolean // the route definition uses an `id` field instead of the `path` field
+  isNonLayout?: boolean // the route definition shouldn't be nested under an existing route grouping
+  isParentRoot?: boolean
+  isLayout?: boolean
+  isVirtualParentRequired?: boolean
   isRoute?: boolean
   isLoader?: boolean
   isComponent?: boolean
@@ -89,6 +92,9 @@ async function getRouteNodes(config: Config) {
           let isPendingComponent = routePath?.endsWith('/pendingComponent')
           let isLoader = routePath?.endsWith('/loader')
 
+          const segments = (routePath ?? '').split('/')
+          let isLayout = segments[segments.length - 1]?.startsWith('_') || false
+
           ;(
             [
               [isComponent, 'component'],
@@ -126,6 +132,7 @@ async function getRouteNodes(config: Config) {
             isPendingComponent,
             isLoader,
             isLazy,
+            isLayout,
           })
         }
       }),
@@ -148,6 +155,7 @@ type RouteSubNode = {
   pendingComponent?: RouteNode
   loader?: RouteNode
   lazy?: RouteNode
+  virtualParent?: RouteNode
 }
 
 export async function generator(config: Config) {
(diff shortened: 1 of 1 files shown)


packages/router-generator/src/generator.ts
```diff
@@ -312,9 +332,40 @@ export async function generator(config: Config) {
       return
     }
 
+    node.isVirtualParentRequired =
+      removeGroups(node.path ?? '')
+        .split('')
+        .filter((char) => char === '/').length >= 2 && // check that the parent route wouldn't be the root route
```
schiller-manuel, line 338: can the "check that the parent route wouldn't be the root route " be simplified to this?

```tsx
const node.parent?.parent
```

packages/router-generator/src/generator.ts
```diff
@@ -637,6 +690,51 @@ function removeGroups(s: string) {
...
+ * @param {string} variableName - The variable name from which to remove the last segment. Defaults to an empty string.
+ * @returns {string} The variable name with the last segment removed.
+ * @example
+ * removeLastSegmentFromVariableName('RouteBBarTanRoute') // 'RouteBBarTan'
+ * removeLastSegmentFromVariableName('RouteAFoo') // 'RouteA'
+ */
+export function removeLastSegmentFromVariableName(
+  variableName: string = '',
+): string {
+  const segments = variableName.split(/(?=[A-Z])/)
```
schiller-manuel, line 719: can't we build the variable name again without the last segment instead relying on string manipulations?

packages/router-generator/src/generator.ts
```diff
@@ -312,9 +332,40 @@ export async function generator(config: Config) {
...
+        isVirtualParentRoute: true,
+        isVirtualParentRequired: false,
+      }
+
+      parentNode.children = parentNode.children ?? []
+      parentNode.children.push(node)
+
+      node.parent = parentNode
+
+      await handleNode(parentNode)
```
schiller-manuel, line 361: can it be a problem that this parentNode is created "on-the-fly"?
e.g. what happens if two routes need the same virtual parent?

packages/router-generator/src/generator.ts
```diff
@@ -216,61 +229,68 @@ export async function generator(config: Config) {
...
     node.isNonLayout = first.endsWith('_')
-    node.cleanedPath = removeGroups(removeUnderscores(node.path) ?? '')
 
-    // Ensure the boilerplate for the route exists
-    const routeCode = fs.readFileSync(node.fullPath, 'utf-8')
-
-    const escapedRoutePath = removeTrailingUnderscores(
-      node.routePath?.replaceAll('$', '$$') ?? '',
+    node.cleanedPath = removeGroups(
+      removeUnderscores(removeLayoutSegments(node.path)) ?? '',
```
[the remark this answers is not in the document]
SeanCassiere: Remove layout segments entirely removes blocks that start with `_`, but does not capture those that end with underscores like `/foo_/bar.tsx`.

packages/router-generator/src/generator.ts
```diff
@@ -216,61 +229,68 @@ export async function generator(config: Config) {
...
     )
+    // node.cleanedPath = removeGroups(removeUnderscores(node.path) ?? '')
```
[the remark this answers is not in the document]
SeanCassiere: Will do!

packages/router-generator/src/generator.ts
SeanCassiere: The problem is that this condition never comes to pass.

Since a nested layout route (with no `.route.tsx` configuration file) wouldn't ever be detected as having a parent route because the `routeNodes` array wouldn't already have a virtual parent route to link it to.

In the next chunk of code below is where a virtual parent route is created, the current route is added as its child, and the newly created parent node is mapped to the current node's `.parent` prop.

SeanCassiere: There's an alternative that works.

```tsx
const cleanedPathIsEmpty = (node.cleanedPath || '').length === 0
node.isVirtualParentRequired = !cleanedPathIsEmpty && !node.parent && node.isLayout
```

Issue #1268, opened 2024-03-06
Link component doesn't update URL with search params.


01Boris:
### Describe the bug

Link component in memorized sub-components doesn't update URL search params, without parent rerender.

- In dashboard like layout, we have multiple components that are memorized using `React.memo` to prevent rerenders. In this memorized components are used router `Link` component. 
- When user clicks on the link, search params are changed to show modals, etc... using old search params like:
    - `<Link search={old => ({...old, modalState: "visible" })} replace>Show modal</Link>`
- When the search params are changed, the child component isn't rerendered, and the old search in `Link` component neither.
- So we end up with mish-mash URLs in different links component:
    - real page `/about?param1=menu-prev`
    - Link1 `/about?param1=menu-prev`, sets `{param1: 'menu-prev'}`
    - LInk2 `/about?param1=menu-next&param2=menu-2`,  sets `{param2: 'menu-2'}`
- When using `useNavigate` hook, this works as intended.

### Your Example Website or App

https://stackblitz.com/edit/github-okg5qa?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

1. Go to about page
2. On about page click on link `About + Prev`
3. On about page click on link `About + Param2`

### Expected behavior

Both `Link` components are updated to reflect the real search params so:
    - Link1 `/about?param1=menu-prev`
    - LInk2 `/about?param1=menu-prev&param2=menu-2`

### Platform

- OS: macOS
- Browser:  Chrome, Safari, Firefox
- Version: @tanstack/react-router@^1.19.1

thomas-ndlss:
Hello, 

It seems that I'm having the same issue with the latest version:
When I add a `validateSearch` in a `createFileRoute`  
```typescript
/* eslint-disable unicorn/prefer-top-level-await */
import { z } from 'zod';

import { createFileRoute } from '@tanstack/react-router';

const searchValidators = z
	.object({ toto: z.number().optional().default(10).catch(20) })
	.optional()
	.default({ toto: 30 })
	.catch({ toto: 40 });

export const Route = createFileRoute('/_dashboard/dashboard/data/')({
	validateSearch: (search: Record<string, unknown>): z.infer<typeof searchValidators> => {
		return searchValidators.parse(search);
	},
	loader: () => {
		console.log('Inside loader: /_dashboard/dashboard/data/');
	},
});
```
In 1.16.0, it updates the url with the correct search params.
[...]

PR #1293, opened 2024-03-12
(fix) routers - update url search params according to validate search

thomas-ndlss:
Should fix: #1268 

With this fix, it should update the url search params according to the validateSearch zod schema so the visible url should match the router search state (except if masked is activated).

Update on` packages/react-router/src/RouterProvider.tsx` is a "revert" from changes made between `1.18.3` & `1.18.4`.
I don't know why this was deleted, maybe it's part of something I miss. If it's the case, ignore this PR.

If you need more precisions, let me know.


schiller-manuel:
did you test it locally?

PR #1303, opened 2024-03-12
fix: don't hydrate assets for notFound or redirected matches

iDarkLightning:
We currently add `meta` `links` and `scripts` on the client even if the route is redirected or not found. This causes hydration errors, because the server does not add those properties. 

(Aside from that it also just doesn't make sense to load assets for a route that was not found and can potentially reveal unwanted info)


iDarkLightning:
The format is failing on a file I didn't touch, don't know if I should run the format to fix it?

tannerlinsley:
Excellent!

PR #1306, opened 2024-03-13
fix: use transformer to parse data

iDarkLightning:
We're currently stringifying the data using the transformer, but not parsing it back

### DIFF
--- a/packages/react-router/src/awaited.tsx
+++ b/packages/react-router/src/awaited.tsx
@@ -23,7 +23,7 @@ export function useAwaited<T>({ promise }: AwaitOptions<T>): [T] {
     const streamedData = (window as any)[`__TSR__DEFERRED__${state.uid}`]
 
     if (streamedData) {
-      Object.assign(state, streamedData)
+      Object.assign(state, router.options.transformer.parse(streamedData))
     } else {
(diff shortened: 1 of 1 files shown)


schiller-manuel:
1. please rebase, formatting was fixed
2. why do we have to seeminly double stringify? `{JSON.stringify(router.options.transformer.stringify`

iDarkLightning:
> 1. please rebase, formatting was fixed
> 2. why do we have to seemingly double stringify? `{JSON.stringify(router.options.transformer.stringify`

Rebased. 

We need to double stringify because the transformer stringify returns a string that we inject into the template string. 

e.g using `superjson`:

```ts
router.options.transformer.stringify({ date: new Date(0) })
// ^ Returns '{"json":{"date":"1970-01-01T00:00:00.000Z"},"meta":{"values":{date:"Date"}}}'
```

Put inside the template string, this becomes:

```ts
window.__TSR_DEFERRED__uid = {"json":{"date":"1970-01-01T00:00:00.000Z"},"meta":{"values":{date:"Date"}}}
```

which then can't be parsed because `[object Object] is not valid JSON` yada yada. By double stringifying it, we make sure the value is actually stored as a string, which then lets us parse it again.

PR #1310, opened 2024-03-14
feat: add prevent creation of specific file format

goldfrosch:
I want to prevent files matching certain regular expressions in the routes directory from being generated at the path generation stage for convenient use of file-based libraries like vanilla-extract.

**Problem**
- I find the current prefix usage a bit inconvenient.
- Sometimes, I accidentally create templates when I forget the rules and create files.

**Suggest**
- I would like to prevent path generation based on specific regular expressions to avoid unwanted template creation


schiller-manuel:
I don't think we need a separate example for this, I would rather add a file that is ignored to one of the existing example (e.g. basic-file-based)

packages/router-generator/src/config.ts
```diff
@@ -5,6 +5,7 @@ import { z } from 'zod'
 export const configSchema = z.object({
   routeFilePrefix: z.string().optional(),
   routeFileIgnorePrefix: z.string().optional().default('-'),
+  routeFileIgnoreType: z.string().optional(),
```
schiller-manuel, line 8: naming should be improved, how about `routeFileIgnorePattern`?

packages/router-generator/src/generator.ts
```diff
@@ -55,6 +55,10 @@ async function getRouteNodes(config: Config) {
         return d.name.startsWith(routeFilePrefix)
       }
 
+      if (routeFileIgnoreType) {
+        return !d.name.match(new RegExp(routeFileIgnoreType, 'g'))
```
schiller-manuel, line 59: create the regex only once

docs/framework/react/guide/file-based-routing.md
```diff
@@ -91,6 +91,8 @@ The following options are available for configuration via the `tsr.config.json`
   - (Optional) If set, only route files and directories that start with this string will be considered for routing.
 - **`routeFileIgnorePrefix`**
   - (Optional, **Defaults to `-`**) Route files and directories that start with this string will be ignored. By default this is set to `-` to allow for the use of directories to house related files that do not contain any route files.
+- **`routeFileIgnoreType`**
+  - (Optional) ignore specific file formats in the route directory. It can be used in regular expression format. For example `.((css|const).ts)` can ignore `.css.ts` and `.const.ts` file
```
schiller-manuel, line 95: this allows to ignore not only based on file extension, instead the whole filename can be used in the pattern

schiller-manuel (changes requested):

docs/framework/react/guide/file-based-routing.md
goldfrosch: right.

what I originally wanted was to simply exclude files that matched a specific file pattern, such as '.css.ts' or '.const.ts'. however, this code excludes all files and directories that match a certain regular expression, regardless of their type

although the initial setup may be more cumbersome and time-consuming to understand than using prefixes, it seems to have the advantage of being able to set more specific conditions for ignoring certain files and directories.

packages/router-generator/src/generator.ts
goldfrosch: https://github.com/TanStack/router/pull/1310/commits/c4e21de9620695d1f3329f773de3be499de36b5c
thx

goldfrosch:
> I don't think we need a separate example for this, I would rather add a file that is ignored to one of the existing example (e.g. basic-file-based)

I move example to basic-file-based

docs/framework/react/guide/file-based-routing.md
goldfrosch: https://github.com/TanStack/router/pull/1310/commits/81872bab5dd3f646c081efa9d34c393e3c0eec84

I changed description

PR #1316, opened 2024-03-15
fix: Type error when passing props to `Link` via destructed `LinkProps`.

ryanstaniforth:
Fixes a bug where [destructing `LinkProps` into the props for a `Link` causes a type error](https://stackblitz.com/edit/tanstack-router-qmueor?file=src%2Fmain.tsx).

Similar to #1307, this omits the `preload` prop when merging `HTMLProps` and `LinkProps`.

I've also added a test to `examples/react/basic/src/main.tsx` with the intention of moving it to a more appropriate location once a testing framework is setup.

### DIFF
--- a/examples/react/basic/src/main.tsx
+++ b/examples/react/basic/src/main.tsx
@@ -176,5 +176,5 @@ if (!rootElement.innerHTML) {
 
 // Test that `Link` can be supplied its properties from a destructed `LinkProps` object without causing a type error.
 // TODO: Move into a defined test once a testing framework is set up.
-const testLinkOptions: LinkProps = { to: '/' }
-const testLink = <Link {...testLinkOptions}>test</Link>
+const testLinkProps: LinkProps = { to: '/' }
+const testLink = <Link {...testLinkProps}>test</Link>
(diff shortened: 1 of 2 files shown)


SeanCassiere:
Could your test be changed to use `LinkProps`? or possibly add another option with that type?

Since the `LinkProps` type is what is meant to be the base for what is accepted by the `<Link>` component.

examples/react/basic/src/main.tsx
```diff
@@ -172,3 +173,8 @@ if (!rootElement.innerHTML) {
 
   root.render(<RouterProvider router={router} />)
 }
+
+// Test that `Link` can be supplied its properties from a destructed `LinkOptions` object without causing a type error.
+// TODO: Move into a defined test once a testing framework is setup.
+const testLinkOptions: LinkOptions = { to: '/' }
```
SeanCassiere, line 179: Maybe change this to use LinkProps? or possibly add another option with that type?

Since the LinkProps type is what is meant to be the base for what is accepted by the <Link> component.

SeanCassiere (changes requested):

examples/react/basic/src/main.tsx
ryanstaniforth: Makes sense, update to `LinkProps` :)

PR #1318, opened 2024-03-15
docs: remove formatted {' '} on Quick Start 

j9141997:
remove formatted {' '} on https://tanstack.com/router/latest/docs/framework/react/quick-start#srcroutes__roottsx

If we do not need it, close it!


schiller-manuel:
this space is necessary

PR #1337, opened 2024-03-19
tests(router-generator): add tests to check the generated output of the `router-generator`

SeanCassiere:
Added some tests to the `router-generator` package so we can have some confidence in making changes to it without borking it for everyone.

The current results against which the test cases are being run may not even be correct, but with *this* we can predictably make those changes going forward.


packages/router-generator/tests/generator.test.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+async function getExpectedRouteTreeFileText(folder: string) {
+  const dir = getFolderDir(folder)
+  const location = dir + '/routeTree.expected.ts'
+  const text = await fs.readFile(location, 'utf-8')
+  return text
+}
+
+// Write a test
```
schiller-manuel, line 33: remove comment? :D

packages/router-generator/tests/generator.test.ts
```diff
@@ -0,0 +1,90 @@
...
+describe('with a default config', () => {
+  it('should wire-up the routes for a "single-level" tree', async () => {
+    const folderName = 'single-level'
+    const config = await setupConfig(folderName)
+
+    await generator(config)
+
+    const [expectedRouteTree, generatedRouteTree] = await Promise.all([
```
schiller-manuel, line 41: can we reduce the verbosity one more step?

```tsx
describe('with a default config', () => {
it.each([
  ['single-level],
  ['flat'],
  // etc
])(''should wire-up the routes for a "%s" tree', async (folderName) => {
    const config = await setupConfig(folderName)
    await generator(config)
    // ...
});
}
```

OR
we could iterate over all folders instead of enumerating them in the test?

SeanCassiere: I'd go for the first option since this is currently just testing the `generator` function and this could change should we decide to test the `config` related stuff as well.

Also, didn't know that `it.each` was a thing...
PS: Literally my first time ever writing tests... Pretty much yolo'd till now.

SeanCassiere: I'll dump all the dirs inside a folder and get the foldernames from there.

PR #1345, opened 2024-03-20
fix(react): make sure useMatch errors with a better error message

TkDodo:
and build `useParams` on top of `useMatch`


packages/react-router/src/useParams.tsx
```diff
@@ -19,10 +18,10 @@ export function useParams<
...
 ): TSelected {
-  return useRouterState({
-    select: (state: any) => {
-      const params = (last(getRenderedMatches(state)) as any)?.params
-      return opts?.select ? opts.select(params) : params
+  return useMatch({
+    ...opts,
+    select: (match) => {
+      return opts?.select ? opts.select(match.params) : match.params
     },
```
TkDodo, line 25: I am not too confident in this refactoring. Is there a way to try out preview builds? In query, we have them set up with codesandbox.

It's curious that `opts` wasn't used at all before (apart from `select`), so `StrictOrFrom` didn't really do anything at runtime?

I mainly made this refactoring so that `useParams()` will also fail with the good error messages we get from `useMatch`.

schiller-manuel: we don't have preview builds yet. however, you can run `pnpm run dev`

```shell
pnpm run dev
```

in the root router folder, and the same command in one of the examples to test your changes locally

schiller-manuel: > It's curious that `opts` wasn't used at all before (apart from `select`), so `StrictOrFrom` didn't really do anything at runtime?

yeah looks strange, probably not meant that way. good catch.

SeanCassiere (approved):

packages/react-router/src/useParams.tsx
TkDodo: I've confirmed that it works locally. One thing I'm not fully happy with is the error message itself - since `useParams` and `useSearch` are built on top of `useMatch`, the error is now:

```
Invariant failed: No match found while rendering useMatch() in a component that is meant to render the '/about' route. Did you mean to 'useMatch({ strict: false })' instead?
```

However, we haven't called `useMatch()` ourselves, so the suggestion is not quite right. Do you think I should improve that and if so, how ?

schiller-manuel: how about logging a stack trace? then it should be clear what's going on

TkDodo: I think I found a better error message:

```
Invariant failed: No match found for route '/' while rendering the '/about' route. Did you mean to pass '{ strict: false }' instead?
```

this will work for `useParams`, `useSearch` and `useMatch`.

70f333f1

PR #1346, opened 2024-03-20
tests: add basic type tests for Link

chorobin:
I've added some basic type tests for `Link`.

- Only testing public interface like `Link` and not internal types. We should be able to refactor the internal types and `Link`'s behavior should be the same
- I've avoided nesting in the tests. I personally find nesting hard to read and I don't mind repetition in tests.
- When I write a test case name I usually try to answer the following:  What scenario is being tested? What do I expect the result to be?

### DIFF
--- a/packages/react-router/package.json
+++ b/packages/react-router/package.json
@@ -35,7 +35,7 @@
   "scripts": {
     "clean": "rimraf ./dist && rimraf ./coverage",
     "test:types": "tsc",
-    "test:lib": "vitest",
+    "test:lib": "vitest --typecheck",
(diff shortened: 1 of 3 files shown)


packages/react-router/package.json
```diff
@@ -35,7 +35,7 @@
   "scripts": {
     "clean": "rimraf ./dist && rimraf ./coverage",
     "test:types": "tsc",
-    "test:lib": "vitest",
+    "test:lib": "vitest --typecheck",
```
chorobin, line 38: We should also run type checker when we run vitest

tannerlinsley:
It's a wonderful start!

PR #1355, opened 2024-03-22
fix: do not use CatchBoundary if neither route nor router default error component is configured

### LINKED ISSUE #724: Route loader errors stopped bubbling up to nearest error component
### Describe the bug

Up until a recent-ish beta version (not sure which specifically), errors would bubble up the route tree until it hit a route with an `errorComponent` or the root. Now it doesn't bubble and uses some boilerplate UI unless I put the custom error component where the error actually happens.

https://stackblitz.com/edit/github-gtqdb6

1. open `/about` route

### Expected behavior

The error UI div should be rendered instead of the boilerplate

### Screenshots or Videos

_No response_

### Platform

Everywhere

### Additional context

_No response_

schiller-manuel:
fixes #724

### DIFF
--- a/packages/react-router/src/Matches.tsx
+++ b/packages/react-router/src/Matches.tsx
@@ -132,9 +132,7 @@ export function Match({ matchId }: { matchId: string }) {
   const pendingElement = PendingComponent ? <PendingComponent /> : null
 
   const routeErrorComponent =
-    route.options.errorComponent ??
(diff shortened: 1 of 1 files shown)


TkDodo (approved):

ahkohd:
I can confirm this bug has been fixed in @tanstack/react-router@1.22.3

PR #1358, opened 2024-03-23
feat: add reset prop to errorComponent

TkDodo:


packages/react-router/src/router.ts
```diff
@@ -1534,9 +1535,10 @@ export class Router<
   }
 
   invalidate = () => {
-    const invalidate = (d: any) => ({
+    const invalidate = (d: RouteMatch<TRouteTree>) => ({
       ...d,
       invalid: true,
+      ...(d.status === 'error' ? ({ status: 'pending' } as const) : {}),
```
TkDodo, line 1541: some questions here @tannerlinsley:

- is there anything else we need to reset - like the `error` itself back to `undefined` ?
- do we need to do the reset conditionally? what if the route was in a state where loading data worked once and then error'd on a refetch?

TkDodo:
I can confirm that the `reset` works as expected, even when integrated with react-query. What I'm doing now is:

```
function PostErrorComponent({ error, reset }: ErrorRouteProps) {
  const router = useRouter()
  const queryErrorResetBoundary = useQueryErrorResetBoundary()

  return (
    <div>
      { error.message }
      <button
        onClick={() => {
          reset()
          queryErrorResetBoundary.reset()
          router.invalidate()
        }}
      >
        reset
      </button>
    </div>
  )
}
```

- `reset` from the `ErrorComponent` is new and will reset the internal error boundary. Without this, nothing works.
- `router.invalidate` makes sure that the loader is re-run and the current route is set back into `pending` state. You would only need this if your error is from a route loader and you want the loader to run again.
- `queryErrorResetBoundary.reset()` is necessary to let queries run again, in case the error caught by the boundary is _not_ from a loader, but from e.g. a `useSuspenseQuery` component that fetches data at component level, Since our router error boundary can will catch those errors.

If you know where your error comes from (component or loader), you can call the respective methods conditionally. But I think this gives enough control to user-land. I think docs could use a bit more love. Is there a dedicated page around error handling ?

schiller-manuel:
can you please integrate this into one of the examples, e.g. kitchen-sink based on react query?

TkDodo:
> can you please integrate this into one of the examples, e.g. kitchen-sink based on react query?

done - I integrated it into the `query-basic` examples (file and code based), and also amended the docs.

but I think there's one more thing that's missing: When we navigate away from a route, the `CatchBoundary` of the router gets reset because of how `getResetKey` is implemented:

https://github.com/TanStack/router/blob/1dd3137acf1582d145118eddb842fffb9367f005/packages/react-router/src/Matches.tsx#L163-L164

However, this then doesn't reset the `QueryErrorResetBoundary`, so TanStack Query would still have the error cached and wouldn't refetch.

What we would need is an `onReset` callback on the `CatchBoundary`, but since users don't render that manually (they just render the Error component that's used as a fallback), I'm not sure how to add it here ... 🤔

TkDodo:
okay I think I figured it out, but we need

- #1363

to make it work. Once we merged this, re-setting the Query ErrorBoundary in an effect covers all cases. I've updated the examples and docs.

schiller-manuel:
can we merge this?

TkDodo:
imo yes

PR #1359, opened 2024-03-23
fix: useNavigate subscriptions

TkDodo:
there shouldn't be a need to subscribe useNavigate to useMatch to read the pathname, because we just need it inside the callback, we can read it imperatively

also, there was likely a stale closure problem if `location` changed, as `useMatch` would re-render the component, but `useCallback` wouldn't see the fresh value

fixes #1351


packages/react-router/src/useNavigate.tsx
```diff
@@ -21,22 +21,18 @@ export function useNavigate<
...
-  }
+}): UseNavigateResult<TDefaultFrom> {
+  const { navigate, state } = useRouter()
 
-  return React.useCallback(result, [])
+  return React.useCallback(
+    (options: NavigateOptions) => {
+      return navigate({
+        ...options,
+        from: options?.to ? state.location.pathname : undefined,
```
TkDodo, line 31: I am not too confident if `state.location.pathname` is correct here, or if we should use `state.location.resolvedPathname` instead? What's the difference ?

TkDodo: also, question: could we rename `router.state` to `router.getState()` and make it a normal getter function? I was pretty confused to see that `router.state` is an imperative access. Looking at other state managers, there is often a `getState` method, like `store.getState()` in redux.

https://github.com/tanstack/router/blob/c9ef45761037f9807b7d6514b636a51f9079b9e4/packages/react-router/src/router.ts#L395-L397

To be non breaking, we would probably need to keep the `.state` accessor around

TkDodo:
I think there is a similar problem in `useMatchRoute`:

https://github.com/TanStack/router/blob/c5c325d817a75c7e964b83bf051262af01835822/packages/react-router/src/Matches.tsx#L388-L392

Imho the `useRouterState` subscription doesn't do anything: it will re-render all components that use `useMatchRoute` when location changes, but the returned callback function will always be the same. Even if `matchRoute` from `useRouter` changes, that wouldn't do anything.

@tannerlinsley you explicitly added this here:
https://github.com/TanStack/router/commit/0d4092fdec0d2949a38e1f2381d9c52910a01f21#diff-4cfd6abe2272d72c1dd073ec37936ad823bcd1f9418b466a9dea8b50214448f8R223

but I think it should be safe to remove it - all it does is create re-renders of components, which is not what this hook is for. Consumers should use `useMatch` if they want that.

packages/react-router/src/useNavigate.tsx
SeanCassiere: > ... What's the difference ?

If I'm not mistaken `location` was for the intended destination and `resolvedLocation` was what the user was currently on.
I'd have to more thoroughly go through this (not on my phone) to confirm.

SeanCassiere: > also, question: could we rename `router.state` to `router.getState()`...

There's honestly no real reason not to, but...

This would only be internal right? the public API for the existing hooks (useRouter and useRouterState) wouldn't see any changes right (like with redux, in user-land inside a component, I don't need to call a getter, just use the value returned from the hook)?

Maybe, we discuss this a bit more and add it in its own standalone release.
Depending on the final API, we can also add it onto the [v2 list](https://github.com/TanStack/router/discussions/975).

TkDodo: Okay then I would switch to `resolvedLocation` here probably?

> This would only be internal right?

the router api is public via `useRouter`, so it would be breaking. But we can add the method and deprecate the getter and remove it in v2. Hooks wouldn't change.

TkDodo: using `resolvedLocation` now: 03cfcc85

SeanCassiere: Just for reference, the implementation for the `router.load` [method](https://github.com/TanStack/router/blob/876b887589b14fb4bce0773eb520417682a741e2/packages/react-router/src/router.ts#L1552), may be worth looking at since it's the only real place that it is functionally used (other than being a dependency for rendering).

PR #1361, opened 2024-03-24
docs: explain `getRenderedMatches` function

Dragomir-Ivanov:
Add new guide using `getRenderedMatches` function

### DIFF
--- a/docs/framework/react/api/router/getRenderedMatches.md
+++ b/docs/framework/react/api/router/getRenderedMatches.md
@@ -1,0 +1,34 @@
+---
+id: getRenderedMatchesFunction
+title: getRenderedMatches function
+---
+
+The `getRenderedMatches` function returns an array of `RouteMatch` objects for all routes that are currently rendered, starting from the root route.
+
+## getRenderedMatches options
+
+The `getRenderedMatches` function accepts a single argument, the `state` which is the `RouterState` object of router.
+
+- Type: [`RouterState`](./api/router/RouterStateType)
+- Required
+
+## getRenderedMatches returns
+
+- Array of [`RouteMatch`](./api/router/RouteMatchType) objects for all rendered routes, starting from the root route.
+
+## Examples
+
+```tsx
+import { redirect } from '@tanstack/react-router'
+
+function App() {
+  const router = useRouter()
+  const matches = getRenderedMatches(router.state)
+  const lastMatch = matches[matches.length - 1]
+
+  // do something with lastMatch
+
+  // print the last route search params
+  console.log(lastMatch.search)
+}
+```
--- a/docs/framework/react/guide/imperative-state-access.md
+++ b/docs/framework/react/guide/imperative-state-access.md
@@ -1,0 +1,39 @@
+---
+title: Imperative State Access
+---
+
+Sometimes you need to access the state of the router without subscribing to any hooks like `useLoaderDeps`,`useSearch`,`useParams`, etc. This will allow you to prevent unwanted re-renders.
+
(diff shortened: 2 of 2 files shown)


SeanCassiere:
@schiller-manuel is this meant to be exposed to the end-user?

Dragomir-Ivanov:
@SeanCassiere 
It is very useful one, and if not this particular function, something like this needs to be exported.

SeanCassiere:
@Dragomir-Ivanov I don't disagree that this function is useful, but it's a balance between what's useful for a user vs what can be easily implemented by them in a few lines.

The `getRenderedMatches` helper function is something we use internally, for showing matches using the router's state. It does something simple and predictably based on what the package needs for its internals to function.

Documenting and exposing this function to the public, opens up the package's API surface, which we then need to maintain going forward.

It also restricts what we can do with then function:

1. What happens, if we need to change/augment the function based on a new feature being introduced?
If so, it suddenly all **has** to become *optional* arguments, otherwise we'd end up introducing a breaking change.

2. Or what happens, when someone files a feature request, for some changes to the function?
Should decide on adding the feature, we'd then have to introduce added complexity to the function, to satisfy user needs, for what is an internal function.

I don't believe we ever intended for the user to be consuming this function, and adding this to the package's documentation makes it official and locks us in.

@schiller-manuel up to you, we were on the mindset of reducing the public api-surface we needed to guarantee and maintain.

tannerlinsley:
I vote no.

Dragomir-Ivanov:
Okay, is there any other way to get router matches state without subscribing to anything?

If not a documented function, then some recipe somewhere on the site, with a big fat warning, that the recipe might break in the future.

SeanCassiere:
@Dragomir-Ivanov You could use `useMatches` [hook](https://tanstack.com/router/latest/docs/framework/react/api/router/useMatchesHook).
It's [implementation](https://github.com/TanStack/router/blob/876b887589b14fb4bce0773eb520417682a741e2/packages/react-router/src/Matches.tsx#L497) basically just consumes the `getRenderedMatches` helper anyways.

Dragomir-Ivanov:
@SeanCassiere 
Thank you for your fast response! However this hook is reactive, that means it triggers every time there is a route change, like `search` param change. I would like a way to get router matches state imperatively, so my component doesn't re-render every time. There is a use-case in the PR, with `uncontrolled` `<input>` element, that needs a `search param` only on first render in order to initialize its `defaultValue` prop.

SeanCassiere:
> @SeanCassiere 
> Thank you for your fast response! However this hook is reactive, that means it triggers every time there is a route change, like `search` param change. I would like a way to get router matches state imperatively, so my component doesn't re-render every time. There is a use-case in the PR, with `uncontrolled` `<input>` element, that needs a `search param` only on first render in order to initialize its `defaultValue` prop. 

You can fine-tune what is returned using the optional `select` argument to ensure the correct dependencies trigger a return of data via the hook.

By default, all hooks will be reactive their scope and will re-render as is the norm with React. TanStack Router allows, you to use the `select` key to control this to an extent on most hooks, but this is React and that requires reactivity.

Worst comes to worst, if the `getRenderedMatches` filter function is all that works for you, you can always copy-paste it's [implementation](https://github.com/TanStack/router/blob/876b887589b14fb4bce0773eb520417682a741e2/packages/react-router/src/Matches.tsx#L454) into your codebase.

Going forward, in the future we'll be controlling exactly what we export from the router package, so we don't expose internals.

Dragomir-Ivanov:
Okay, I have no issue accessing `state.matches` directly, without `getRenderedMatches`.
Thank you.

PR #1372, opened 2024-03-26
fix: trailing slash shouldn't fail the path matching

hansyulian:
discussion:
- https://github.com/TanStack/router/discussions/1364

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 import { useMatch } from './Matches'
 import { useRouterState } from './useRouterState'
 import { useRouter } from './useRouter'
-import { deepEqual, functionalUpdate } from './utils'
+import { deepEqual, exactPathTest, functionalUpdate } from './utils'
 import type { HistoryState } from '@tanstack/history'
 import type { Trim } from './fileRoute'
 import type { AnyRoute, RootSearchSchema } from './route'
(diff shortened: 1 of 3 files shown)


packages/react-router/src/link.tsx
```diff
@@ -320,35 +320,35 @@ export type LinkOptions<
...
+  ? ToRest extends ['/']
+  ? Join<['/', ...FromRest, '/']>
+  : ResolveRelativePath<Join<FromRest>, Join<ToRest>>
+  : never
+  : Split<TTo> extends ['.', ...infer ToRest]
+  ? ToRest extends ['/']
+  ? Join<[TFrom, '/']>
+  : ResolveRelativePath<TFrom, Join<ToRest>>
+  : CleanPath<Join<['/', ...Split<TFrom>, ...Split<TTo>]>>
+  : never
```
SeanCassiere, line 351: This shouldn't be getting flattened like this. Could you check if your prettier set-up is correctly picking up the prettier config at the root.

SeanCassiere (changes requested):
Just prettier fixes for me.

packages/react-router/src/link.tsx
hansyulian: hi @SeanCassiere thx for the feedback, have fixed the prettier issue

SeanCassiere (approved):
LGTM

mudasirpandith:
@SeanCassiere i am getting same issue, i am having `v1.28.1`

TkDodo:
@SeanCassiere @mudasirpandith I observed this as well with `trailingSlash:always` for sub-routes. Reproduction would be this:

https://stackblitz.com/edit/tanstack-router-kh6hmr

and I think this PR should fix it:
- https://github.com/TanStack/router/pull/1711

mudasirpandith:
@TkDodo What should be my approach to overcome this in my project?

Issue #1271, opened 2024-03-07
ToOptions doesnt't satisfy <Link> props.


leqwasd:
### Describe the bug

1. thing I need - a Component, that wraps <Link>, so I can add predefined styles to all of my links. To use "more advanced" paths with them, I found a Type - ToOptions (https://tanstack.com/router/v1/docs/framework/react/api/router/ToOptionsType), that I can pass to Link.
So I created my own Link component that in its basic form looks like this:
```tsx
import { Link, ToOptions } from '@tanstack/react-router';
import * as React from 'react';
type MyLinkProps = {
  toOptions: ToOptions;
};
const MyLink: React.FC<React.PropsWithChildren<MyLinkProps>> = ({
  children,
  toOptions,
}) => {
  return <Link {...toOptions}>{children}</Link>;
};
export default MyLink;
```
It works - I get a nice help from TypeScript autocompletes. It says - that I need "from" property in some situations.

2. I want routes that looks like this:
posts/$id
posts/$id/A
posts/$id/B

In posts/$id.tsx - I display <Outlet /> so I can see the subroutes. 

If you run attached project like this - it works, MyLink component is fine.
(See here) https://stackblitz.com/edit/github-rb4ewj?file=src%2FMyLink.tsx

When I navigate to posts/$id - <Outlet /> contains nothing... So - I was wondering - and added a new file posts/$id/index.tsx.
(I created a Fork from the previous stackblitz here: https://stackblitz.com/edit/github-rb4ewj-tg9tx2?file=src%2FMyLink.tsx)
And this is where things go wrong - Link now complains about something not satisfying something...

```
Type '{ children: ReactNode; to: "/" | "/posts/$id" | "/posts/$id/A" | "/posts/$id/B" | "/posts/$id/"; hash?: true | Updater<string> | undefined; state?: true | NonNullableUpdater<HistoryState> | undefined; from?: RoutePathsAutoComplete<...> | undefined; search?: true | ... 1 more ... | undefined; params?: true | ... 1 mo...' is not assignable to type 'IntrinsicAttributes & ({ to?: ToPathOption<Route<any, "/", "/", string, "__root__", RootSearchSchema, RootSearchSchema, RootSearchSchema, ... 12 more ..., any>, string, "/" | ... 3 more ... | "/posts/$id/"> | undefined; hash?: true | ... 1 more ... | undefined; state?: true | ... 1 more ... | undefined; from?: Route...'.
[...]

PR #1373, opened 2024-03-26
fix: rename component created by createLink

TkDodo:
because if it's named `Link`, the world explodes

refs #1271


tannerlinsley:
😂

levrik:
Oh. This explains a lot 😅 I just created my own `createLink` to work around this because I didn't get why my browser is freezing when using the official API. Implementations seemed identical but I used an arrow function 😬

Issue #1312, opened 2024-03-14
Vite Plugin Route file generation does not work while Vite is running on Windows machines


natejboucher:
### Describe the bug

We just implemented router in our application and developers with Mac machines have the code generation feature working while Vite is running, it picks up new files and file changes, like the documentation says ([docs](https://tanstack.com/router/latest/docs/framework/react/guide/file-based-routing))
However developers with Windows machines do not have that feature. The code generation only seems to check the folder when starting or restarting the application. 

We use the default configuration for the plugin with the `src/routes` folder structure. Currently running `router-vite-plugin: V1.19.0`  and  `tanstack/react-router: V1.19.1`. We tested the most recent release and the bug was still present, but decided to roll back due to known typescript issues.

This also does not work on the kitchen sink examples provided, however it is worth noting that it doesn't work there for Mac users either. 

### Your Example Website or App

https://tanstack.com/router/latest/docs/framework/react/examples/kitchen-sink-react-query-file-based

### Steps to Reproduce the Bug or Issue

1. Have a Windows machine
2. Have Tanstack Router in your application with the `router-vite-plugin` added. 
3. While the app is running, add a file to the `routes` folder 
4. The code generation feature should not have picked up the change and the file should remain empty, along with no changes to the `routeTree`
5. Stop the application and then start it
6. The code generation should pickup the new file, generate the code and update the `routeTree`

### Expected behavior

The code generation feature should see the new file or file change while Vite is running and run the code generation feature.

### Platform

- OS: Windows
- Browser: Chrome
- Version: React Router: `1.19.1` Router Vite Plugin: `1.19.0`

SeanCassiere:
I made some changes to the hook that our vite plugin was using to watch for changes. It went out with [release 1.24.1](https://github.com/TanStack/router/releases/tag/v1.24.1).

@natejboucher - could you confirm if this has made any impact to the experience on Windows?
@nilshartmann - could you also confirm this against your scenarios that you listed?

nilshartmann:
@SeanCassiere thanks a ton, it's working now 🎉  

Happy routing! 🐣

SeanCassiere:
@natejboucher when you have chance, could you confirm if this change made any impact to the Windows experience?

You can download the [Router Stackblitz Starter](https://stackblitz.com/github/tanstack/router/tree/main/examples/react/quickstart-file-based) and quickly test it with the latest version of TanStack Router with the Vite Plugin.

SeanCassiere:
[...]

PR #1378, opened 2024-03-27
feat(router-vite-plugin): trigger the route-tree generation using `watchChange` hook

SeanCassiere:
Something I've noticed, as stated in [this comment](https://github.com/TanStack/router/issues/1312#issuecomment-1999304792), is that the `router-vite-plugin` doesn't actually perform the whole route-tree generation process on file create and delete events.
* On create: You need to create the file, and then save it, to have the plugin actually do anything.
* On delete: You need to delete the file, and then go to another file and hit *save*, to have the plugin do anything.

This change uses the `watchChange` hook to trigger the route-tree generation process based on `create`, `update`, and `delete` events.

### DIFF
--- a/packages/router-vite-plugin/src/index.ts
+++ b/packages/router-vite-plugin/src/index.ts
@@ -37,9 +37,6 @@ export function TanStackRouterVite(inlineConfig: Partial<Config> = {}): Plugin {
       userConfig = await getConfig(inlineConfig, ROOT)
       await generate()
     },
-    handleHotUpdate: async ({ file }) => {
-      await handleFile(file)
-    },
     watchChange: async (file, context) => {
       if (['create', 'update', 'delete'].includes(context.event)) {
         await handleFile(file)
--- a/packages/router-vite-plugin/src/index.ts
+++ b/packages/router-vite-plugin/src/index.ts
@@ -17,6 +17,20 @@ export function TanStackRouterVite(inlineConfig: Partial<Config> = {}): Plugin {
     }
   }
 
+  const handleFile = async (file: string) => {
+    const filePath = normalize(file)
+    if (filePath === join(ROOT, CONFIG_FILE_NAME)) {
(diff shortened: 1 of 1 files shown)


packages/router-vite-plugin/src/index.ts
```diff
@@ -17,23 +17,32 @@ export function TanStackRouterVite(inlineConfig: Partial<Config> = {}): Plugin {
...
+    }
+  }
+
   return {
     name: 'vite-plugin-tanstack-router',
     configResolved: async () => {
       userConfig = await getConfig(inlineConfig, ROOT)
       await generate()
     },
     handleHotUpdate: async ({ file }) => {
```
SeanCassiere, line 40: Should we still be doing this?

The jsdoc comment for it states:
```
Perform custom handling of HMR updates. The handler receives a context containing changed filename, timestamp, a list of modules affected by the file change, and the dev server instance.

- The hook can return a filtered list of modules to narrow down the update. e.g. for a Vue SFC, we can narrow down the part to update by comparing the descriptors.

- The hook can also return an empty array and then perform custom updates by sending a custom hmr payload via server.ws.send().

- If the hook doesn't return a value, the hmr update will be performed as normal.
```

tannerlinsley: Have you tried using it with this removed? I am just as unsure as you what would happen. But, if we remove it, and it does nothing, then I think it’s the right choice. if you are relatively sure that it will have no side effects, then let’s just remove it and deal with any repercussions that come up. It’s worth the risk.

SeanCassiere: With all the local tests I've done (with it removed), I couldn't find a scenario in which it broke any existing functionality both during `dev` and `build`.
I'll push this today and keep an eye on the support channels, given that the change is so small, reverting this one shouldn't take more than a couple of minutes.

PR #1399, opened 2024-03-31
fix(react-router-server): do not spread a `key` prop inside the `<Asset>` component

### LINKED ISSUE #1138: <Scripts /> throws warning of props object containing a "key" prop being spread into JSX.
### Describe the bug

I upgraded to v1.15.12 using Vinxi w/ React SSR streaming and got errors about improperly spread props in <Scripts/>. The errors started to appear since this commit: https://github.com/TanStack/router/commit/3aac5c6971f5fb775fd6b57b49ff7dac21766e81

Important to note though that I understand these changes are experimental, and I assume that how a Vinxi + Tanstack Router setup should be done is being simplified/refactored for Tanstack Start at the moment.

I made my best guess in figuring out how to plug in the new utility functions in the commit above and pasted a screenshot of my setup below.

See screenshots below.

image
image

### Expected behavior

For the warnings to not appear.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: v1.15.12

### Additional context

_No response_

SeanCassiere:
The props/attrs being spread into the `<Asset>` component of the `@tanstack/react-router-server` package contains a `key` attribute which when spread goes against how React expects that attribute to be used.

This sets the key to the asset component itself at render.

Could look at further optimization to not rely on the array index position, rather to be calculated based on the individual identifiable values.

Closes #1138

### DIFF
--- a/packages/react-router-server/src/client/Meta.tsx
+++ b/packages/react-router-server/src/client/Meta.tsx
@@ -7,29 +7,6 @@ import * as React from 'react'
 import { Asset } from './Asset'
 import type { RouterManagedTag } from './RouterManagedTag'
 
-function getMetaKey(asset: RouterManagedTag, index: number) {
-  if (asset.tag === 'title') {
-    return `tsr-meta-title-${index}`
-  }
-
-  if (asset.tag === 'meta') {
-    const ident = [
-      asset.attrs.name,
-      asset.attrs.content,
-      asset.attrs.httpEquiv,
-      asset.attrs.charSet,
-    ].join('')
-    return `tsr-meta-meta-${ident || index}`
-  }
-
-  if (asset.tag === 'link') {
-    const ident = [asset.attrs.rel, asset.attrs.href].join('')
-    return `tsr-meta-link-${ident || index}`
-  }
-
-  return `tsr-meta-${asset.tag}-${index}`
-}
-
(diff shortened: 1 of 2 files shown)


advancedtw:
Hi, I regret to inform you that this PR doesn't fully resolve the issue. While it appeared promising initially, the hydration warning can still occur. It seems that there's still a key present in 'attrs' when it reaches Asset.tsx. My suspicion is that the Asset component might be created without Script.tsx or Meta.tsx being invoked, or more likely, a key was already present before reaching Script/Meta.

I have tried a simple test by deleting the key and the warning is indeed gone.
```ts
//Asset.tsx
import { jsx } from "react/jsx-runtime";
function Asset({ tag, attrs, children }) {
  delete attrs.key
...
``` 

in my case when I reach Asset.tsx I have these key still present 
```
key                         type
----------------------------------
vite-plugin:stylex.css      style 
plugin-0                    script 
plugin-1                    script 
vite-client                 script 
```

advancedtw:
ok I think I found the root cause at least for the script keys.  manifestScripts contains keys and is being merged directly

SeanCassiere:
@advancedtw are the keys being set by the router, or from user land?

If its from the router, could you please put up a PR?

PR #1405, opened 2024-04-01
fix(react-router): incorrectly splitting the key-values pairs from the querystring

### LINKED ISSUE #1404: SearchParams ending in "=" strips the equals sign
### Describe the bug

When using either validateSearch or useSearch() to access the search params, for search params that end in an equals sign, this is being stripped.

For example, below are some examples of what I get

URL: `http://localhost:3000?token=somerandomtoken=`
useSearch(): `{ token=somerandomtoken }`

URL: `http://localhost:3000?token=somerandomtoken=&test=somethingelse`
useSearch(): `{ token: "somerandomtoken", test: "somethingelse" }`

I've checked these same examples against URLSearchParams and this does include the equals sign.

My understanding is that the aim, at least for flat search params, is to behave identical to URLSearchParams to ensure that encoding works with other libraries, therefore this is likely a bug.

https://stackblitz.com/edit/tanstack-router-s5pt7g?file=src%2Froutes%2Findex.tsx

1. Go to the index route
2. Type in the following search param `?token=test=`
3. Check the console and note that the "=" is missing

### Expected behavior

I would expect the returned value in validateSearch and useSearch to be the following:
```ts
{
  token: "test="
}
```

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 91.1]

### Additional context

_No response_

SeanCassiere:
Updates the ﻿decode function to address parsing issues with querystrings with values containing trailing equal signs.

Previously, query strings like ﻿`token=foo=` were parsed incorrectly as ﻿`{ token: 'foo' }` instead of `﻿{ token: 'foo=' }`.

Closes #1404.

### DIFF
--- a/packages/react-router/src/qss.ts
+++ b/packages/react-router/src/qss.ts
@@ -2,6 +2,17 @@
 
 // qss has been slightly modified and inlined here for our use cases (and compression's sake). We've included it as a hard dependency for MIT license attribution.
 
+/**
+ * Encodes an object into a query string.
+ * @param obj - The object to encode into a query string.
+ * @param [pfx] - An optional prefix to add before the query string.
+ * @returns The encoded query string.
+ * @example
+ * ```
+ * // Example input: encode({ token: 'foo', key: 'value' })
+ * // Expected output: "token=foo&key=value"
+ * ```
+ */
 export function encode(obj, pfx?: string) {
   let k,
     i,
@@ -25,6 +36,14 @@ export function encode(obj, pfx?: string) {
   return (pfx || '') + str
 }
 
+/**
+ * Converts a string value to its appropriate type (string, number, boolean).
+ * @param mix - The string value to convert.
+ * @returns The converted value.
+ * @example
+ * // Example input: toValue("123")
+ * // Expected output: 123
+ */
 function toValue(mix) {
   if (!mix) return ''
   const str = decodeURIComponent(mix)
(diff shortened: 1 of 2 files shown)


packages/react-router/src/qss.ts
```diff
@@ -33,18 +52,33 @@ function toValue(mix) {
   return +str * 0 === 0 && +str + '' === str ? +str : str
 }
 
+/**
+ * Decodes a query string into an object.
+ * @param str - The query string to decode.
+ * @param [pfx] - An optional prefix to filter out from the query string.
```
SeanCassiere, line 58: What's the actual intended use-case for the `prefix` variable here?

Is it meant to:

1. Remove a prefix from each key in the string.
Like `prefix = "pre_"`, so `"pre_foo=bar&pre_john=mary"` -> `"{ foo: 'bar', john: 'mary' }"`

2. Or it is just meant to remove some prefix in the string.
Like `prefix = "pre_"`, so its `"pre_foo=bar&john=mary"` -> `"{ foo: 'bar', john: 'mary' }"`

It's not actually used anywhere in the codebase as far as I can tell, if so I can remove the commented tests for it so it only tests for use-cases we've verified it for.

packages/react-router/src/qss.ts
```diff
@@ -33,18 +52,33 @@ function toValue(mix) {
...
+      k = tmp.slice(0, equalIndex)
+      const value = tmp.slice(equalIndex + 1)
+      if (out[k] !== void 0) {
+        out[k] = [].concat(out[k], toValue(value))
+      } else {
+        out[k] = toValue(value)
+      }
     } else {
-      out[k] = toValue(tmp.shift())
+      k = tmp
```
SeanCassiere, line 80: Do we need this?

What happens is that if the key is empty ("?token"), an empty string is assigned to the key in the `out` object,

packages/react-router/src/qss.ts
SeanCassiere: Went through the qss documentation and figured it out. I've added to test to ensure we don't break anything on our end, but we can safely resolve this comment for now.

packages/react-router/src/qss.ts
schiller-manuel: it's not clear-cut
https://stackoverflow.com/a/4557763

I'd leave it in.

SeanCassiere: I agree.

Also, considering this is for a library meant for the web, it makes sense to emulate what *URLSearchParams* do in the same situation.

image

PR #1418, opened 2024-04-02
tests: fix type tests for Link, make sure type tests fail in ci and add type tests for union `to` and difference in params

chorobin:
- Tests were failing and were not type checking in ci
- Added test cases for `to` being a union, params being optional if `from` and `to` contain the same params


packages/react-router/tests/link.test-d.tsx
```diff
@@ -216,6 +230,116 @@ test('when navigating from a route with no params to a route with params, params
...
+  const params = expectTypeOf(TestLink)
+    .parameter(0)
+    .toHaveProperty('params')
+    .extract<Function>()
+
+  params.returns.branded.toEqualTypeOf<{ invoiceId?: string | undefined }>()
+  params.parameter(0).toEqualTypeOf<{ invoiceId: string }>()
+})
+
+test('when navigating to a union of routes, params is optional', () => {
```
schiller-manuel, line 253: is this correct though? shouldn't `params` be `{ invoiceId: string } | `{ postId: string}`?

packages/react-router/tests/link.test-d.tsx
```diff
@@ -216,6 +230,116 @@ test('when navigating from a route with no params to a route with params, params
...
+  const TestLink = Link<
+    RouteTree,
+    string,
+    '/invoices/$invoiceId/' | '/posts/$postId/'
+  >
+  expectTypeOf(TestLink).parameter(0).not.toMatchTypeOf<{ params: unknown }>()
```
schiller-manuel, line 259: is this the type check for "params is optional"? how? and is there a more explicit way?

packages/react-router/tests/link.test-d.tsx
schiller-manuel: yeah I was expecting this test:

https://github.com/TanStack/router/pull/1418/files#diff-921603b68301ae5e52534582aeb75f7defc2946a96ad3b268b28ad34ae5fec04R262

why would we need this test here? the test below already checks everything, right?

packages/react-router/tests/link.test-d.tsx
```diff
@@ -216,6 +230,116 @@ test('when navigating from a route with no params to a route with params, params
...
+  params.returns.branded.toEqualTypeOf<
+    { invoiceId: string } | { postId: string }
+  >()
+
+  params
+    .parameter(0)
+    .toEqualTypeOf<{} | { invoiceId: string } | { postId: string }>()
+})
+
+test('when navigating to a union of routes including the route, params is optional', () => {
```
schiller-manuel, line 297: "including the route" .. which one?

schiller-manuel: ah you meant "the root", right?

packages/react-router/tests/link.test-d.tsx
```diff
@@ -304,3 +428,83 @@ test('when navigating to a route with search params, search params is a function
...
+    .parameter(0)
+    .toHaveProperty('search')
+    .extract<Function>()
+
+  params.returns.branded.toEqualTypeOf<{ page: number } | {}>()
+
+  params.parameter(0).toEqualTypeOf<{} | { page: number }>()
+})
+
+test('when navigating to a union of routes including the route, search params is optional', () => {
```
schiller-manuel, line 472: "the route" => "the root"

packages/react-router/tests/link.test-d.tsx
chorobin: I will double check but I believe this is slightly different as we're checking if it's optional. But it's a good point if this is intentional behaviour or not. Should it be optional for a union of different routes?

I find the unions a harder use case to think about if it should be optional or not. For example the default case `string` falls under the same code path and this makes sense to be optional

chorobin: `string` is basically considered to be any route and is therefore a union of everything

packages/react-router/tests/link.test-d.tsx
chorobin: Docs cover this

https://github.com/mmkal/expect-type#error-messages

packages/react-router/tests/link.test-d.tsx
chorobin: Yep. Sorry, bad spelling

packages/react-router/tests/link.test-d.tsx
chorobin: I'm going to fix this comments tomorrow so don't merge. But you have me doubting myself so I will check

packages/react-router/tests/link.test-d.tsx
chorobin: Yep. Same mistake!

SeanCassiere (commented):
Correct me if I'm wrong, but I believe we aren't testing for external links.

The implementation for the `<Link>` component checks for external links and sets the `href` property to whatever the input of `to` was.

Not sure if that's something that'll be covered by this PR or not, just putting it out there since the types for Link are being touched here.

chorobin:
> Correct me if I'm wrong, but I believe we aren't testing for external links.
> 
> The implementation for the `<Link>` component checks for external links and sets the `href` property to whatever the input of `to` was.
> 
> Not sure if that's something that'll be covered by this PR or not, just putting it out there since the types for Link are being touched here.

Maybe we can do sanity check that we allow any string literal type in `to`. But bare in mind these are only type tests and doesn't test the runtime behaviour. We need both but since I'm touching the types most of the time, that's what I see break the most so I wanted to add some

packages/react-router/tests/link.test-d.tsx
chorobin: I'm trying to think of a different way but it's difficult as the types can be quite complex for `params`

PR #1425, opened 2024-04-05
feat: derive more Route props on instantiation

### LINKED ISSUE #1423: `Route.id` is `undefined` immediately after route creation
### Describe the bug

Immediately after creating a new route, the `Route.id` is `undefined`.

```ts
const myRoute = createRoute(opts)

myRoute.id
//      ^ undefined
```

https://stackblitz.com/edit/tanstack-router-n22e5x?file=src%2Fmain.tsx

1. Go to the url provided above.
2. Open the console.
3. Observe the logs that show what the aboutRoute id is. (see image below)

### Expected behavior

I expect that when I have created a route, I can immediately use the route's id property.

In the context of my repro, I expect that both logs in the console will say that the route id is `/about`.
### Screenshots or Videos

CleanShot 2024-04-04 at 10 07 43@2x

(Don't come at be for the bad function names! 😜 I refactored and didn't rename, ok?!)

### Platform

- OS: macOS
- Browser: Chrome & Safari
- Version: 
  - Chrome Version 123.0.6312.107 (Official Build) (arm64)
  - Safari Version 17.4.1 (19618.1.15.11.14)

### Additional context

This is a bit of a tangent, so here we go...

For some of my components, I need to fake out TS to tell the router "you are definitely at _one of_ these routes", so I do stuff like this:

```ts
const route = useMatch({ strict: false } as unknown as {
	from: typeof someRoute.id | typeof anotherRoute.id
})
```

Works fine, since that is just type information. But if I want to be sure I am safe at runtime, I can do this

```tsx
const someRoute = createRoute({/* ... route def ...*/})
const anotherRoute = createRoute({/* ... route def ...*/})

const myRouteIds = [someRoute.id, anotherRoute.id] as const
type MyRouteId = typeof myRouteIds[number]

export function useMyRouteId() {
	const routeId = useMatch({ strict: false, select: x => x.routeId })
	if (!myRouteIds.has(routeId as MyRouteId)) {
		throw new Error("You are not on one of my routes!!11!!")
	}
	return routeId as MyRouteId
}
```

(I don't do it quite like that, but this was simpler for an example.)

Except... I can't. Because `myRouteIds` captures a bunch of `undefined` values, since they are not available immediately.

Of course, you can work around this, but it sure was surprising to discover.

happycollision:
> fixes #1423
> 
> There are a few things going on in this commit. First and foremost, I've ensured that several route properties (id, path, fullPath, parentRoute, to) are actually created inside the constructor so they are available immediately after instantiation.
> 
> The logic for creating these properties was involved enough that I thought it was useful to keep it in a separate function. However, I did opt to return the properties from the function so that they can still be set on the instance inside the constructor. That allows use to not lie to TypeScript about the availability of these properties. (So I removed the `!` from the type declarations.)
> 
> Finally, I simplified a lot of the logic that was used to set all the properties themselves. By declaring the variables early, then setting all of them inside a single if/else block based on if we are in the root route or not, you can see much more clearly the difference between how the root route properties are created and how all other route properties are created. There were also several places where a value was set on a variable, used to derive a separate variable, then re-set again to finalize the original variable. That just isn't necessary when you branch once based on whether or not you are in the root route.

Now... let me know if I am totally missing some reason that these cannot be created until the router is instantiated! (But the tests seemed to pass just fine, so I hope all this works.)

### DIFF
--- a/packages/react-router/src/route.ts
+++ b/packages/react-router/src/route.ts
@@ -621,13 +621,12 @@ export class Route<
     TLoaderData
   >
 
-  // Set up in this.init()
-  parentRoute!: TParentRoute
-  id!: TId
-  // customId!: TCustomId
-  path!: TPath
-  fullPath!: TFullPath
-  to!: TrimPathRight<TFullPath>
+  parentRoute: TParentRoute
+  id: TId
+  // customId: TCustomId
+  path: TPath
+  fullPath: TFullPath
+  to: TrimPathRight<TFullPath>
 
   // Optional
   children?: TChildren
@@ -668,6 +667,15 @@ export class Route<
       `Route cannot have both an 'id' and a 'path' option.`,
     )
     ;(this as any).$$typeof = Symbol.for('react.memo')
+
+    const derived = this.deriveProperties()
+
+    this.parentRoute = derived.parentRoute
+    this.path = derived.path
+    this.id = derived.id
+    // this.customId = paths.customId
+    this.fullPath = derived.fullPath
+    this.to = derived.to
   }
 
   types!: {
@@ -695,7 +703,9 @@ export class Route<
 
   init = (opts: { originalIndex: number }) => {
     this.originalIndex = opts.originalIndex
+  }
 
+  private deriveProperties = () => {
     const options = this.options as
       | (RouteOptions<
           TParentRoute,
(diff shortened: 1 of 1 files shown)


packages/react-router/src/route.ts
```diff
@@ -668,6 +667,15 @@ export class Route<
       `Route cannot have both an 'id' and a 'path' option.`,
     )
     ;(this as any).$$typeof = Symbol.for('react.memo')
+
+    const paths = this.deriveProperties()
```
happycollision, line 671: Maybe `paths` wasn't the best name here. `derived` is probably better.

packages/react-router/src/route.ts
```diff
@@ -722,50 +732,42 @@ export class Route<
...
       invariant(
-        this.parentRoute,
+        parentRoute,
         `Child Route instances must pass a 'getParentRoute: () => ParentRoute' option that returns a Route instance.`,
       )
-    }
-
-    let path: undefined | string = isRoot ? rootRouteId : options.path
 
-    // If the path is anything other than an index path, trim it up
```
happycollision, line 738: FYI, the `trimPathLeft` function already ignores index paths.

packages/react-router/src/route.ts
```diff
@@ -722,50 +732,42 @@ export class Route<
...
-          customId,
-        ])
-
-    if (path === rootRouteId) {
-      path = '/'
+    return {
+      parentRoute: parentRoute as TParentRoute,
+      path: path as TPath,
+      id: id as TId,
+      // customId: customId as TCustomId,
```
happycollision, line 767: I assume the original `customId` stuff was commented out (instead of just removed) for a reason, so I kept it in as a comment.

packages/react-router/src/route.ts
```diff
@@ -722,50 +732,42 @@ export class Route<
     const isRoot = !options?.path && !options?.id
 
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-    this.parentRoute = this.options?.getParentRoute?.()
+    const parentRoute = options?.getParentRoute?.()
+
+    let path: undefined | string
+    let customId: string | undefined
```
happycollision, line 738: Heh. I guess I could have been more consistent with the order of the types.

SeanCassiere:
This works for a code-based routing setup, but does **NOT** work for file-based routing or even normal code-splitting using `createLazyRoute`. Currently, it crashes the app for the other use-cases.

From what I can gather, it comes down to a problem in the order of execution for file-based routing and code-splitting when using these changes.

1. App.tsx is run, which imports and evaluates the route-tree.
2. The route-tree, imports and evaluates the configuration files for each route.
3. When a route is imported/evaluated, then its route-id has already been initialized as `__root`.

Point 3, happens because the Route wouldn't have a path (since the `getParentRoute` getter wouldn't be defined at this point) or an ID. As a result, the route's ID gets set to the root route, and upon the Router's boot-up, you face an ID collision and errors out.

This process of setting the `getParentRoute` getter on the Routes happens through the update method in the `routeTree.gen.ts` (see [here](https://github.com/TanStack/router/blob/main/packages/router-generator/tests/generator/nested-layouts/routeTree.expected.ts)), which we wouldn't be able to ask for users to start setting the `getParentRoute` getter since that'd both be a breaking change and wouldn't be that much different from code-based routing.

SeanCassiere:
@happycollision have you had a chance to look at this?

And is this something that's absolutely critical for you? I'm asking since this pattern of accessing the `Route.id` outside of a React component is a bit of an anti-pattern here.

happycollision:
I haven’t yet.

I wouldn’t say this is critical, but it is a surprising discovery that I thought was a bug somewhere in my code rather than a mismatch between my expectations of what Typescript tells me about a Route instance and the reality of that Route instance.

I figured if it were easy to supply the values immediately, then there’s no need to explain the confusing behavior anywhere. I just haven’t had time to dig into the file router side of things yet. (I am surprised that no tests broke when I made my change… though one never completed on my machine, so maybe that is exactly what happened!)

if you’d like to flush things out, you can close this PR, but TanStack Router is going to be pretty key for some functionality in the app my team is creating, so I will definitely take time to look at this more in the near future.

Another option that would also have helped me when I inadvertently did something unsupported would be to throw when accessing the unassigned property. Perhaps only in development?

In any case, feel free to close this, and with your permission I can reopen later, or make a new one.

Or you can tell me to bug off! 😄

SeanCassiere:
I like the idea of possibly throwing dev warnings when trying to access these values outside of React components!

From the current trajectory of things, Router should be seeing some changes in the coming months, as we start flushing out some bugs and edge cases (especially with SSR and whatnot).

Happy to have you back here!

PR #1428, opened 2024-04-05
tests: add fileRoute and route type tests

chorobin:
- add type tests for code based routes
- add basic type tests for file based routes

### DIFF
--- a/packages/react-router/tests/fileRoute.test-d.tsx
+++ b/packages/react-router/tests/fileRoute.test-d.tsx
@@ -15,8 +15,6 @@ const postsRoute = createFileRoute('/_postLayout/posts')()
 
 const postRoute = createFileRoute('/_postLayout/posts/$postId_')()
 
-const authRoute = createFileRoute('/(auth)')()
-
 const protectedRoute = createFileRoute('/(auth)/protected')()
 
 declare module '../src/fileRoute' {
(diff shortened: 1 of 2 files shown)


packages/react-router/tests/fileRoute.test-d.tsx
```diff
@@ -0,0 +1,87 @@
...
+const postsRoute = createFileRoute('/_postLayout/posts')()
+
+const postRoute = createFileRoute('/_postLayout/posts/$postId_')()
+
+const authRoute = createFileRoute('/(auth)')()
+
+const protectedRoute = createFileRoute('/(auth)/protected')()
+
+declare module '../src/fileRoute' {
+  interface FileRoutesByPath {
```
chorobin, line 23: I think we should refractor the types for file based to allow us to pass `FileRoutesByPath` like we do for `RouteTree` in `Link`

schiller-manuel: which changes would this involve?

packages/react-router/tests/fileRoute.test-d.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const invoiceRoute = createFileRoute('/invoices/$invoiceId')()
+
+const postLayoutRoute = createFileRoute('/_postLayout')()
+
+const postsRoute = createFileRoute('/_postLayout/posts')()
+
+const postRoute = createFileRoute('/_postLayout/posts/$postId_')()
+
+const authRoute = createFileRoute('/(auth)')()
```
schiller-manuel, line 18: is this even possible? I thought a route group was merely a way to prevent a folder to show up in the url, but it cannot hold routes itself.

packages/react-router/tests/fileRoute.test-d.tsx
chorobin: We would need to add a generic like `TRouteTree` so it can easily be more mocked in the tests

packages/react-router/tests/fileRoute.test-d.tsx
chorobin: Probably not. I will make it just the child

PR #1436, opened 2024-04-06
perf: improve performance of search params merging

chorobin:
We need to avoid `TFullSearchParams` from being eagerly evaluated. This happened for a number of reasons.

- Generic constraints on `TFullSearchParams`. We need to get rid of these.
- Do not use `Omit` for `Assign` 
- Do not use `Expand`

PR #1439, opened 2024-04-07
fix(react-router): revert the `Assign` type to what was used prior in `1.26.11`

SeanCassiere:
The new `Assign` type introduced in `1.26.12` (#1436) was removing the `optional` part of keys in the route search schema.

image


SeanCassiere:
Looks like simple fixes aren't the solution for this.

Will wait till Chris is available to administer the right fix for this.

PR #1440, opened 2024-04-07
fix: optional search params not being optional

chorobin:


chorobin:
Don't merge. This regressed the previous issue.

chorobin:
> Don't merge. This regressed the previous issue.

Can merge now

rudzky:
Really nice improvement, but is it possible to make search parameter optional and set him a default value as well? If I don't provide a value I should get this default one. That would be great with things like pagination

TkDodo:
@rudzky what would the `type` of `page` be then when you read it with `useSearch()`? `number` or `number | undefined` ?

if you expect it to be `number`, but still want links and navigations to treat the parameter as optional, then there isn't a really good way to do it. The current options are to use the `SearchSchemaInput` type like so:

```
validateSearch: (input: { page?: number } & SearchSchemaInput) =>
  z.object({
    page: page: z.number().optional().default(1),
  }).parse(input)
```

because input is tagged as `SearchSchemaInput`, it will be used for input to Link and navigations, so you can omit `page`. But when reading with `useSearch()`, you'll get `number` because of how the zod schema is defined. Here is one example from the docs:

https://github.com/TanStack/router/blob/fcc278ffc8ef2bba50c4ed4b08a19c7248eb95c2/examples/react/basic-default-search-params/src/main.tsx#L145-L149

PR #1443, opened 2024-04-08
fix: widen signature of event

TkDodo:
There are situations where persist doesn't exist, see: https://github.com/TanStack/router/pull/1365#discussion_r1552986706


levrik:
@TkDodo Thanks a lot! This should increase compatibility with the ecosystem.

PR #1451, opened 2024-04-09
fix history when hash history is used with search params

### LINKED ISSUE #1449: When using hash history router reads and writes search params from fragment instead of query
### Describe the bug

There is a discrepancy between `window.location.search` and the search parameters used
by the router when using hash routes. It seams that react-router is reading and writing search parameters in the hash fragment. Search parameters written in the normal query location are ignored.

URI  = scheme ":" hier-part [ "?" query ] [ "#" fragment ]

According to he specification the query string should be before the hash fragment (rfc3986)[https://www.rfc-editor.org/rfc/rfc3986#section-3].

https://stackblitz.com/~/github.com/owodunni/router-test

1. On the / page click the link `About with router search`, see that the component got the page search parameter. But `window.location.search` is empty.
2. On the / page click the link `About with 'normal' search`, see that the `window.location.search` is set.

### Expected behavior

As a developer I expect react-router websites to adhere to standard URI formats. It should be straight forward for me to communicate how others should link to my website with query parameters.

### Screenshots or Videos

_No response_

### Platform

- OS: any
- Browser: any
- Version: 1.26.16

### Additional context

The website I am developing uses query parameters outside of react-router to enable dev mode. This has to be set differently from other query parameters since its not for react-router to use.

owodunni:
When search params where used with hash history then the search parameters where not read nore writen to the correct location in the href. The correct location for search params according to [RFC3986](https://www.rfc-editor.org/rfc/rfc3986#section-3) is:

`URI = scheme ":" hier-part [ "?" query ] [ "#" fragment ]`

Those are the parameters found in `window.location.search` but the hash
history read and wrote them here:

`URI = scheme ":" hier-part [ "#" fragment ] [ "?" query ]`

To fix that we need to account for the search parameters when
transforming the href to and from the hash route url.

The fix is split into two commits, the first commit makes the router compatible with RFC 3986 but introduces a breaking change as seen bellow. The second commit patches the breaking change but make the overall behaviour more hacky.

BREAKING CHANGE: Any website that relies on the previous none RFC 3986
compatible format and use that format when linking to a page with search
parameters and a hash route will no longer get their search parameters
parsed.

closes #1449

### DIFF
--- a/packages/history/package.json
+++ b/packages/history/package.json
@@ -36,6 +36,8 @@
     "clean": "rimraf ./dist && rimraf ./coverage",
     "test:eslint": "eslint --ext .ts,.tsx ./src",
     "test:types": "tsc --noEmit",
+    "test:unit": "vitest --typecheck",
+    "test:unit:dev": "pnpm run test:unit --watch",
     "test:build": "publint --strict",
     "build": "vite build"
   },
--- a/packages/history/src/index.ts
+++ b/packages/history/src/index.ts
@@ -338,11 +338,14 @@ export function createHashHistory(opts?: { window?: any }): RouterHistory {
   return createBrowserHistory({
     window: win,
     parseLocation: () => {
+      const search = win.location.search
       const hashHref = win.location.hash.split('#').slice(1).join('#') ?? '/'
-      return parseHref(hashHref, win.history.state)
+      return parseHref(`${hashHref}${search}`, win.history.state)
     },
-    createHref: (href) =>
-      `${win.location.pathname}${win.location.search}#${href}`,
+    createHref: (href) => {
+      const searchHref = href.split('?').slice(1).join()
+      return `${win.location.pathname}${searchHref ? `?${searchHref}` : ''}#${href}`
+    },
   })
 }
 
--- a/packages/history/src/index.ts
+++ b/packages/history/src/index.ts
@@ -346,8 +346,10 @@ export function createHashHistory(opts?: { window?: any }): RouterHistory {
       // with how createHashHistory worked before
       const [cleanHashHref, ...otherSearch] = hashHref.split('?')
 
-      if (search && otherSearch) search = [search, ...otherSearch].join('&')
-      else if (!search && otherSearch) search = `?${otherSearch.join()}`
+      if (search && otherSearch.length > 0)
+        search = [search, ...otherSearch].join('&')
+      else if (!search && otherSearch.length > 0)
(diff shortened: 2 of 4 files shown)


SeanCassiere:
I'm not sure if we are looking to merge this in, but there are some weird artefacts that are present in the URL when using this change.

When visiting *index*, the URL looks like this.
image
`http://localhost:3001/?/,#/?/?`

When using *URL Search Params*, it looks to be entered twice into the URL.
image
`http://localhost:3001/?usersView=%7B%22filterBy%22%3A%22ervin%22%7D#/dashboard/users/?usersView=%7B%22filterBy%22%3A%22ervin%22%7D`

owodunni:
Woop! Those where introduced in the second commit. That tried to make this backwards compatible. I'll add more tests and fix it.

owodunni:
The first issue regarding weird characters is addressed. However the duplicate part is tricky, I was hoping to keep the changes local to `createHashHistory` but I think we need to rewrite `parseHref` or have a separate `parseHref` function for this case. 

Regarding if this will be merged or not. I'm new to this project where is the best place for those types of discussions, here, issues or discord?

SeanCassiere:
> Regarding if this will be merged or not. I'm new to this project where is the best place for those types of discussions, here, issues or discord?

Probably would be Discord.

@tannerlinsley @schiller-manuel is this something that we need in the router?
There's certainly a question as to whether it's just us that's doing this wrong, or are all the routers doing the same thing no according to spec.

SeanCassiere:
Just tested this using Remix React-Router, and they look to have the same behaviour that we have (or more likely the other way around since we are big fans of Remix over here).

Did my testing on the about page with Remix React-Router: [https://stackblitz.com/edit/github-z72eqz?file=src%2FApp.tsx](https://stackblitz.com/edit/github-z72eqz?file=src%2FApp.tsx)

So (@tannerlinsley @schiller-manuel), what are your thoughts on possibly breaking this behaviour to be spec-compliant? or shall just stick with what we have (and what others in the ecosystem are doing)?

Personally, I'd rather have it at this stage where shifting to TSR shouldn't be too much of a break for people already using Remix's react-router.

SeanCassiere:
Thank you for submitting your PR to TSR.
 
However, as Tanner [commented](https://github.com/TanStack/router/issues/1449#issuecomment-2048333214) on the original issue, we won't be making this change at this time due to both historical reasons and the breaking change that it'd introduce.

PR #1453, opened 2024-04-09
perf: improve performance of Link and routes

chorobin:
## Purpose

Type checking large code bases with many `Route`'s and `Link`'s can get quite sluggish. This PR introduces a number of performance optimizations I've discovered while profiling on the large file based example. The purpose of this PR is to improve suggestion times for such code bases and build times

## Extended Diagnostics

On main the following diagnostics are present for the example large file based

```
> tsc --extendedDiagnostics

Files:                         589
Lines of Definitions:        77132
Lines of TypeScript:         12777
Identifiers:                113478
Symbols:                    351283
Types:                       62925
Instantiations:            1541766
Memory used:               311542K
Assignability cache size:   113357
Identity cache size:          3707
Subtype cache size:              0
Strict subtype cache size:  162006
Parse time:                  0.55s
ResolveModule time:          0.11s
Program time:                0.84s
Bind time:                   0.24s
Check time:                  5.64s
Total time:                  6.72s
```

After the changes introduced in this PR the instantiations and check time are reduced

```
> tsc --extendedDiagnostics

Files:                         589
Lines of Definitions:        77133
Lines of TypeScript:         12777
Identifiers:                113461
Symbols:                    189848
Types:                       50776
Instantiations:             592405
Memory used:               222842K
Assignability cache size:    34415
Identity cache size:          3718
Subtype cache size:              0
I/O Read time:               0.07s
Parse time:                  0.54s
ResolveModule time:          0.11s
Program time:                0.83s
Bind time:                   0.23s
Check time:                  3.82s
Total time:                  4.88s
```

## Profiling

This is the profile for the large file route on main.

image

Going deep into each wall of the profile. The biggest wall is the route tree definition. There is a number of things to improve here.

- Performance of each individual file route types
- Variance checks on each Route
- Type parameter constraints on each Route
- Improving performance of merging properties from each parent route
- Only `Expand` when necessary and not eagerly

image

And the next significant thing is each `Link`. Each `Link` can be improved by the following:

- Separating `LinkComponentProps` to cache props for a component passed to `createLink`
- Use the resolved route for relative paths to narrow down the union of paths relevant for autocomplete
- Use `CheckPath` directly on the `to` prop to avoid an intersection
- Evaluate less of `params` and `search` until the user use these props

image

After the performance improvements the route tree looks more like this

image

and each `Link` looks like this

image

`Link` may seem like a small improvement for one `Link` but when you have many in large code bases this quickly accumulates

### DIFF
--- a/packages/react-router/src/Matches.tsx
+++ b/packages/react-router/src/Matches.tsx
@@ -603,10 +603,10 @@ export function useLoaderData<
     ...opts,
     select: (s) => {
       return typeof opts.select === 'function'
-        ? opts.select(s.loaderData)
+        ? opts.select(s.loaderData as TRouteMatch)
         : s.loaderData
     },
-  })
+  }) as TSelected
 }
 
 export function isServerSideError(error: unknown): error is {
--- a/packages/react-router/src/fileRoute.ts
+++ b/packages/react-router/src/fileRoute.ts
@@ -24,7 +24,7 @@ import type {
   TrimPathLeft,
   UpdatableRouteOptions,
 } from './route'
-import type { Assign, Expand, IsAny } from './utils'
+import type { Assign, IsAny } from './utils'
 import type { RouteMatch } from './Matches'
 import type { NoInfer } from '@tanstack/react-store'
 import type { RegisteredRouter } from './router'
@@ -67,7 +67,7 @@ export type RemoveUnderScores<T extends string> = Replace<
 >
 
 type RemoveRouteGroups<T extends string> =
-  T extends `${infer Before}(${infer RouteGroup})${infer After}`
+  T extends `${infer Before}(${string})${infer After}`
     ? RemoveRouteGroups<`${Before}${After}`>
     : T
 
@@ -76,32 +76,22 @@ type NormalizeSlashes<T extends string> =
     ? NormalizeSlashes<`${Before}/${After}`>
     : T
 
-type ReplaceFirstOccurrence<
-  TValue extends string,
-  TSearch extends string,
-  TReplacement extends string,
-> = TValue extends `${infer Prefix}${TSearch}${infer Suffix}`
-  ? `${Prefix}${TReplacement}${Suffix}`
-  : TValue
-
 export type ResolveFilePath<
   TParentRoute extends AnyRoute,
   TFilePath extends string,
 > = TParentRoute['id'] extends RootRouteId
   ? TrimPathLeft<TFilePath>
-  : ReplaceFirstOccurrence<
-      TrimPathLeft<TFilePath>,
-      TrimPathLeft<TParentRoute['types']['customId']>,
-      ''
-    >
+  : TFilePath extends `${TParentRoute['types']['customId']}${infer TRest}`
+    ? TRest
+    : TFilePath
 
 export type FileRoutePath<
   TParentRoute extends AnyRoute,
   TFilePath extends string,
   TResolvedFilePath = ResolveFilePath<TParentRoute, TFilePath>,
-> = TResolvedFilePath extends `_${infer _}`
+> = TResolvedFilePath extends `_${string}`
   ? ''
-  : TResolvedFilePath extends `/_${infer _}`
+  : TResolvedFilePath extends `/_${string}`
     ? ''
     : TResolvedFilePath
 
@@ -154,34 +144,27 @@ export class FileRoute<
   createRoute = <
     TSearchSchemaInput extends RouteConstraints['TSearchSchema'] = {},
     TSearchSchema extends RouteConstraints['TSearchSchema'] = {},
-    TSearchSchemaUsed extends Record<
-      string,
-      any
-    > = TSearchSchemaInput extends SearchSchemaInput
+    TSearchSchemaUsed = TSearchSchemaInput extends SearchSchemaInput
       ? Omit<TSearchSchemaInput, keyof SearchSchemaInput>
       : TSearchSchema,
-    TFullSearchSchemaInput extends
-      RouteConstraints['TFullSearchSchema'] = ResolveFullSearchSchemaInput<
+    TFullSearchSchemaInput = ResolveFullSearchSchemaInput<
       TParentRoute,
       TSearchSchemaUsed
     >,
     TFullSearchSchema = ResolveFullSearchSchema<TParentRoute, TSearchSchema>,
-    TParams extends RouteConstraints['TParams'] = Expand<
-      Record<ParsePathParams<TPath>, string>
-    >,
-    TAllParams extends RouteConstraints['TAllParams'] = MergeFromFromParent<
+    TParams = Record<ParsePathParams<TPath>, string>,
+    TAllParams = MergeFromFromParent<
       TParentRoute['types']['allParams'],
       TParams
     >,
     TRouteContextReturn extends
       RouteConstraints['TRouteContext'] = RouteContext,
-    TRouteContext extends RouteConstraints['TRouteContext'] = [
-      TRouteContextReturn,
-    ] extends [never]
+    TRouteContext = [TRouteContextReturn] extends [never]
       ? RouteContext
       : TRouteContextReturn,
-    TAllContext = Expand<
-      Assign<IsAny<TParentRoute['types']['allContext'], {}>, TRouteContext>
+    TAllContext = Assign<
+      IsAny<TParentRoute['types']['allContext'], {}>,
+      TRouteContext
     >,
     TRouterContext extends RouteConstraints['TRouterContext'] = AnyContext,
     TLoaderDeps extends Record<string, any> = {},
@@ -190,7 +173,6 @@ export class FileRoute<
       ? undefined
       : TLoaderDataReturn,
     TChildren extends RouteConstraints['TChildren'] = unknown,
-    TRouteTree extends RouteConstraints['TRouteTree'] = AnyRoute,
   >(
     options?: FileBaseRouteOptions<
       TParentRoute,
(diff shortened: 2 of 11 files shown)


packages/react-router/src/Matches.tsx
```diff
@@ -603,10 +603,10 @@ export function useLoaderData<
     ...opts,
     select: (s) => {
       return typeof opts.select === 'function'
-        ? opts.select(s.loaderData)
+        ? opts.select(s.loaderData as TRouteMatch)
```
schiller-manuel, line 606: shouldn't this be `TRouteMatch['loaderData']`?

chorobin: Yeah. This is weird. I will have to double check why this was complaining

packages/react-router/src/fileRoute.ts
```diff
@@ -154,34 +144,27 @@ export class FileRoute<
   createRoute = <
     TSearchSchemaInput extends RouteConstraints['TSearchSchema'] = {},
     TSearchSchema extends RouteConstraints['TSearchSchema'] = {},
-    TSearchSchemaUsed extends Record<
-      string,
-      any
-    > = TSearchSchemaInput extends SearchSchemaInput
+    TSearchSchemaUsed = TSearchSchemaInput extends SearchSchemaInput
```
chorobin, line 147: For type parameters which act more like variables, its better not to use generic constraints on them. This is because typescript will do each checking on the value you assign to it.

packages/react-router/src/fileRoute.ts
```diff
@@ -190,7 +173,6 @@ export class FileRoute<
       ? undefined
       : TLoaderDataReturn,
     TChildren extends RouteConstraints['TChildren'] = unknown,
-    TRouteTree extends RouteConstraints['TRouteTree'] = AnyRoute,
```
chorobin, line 193: I don't think this was needed

packages/react-router/src/link.tsx
```diff
@@ -84,67 +84,78 @@ export type RemoveLeadingSlashes<T> = T extends `/${infer R}`
   ? RemoveLeadingSlashes<R>
   : T
 
+export type ResolvePaths<TRouteTree extends AnyRoute, TSearchPath> =
```
chorobin, line 87: For relative routes starting with '../' and '.' we can narrow the union of the paths down to the children of the resolved path. 

Basically, because we know `TFrom` is a route. We can actually only get descendants of the resolved route

packages/react-router/src/link.tsx
```diff
@@ -84,67 +84,78 @@ export type RemoveLeadingSlashes<T> = T extends `/${infer R}`
...
   : never
 
 export type SearchRelativePathAutoComplete<
+  TRouteTree extends AnyRoute,
   TTo extends string,
   TSearchPath extends string,
-  TPaths,
-  TSearchedPaths = SearchPaths<TPaths, TSearchPath>,
-> = TSearchedPaths extends string ? `${TTo}/${TSearchedPaths}` : never
+> = `${TTo}/${SearchPaths<TRouteTree, TSearchPath>}`
```
chorobin, line 104: This was a silly mistake by me before. Template literal types automatically distribute. Its a small optimization but it does help a bit

packages/react-router/src/link.tsx
```diff
@@ -189,13 +200,13 @@ export type ToSubOptions<
   // The source route path. This is automatically set when using route-level APIs, but for type-safe relative routing on the router itself, this is required
   from?: RoutePathsAutoComplete<TRouteTree, TFrom>
   // // When using relative route paths, this option forces resolution from the current path, instead of the route API's path or `from` path
-} & CheckPath<TRouteTree, {}, TFrom, TTo> &
```
chorobin, line 192: Intersections are slow because we have to merge a union with another union. Better just to change `to` prop directly

packages/react-router/src/link.tsx
```diff
@@ -224,49 +235,113 @@ type PostProcessParams<
...
+  TParamVariant
+>
+
+type MakeOptionalParams<
+  TRouteTree extends AnyRoute,
+  TParamVariant extends ParamVariant,
+  TFrom,
+  TToParams,
+> = TParamVariant extends 'SEARCH'
+  ? Expand<{
```
chorobin, line 258: Think I need to do a few more `Expand`

packages/react-router/src/link.tsx
```diff
@@ -644,30 +713,32 @@ export type LinkProps<
     | ((state: { isActive: boolean }) => React.ReactNode)
 }
 
+type LinkComponentProps<TComp> = React.PropsWithoutRef<
```
chorobin, line 716: This is nice. Basically because `TComp` doesn't change on every `Link` we can seperate these props into a seperate type and TS will cache this for all `Link`'s

packages/react-router/src/routeInfo.ts
```diff
@@ -14,7 +14,7 @@ export type RoutesById<TRouteTree extends AnyRoute> = {
 }
 
 export type RouteById<TRouteTree extends AnyRoute, TId> = Extract<
-  Extract<ParseRoute<TRouteTree>, { id: TId }>,
+  RoutesById<TRouteTree>[TId],
```
chorobin, line 17: Turns out mapped types are faster than `Extract` on a union ;). Makes sense when you think about it. But it only shows when you start to have larger code bases

packages/react-router/src/router.ts
```diff
@@ -260,9 +260,9 @@ export function createRouter<
 }
 
 export class Router<
-  TRouteTree extends AnyRoute = AnyRoute,
-  TDehydrated extends Record<string, any> = Record<string, any>,
-  TSerializedError extends Record<string, any> = Record<string, any>,
+  in out TRouteTree extends AnyRoute = AnyRoute,
```
chorobin, line 263: We can make these invariant to stop TS from doing variance checks on them. I don't think they're necessary in this case

packages/react-router/src/utils.ts
```diff
@@ -36,10 +38,12 @@ export type DeepPartial<T> = T extends object
     }
   : T
 
-// eslint-disable-next-line @typescript-eslint/naming-convention
-export type MakeDifferenceOptional<T, U> = Omit<U, keyof T> &
-  Partial<Pick<U, keyof T & keyof U>> &
-  PickRequired<Omit<U, keyof PickRequired<T>>>
```
chorobin, line 42: I'm not sure why this existed but it works without it?

packages/react-router/tests/link.test-d.tsx
```diff
@@ -242,11 +236,9 @@ test('when navigating from a route with no params to a route with params', () =>
 
   const params = expectTypeOf(TestLink).parameter(0).toHaveProperty('params')
 
-  params
-    .exclude<Function | boolean>()
-    .branded.toEqualTypeOf<{ invoiceId: string }>()
+  params.exclude<Function | boolean>().toEqualTypeOf<{ invoiceId: string }>()
 
-  params.returns.branded.toEqualTypeOf<{ invoiceId: string }>()
```
chorobin, line 249: We should always `Expand` instead of relying on `branded` in tests. It means we have complicated intersections

packages/react-router/tests/route.test-d.tsx
```diff
@@ -102,13 +101,12 @@ test('when creating a child route with a loader from the root route with context
   const invoicesRoute = createRoute({
     path: 'invoices',
     getParentRoute: () => rootRoute,
-    loader: async () => [{ id: 'invoice1' }, { id: 'invoice2' }] as const,
+    loader: async (opts) => {
+      expectTypeOf(opts).toMatchTypeOf<{ context: { userId: string } }>()
```
chorobin, line 105: I decided this approach is better because its testing the inference flow

packages/react-router/src/utils.ts
schiller-manuel: I added this here: 

https://github.com/TanStack/router/commit/92923cc4e08112f429b06f4b7120dc7241e8a2d8#diff-5fd7a0f05758a78cf8b73c09e4d71d21af63fe3d9b312b0ee75b04041ef078e7R222

do we have a test case that tests if when navigating e.g. from `/foo/$bar` to `/foo/$bar/baz/$id`, `$bar` is optional whereas `$id` is not?

packages/react-router/src/fileRoute.ts
schiller-manuel: should we remove the `RouteConstraints` type then fully?

packages/react-router/src/utils.ts
chorobin: Yes. I thought this was covered by the code already here.

chorobin: This is the closest test case but maybe I should write a specific one!
https://github.com/TanStack/router/blob/10c717513d28ae2de918ff6f1cd74ff5734d6731/packages/react-router/tests/link.test-d.tsx#L253

schiller-manuel: yeah would be cool to have that

packages/react-router/src/fileRoute.ts
chorobin: I don't want to say constraints are bad because you need them to check generics inferred by the user

chorobin: But if it's not inferred by the user, I assume it can be trusted

packages/react-router/src/link.tsx
```diff
@@ -183,19 +194,19 @@ export type ToSubOptions<
   TFrom extends RoutePaths<TRouteTree> | string = string,
   TTo extends string = '',
 > = {
-  to?: ToPathOption<TRouteTree, TFrom, TTo>
+  to?: ToPathOption<TRouteTree, TFrom, TTo> & {}
```
chorobin, line 197: This is a simple way to expand the complex type so it appears nice

PR #1495, opened 2024-04-19
fix: only allow leaf nodes as `to` suggestions

chorobin:
- Introduce `RouteLeaves` type to only autocomplete leaves
- We have to change `CheckPath` as well (yuck, I can't think of a better way)
- Add some type tests to cover only leaves

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -10,6 +10,7 @@ import type { Trim } from './fileRoute'
 import type { AnyRoute, RootSearchSchema } from './route'
 import type {
   RouteByPath,
+  RouteLeaves,
   RoutePaths,
   RoutePathsAutoComplete,
 } from './routeInfo'
@@ -88,8 +89,8 @@ export type RemoveLeadingSlashes<T> = T extends `/${infer R}`
 
 export type ResolvePaths<TRouteTree extends AnyRoute, TSearchPath> =
   RouteByPath<TRouteTree, RemoveTrailingSlashes<TSearchPath>> extends never
-    ? RoutePaths<TRouteTree>
-    : RoutePaths<RouteByPath<TRouteTree, RemoveTrailingSlashes<TSearchPath>>>
+    ? RouteLeaves<TRouteTree>
+    : RouteLeaves<RouteByPath<TRouteTree, RemoveTrailingSlashes<TSearchPath>>>
 
 export type SearchPaths<
   TRouteTree extends AnyRoute,
@@ -137,7 +138,7 @@ export type AbsolutePathAutoComplete<
           ? never
           : './')
   | (string extends TFrom ? '../' : TFrom extends `/` ? never : '../')
-  | RoutePaths<TRouteTree>
+  | RouteLeaves<TRouteTree>
   | (TFrom extends '/' ? never : SearchPaths<TRouteTree, TFrom>)
 
 export type RelativeToPathAutoComplete<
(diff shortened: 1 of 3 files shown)


packages/react-router/src/link.tsx
```diff
@@ -432,7 +433,15 @@ export type LinkOptions<
 }
 
 export type CheckPath<TRouteTree extends AnyRoute, TPass, TFail, TFrom, TTo> =
-  ResolveRoute<TRouteTree, TFrom, TTo> extends never ? TFail : TPass
+  ResolveRoute<TRouteTree, TFrom, TTo> extends infer TRoute extends AnyRoute
+    ? [TRoute] extends [never]
+      ? TFail
```
schiller-manuel, line 438: is this really necessary?
I remove it, and type tests still work (and it reduces instantiations from 573059 to 572245)

chorobin: You're right atm and thats because we're not strictly checking `from`. But this first checks if the route exists at all and if it does not then fail.

chorobin:
Before:
```
> tsc --extendedDiagnostics

Files:                         589
Lines of Definitions:        77165
Lines of TypeScript:         13986
Identifiers:                114743
Symbols:                    191235
Types:                       50725
Instantiations:             564447
Memory used:               228692K
Assignability cache size:    34134
Identity cache size:          3717
Subtype cache size:              0
Parse time:                  0.55s
ResolveModule time:          0.12s
Program time:                0.86s
Bind time:                   0.24s
Check time:                  4.26s
Total time:                  5.36s
```

After:

```
> tsc --extendedDiagnostics

Files:                         589
Lines of Definitions:        77166
Lines of TypeScript:         13986
Identifiers:                114762
Symbols:                    200101
Types:                       51716
Instantiations:             567375
Memory used:               229438K
Assignability cache size:    35133
Identity cache size:          3717
Subtype cache size:              0
I/O Read time:               0.07s
Parse time:                  0.56s
ResolveModule time:          0.12s
Program time:                0.86s
Bind time:                   0.29s
Check time:                  4.67s
Total time:                  5.82s
```

schiller-manuel (approved):

jdavidferreira:
@chorobin @schiller-manuel 

I have some routes like the following:

`/parent`
`/parent/child1`

After this change, I cannot do this:

```
<Link to="/parent" >Parent</Link>
```

Because TS is complaining, and it says that `"/parent"` is not assignable to the `to` type. Parent is not a leaf route, but I should be able to navigate to it, I do have a component/page for that route.

schiller-manuel:
You need to define an index route at `/parent/` if you want to navigate to this route

nklhtv:
@chorobin @schiller-manuel
Ouch, that completely destroys the structure of my app.
I've built an app with a route tree similar to the [location masking example](https://github.com/tanstack/router/tree/main/examples/react/location-masking) (without the actual masking)
```
const routeTree = rootRoute.addChildren([
  photoRoute,
  photosRoute.addChildren([photoModalRoute]),
  indexRoute,
])
```
where i have a:
- route similar to photosRoute which shows a long list and an Outlet at the end
- route similar to photoModalRoute which shows the details on individual element in modal

If i refactor my route tree to something like `photosRoute.addChildren([photosIndexRoute, photoModalRoute])` and render my list of items inside photosIndexRoute then i wont be able to show the modal on top of the list of items.

I've tried to open the location masking example in codesandbox and typescripts complain for using [<Link to={'/photos'}](https://github.com/TanStack/router/blob/main/examples/react/location-masking/src/main.tsx#L93)

PR #1496, opened 2024-04-20
feat(router-generator): add a biome-ignore comment on the generated route-tree file

k35o:
Errors do occur in `routeTree.gen.ts` in projects that use biome to check code.
I would like to ignore the errors in the generated files as well as in the response to eslint.


SeanCassiere (approved):

SeanCassiere:
@k35o can you please update the test snapshots in `packages/router-generator/tests/generator/**/routeTree.expected.ts` with the new line being added in.

k35o:
@SeanCassiere 
Thank you! I update it!

SeanCassiere (approved):

schiller-manuel:
instead of adding each possible ignore instruction separately, I would rather add a possibility to let the user customize the header.

SeanCassiere:
@schiller-manuel an array of strings would probably make the most sense.

Also, this change would require the ability to both *prepend* and *append* to the file in that case, since some tools have their comments set up in a way to `pause tracking` and `resume tracking`.

SeanCassiere:
@schiller-manuel PR is up!

https://github.com/TanStack/router/pull/1502

SeanCassiere:
Superseded by #1502.

@k35o thank you for your contribution which led to a user-customizable feature being added in #1502.

You can upgrade the router-generator to version [1.30.0](https://github.com/TanStack/router/releases/tag/v1.30.0) and configure the header content of the generated route-tree file using the `routeTreeFileHeader` field (see documentation here: https://tanstack.com/router/latest/docs/framework/react/guide/file-based-routing#options)

PR #1502, opened 2024-04-20
feat(router-generator): prepend and append user content to the generated route-tree file

SeanCassiere:
From this comment: https://github.com/TanStack/router/pull/1496#issuecomment-2067631665

This change allows a user to prepend and append lines to the generated route-tree file by letting them configure an array of strings for each property.


packages/router-generator/src/config.ts
```diff
@@ -13,6 +13,8 @@ export const configSchema = z.object({
   disableTypes: z.boolean().optional().default(false),
   addExtensions: z.boolean().optional().default(false),
   disableLogging: z.boolean().optional().default(false),
+  prependToRouteTreeFile: z.array(z.string()).optional().default([]),
```
SeanCassiere, line 16: Let me know if you've got a better name for this configuration field.

packages/router-generator/src/generator.ts
```diff
@@ -462,6 +462,7 @@ export async function generator(config: Config) {
     '// @ts-nocheck',
     '// noinspection JSUnusedGlobalSymbols',
     '// This file is auto-generated by TanStack Router',
+    ...config.prependToRouteTreeFile,
```
SeanCassiere, line 465: Do we want this to actually be prepended to the very top? or should it come after the stuff we are shipping in the file?

packages/router-generator/src/config.ts
schiller-manuel: how about `routeTreeFileHeader` ?

packages/router-generator/src/generator.ts
```diff
@@ -462,6 +462,7 @@ export async function generator(config: Config) {
     '// @ts-nocheck',
     '// noinspection JSUnusedGlobalSymbols',
     '// This file is auto-generated by TanStack Router',
```
schiller-manuel, line 464: I would put all this into the default value of `routeTreeFileHeader`

packages/router-generator/src/config.ts
```diff
@@ -13,6 +13,8 @@ export const configSchema = z.object({
...
+  appendToRouteTreeFile: z.array(z.string()).optional().default([]),
```
schiller-manuel, line 17: how about `routeTreeFileFooter`?

packages/router-generator/src/generator.ts
```diff
@@ -603,6 +604,7 @@ export async function generator(config: Config) {
         ]),
     '// Create and export the route tree',
     `export const routeTree = rootRoute.addChildren([${routeConfigChildrenText}])`,
+    ...config.appendToRouteTreeFile,
     '/* prettier-ignore-end */',
```
schiller-manuel, line 608: put `/* prettier-ignore-end */` into the default value of `routeTreeFileFooter`

packages/router-generator/src/generator.ts
SeanCassiere: Will do, although I think its worth keeping this one outside the default list and something we ship with the generator always instead: // This file is auto-generated by TanStack Router

schiller-manuel (approved):

PR #1509, opened 2024-04-22
feat: `trailingSlashes` option on the type level

chorobin:
We can infer `trailingSlashes` from the router and use it to enforce the `to` prop when navigating.

`trailingSlashes` currently has three options: always | never | preserve.

always - we always have a trailing slash on leaves autocompleted with to. Branches are not affected in from as you might want a loose match on this for any leaves under a branch instead of an index route.

never - we never have a trailing slash on leaves when autocompleted on to. from does not change as mentioned in always

preserve - to autocompletes to include both  trailing slash or without in the union. It can be either.


chorobin:
Don't merge yet. I want to add more type tests to cover over parts that could have regressed

packages/react-router/src/link.tsx
```diff
@@ -242,25 +248,22 @@ type PostProcessParams<
...
   TParamVariant extends ParamVariant,
   TFrom,
-  TFromRouteType extends
-    | 'allParams'
-    | 'fullSearchSchema' = TParamVariant extends 'PATH'
-    ? 'allParams'
-    : 'fullSearchSchema',
 > = PostProcessParams<
-  RouteByPath<TRouteTree, TFrom>['types'][TFromRouteType],
+  RouteByPath<TRouter['routeTree'], TFrom>['types'][TParamVariant extends 'PATH'
```
schiller-manuel, line 255: is this a typescript performance optimization (moving out of the type params)?

packages/react-router/tests/link.test-d.tsx
```diff
@@ -68,33 +68,183 @@ const routeTree = rootRoute.addChildren([
...
+
+type RouterAlwaysTrailingSlashes = typeof routerAlwaysTrailingSlashes
+
+type RouterNeverTrailingSlashes = typeof routerNeverTrailingSlashes
+
+type RouterPreserveTrailingSlashes = typeof routerPreserveTrailingSlashes
 
 test('when navigating to the root', () => {
-  expectTypeOf(Link<RouteTree, string, '/'>)
+  type hi = DefaultRouter['options']['trailingSlash']
```
schiller-manuel, line 99: remove?

chorobin: Oops. Ye. I'm changing the type tests atm

packages/react-router/src/link.tsx
chorobin: No. I thought about it quite a bit. I now need access to the options on the router. Which would be fine with `RegisteredRouter` if it weren't for the type tests. Module augmentation is awkward with type tests. It's easier to mock it out if the whole router is a type parameter

chorobin:
Should be ok to merge when approved

PR #1548, opened 2024-05-02
fix(react-router): changing path params without specifying the `to` destination

### LINKED ISSUE #1525: Navigation without to param does not work since 1.30.1
### Describe the bug

I updated to 1.30.1, because fix for bug with path params was introduced and it works fine. But, it broke other thing. Before 1.30.1 I was able to make link, or use navigate(), or throw redirect() without specifing to. I did it, because I only need to change path param(for example language code in path param), but I want to leave user on the same page and it worked fine. Now it ignores new path params and generates href with old params in this case.

everywhere

1. Create <Link> component and specify only from and params
2. Press/hover on it. Nothing happens

### Expected behavior

I expected to generate href with new params, but previous path

### Screenshots or Videos

_No response_

### Platform

- OS: all
- Browser: all
- Version: 1.30.1

### Additional context

_No response_

DogPawHat:
Add fixes and tests for `router.navigate` when using the `params: (prevParams) => newParams` option (with no extra `to` or `from` option).

Thanks to @SeanCassiere for the suggestion to patch `router.ts`

Fixes #1525
Should also fix https://github.com/TanStack/tanstack.com/issues/218

### DIFF
--- a/packages/react-router/tests/router.test.ts
+++ b/packages/react-router/tests/router.test.ts
@@ -301,47 +301,4 @@ describe('encoding: splat param for /$', () => {
       'framework/react/guide/file-based-routing tanstack',
     )
   })
-
-  it('router.navigate with `params` and `to`, should transition from "/posts/tanner" to "/posts/tkdodo"', async () => {
-    const { router, routes } = createTestRouter(
-      createMemoryHistory({ initialEntries: ['/posts/tanner'] }),
-    )
-
-    await router.load()
-
-    expect(router.state.location.pathname).toBe('/posts/tanner')
-
-    await router.navigate({
-      to: routes.postIdRoute.to,
-      params: (prev: any) => {
-        console.log(prev)
-        return { ...prev, slug: 'tkdodo' }
-      },
-    })
-    // Not sure why I have to do this
-    await router.invalidate()
-
-    expect(router.state.location.pathname).toBe('/posts/tkdodo')
-  })
-
-  it('router.navigate with `params` and no `to`, should transition from "/posts/tanner" to "/posts/tkdodo"', async () => {
-    const { router } = createTestRouter(
-      createMemoryHistory({ initialEntries: ['/posts/tanner'] }),
-    )
-
-    await router.load()
-
-    expect(router.state.location.pathname).toBe('/posts/tanner')
-
-    await router.navigate({
-      params: (prev: any) => ({ ...prev, slug: 'tkdodo' }),
-    })
-    await router.invalidate()
-
-    // This fails but should succeed
-    expect(router.state.location.pathname).toBe('/posts/tkdodo')
-
-    // This succeeds but should fail
-    // expect(router.state.location.pathname).toBe('/posts/tanner')
-  })
 })
--- a/packages/react-router/tests/router.test.ts
+++ b/packages/react-router/tests/router.test.ts
@@ -316,7 +316,7 @@ describe('encoding: splat param for /$', () => {
       params: (prev: any) => {
         console.log(prev)
         return { ...prev, slug: 'tkdodo' }
-      }
+      },
     })
     // Not sure why I have to do this
     await router.invalidate()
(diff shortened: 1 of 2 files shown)


packages/react-router/tests/router.test.ts
```diff
@@ -301,4 +301,45 @@ describe('encoding: splat param for /$', () => {
       'framework/react/guide/file-based-routing tanstack',
     )
   })
+
+  it('router.navigate with `params` and `to`, should transition from "/posts/tanner" to "/posts/tkdodo"', async () => {
```
SeanCassiere, line 305: Move your tests to it's own `redirects.test.tsx` file.

packages/react-router/tests/router.test.ts
```diff
@@ -301,4 +301,45 @@ describe('encoding: splat param for /$', () => {
       'framework/react/guide/file-based-routing tanstack',
     )
   })
+
```
SeanCassiere, line 304: Try adding this to [LN938 in router.ts](https://github.com/TanStack/router/blob/61ecd34f05c6dd2f13b2ad6490b1725cee2ff799/packages/react-router/src/router.ts#L938), it may just do the trick.

```tsx
      let pathname = dest.to
        ? this.resolvePathWithBase(fromPath, `${dest.to}`)
        : this.resolvePathWithBase(
            fromPath,
            stayingMatches?.find((d) => d.pathname === fromPath)?.routeId ||
              fromPath,
          )
```

packages/react-router/tests/router.test.ts
```diff
@@ -301,4 +301,45 @@ describe('encoding: splat param for /$', () => {
...
+        return { ...prev, slug: 'tkdodo' }
+      }
+    })
+    // Not sure why I have to do this
+    await router.invalidate()
+
+    expect(router.state.location.pathname).toBe('/posts/tkdodo')
+  })
+
+  it('router.navigate with `params` and no `to`, should transition from "/posts/tanner" to "/posts/tkdodo"', async () => {
```
SeanCassiere, line 327: With the tests in a new file, can your tests be expanded to include both a single path param (like you have now: `/posts/$slug`), and a more complicated setup like `/p/$project/$version/$framework` (URL like: `/p/router/latest/react`) and see if you can change only the $framework to "vue".

SeanCassiere (changes requested):
Left some comments

packages/react-router/tests/router.test.ts
DogPawHat: It works! Thanks!

DogPawHat:
I think were golden, the fix has been added.

SeanCassiere:
It's 4.30am here. I'll check this out later and run some tests in the examples.

schiller-manuel:
please update the title of this PR since now this not only adds tests but also fixes a abug

SeanCassiere (approved):

SeanCassiere:
Tested this locally and can confirm it works.

SeanCassiere (approved):

ruiaraujo012:
After this release, I'm getting an error when changing the search of a route:
```ts
navigate({
    search: { tab: newTab },
});
```
I have a screen with multiple tabs that are controlled by the search param `tab`. The first render (navigation) works well `/d0efbc00-7ac7-4869-b3af-f6115a9b52ef?tab=estado-pedido`, but when I change tabs, it throws an error:

```json
{
    "status": "resolved",
    "routerCode": "BEFORE_LOAD"
}
```

image

@DogPawHat @SeanCassiere any advice?

SeanCassiere:
@ruiaraujo012 please reference this PR and create an issue with a reproduction. I'll check it out over the weekend.

ruiaraujo012:
Ok, thanks, I'll do it tomorrow.

ruiaraujo012:
I'm unable to replicate it, I have a complex UI with multiple tabs. Maybe the problem is mine, but it only appears after I update the package, if I roll back it works again. If I put the to prop it works as well.

If I have more time, I'll try to replicate it again closer to what I have on my codebase.

ruiaraujo012:
Forget what I said, If I use a layout, it causes the error, I'll open an issue with the reproduction.

PR #1556, opened 2024-05-03
feat(react-router): same url navigation

rflban:


SeanCassiere:
Closing this since much of what this looks to be covering doesn't seem to be going in.

It'd probably be better if this was discussed in an issue before changing this behaviour.

PR #1559, opened 2024-05-04
fix(react-router): on initial load the `router.state.status` does not transition to "idle"

### LINKED ISSUE #1552: router status never changes from "pending"
### Describe the bug

From a Discord thread: https://discord.com/channels/719702312431386674/1235498104317284372

The Router's status never changes from "pending".

So when you use this code-snippet, it always stays as "pending" even after the page has loaded.
```tsx
const status = useRouterState({ select: (s) => s.status });
```

https://stackblitz.com/edit/github-zdscfw-ekwvlx?file=src%2Froutes%2Findex.tsx

1. Go to `/`
2. Observe the status remains as "pending" even after the page has completed loading.

### Expected behavior

As per the reference documentation, it should return to "idle" once done.

https://tanstack.com/router/latest/docs/framework/react/api/router/RouterStateType#status-property

### Screenshots or Videos

_No response_

### Platform

- TSR Version: 1.31.7

### Additional context

> "I bisected the versions, and it looks like it was introduced in v1.28.8. We've pinned to v1.28.1, so I didn't notice before" - Jakob Norlin

SeanCassiere:
On the initial load, the router's status does not transition back to "idle". This was working correctly prior to the [1.28.8](https://github.com/TanStack/router/releases/tag/v1.28.8) release.

Turns out, that the `usePrevious` hook used by the `<RouterProvider>`, wasn't actually storing the previous value so the conditions for firing the events were never met during the initial load.

Fixes #1552

### DIFF
--- a/packages/react-router/src/RouterProvider.tsx
+++ b/packages/react-router/src/RouterProvider.tsx
@@ -284,15 +284,14 @@ export type RouterProps<
 //   return ref.current
 // }
 
-function usePrevious<T>(value: T): T | undefined {
+function usePrevious<T>(value: T): T {
   const ref = React.useRef<T>(value)
 
-  if (typeof ref.current === 'undefined') {
-    ref.current = value
-    return ref.current
-  } else {
+  if (ref.current !== value) {
     const prevValue = ref.current
     ref.current = value
     return prevValue
+  } else {
+    return ref.current
   }
 }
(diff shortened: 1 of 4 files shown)


SeanCassiere:
Tanner mentioned this should be done in the router events and its possibly an event that's not firing correctly.

PR #1564, opened 2024-05-06
perf: geting lazy?, doing less work? generate more (plus some recommendations)

chorobin:
# Getting lazy? Doing less work? and generating more (plus some recommendations)

This PR focuses more on complex routes and to some degree many routes. When I say a complex route, I mean routes which have `search`, `params`, `loaders` and external dependencies such as `zod` or other heavy TS related dependencies.

A bigger issue I've noticed in TSR is in order to get suggestions, the route tree needs to be evaluated. If your routes contain complex things that TS needs to infer like validation schemas then this can slow down TS evaluating the route tree because its hard at work on the types of the validation libraries. But for most suggestions, we only need the `fullPath`, we don't need to work out the search schema or the params in order to get suggestions for `Link` for example until `search` is used. The ideal scenario is we do as little as possible in order to get the suggestions we need

## Show me the results!

I've updated our large filed based example to include:
- 100 routes for absolute pathing
- 100 routes for relative pathing
- 100 routes for params with loader
- 100 routes for search with loader

These include external dependencies which also do TS inference. Note the results may vary depending on what dependencies you use and how the types are written

On main:
```
> tsc --extendedDiagnostics

Files:                         636
Lines of Definitions:        80134
Lines of TypeScript:         18481
Identifiers:                132113
Symbols:                    345367
Types:                      119980
Instantiations:            2720561
Memory used:               345796K
Assignability cache size:    56379
Identity cache size:          8114
Subtype cache size:           2022
I/O Read time:               0.09s
Parse time:                  0.67s
ResolveModule time:          0.15s
Program time:                1.03s
Bind time:                   0.29s
Check time:                  7.25s
Total time:                  8.57s
```
On current branch:
```
> tsc --extendedDiagnostics

Files:                         636
Lines of Definitions:        80143
Lines of TypeScript:         19712
Identifiers:                133346
Symbols:                    322837
Types:                      111429
Instantiations:             773068
Memory used:               327186K
Assignability cache size:    45204
Identity cache size:          8109
Subtype cache size:           2022
Parse time:                  0.59s
ResolveModule time:          0.12s
Program time:                0.92s
Bind time:                   0.28s
Check time:                  5.51s
Total time:                  6.71s
```

## Getting lazy? Doing less work?

In TSR one of the largest performance bottlenecks is the route tree building. This node is important to reduce because it is directly linked to TS server suggestion times. What is contributing to this node being so slow is firstly, a lot of work upfront to evaluate `search`, `loader`, `context` and the others, which are possibly inferred from external libraries.

The first interesting change this PR introduces is not a breaking one but introducing a new signature to `addChildren`. `addChildren` accepts a tuple of routes like the following:

`route.addChildren([childRoute1, childRoute2])`

`addChildren` also in addition can now accept an object `route.addChildren({ childRoute1, childRoute2 })`.

Why this change? Well, it looks like tuples seem to be forcing TS to be doing something in each route which is quite slow and it gets worse with different external dependencies, where objects seem to be more cheap or perhaps more lazy in nature, maybe there's also caching involved? Depending on the libraries used in for example `validateSearch`, you may notice a difference in performance improvements.

For example, using `valibot` instead of `zod` in the above example with `addChildren` that accept a tuple
```
> tsc --extendedDiagnostics

Files:                          638
Lines of Library:             38680
Lines of Definitions:         86278
Lines of TypeScript:          19809
Lines of JavaScript:              0
Lines of JSON:                    0
Lines of Other:                   0
Identifiers:                 141166
Symbols:                     270865
Types:                        90876
Instantiations:            22514251
Memory used:                310778K
Assignability cache size:     33466
Identity cache size:           7787
Subtype cache size:            2830
Strict subtype cache size:        1
I/O Read time:                0.09s
Parse time:                   0.63s
ResolveModule time:           0.14s
ResolveTypeReference time:    0.01s
ResolveLibrary time:          0.02s
Program time:                 1.03s
Bind time:                    0.30s
Check time:                  14.66s
printTime time:               0.00s
Emit time:                    0.00s
Total time:                  15.99s
```
Now with `addChildren` that accepts an object

```
> tsc --extendedDiagnostics

Files:                         638
Lines of Library:            38680
Lines of Definitions:        86278
Lines of TypeScript:         19813
Identifiers:                141170
Symbols:                    269855
Types:                       90640
Instantiations:            1093144
Memory used:               316056K
Assignability cache size:    33455
Identity cache size:          7787
Subtype cache size:           2830
Strict subtype cache size:       1
I/O Read time:               0.09s
Parse time:                  0.68s
Program time:                1.09s
Bind time:                   0.35s
Check time:                  5.35s
Total time:                  6.79s
```

That's a big difference between tuples and objects. I'm still not totally sure why tuples behave this way or if I'm even correct about my assumptions but to me it looks like tuples are forcing some kind of extra work the TS compiler is doing and its related to `validateSearch`  and how the types of the external library are likely written.

### Profile

The first link that is checked by TS has to evaluate the route tree and it looks like this with `addChildren` accepting a tuple

image

The same node with `addChildren` accepting an object looks like this

image

But thats not the end of the story. Each individual Link is also faster with children as an object. Each Link is checked with the following time when `addChildren` accepts a tuple

image

And each Link is checked with the following time with `addChildren` accepts an object

image

As mentioned before. 30ms might not seem like much. But if you have 30ms for every `Link` in your code base, it quickly accumulates. 4ms is a huge difference in these cases

In this PR I have switched file based routing over to using objects automatically for a default performance boost. Code based can opt into the signature.

### Generating more

File based routing previously has done a lot of string interpretation on the type level in order to calculate things like `fullPath`, `path`, `id` etc. Template literal types are expensive for the compiler and generally in file based routing, these strings can be generated directly by the generator without hurting inference. This speeds up creating routes

### Improve params parsing

A more direct implementation of path params parsing without any additional splitting or mapped types

## What can blow up your TS Performance

**With TSR you want to narrow to the most relevant routes. This means using `from` and `to` on `Link` or other API's**.
 
For example, `<Link search={{ page: 0 }} />` will increase the type checking time on this link because TSR will resolve `search` to all possible routes (a union of all search params) and will therefore spend longer type checking it. This depends on the size of your route tree and how many search params you have but on our large example, I noticed a single `Link` would type check in roughly 100ms.

It is better to narrow the route as much as you can `<Link from={Route.fullPath} search={{ page: 0 }} />` with `from` or `to`. In this case the time type checking `Link` went down to 9ms.

This same logic also applies to `LinkOptions`. If you're using types from TSR which are by default not narrowed to a route then this can also blow up TS performance when merging this props with `Link`. Just remember that when `from` or `to` is not specified then things can start slowing down, so definitely try to do it as little as possible.


chorobin:
Still some work needs to be done so not ready just yet

packages/router-generator/src/generator.ts
```diff
@@ -599,7 +602,7 @@ export async function generator(config: Config) {
 }`,
         ]),
     '// Create and export the route tree',
-    `export const routeTree = rootRoute.addChildren([${routeConfigChildrenText}])`,
+    `export const routeTree = rootRoute.addChildren({${routeConfigChildrenText}})`,
```
schiller-manuel, line 605: should the object vs array syntax be configurable?

packages/react-router/src/route.ts
```diff
@@ -437,7 +459,10 @@ export interface AnyRoute
   > {}
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
-export type MergeFromFromParent<T, U> = IsAny<T, U, T & U>
+export type MergeFromFromParent<
```
schiller-manuel, line 462: maybe should be renamed? since it now only handles `allParam`

schiller-manuel:
we should add that `What can blow up your TS Performance` section into the documentation

chorobin:
> we should add that `What can blow up your TS Performance` section into the documentation

Definitely. Think we need to write a section on it so they can go to it and be able to resolve common problems

packages/react-router/src/fileRoute.ts
```diff
@@ -142,37 +139,27 @@ export class FileRoute<
   }
 
   createRoute = <
-    TSearchSchemaInput extends RouteConstraints['TSearchSchema'] = {},
-    TSearchSchema extends RouteConstraints['TSearchSchema'] = {},
-    TSearchSchemaUsed = TSearchSchemaInput extends SearchSchemaInput
-      ? Omit<TSearchSchemaInput, keyof SearchSchemaInput>
-      : TSearchSchema,
+    TSearchSchemaInput = {},
```
chorobin, line 142: I realised these constraints were not necessary because they were just constrained to `{}` which is already extremely loose

packages/router-generator/src/generator.ts
chorobin: I'm not sure. I like the idea that if we can confirm one way of building the route tree is more performance then we should prefer that for file based because I'm not sure if users care in this case if it is a tuple or object?

packages/router-generator/tests/generator/nested-layouts/routeTree.expected.ts
```diff
@@ -138,78 +144,142 @@ const JestedLayoutB3LayoutC2BarRoute = JestedLayoutB3LayoutC2BarImport.update({
 declare module '@tanstack/react-router' {
```
chorobin, line 144: @schiller-manuel I've added conditions for the tests in the generator for `fullPath`, `id`, `path`. Are there any situations missing here that I should add?

packages/router-generator/src/generator.ts
```diff
@@ -583,7 +583,16 @@ export async function generator(config: Config) {
   interface FileRoutesByPath {
     ${routeNodes
       .map((routeNode) => {
-        return `'${removeTrailingUnderscores(routeNode.routePath)}': {
+        const filePathId = removeTrailingUnderscores(routeNode.routePath)
+        const id = filePathId
+        const fullPath = removeGroups(
```
chorobin, line 588: @schiller-manuel Just want to check with you that this logic is correct and I'm not missing something?

packages/router-generator/src/generator.ts
schiller-manuel: ok, let's default to objects for filebased routing

packages/router-generator/src/generator.ts
```diff
@@ -583,7 +583,16 @@ export async function generator(config: Config) {
...
+          removeUnderscores(removeLayoutSegments(routeNode.routePath)) ?? '',
+        )
+
+        return `'${filePathId}': {
+          id: '${id}'
```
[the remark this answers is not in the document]
chorobin: I thought `id` was slightly different?

packages/react-router/src/fileRoute.ts
```diff
@@ -142,37 +74,27 @@ export class FileRoute<
...
-      RouteConstraints['TRouteContext'] = RouteContext,
-    TRouteContext = [TRouteContextReturn] extends [never]
-      ? RouteContext
-      : TRouteContextReturn,
-    TAllContext = Assign<
-      IsAny<TParentRoute['types']['allContext'], {}>,
-      TRouteContext
-    >,
-    TRouterContext extends RouteConstraints['TRouterContext'] = AnyContext,
+    TAllParams = ResolveAllParamsFromParent<TParentRoute, TParams>,
```
chorobin, line 89: It's a bit faster to name all complex types

chorobin:
Don't merge yet. Need to double check some things. Ideas 💡

schiller-manuel:
@SeanCassiere in light of https://github.com/TanStack/router/pull/1584, can you please review this?

packages/router-generator/src/generator.ts
SeanCassiere: Yeah, I agree. Plus we've been consistent about the fact that whatever is output from the router-generator is in our control anyway.

packages/router-generator/src/generator.ts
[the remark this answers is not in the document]
SeanCassiere: Correct me if I'm wrong, but you are explicitly setting these values on the declaration so you don't need to resolve the inference again via the TS-resolve types, yes?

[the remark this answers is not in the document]
SeanCassiere: Also, just want to confirm that we don't end up in a scenario that a created route configuration (not route type declaration) doesn't ever have both an `id` and a `path`. 

Like:
```tsx
const fooRoute = fooImport.update({
  id: 'foo',
  path: '/foo'
})
```

packages/router-generator/src/generator.ts
SeanCassiere: Checking the "kitchen-sink-file-based" example, currently the `/router-group` route has its id showing up as `/route-group` and NOT `/(this-folder-is-not-in-the-url)/route-group`.

@chorobin not sure if you want the `id` value here to represent the what the route id is at runtime in the router, but if so, you'd probably want to remove the groups here as well.

packages/react-router/src/link.tsx
```diff
@@ -56,13 +58,16 @@ export type Split<TValue, TIncludeTrailingSlash = true> = TValue extends unknown
       : never
   : never
 
-export type ParsePathParams<T extends string> = keyof {
-  [K in Trim<Split<T>[number], '_'> as K extends `$${infer L}`
-    ? L extends ''
+export type ParsePathParams<
```
SeanCassiere, line 61: This comment may not really belong here at this line, rather it's something I noticed about the link component when testing out the PR.

In the "kitchen-sink-file-based" example, using a `<Link>` component it accepts the following `to` values.
1. `/route-group`
2. `route-group`

`1` works as expected, but `2` then says it requires params.
```
Property 'params' is missing in type '{ to: "route-group"; }' but required in type 'MakeRequiredPathParams<Router<Route<any, "/", "/", string, "__root__", RootSearchSchema, RootSearchSchema, RootSearchSchema, RootSearchSchema, ... 10 more ..., { ...; }>, TrailingSlashOption, Record<...>, Record<...>>, string, "route-group">'.ts(2741)
```

Also, something else, I noticed is that the `from` property doesn't accept just `route-group`, only allows `/route-group` unlike the `to` property shown above.

SeanCassiere (commented):
I don't have the best understanding of all the ts-stuff we do with the router, but trying it out on my machine with out examples, the functionality works identical to what it was like before.

Left some comments, but nothing too crazy, all-in-all amazing, amazing, amazing, work @chorobin 🙌🏼!

SeanCassiere:
Also, we probably want to using the `feat:` or `refactor:` prefix on the commit name. Not sure if we are using conventional commits or angular commits, but if I've not mistaken on one of them using `perf:` triggers a major version bump.

SeanCassiere:
> @SeanCassiere in light of #1584, can you please review this?

Trying out the changes from my previous PR on this yielded no collisions that I could observe.

That being said, @chorobin once the merge conflict has been resolved, you'd want to rebase and confirm the tests I introduced in that PR still work.

docs/framework/react/guide/type-safety.md
```diff
@@ -110,3 +110,90 @@ const router = createRouter({
   },
 })
 ```
+
+## Performance Recommendations
+
+As your application scales, typescript check times will naturally increase. There are a few things to keep in mind when your application scales to keep your TS check times down.
```
schiller-manuel, line 116: typescript => TypeScript

docs/framework/react/guide/type-safety.md
```diff
@@ -110,3 +110,90 @@ const router = createRouter({
...
+
+Consider the following usage of `Link`
+
+```tsx
+<Link to=".." search={{ page: 0 }} />
+<Link to="." search={{ page: 0 }} />
+<Link search={{ page: 0 }} />
+```
+
+**These examples are bad for TS performance**. Thats because `search` resolves to a union of all `search` params for all routes and TS has to check whatever you pass to the `search` prop against this potentially big union. As your application grows, this check time will increase linearly to number of routes and search params. We have done our best to optimise for this case (typescript will typically do this work once and cache it) but the initial check against this large union is expensive. This also applies to `params` and other API's such as `useSearch`, `useParams`, `useNavigate` etc
```
schiller-manuel, line 128: Thats => That's

docs/framework/react/guide/type-safety.md
```diff
@@ -110,3 +110,90 @@ const router = createRouter({
...
+
+Instead you should try to narrow to relevant routes with `from` or `to`
+
+```tsx
+<Link from={Route.fullPath} to=".." search={{page: 0}} />
+<Link from="/posts" to=".." search={{page: 0}} />
+```
+
+Remember you can always pass a union to `to` or `from` to narrow the routes you're interested in. You can also pass branches to `from` to be more loose.
```
schiller-manuel, line 137: `You can also pass branches to `from` to be more loose.`

This probably needs more explanation for the standard user, especially the terminology of "branches" in the route tree.

docs/framework/react/guide/type-safety.md
```diff
@@ -110,3 +110,90 @@ const router = createRouter({
...
+  postsRoute: postsRoute.addChildren({ postRoute, postsIndexRoute }),
+  indexRoute,
+})
+```
+
+Note this syntax is more verbose but has better TS performance. With file based routing, the route tree is generated for you so a verbose route tree is not a concern
+
+### Avoid internal types without narrowing
+
+Its common you might want to re-use types exposed. For example you might be tempted to use `LinkProps` like so
```
schiller-manuel, line 161: Its => It is

docs/framework/react/guide/type-safety.md
```diff
@@ -110,3 +110,90 @@ const router = createRouter({
...
+```tsx
+const props = {
+  to: '/posts/',
+} as const satisfies LinkProps<RegisteredRouter, string '/posts/'>
+
+return (
+  <Link {...props}>
+)
+```
+
```
schiller-manuel, line 198: can we provide instructions how to type the params of functions that want to receive `LinkProps`?

examples/react/basic-file-based/src/routeTree.gen.ts
```diff
@@ -73,38 +73,65 @@ const LayoutLayout2LayoutARoute = LayoutLayout2LayoutAImport.update({
 declare module '@tanstack/react-router' {
```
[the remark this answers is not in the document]
schiller-manuel: while I agree we should re-generate the route-trees of file based examples, I don't think we MUST migrate the code based ones.

packages/react-router/src/index.tsx
```diff
@@ -71,6 +64,7 @@ export {
   type UseLinkPropsOptions,
   type ActiveLinkOptions,
   type LinkProps,
+  type LinkComponentProps,
```
schiller-manuel, line 67: do we really want to export this type? what would this be used for?

examples/react/basic-file-based/src/routeTree.gen.ts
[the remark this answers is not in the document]
chorobin: Let's slow down a bit. I'm exploring another idea that I really hope means we can not introduce a new signature. But we will see! I'm not sure if it'll work yet

packages/router-generator/src/generator.ts
chorobin: Thanks for catching that. I've updated the tests and the code to remove the route groups

packages/router-generator/src/generator.ts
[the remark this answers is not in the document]
chorobin: I'm not sure if I get your point. I think file based routing always had `id` and `path` in the types?

examples/react/basic-file-based/src/routeTree.gen.ts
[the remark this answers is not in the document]
chorobin: Ok. Turns out the idea is again limited by the children being tuples, it likely forcing unnecessary instantiations...

We need to keep the signature to allow objects for children

chorobin:
> Also, we probably want to using the `feat:` or `refactor:` prefix on the commit name. Not sure if we are using conventional commits or angular commits, but if I've not mistaken on one of them using `perf:` triggers a major version bump.

I've never heard of this. In conventional commits I thought you had to explicitly mention `!` or `BREAKING CHANGE` to bump major version

schiller-manuel:
this can be seen here: https://github.com/TanStack/config/blob/main/src/publish/index.js#L144

however, it only matters that the squashed commit has that prefix, so it can be set when this is merged

packages/react-router/src/route.ts
```diff
@@ -763,7 +788,11 @@ export class Route<
     this.to = fullPath as TrimPathRight<TFullPath>
   }
 
-  addChildren = <const TNewChildren extends ReadonlyArray<AnyRoute>>(
+  addChildren = <
+    const TNewChildren extends
+      | Record<string, AnyRoute>
```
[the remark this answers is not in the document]
chorobin: Added to the performance recommendations

packages/react-router/src/link.tsx
chorobin: Yeah, its a separate issue I'm going to be fixing in a separate PR. Its a bug on main as well.

packages/react-router/src/link.tsx
```diff
@@ -184,7 +191,13 @@ export type ToOptions<
   TTo extends string = '',
   TMaskFrom extends RoutePaths<TRouter['routeTree']> | string = TFrom,
   TMaskTo extends string = '',
-> = ToSubOptions<TRouter, TFrom, TTo> & {
+> = ToSubOptions<TRouter, TFrom, TTo> & MaskOptions<TRouter, TMaskFrom, TMaskTo>
+
+export interface MaskOptions<
```
chorobin, line 196: Naming all of these types helps TS cache stuff when passing `LinkProps` to a `Link`. Slightly more optimised

packages/router-generator/src/generator.ts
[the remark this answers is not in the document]
SeanCassiere: Check more into this, it looks good.

Just wanted to make sure we weren't modifying anything in route configuration, and only touching stuff on the type declaration.

PR #1577, opened 2024-05-09
feat(react-router): add a `useLocation` hook to get the current `ParsedLocation`

MrPancakes39:
The `useLocation` is a useful hook in [_React Router_](https://reactrouter.com/en/main/hooks/use-location).
So I decided to add it to the _TanStack Router_ too.


packages/react-router/src/useLocation.ts
```diff
@@ -0,0 +1,10 @@
+import { useRouter } from "./useRouter";
+import type { ParsedLocation } from "./location";
+import type { FullSearchSchema } from "./routeInfo";
+import type { AnyRoute } from "./route";
+
+
+export function useLocation<TRouteTree extends AnyRoute = AnyRoute>(): ParsedLocation<FullSearchSchema<TRouteTree>> {
+  const router = useRouter();
```
SeanCassiere, line 8: For perf, it'd be better if the selector in the useRouterState hook was used.

```tsx
const location = useRouterState({ select: (s) => s.location })
```

docs/framework/react/api/router/useLocationHook.md
```diff
@@ -0,0 +1,23 @@
+---
+id: useLocationHook
+title: useLocation hook
+---
+
+The `useLocation` method is a hook that returns the current [`location`](./ParsedLocationType.md) object. This hook is useful for when you want to perform some side effect whenever the current location changes.
+
+## useLocation returns
+
+- The current [`location`](./ParsedLocationType.md) object.
```
SeanCassiere, line 10: The `.md` extension shouldn't be used.

docs/framework/react/api/router/useLocationHook.md
```diff
@@ -0,0 +1,23 @@
+---
+id: useLocationHook
+title: useLocation hook
+---
+
+The `useLocation` method is a hook that returns the current [`location`](./ParsedLocationType.md) object. This hook is useful for when you want to perform some side effect whenever the current location changes.
```
SeanCassiere, line 6: The `.md` extension shouldn't be used.

SeanCassiere (changes requested):
Some tests would also be needed here. You can see the setup in `packages/react-router/tests/router.test.tsx` and grab what you need from there.

SeanCassiere:
Not sure if we are going to add this hook into the library or not, but this PR would need some changes should we decide to add it in.

packages/react-router/src/useLocation.ts
```diff
@@ -0,0 +1,10 @@
+import { useRouter } from "./useRouter";
+import type { ParsedLocation } from "./location";
+import type { FullSearchSchema } from "./routeInfo";
+import type { AnyRoute } from "./route";
+
+
+export function useLocation<TRouteTree extends AnyRoute = AnyRoute>(): ParsedLocation<FullSearchSchema<TRouteTree>> {
```
SeanCassiere, line 7: Should we be supporting the select syntax here? Like

```tsx
const pathname = useLocation({ select: s => s.pathname })
```

tannerlinsley: Yes probably

SeanCassiere (approved):

SeanCassiere:
Good to go.

MrPancakes39:
I apologize for not seeing the needed changes sooner and resolving them myself.
Thank you @SeanCassiere for your consideration and work.

PR #1590, opened 2024-05-13
tests(react-router): add some basic runtime tests for `Link`

chorobin:
Again quite high level to test navigation and not internals. I will add some more but wanted to know if you guys are ok with basic high level tests like this

### DIFF
--- a/packages/react-router/tests/link.test.tsx
+++ b/packages/react-router/tests/link.test.tsx
@@ -1,44 +1,878 @@
 import React from 'react'
-import { describe, it, expect, afterEach } from 'vitest'
-import { render, cleanup } from '@testing-library/react'
+import '@testing-library/jest-dom/vitest'
+import { afterEach, describe, expect, it, test, vi } from 'vitest'
+import {
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from '@testing-library/react'
 
 import {
-  createRouter,
+  createLink,
+  createMemoryHistory,
   createRootRoute,
   createRoute,
-  createMemoryHistory,
-  RouterProvider,
-  createLink,
+  createRouter,
   Link,
+  RouterProvider,
+  useLoaderData,
+  useSearch,
+  redirect,
+  useRouteContext,
+  useParams,
+  Outlet,
+  createRootRouteWithContext,
 } from '../src'
 
 afterEach(() => {
+  window.history.replaceState(null, 'root', '/')
   cleanup()
 })
 
 describe('Link', () => {
-  it('should NOT pass the "disabled" prop to the rendered Link component', async () => {
+  test('when a Link is disabled', async () => {
     const rootRoute = createRootRoute()
     const indexRoute = createRoute({
       getParentRoute: () => rootRoute,
       path: '/',
       component: () => (
-        <Link to="/" disabled>
-          Index
-        </Link>
+        <React.Fragment>
+          <h1>Index</h1>
+          <Link to="/posts" disabled>
+            Posts
+          </Link>
+        </React.Fragment>
       ),
     })
-    const router = createRouter({
-      routeTree: rootRoute.addChildren([indexRoute]),
-      history: createMemoryHistory({ initialEntries: ['/'] }),
+
+    const postsRoute = createRoute({
+      getParentRoute: () => rootRoute,
+      path: '/posts',
+      component: () => (
+        <React.Fragment>
+          <h1>Posts</h1>
+        </React.Fragment>
+      ),
     })
 
-    await router.load()
+    const router = createRouter({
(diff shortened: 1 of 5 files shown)


packages/react-router/tests/link.test.tsx
```diff
@@ -10,35 +11,449 @@ import {
...
+        )
+      },
+    })
+
+    const router = createRouter({
+      routeTree: rootRoute.addChildren([indexRoute, postsRoute]),
+      history: createMemoryHistory({ initialEntries: ['/'] }),
+    })
+
+    await router.load()
```
SeanCassiere, line 117: For real-life testing, should we be doing the `router. load()` stuff? Because that seems like a departure from what's really happening at runtime

TkDodo: How about rendering the router provider and then just waiting for something to be visible on the screen? Shouldn't the loaders be triggered automatically?

chorobin: Yeah. I just thought I can do this

packages/react-router/tests/link.test.tsx
```diff
@@ -1,44 +1,805 @@
...
 } from '../src'
 
 afterEach(() => {
+  window.history.replaceState(null, 'root', '/')
   cleanup()
 })
 
 describe('Link', () => {
-  it('should NOT pass the "disabled" prop to the rendered Link component', async () => {
+  test('when a Link is disabled', async () => {
```
SeanCassiere, line 28: Is there any reason `test` is being used instead of `it`?

packages/react-router/tests/link.test.tsx
```diff
@@ -1,44 +1,805 @@
...
+    fireEvent.click(postsLink)
+
+    expect(await screen.findByText('Posts Index')).toBeInTheDocument()
+
+    const postLink = await screen.findByRole('link', {
+      name: 'To the first post',
+    })
+
+    expect(postLink).toHaveAttribute('href', '/posts/id1')
+    expect(postLink).toHaveAttribute('from', '/posts/')
```
SeanCassiere, line 701: Is this checking the `<Link>` component as React node? or the actual HTML that's printed to the document?

If it is the `<Link>` component, then cool 👍🏼.

But if it is the HTML anchor element being printed to the document, then maybe that's actually a bug/unintended behaviour? Since `from` is not a valid HTML Anchor attribute, then it's probably not something we should then be *expecting* to available in the DOM.

packages/react-router/tests/link.test.tsx
```diff
@@ -1,44 +1,805 @@
...
+    expect(postsLink).toHaveAttribute('href', '/posts?page=2')
+
+    fireEvent.click(postsLink)
+
+    expect(await screen.findByText('Something went wrong!'))
+
+    expect(onError).toHaveBeenCalledOnce()
+  })
+
+  test('when navigating to /posts with a beforeLoad that redirects', async () => {
```
SeanCassiere, line 443: In the future perhaps, this should be moved to `redirects.tests.tsx` and be tested at runtime using both `<Link>` and a button wired up to `useNavigate()`.

packages/react-router/tests/link.test.tsx
```diff
@@ -1,44 +1,805 @@
...
+    expect(indexLink).toHaveAttribute('aria-current', 'page')
+    expect(indexLink).toHaveClass('active')
+    expect(indexLink).toHaveAttribute('data-status', 'active')
+    expect(indexLink).toHaveAttribute('href', '/')
 
-    expect(customElement!.hasAttribute('disabled')).toBe(false)
+    const postsLink = await screen.findByRole('link', { name: 'Posts' })
+
+    expect(postsLink).toHaveClass('inactive')
+    expect(postsLink).toHaveAttribute('href', '/posts')
```
SeanCassiere, line 123: Could a test also be added in, testing the href value when a `basePath` value is set on the router.

```tsx
const router = createRouter({
  basePath: '/app'
})
```

Seems like a good runtime check to guarantee.

packages/react-router/tests/link.test.tsx
chorobin: I personally prefer actually no nesting in tests. And to juse use `test`. I like my tests super simple

packages/react-router/tests/link.test.tsx
chorobin: Yeah, I guess this is a bug!. It's the `a` tag

packages/react-router/tests/link.test.tsx
chorobin: I guess its hard to know where it truly lives. But I wanted to test the user pressing a link and it redirecting them because I think that's how it's actually used

PR #1595, opened 2024-05-15
fix: take from into account when building urls

TkDodo:


packages/react-router/src/router.ts
```diff
@@ -919,7 +919,7 @@ export class Router<
       } = {},
       matches?: Array<MakeRouteMatch<TRouteTree>>,
     ): ParsedLocation => {
-      let fromPath = this.latestLocation.pathname
+      let fromPath = dest.from || this.latestLocation.pathname
```
TkDodo, line 922: this is the fix - the rest is tests :)

chorobin (approved):

SeanCassiere (approved):

PR #1597, opened 2024-05-15
fix(router-generator): generator should not infer trailing slash is not necessary for index routes

### LINKED ISSUE #1596: `route.tsx` changes whether trailing slashes are required in links to `index.tsx`
### Describe the bug

With file-based routing, `foo/route.tsx` changes whether links to `foo/index.tsx` must be `/foo/` or `/foo`:

1. If `foo/route.tsx` does not exist, `foo/index.tsx` can be linked to only with `/foo/`, not `/foo`.
1. If `foo/route.tsx` does exist, `foo/index.tsx` can be linked to only with `/foo`, not `/foo/`.

https://stackblitz.com/edit/github-fzqvge?file=src%2Froutes%2F__root.tsx

1. Use file-based routing.
2. In a directory that has an `index.tsx`, create a `route.tsx`.

### Expected behavior

In the given StackBlitz example, I expected either

- the trailing slash to always be required, or
- the trailing slash to never be required,

i.e. I did not expect the presence of `route.tsx` to change the links to `index.tsx`.

### Screenshots or Videos

image

### Platform

StackBlitz

### Additional context

_No response_

chorobin:
Trailing slashes should only be there if the index route has to be uniquely identified by it.

fixes #1596

### DIFF
--- a/packages/router-generator/tests/generator/flat/routeTree.expected.ts
+++ b/packages/router-generator/tests/generator/flat/routeTree.expected.ts
@@ -123,15 +123,15 @@ declare module '@tanstack/react-router' {
     }
     '/blog/$slug/': {
       id: '/blog/$slug/'
-      path: '/$slug/'
-      fullPath: '/blog/$slug/'
+      path: '/$slug'
+      fullPath: '/blog/$slug'
       preLoaderRoute: typeof BlogSlugIndexImport
       parentRoute: typeof BlogRouteImport
     }
     '/posts/$postId/': {
       id: '/posts/$postId/'
-      path: '/$postId/'
-      fullPath: '/posts/$postId/'
+      path: '/$postId'
+      fullPath: '/posts/$postId'
       preLoaderRoute: typeof PostsPostIdIndexImport
       parentRoute: typeof PostsRouteImport
     }
--- a/packages/router-generator/src/generator.ts
+++ b/packages/router-generator/src/generator.ts
@@ -585,14 +585,11 @@ export async function generator(config: Config) {
       .map((routeNode) => {
         const filePathId = removeTrailingUnderscores(routeNode.routePath)
         const id = removeGroups(filePathId ?? '')
-        const fullPath = removeGroups(
-          removeUnderscores(removeLayoutSegments(routeNode.routePath)) ?? '',
-        )
 
         return `'${filePathId}': {
           id: '${id}'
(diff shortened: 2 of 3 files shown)


packages/router-generator/tests/generator/flat/routeTree.expected.ts
```diff
@@ -123,15 +123,15 @@ declare module '@tanstack/react-router' {
     }
     '/blog/$slug/': {
       id: '/blog/$slug/'
```
chorobin, line 125: Not sure if the id should have a trailing slash or not? But at least this fixes the `fullPath` and `path`. TS used to normalize this but its right to move this work to the generator as we don't want the compiler spending time on stuff the generator can do

TkDodo (commented):
Do we have a test for different `trailingSlash` settings too?

TkDodo (approved):

schiller-manuel (approved):

packages/router-generator/tests/generator/flat/routeTree.expected.ts
SeanCassiere: Where is this `id` field meant to be used in Router?

If it is being used in places like useSearch, and hooks that use `from`, then trailing-slashes should be removed (with the exception to index routes).

For example, in the basic example, `from` accepts the following options regardless of the `trailingSlash` setting.
```
/
/posts
/posts/
/posts/$postId
__root__
```

SeanCassiere (approved):

SeanCassiere:
@chorobin looks good on my-side, just left [a comment](https://github.com/TanStack/router/pull/1597#discussion_r1602331479) regarding the `id` value.

packages/router-generator/tests/generator/flat/routeTree.expected.ts
chorobin: Let me clarify. These are index routes that I'm removing trailing slashes from. It's index routes without a `route.tsx` which was causing the problem. Before I guess TS was removing this slash, then I moved it to the generator and `fullPath` got a trailing slash in this situation.

So do you think `id` should have a trailing slash or not in this situation? Because `TFilePath` does have one and I guess always had one

chorobin: I just checked previous versions and `id` always had a trailing slash in this situation. I'm going to merge

PR #1611, opened 2024-05-17
fix(react-router): allow root route to be wrapped in suspense

### LINKED ISSUE #1601: PendingComponent isn't shown while beforeLoad is pending on root route
### Describe the bug

Since 1.31.28, `PendingComponent` on the root route isn't shown while `beforeLoad` is pending. This looks like it was introduced by https://github.com/TanStack/router/commit/854e52226d4c9241ce787dd82fc69c3e19c57a1a#diff-4cfd6abe2272d72c1dd073ec37936ad823bcd1f9418b466a9dea8b50214448f8R186. 
Instead, a white page is shown as nothing gets mounted.

https://stackblitz.com/edit/github-khgssk-d8fyff

1. Open app
2. Wait for root `beforeLoad` to resolve
3. Observe that the `pendingComponent` is never shown

### Expected behavior

PendingComponent should be shown.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- TSR Version: 1.32.12

### Additional context

_No response_

freshgiammi:
By default, the root route isn't wrapped in a `Suspense` which causes the pendingComponent to not be shown while beforeLoad is resolving. 
This is needed for SSR, but leaves SPA users unable to show a loading component while loaders resolve.

This patch checks the `wrapInSuspense` option to forcefully enable React.Suspense in the root route.

Fixes #1601

### DIFF
--- a/packages/react-router/src/Matches.tsx
+++ b/packages/react-router/src/Matches.tsx
@@ -184,7 +184,8 @@ export function Match({ matchId }: { matchId: string }) {
     : route.options.notFoundComponent
 
   const ResolvedSuspenseBoundary =
-    !route.isRoot &&
+    // If we're on the root route, allow forcefully wrapping in suspense
+    (!route.isRoot || route.options.wrapInSuspense) &&
     (route.options.wrapInSuspense ??
       PendingComponent ??
       (route.options.errorComponent as any)?.preload)


SeanCassiere:
Adding a Suspense Boundary above the RouterProvider would also work yes?

```tsx
function App() {
  return (
  <React.Suspense fallback={<DefaultPendingComponent />}>
    <RouterProvider router={router} />
  </React.Suspense>
  )
}
```

freshgiammi:
> Adding a Suspense Boundary above the RouterProvider would also work yes?
> 
> ```tsx
> function App() {
>   return (
>   <React.Suspense fallback={<DefaultPendingComponent />}>
>     <RouterProvider router={router} />
>   </React.Suspense>
>   )
> }
> ```

It doesn't look like it's working, neither on my app nor the [reproduction](https://stackblitz.com/edit/github-khgssk-d8fyff) linked in the issue. Besides if it were, we'd still have two options in `createRootRoute/createRootRouteWithContext` which don't do anything, as the checks for suspense are skipped on root routes: `pendingComponent` and `wrapInSuspense`.

PR #1613, opened 2024-05-17
fix(router-devtools): correctly compose button toggle styles

freshgiammi:
Currently, router-devtools fails to correctly compose the provided `toggleButtonProps` with the toggle button styles, resulting in provided properties not being applied to the button element.

This patch correctly composes `style` and `className` back into the toggle.


freshgiammi:
On a side note, it looks like `closeButtonProps` aren't applied correctly too, but they can't be fixed without a bit of refactoring.
That prop makes less sense than the toggle one IMHO, but I haven't touched it yet (even though I'm guessing removing it entirely should be fine).

PR #1625, opened 2024-05-18
refactor: don't disable rules-of-hooks linter

TkDodo:


packages/react-router/src/Transitioner.tsx
```diff
@@ -66,8 +57,16 @@ export function Transitioner() {
...
+    const tryLoad = async () => {
+      try {
+        await router.load()
+      } catch (err) {
+        console.error(err)
+      }
+    }
+
     tryLoad()
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
TkDodo, line 70: moving `tryLoad` into the effect fixes the issue as it's not used outside.

packages/react-router/src/Transitioner.tsx
```diff
@@ -46,14 +38,13 @@ export function Transitioner() {
       state: true,
     })
 
-    if (routerState.location.href !== nextLocation.href) {
+    if (router.state.location.href !== nextLocation.href) {
```
TkDodo, line 41: if we don't want to be reactive towards `routerState` coming from `useRouterState()` (which we don't want as we disabled the linter for it), we can just use `router.state` (which really is `router.getState()`) to read a fresh value inside the effect.

TkDodo: I don't know. If we wanted that, the correct fix is to add it to the dependency array. But since I don't know the impact of this change, I wanted to keep the refactoring without any runtime changes.

Because the dependency array before was:

```
// eslint-disable-next-line react-hooks/exhaustive-deps
}, [router, router.history])
```

it wasn't reacting to `router.state.location.href` either. When we just remove the warning, the linter wants to add it to the dependency array. But again, that changes the behaviour because the effect will potentially run more often.

tannerlinsley (approved):

PR #1626, opened 2024-05-18
refactor: fix type errors in eliminateUnreferencedIdentifiers.ts

TkDodo:


packages/router-vite-plugin/src/eliminateUnreferencedIdentifiers.ts
```diff
@@ -106,35 +106,6 @@ export const eliminateUnreferencedIdentifiers = (
...
-              property.remove()
-            }
-          })
-
-          if (
-            beforeCount !== referencesRemovedInThisPass &&
-            pattern.get('properties').length < 1
-          ) {
-            path.remove()
-          }
```
TkDodo, line 137: this seems to be a leftover as `path.node.id.type === 'ObjectPattern'` is checked one line above, which means this should be dead code

packages/router-vite-plugin/src/eliminateUnreferencedIdentifiers.ts
```diff
@@ -69,16 +69,16 @@ export const eliminateUnreferencedIdentifiers = (
...
           }
         } else if (t.isObjectPattern(value)) {
-          handleObjectPattern(value as NodePath<BabelTypes.ObjectPattern>)
+          handleObjectPattern(value as any)
         }
       } else if (t.isRestElement(property.node)) {
         const argument = property.get('argument')
         if (
-          t.isIdentifier(argument) &&
+          t.isIdentifier(argument as any) &&
```
TkDodo, line 81: not sure what else to do here - I think it's more important to get `main` into a state that has no type-test errors

packages/react-router/src/root.ts
```diff
@@ -0,0 +1,2 @@
+export const rootRouteId = '__root__' as const
```
SeanCassiere, line 1: Looks like its just the import for this in `src/fileRoute` that needs to be removed.

SeanCassiere: Fixed these

TkDodo: Thx

PR #1639, opened 2024-05-20
docs: Update `useBlocker` function signature

eliamaino-fp:
According to the [signature](https://github.com/TanStack/router/blob/05941e5ef2b7d2776e885cf473fdcc3970548b22/packages/react-router/src/useBlocker.tsx#L7) the blocking function has `BlockerFn` type and not `() => string`


schiller-manuel:
the prop is not called `message` anymore, can you fix this as well please?
I also noticed we don't have API docs of the `useBlocker` hook ...

Issue #1647, opened 2024-05-21
Throwing an error in a `beforeLoad` handler does not render error component


Zebeqo:
### Describe the bug

https://github.com/TanStack/router/issues/724#issuecomment-2122566407

### Your Example Website or App

https://stackblitz.com/edit/github-s4ju8g?file=src%2Froutes%2Ftest.tsx

### Steps to Reproduce the Bug or Issue

1. Click on `Test` link
2. Not showing error component from `_root.tsx`
3. Move error from `beforeLoad` to `loader`
4. Error component is showed as expected

### Expected behavior

Throwing an error in a `beforeLoad` handler should render error component

### Platform

- OS: macOS
- Browser: Chrome
- Version: 124.0

PR #1648, opened 2024-05-21
fix: should render error component when beforeLoad throws an error

chorobin:
beforeLoad wasn't rendering the errorComponent. Added unit tests to cover this case.

### DIFF
--- a/packages/react-router/tests/link.test.tsx
+++ b/packages/react-router/tests/link.test.tsx
@@ -77,7 +77,7 @@ describe('Link', () => {
     ).rejects.toThrow()
   })
 
-  it('when the current route is the root', async () => {
+  test('when the current route is the root', async () => {
     const rootRoute = createRootRoute()
     const indexRoute = createRoute({
       getParentRoute: () => rootRoute,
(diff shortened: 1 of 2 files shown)


packages/react-router/tests/link.test.tsx
```diff
@@ -619,6 +668,64 @@ describe('Link', () => {
     expect(await screen.findByText('UserId: userId')).toBeInTheDocument()
   })
 
+  test('when navigating to /posts with a beforeLoad that throws an error', async () => {
+    const onError = vi.fn()
+    const rootRoute = createRootRouteWithContext<{
```
schiller-manuel, line 673: why do we need router context here?
If we don't need it for this test, please remove to simplify the test

packages/react-router/tests/link.test.tsx
```diff
@@ -619,6 +668,64 @@ describe('Link', () => {
     expect(await screen.findByText('UserId: userId')).toBeInTheDocument()
   })
 
+  test('when navigating to /posts with a beforeLoad that throws an error', async () => {
```
schiller-manuel, line 671: we would also need a test that shows that the error is bubbling up from e.g. in `posts` to `__root`

schiller-manuel:
fixes #1647

PR #1649, opened 2024-05-21
feat(react-router): improve the blocker functionality for use with custom UI

Balastrong:
We were talking about this on discord, trying to find a handy way to display custom UI when blocking navigation.

I adapted a bit @freshgiammi's first draft and expanded it to the `Blocker` component to also have `allow`/`deny` in its children.
I also updated the example to validate it works as expected.

Basically we get these two functions to allow the user to show some custom UI (a modal?) when blocking the navigation and giving full control if resuming or blocking it.

If the API makes sense I'm happy to also update the docs.

cc @schiller-manuel

### DIFF
--- a/examples/react/navigation-blocking/src/main.tsx
+++ b/examples/react/navigation-blocking/src/main.tsx
@@ -67,15 +67,13 @@ const editor1Route = createRoute({
 function Editor1Component() {
   const [value, setValue] = React.useState('')
   const [useCustomBlocker, setUseCustomBlocker] = React.useState(false)
-  const [showConfirm, setShowConfirm] = React.useState(false)
 
-  const { allow, deny } = useBlocker(
-    () =>
-      useCustomBlocker
-        ? setShowConfirm(true)
-        : window.confirm('Are you sure you want to leave editor 1?'),
-    value,
-  )
+  const { proceed, reset, status } = useBlocker({
+    blockerFn: useCustomBlocker
+      ? undefined
+      : () => window.confirm('Are you sure you want to leave editor 1?'),
+    condition: value,
+  })
 
   return (
     <div className="flex flex-col p-2">
@@ -95,24 +93,18 @@ function Editor1Component() {
           className="border"
         />
       </div>
-      {showConfirm && (
+      {status === 'blocked' && (
         <div className="mt-2">
           <div>Are you sure you want to leave editor 1?</div>
           <button
             className="bg-lime-500 text-white rounded p-1 px-2 mr-2"
-            onClick={() => {
-              setShowConfirm(false)
-              allow()
-            }}
+            onClick={proceed}
           >
             YES
           </button>
           <button
             className="bg-red-500 text-white rounded p-1 px-2"
-            onClick={() => {
-              setShowConfirm(false)
-              deny()
-            }}
+            onClick={reset}
           >
             NO
           </button>
@@ -134,10 +126,10 @@ const editor2Route = createRoute({
 function Editor2Component() {
   const [value, setValue] = React.useState('')
 
-  useBlocker(
-    () => window.confirm('Are you sure you want to leave editor 2?'),
-    value,
-  )
+  useBlocker({
+    blockerFn: () => window.confirm('Are you sure you want to leave editor 2?'),
+    condition: value,
+  })
 
   return (
     <div className="p-2">
--- a/examples/react/navigation-blocking/src/main.tsx
+++ b/examples/react/navigation-blocking/src/main.tsx
@@ -66,18 +66,35 @@ const editor1Route = createRoute({
 
 function Editor1Component() {
   const [value, setValue] = React.useState('')
+  const [useCustomBlocker, setUseCustomBlocker] = React.useState(false)
   const [showConfirm, setShowConfirm] = React.useState(false)
 
-  const { allow, deny } = useBlocker(() => setShowConfirm(true), value)
+  const { allow, deny } = useBlocker(
+    () =>
+      useCustomBlocker
+        ? setShowConfirm(true)
+        : window.confirm('Are you sure you want to leave editor 1?'),
+    value,
+  )
 
   return (
-    <div className="p-2">
+    <div className="flex flex-col p-2">
       <h3>Editor 1</h3>
-      <input
-        value={value}
-        onChange={(e) => setValue(e.target.value)}
-        className="border"
-      />
+      <label>
+        <input
+          type="checkbox"
+          checked={useCustomBlocker}
+          onChange={(e) => setUseCustomBlocker(e.target.checked)}
+        />{' '}
+        Use custom blocker
+      </label>
+      <div>
+        <input
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
+          className="border"
+        />
+      </div>
       {showConfirm && (
         <div className="mt-2">
           <div>Are you sure you want to leave editor 1?</div>
--- a/examples/react/navigation-blocking/src/main.tsx
+++ b/examples/react/navigation-blocking/src/main.tsx
@@ -1,15 +1,15 @@
-import React from 'react'
-import ReactDOM from 'react-dom/client'
 import {
+  Link,
   Outlet,
   RouterProvider,
-  Link,
+  createRootRoute,
+  createRoute,
   createRouter,
   useBlocker,
-  createRoute,
-  createRootRoute,
 } from '@tanstack/react-router'
 import { TanStackRouterDevtools } from '@tanstack/router-devtools'
+import React from 'react'
+import ReactDOM from 'react-dom/client'
 
 const rootRoute = createRootRoute({
   component: RootComponent,
@@ -66,11 +66,9 @@ const editor1Route = createRoute({
 
 function Editor1Component() {
   const [value, setValue] = React.useState('')
+  const [showConfirm, setShowConfirm] = React.useState(false)
 
-  useBlocker(
-    () => window.confirm('Are you sure you want to leave editor 1?'),
-    value,
-  )
+  const { allow, deny } = useBlocker(() => setShowConfirm(true), value)
 
   return (
     <div className="p-2">
@@ -80,6 +78,29 @@ function Editor1Component() {
         onChange={(e) => setValue(e.target.value)}
         className="border"
       />
+      {showConfirm && (
+        <div className="mt-2">
+          <div>Are you sure you want to leave editor 1?</div>
+          <button
+            className="bg-lime-500 text-white rounded p-1 px-2 mr-2"
+            onClick={() => {
+              setShowConfirm(false)
+              allow()
+            }}
+          >
+            YES
+          </button>
+          <button
+            className="bg-red-500 text-white rounded p-1 px-2"
+            onClick={() => {
+              setShowConfirm(false)
+              deny()
+            }}
+          >
+            NO
+          </button>
+        </div>
+      )}
       <hr className="m-2" />
       <Link to="/editor-1/editor-2">Go to Editor 2</Link>
       <Outlet />
--- a/packages/react-router/src/useBlocker.tsx
+++ b/packages/react-router/src/useBlocker.tsx
@@ -9,11 +9,23 @@ type BlockerResolver = {
   reset: () => void
 }
 
-export function useBlocker(opts?: {
+type BlockerOpts = {
   blockerFn?: BlockerFn
-  condition: boolean | any
-}) {
-  const condition = opts?.condition ?? true
+  condition?: boolean | any
+}
+
+export function useBlocker(
+  blockerFnOrOpts?: BlockerFn | BlockerOpts,
+  condition?: boolean | any,
+) {
+  const { blockerFn, blockerCondition } = blockerFnOrOpts
+    ? typeof blockerFnOrOpts === 'function'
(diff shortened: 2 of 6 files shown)


packages/react-router/src/useBlocker.tsx
```diff
@@ -3,25 +3,62 @@ import { useRouter } from './useRouter'
...
+        allow: () => resolve(true),
+        deny: () => resolve(false),
+      })
+    })
+
+  const [promise, setPromise] = React.useState(createPromise)
+
+  React.useEffect(() => {
+    const blockerFnComposed = async () => {
+      const canNavigateSync = blockerFn()
```
schiller-manuel, line 34: what happens if `blockerFn` returns a `Promise`?

packages/react-router/src/useBlocker.tsx
```diff
@@ -3,25 +3,62 @@ import { useRouter } from './useRouter'
 import type { BlockerFn } from '@tanstack/history'
 import type { ReactNode } from './route'
 
+type Resolver = {
+  allow: () => void
+  deny: () => void
+}
+
```
schiller-manuel, line 10: naming: should we maybe consider prior art such as https://reactrouter.com/en/main/hooks/use-blocker ?

schiller-manuel: react-router also has a `state` param, do we need it?

schiller-manuel: I also kinda like `currentLocation` and `nextLocation`

examples/react/navigation-blocking/src/main.tsx
```diff
@@ -80,6 +78,29 @@ function Editor1Component() {
         onChange={(e) => setValue(e.target.value)}
         className="border"
       />
+      {showConfirm && (
```
SeanCassiere, line 81: I think the current setup for the "navigation blocking" example no longer adequately shows off the navigation blocking functionality, since it should demonstrate it using both `window.confirm` and a custom dialog.

We can change this after merge (if need be), but I'd rather go for a flow like this:

```
/
/using-window-confirm
/using-custom-ui
```

Where the `/using-window-confirm` and `/using-custom-ui` pages have a input field (like we have currently). When the fields have some value, it then requires user confirmation before allowing a navigation away.

packages/react-router/src/useBlocker.tsx
```diff
@@ -3,25 +3,62 @@ import { useRouter } from './useRouter'
 import type { BlockerFn } from '@tanstack/history'
 import type { ReactNode } from './route'
 
+type Resolver = {
```
SeanCassiere, line 6: Rename this to something that's more scoped to what it is like `BlockerFnResolver`, `BlockerResolver`, or something to that effect.

packages/react-router/src/useBlocker.tsx
SeanCassiere: I like the `.proceed()` and `.reset()` methods, and especially the `state` value. Though, I'd rename `state` to `status` that'd be more inline with Query.

SeanCassiere: Including `currentLocation` and `nextLocation` also sounds like a good idea!

It'll require the `useBlocker` hook to be aware of the `nextLocation`, which may require the router to keep track of the blockers (which it doesn't do now) since the next location is most easily available in `router.buildAndCommitLocation`.

I'm not opposed to going forward with this PR without `currentLocation` and `nextLocation`, and then adding it in with some strong tests in place.

examples/react/navigation-blocking/package.json
```diff
@@ -9,10 +9,8 @@
     "start": "vite"
   },
   "dependencies": {
-    "@tanstack/react-query": "^5.17.8",
```
SeanCassiere, line 12: I'd keep these packages in.

They were kept in, since in the future we may want to expand these examples to asynchronous-based actions.

SeanCassiere (changes requested):

packages/react-router/src/useBlocker.tsx
freshgiammi: `state` would be helpful: in the example case, it'd avoid the need to manually set the state (and also renders the blockerFn property unnecessary (even though it's still mandatory to provide, but maybe that can change?)

+1 on the currentLocation/nextLocation, but that probably requires changes in the history package ([here](https://github.com/TanStack/router/blob/main/packages/history/src/index.ts#L78)).

Also, considering the similarity of other hooks between the two libraries, it might be a good idea to end up with something that has a similar usage, so `.proceed()` and .`reset()` could be better suited.

packages/react-router/src/useBlocker.tsx
freshgiammi: We could await it, but wouldn't that risk blocking the inner promise? Technically, if we fire `.allow()` before we resolve the `blockerFn`, we could end up with the user being stuck in the page.

Adding a `state` property could make `blockerFn` unnecessary as the user would be able to fire an event on state change (effectively giving the user more control over the block).

examples/react/navigation-blocking/src/main.tsx
Balastrong: The example has two pages, I used the promise approach on a page and left the window.confirm in the other page.

Anyway it's indeed worth having two separate examples to keep one entirely with window.confirm and in the other we can showcase both approaches with the hook and with the component.

SeanCassiere: IMO, having two separate examples mostly just creates more steps for people to have to go through. Ideally, I want to find out about navigation blocking, I either go to its guide or checkout its example. Introducing another decision in that flow is unnecessary.

Having it in a single example, the exact same content and logic, just separated by the two individual routes (with custom UI and without), in my mind makes the most sense. Plus, it also saves on waiting time for waiting for each Stackblitz example to load up.

packages/react-router/src/useBlocker.tsx
Balastrong: Speaking of the state/status, do you manually want to handle it internally on useBlocker or should be something at a router level?

examples/react/navigation-blocking/src/main.tsx
Balastrong: Makes sense! Would you like me to refactor it on this PR or we're keeping this as is and revisit it later on another PR?

SeanCassiere: Probably for the time being have like `useState` toggle between `window.confirm` and the custom UI for simplicity's sake.

This can be made *prettier* afterwards.

packages/react-router/src/useBlocker.tsx
SeanCassiere: Currently, since blockers have no effect on the actual Router, and directly interface with the history package, I'd keep this state local to the `useBlocker` hook and component.

Balastrong: With status exposed seems like `blockerFn` might become optional, I just pushed a proposal. What do you think?

packages/react-router/src/useBlocker.tsx
```diff
@@ -3,25 +3,70 @@ import { useRouter } from './useRouter'
...
-  blockerFn: BlockerFn,
-  condition: boolean | any = true,
-): void {
+type BlockerResolver = {
+  status: 'idle' | 'blocked'
+  proceed: () => void
+  reset: () => void
+}
+
+export function useBlocker(opts?: {
```
SeanCassiere, line 12: Switching to object syntax, makes this a breaking change. I think we may have to stick with the existing API surfaces.

SeanCassiere (changes requested):

packages/react-router/src/useBlocker.tsx
SeanCassiere: > With status exposed seems like `blockerFn` might become optional, I just pushed a proposal and updated the example. What do you think?

🚨 The object syntax would make this a breaking change.

packages/react-router/src/useBlocker.tsx
SeanCassiere: We follow semver and this'd be a breaking change to the user.

Considering we don't want to introduce compexity with function overloads, I think we'd have to go with the alternatives here (like `useBlocker(() => void, value)`), or simply export a new useBlocker hook altogether (with a new name) and preserve the old hook as-is.

@schiller-manuel thoughts on this?

schiller-manuel: we must avoid breaking changes.
instead, I suggest supporting both signatures like this (pseudocode):

```tsx
function useBlocker(
blockerFnOrOpts: BlockerFn | BlockerOpts,
condition?: boolean | any)
{
if (typeof blockerFnOrOpts === 'function') {
// function was passed
}
else {
// opts were passed
}
}
```

then deprecate the old syntax and get rid of it with v2

Balastrong: I updated the API, I don't know why now tests are failing on the start package though... on local tsc doesn't complain

SeanCassiere: Rebasing with main should fix this now.

packages/react-router/src/useBlocker.tsx
```diff
@@ -3,25 +3,83 @@ import { useRouter } from './useRouter'
...
+  proceed: () => void
+  reset: () => void
+}
+
+type BlockerOpts = {
+  blockerFn?: BlockerFn
+  condition?: boolean | any
+}
+
 export function useBlocker(
```
SeanCassiere, line 17: Maybe this needs to be a function overload?

```tsx
export function useBlocker(blockerFn: BlockerFn, condition?: boolean | any): BlockerResolver;
export function useBlocker(blockerOpts: BlockerOpts): BlockerResolver {
  // ...
}
```

And probably should depreciate the old syntax.

SeanCassiere (approved):

packages/react-router/src/useBlocker.tsx
Balastrong: Done! I also updated the docs with the new API

schiller-manuel:
can you please move the API part into a dedicated API reference page similar to the existing hook docs like https://github.com/TanStack/router/blob/main/docs/framework/react/api/router/useSearchHook.md ?
and then link to that API reference from the guide

Balastrong:
> can you please move the API part into a dedicated API reference page similar to the existing hook docs like https://github.com/TanStack/router/blob/main/docs/framework/react/api/router/useSearchHook.md ? and then link to that API reference from the guide

Done! But I replicated the examples on both pages as I couldn't think of a guide without the example or an API reference without examples.
Let me know if you think we shouldn't duplicate it and you want me to remove it from one of the two pages.

packages/react-router/src/useBlocker.tsx
```diff
@@ -3,25 +3,93 @@ import { useRouter } from './useRouter'
...
+  condition?: boolean | any
+}
+
+export function useBlocker(blockerFnOrOpts?: BlockerOpts): BlockerResolver
+
+/**
+ * @deprecated Use the BlockerOpts object syntax instead
+ */
+export function useBlocker(
+  blockerFnOrOpts?: BlockerOpts,
```
schiller-manuel, line 23: should the type be BlockerFn?

schiller-manuel (changes requested):

packages/react-router/src/useBlocker.tsx
Balastrong: This is the deprecated interface so yeah, should be the old function.

schiller-manuel:
thank you for your hard work and patience during the review 😁

Balastrong:
Thank you for your guidance! 🙌

packages/react-router/src/useBlocker.tsx
```diff
@@ -3,25 +3,93 @@ import { useRouter } from './useRouter'
...
+      }))
+      const canNavigateAsync = await promise
+
+      setPromise(createPromise)
+
+      return canNavigateAsync
+    }
+
+    return !blockerCondition ? undefined : history.block(blockerFnComposed)
+  }, [blockerFn, blockerCondition, history, promise])
```
leoyli, line 77: @Balastrong, thank you for the great work!

I noticed that at line 56 you used a `useState` to hold the promise, even there are nothing to be blocked. This can lead to an unexpected bug in the development mode, where React will run the some hooks twice to detect potential development issues, meaning there is a risk that a new promise is created while the previous promise never resolved. 

And in fact, there are indeed bugged because of the async `blockerFnComposed` function in the effect referencing the "old" promise (tested in React 19 beta).

Therefore, I suggest to only create a promise when there are something really to be promised, here is the suggested changes:


```ts
  React.useEffect(() => {
    async function blockerFnComposed() {
      // If a function is provided, it takes precedence over the promise blocker
      if (blockerFn) {
        return await blockerFn();
      }

      const promise = new Promise((resolve) => {
        setResolver({
          status: "blocked",
          proceed: () => resolve(true),
          reset: () => resolve(false),
        });
      });

      const canNavigateAsync = await promise;
      setResolver({
        status: "idle",
        proceed: () => {},
        reset: () => {},
      });

      return canNavigateAsync;
    }

    return !blockerCondition ? undefined : history.block(blockerFnComposed);
  }, [blockerFn, blockerCondition, history]);
  ```

schiller-manuel: @leoyli
can you please create a PR to fix this issue?
ideally with a test 😁

Issue #1659, opened 2024-05-23
Overlapping Keys Cause Type Issues Using Experimental Intersection


rajbirjohar:
### Describe the bug

We have three routes:

`index.tsx` 

```tsx
import { createFileRoute } from "@tanstack/react-router";
import { z } from "zod";

const indexFilters = z.object({
  search: z.string().catch("").optional(),
});

const IndexPage = () => {
  return <div>Index Page</div>;
};

type IndexPageProps = z.infer<typeof indexFilters>;

export const Route = createFileRoute("/")({
  component: IndexPage,
  validateSearch: (search: Record<string, unknown>): IndexPageProps =>
    indexFilters.parse(search),
});
```

`about/index.tsx`

```tsx
import { createFileRoute } from "@tanstack/react-router";
import { z } from "zod";

const aboutFilters = z.object({
  limit: z.number().catch(10).optional(),
  materialTypes: z.array(z.string()).catch([]).optional(),
  // search: z.string().catch("").optional(),
});

const AboutPage = () => {
  return <div>About Page</div>;
};

export const Route = createFileRoute("/about/")({
  component: AboutPage,
  validateSearch: (
    search: Record<string, unknown>
  ): z.infer<typeof aboutFilters> => aboutFilters.parse(search),
});
```

`contact/index.tsx`

```tsx
import { createFileRoute } from '@tanstack/react-router'

export const Route = createFileRoute('/contact/')({
  component: () => <div>Hello /contact/!</div>
})
```

`components/Search/index.tsx`

```tsx
import { useSearch } from "@tanstack/react-router";

const Search = () => {
  const { search, limit, materialTypes } = useSearch({
    strict: false,
    experimental_returnIntersection: true,
  });

  return (
    <div>
      <h1>Search</h1>
      <pre>{JSON.stringify(search, null, 2)}</pre>
    </div>
  );
};

export default Search;
```

And a component that uses the `useSearch` hook with `experimental_returnIntersection`. If both `index.tsx` and `about/index.tsx` have overlapping schema fields, it causes the other non-overlapping fields to throw a type error saying that it does not belong on:

```tsx
Property 'materialTypes' does not exist on type 'Partial<Omit<Partial<Reducer<Reducer<unknown, RootSearchSchema, UnionizeCollisions<unknown, RootSearchSchema>>, { search?: string | undefined; }, UnionizeCollisions<...>>>, "__TRootSearchSchema__">>'.
```

[...]

Issue #1217, opened 2024-02-24
No type safety on useSearch({ strict: false })


RobertOstermann:
### Describe the bug

The type safety is not working as expected for `useSearch` or `useParams` when using `strict: false` in the options.

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-nan1d9?file=src%2Froutes%2Fdashboard.users.tsx

### Steps to Reproduce the Bug or Issue

1. Go to the Kitchen Sink File Base Example (any example with search params would work)
2. Change the `route.useSearch` to `useSearch({ strict: false })`
3. Add a `useParams({ strict: false })`
4. The types are not showing for either function as expected

### Expected behavior

With `strict: false` the type safety should be loosened to show all available types.

### Screenshots or Videos

`useSearch({ strict: false })` no type safety

![image](https://github.com/TanStack/router/assets/46828475/8038e518-97ea-47b6-8632-3b2a9f4476cc)

`useSearch({ strict: false, experimental_returnIntersection: true }) does not work. This does fix the same issue I see with `useParams` though.

![image](https://github.com/TanStack/router/assets/46828475/01b113a0-d60d-4a4f-9d62-c1115ff5c5dd)

`useParams({ strict: false })` no type safety

![image](https://github.com/TanStack/router/assets/46828475/65b8d903-df7b-48b7-ad60-d072fc2a03c0)

`useParams({ strict: false, experimental_returnIntersection: true })` fixes the issue for `useParams`

![image](https://github.com/TanStack/router/assets/46828475/991d251d-452a-451c-9e61-fff1aef4fc66)

### Platform

tanstack router version: 1.16.6
typescript: 5.3.3 (on my local, not sure what stackblitz is on)

schiller-manuel:
> `useParams({ strict: false })` no type safety

This is incorrect, this is type safe. You get back the **union** of all possible search parameters.

see https://stackblitz.com/edit/tanstack-router-kws5zj?file=src%2Froutes%2Fdashboard.users.tsx

```tsx
const search: {} | {} | {} | {
    showNotes?: boolean | undefined;
    notes?: string | undefined;
} | {
    usersView?: {
        sortBy?: "id" | "name" | "email" | undefined;
        filterBy?: string | undefined;
    } | undefined;
} | {
    ...;
} | {
    ...;
} | {
    ...;
}
```

> `useSearch({ strict: false, experimental_returnIntersection: true })` does not work.

Indeed, on stackblitz it does not work.
[...]

PR #1664, opened 2024-05-24
fix(react-router): merge params, context, search and loaderData with `strict: false`

chorobin:
- `useRouteContext`, `useParams`, `useSearch`, `useLoaderData` will merge objects when `strict: false`. Conflicting properties are resolved as a union
- Mapped type is used instead of intersections
- Removed experimental flag for `experimental_returnIntersections`. We no longer use intersections for this which improves the TS performance as we handle the conflicts ourselves


schiller-manuel:
fixes #1659
fixes #1217

packages/react-router/src/route.ts
```diff
@@ -71,10 +71,8 @@ export type RouteOptions<
   TRouterContext = AnyContext,
   TAllContext = AnyContext,
   TLoaderDeps extends Record<string, any> = {},
-  TLoaderDataReturn = unknown,
-  TLoaderData = [TLoaderDataReturn] extends [never]
-    ? undefined
```
schiller-manuel, line 76: why can't we have `undefined` here anymore?
this is the return type in case the loader e.g. only throws a redirect

packages/react-router/src/route.ts
```diff
@@ -1069,10 +1067,8 @@ export function createRootRouteWithContext<TRouterContext extends {}>() {
       ? RouteContext
       : TRouteContextReturn,
     TLoaderDeps extends Record<string, any> = {},
-    TLoaderDataReturn = unknown,
-    TLoaderData = [TLoaderDataReturn] extends [never]
-      ? undefined
```
schiller-manuel, line 1074: same question here about `undefined`

packages/react-router/src/route.ts
```diff
@@ -442,7 +440,7 @@ export type ResolveAllContext<
 export type ResolveLoaderData<TLoaderDataReturn> = [TLoaderDataReturn] extends [
   never,
 ]
-  ? undefined
```
schiller-manuel, line 445: can we still have `undefined`?

packages/react-router/src/route.ts
```diff
@@ -71,10 +71,8 @@ export type RouteOptions<
   TRouterContext = AnyContext,
   TAllContext = AnyContext,
   TLoaderDeps extends Record<string, any> = {},
-  TLoaderDataReturn = unknown,
-  TLoaderData = [TLoaderDataReturn] extends [never]
-    ? undefined
-    : TLoaderDataReturn,
+  TLoaderDataReturn = {},
+  TLoaderData = [TLoaderDataReturn] extends [never] ? {} : TLoaderDataReturn,
```
schiller-manuel, line 75: why don't we use `ResolveLoaderData` here?

packages/react-router/src/route.ts
```diff
@@ -1069,10 +1067,8 @@ export function createRootRouteWithContext<TRouterContext extends {}>() {
       ? RouteContext
       : TRouteContextReturn,
     TLoaderDeps extends Record<string, any> = {},
-    TLoaderDataReturn = unknown,
-    TLoaderData = [TLoaderDataReturn] extends [never]
-      ? undefined
-      : TLoaderDataReturn,
+    TLoaderDataReturn = {},
+    TLoaderData = [TLoaderDataReturn] extends [never] ? {} : TLoaderDataReturn,
```
schiller-manuel, line 1071: why don't we use `ResolveLoaderData` here?

PR #1669, opened 2024-05-25
fix(react-router): only instantiate the confirmation promise when required in the `useBlocker` hook

leoyli:
Open this PR as @schiller-manuel encouraged based on my bug comment in  https://github.com/TanStack/router/pull/1649#discussion_r1614737330.

Since this is my first contribution to this repo, I only strive to make this dead simple.

I manually tested this code in my own project to see it is now working as expected. I checked the PR #1649 to see if I can amend a test suite. But I found it is tricky to add one because the promise is now only a transient instance. 

c.c. @Balastrong

### DIFF
--- a/packages/react-router/src/useBlocker.tsx
+++ b/packages/react-router/src/useBlocker.tsx
@@ -44,17 +44,6 @@ export function useBlocker(
     reset: () => {},
   })
 
-  const createPromise = () =>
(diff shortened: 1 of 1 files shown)


schiller-manuel:
can the examples be enhanced to show the original problem?
and then I can test your fix with the example

PR #1672, opened 2024-05-27
feat(start): add option to disable router vite plugin

iDarkLightning:
Add an option to disable the router vite plugin for code based routing


tannerlinsley:
We may want to point this at the RSC branch too depending on any potential conflicts.

SeanCassiere:
Could you resolve the merge conflicts here? Or feel free to just open a new PR?

I'd like to split the start specific plugin stuff wat from the router plugin logic.

Once the start stuff is pulled away, we can look at making the router plugin work with other bundlers.

iDarkLightning:
Well, initially the objective of the PR would be to add a configuration option to disable the router plugin so that you could use start with code-based routing. Doing this though, I realized that the `createServerFn` directive is a part of the router plugin, which led to [this conversation](https://discord.com/channels/719702312431386674/1240861342475227226/1244707868956164137) over on discord and the eventual conclusion there was to split the two into separate plugins.

I also think this would be good for the future as it would allow people to use the Router plugin independently of Start and vice-versa. Even though currently the Start plugin only does one thing, I think it's pretty easy to imagine a future where it would do more.

SeanCassiere:
> Well, initially the objective of the PR would be to add a configuration option to disable the router plugin so that you could start with code-based routing. Doing this though, I realized that the `createServerFn` directive is a part of the router plugin, which led to [this conversation](https://discord.com/channels/719702312431386674/1240861342475227226/1244707868956164137) over on discord and the eventual conclusion there was to split the two into separate plugins.
> 
> I also think this would be good for the future as it would allow people to use the Router plugin independently of Start and vice-versa. Even though currently the Start plugin only does one thing, I think it's pretty easy to imagine a future where it would do more.

Understood, thanks! I seem to have missed that context in the start channel 😅

If you are able to resolve the merge conflicts or open simply a new PR 😅, I'd be happy to merge it in 😊. If not it's cool, I'm anyways looking at touching the router-plugin this week.

PR #1675, opened 2024-05-28
[Discussion] using Apollo Client's `queryPreloader` in TanStack router loaders

phryneas:
Please don't merge, this is just a PR because it allows me to easily highlight code changes to the default example :)

We talked about support for Apollo Client's `queryPreloader` in TanStack router loaders yesterday, and I promised you a write-up what we need.

Having created this example, I'm not even sure if we need anything more at this point 😅 

Generally, we need these things:

* Hook into serialization of `loader` function return values
* Hook into deserialization of `loader` function values as soon as possible
  * Best would be directly after the loader data made it over the wire, but at the latest when `useLoaderData` is called
  * This deserialization needs access to the current Apollo Client instance and deserialzation triggers a side effect
  * This could be done by closing over `client` in `createRouter`, but it would be even better if the transformer had access to the current `context` in some way - this way, `router.update` could change the client instance without things breaking.

So, this is much less of a feature request than I initially imagined - more of a bunch of questions at this point:

* Can `transformer` return objects that contain `Promise` intances?
* What about streams? That would allow us to support GraphQL features like `@defer`, where the response arrives in chunks, so this would be *really* cool!
* When does deserialization happen? Immediately when data arrives?
* Do you see any chance of exposing `context` to the `transformer`?
* A bit of a quality-of-life thing: could you imagine to allow an array of transformers in `transformer`? Could be useful if more libraries start shipping transformers :)

Adding @jerelmiller to the discussion, too :)

### DIFF
--- a/examples/react/basic-ssr-streaming-file-based/package.json
+++ b/examples/react/basic-ssr-streaming-file-based/package.json
@@ -12,14 +12,15 @@
     "debug": "node --inspect-brk server"
   },
   "dependencies": {
+    "@apollo/client": "^3.10.4",
     "@tanstack/react-router": "^1.34.3",
-    "@tanstack/start": "^1.34.3",
     "@tanstack/router-devtools": "^1.34.4",
     "@tanstack/router-vite-plugin": "^1.34.1",
-    "redaxios": "^0.5.0",
+    "@tanstack/start": "^1.34.3",
     "get-port": "^7.0.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "redaxios": "^0.5.0",
     "superjson": "^2.2.1"
   },
   "devDependencies": {
--- a/examples/react/basic-ssr-streaming-file-based/src/apollo.ts
+++ b/examples/react/basic-ssr-streaming-file-based/src/apollo.ts
@@ -1,0 +1,9 @@
+import { ApolloClient, InMemoryCache } from '@apollo/client'
+
+export function makeClient() {
+  const client = new ApolloClient({
+    uri: 'https://flyby-router-demo.herokuapp.com/',
+    cache: new InMemoryCache(),
+  })
+  return client
+}
--- a/examples/react/basic-ssr-streaming-file-based/src/entry-client.tsx
+++ b/examples/react/basic-ssr-streaming-file-based/src/entry-client.tsx
@@ -2,16 +2,8 @@ import * as React from 'react'
 import ReactDOM from 'react-dom/client'
 
 import { StartClient } from '@tanstack/start'
-import { ApolloProvider } from '@apollo/client'
 import { createRouter } from './router'
-import { makeClient } from './apollo'
 
-const client = makeClient()
-const router = createRouter(client)
+const router = createRouter()
 
-ReactDOM.hydrateRoot(
-  document,
-  <ApolloProvider client={client}>
-    <StartClient router={router} />
-  </ApolloProvider>,
-)
+ReactDOM.hydrateRoot(document, <StartClient router={router} />)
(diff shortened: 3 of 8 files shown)


lithdew:
Things I believe to be true from reading the source code:

1. AFAIK `transformer` can return Promise/Stream objects should the transformer support it. It must be able to hydrate/dehydrate them. Examples of transformers that can do this is i.e. seroval. If the promises are suspended in a React component then React takes care of the serialization/deserialization - see the source code for defer() and useAwaited() where they throw a promise (but add some additional state to aid with keeping track of whether their state is pending/errored/success).
2. Same goes for streams.
3. Happens inside of <Scripts/> where there is a <DehydrateRouter/> which calls the provided `transformer`. Basically while rendering the client.
4. If Apollo has its own transformer for hydration/dehydration, the router provides `hydrate` and `dehydrate` options. You have access to the context while constructing the router.
5. The types for "piping" transformers isn't that nice so I don't think it would be implemented.

Also looking at the changes in the PR:

1. To wrap the router tree with context providers, the router provides `InnerWrap` and `Wrap` options.

lithdew:
I'm not too familiar with how Apollo Client handles hydration/dehydration, but from a quick cursory look at the docs:

Server-side:
```ts
// Add this import to the top of the file
import { getDataFromTree } from "@apollo/client/react/ssr";

// Replace the TODO with this
getDataFromTree(App).then((content) => {
  // Extract the entirety of the Apollo Client cache's current state
  const initialState = client.extract();

  // Add both the page content and the cache state to a top-level component
  const html = <Html content={content} state={initialState} />;

  // Render the component to static markup and return it
  res.status(200);
  res.send(`<!doctype html>\n${ReactDOM.renderToStaticMarkup(html)}`);
  res.end();
});
```

Client-side:
```ts
const client = new ApolloClient({
  cache: new InMemoryCache().restore(window.__APOLLO_STATE__),
  uri: 'https://example.com/graphql'
});
```

If this method also dehydrates the client's cache to support streaming as well, you could call `client.extract()` in `hydrate() { ... }` and `new InMemoryCache.restore(stringFromHydrate)` in `dehydrate() { ... }`.

Otherwise, could you link to any docs or some source code on how to get hydration via. streaming working with Apollo Client?

phryneas:
> Otherwise, could you link to any docs or some source code on how to get hydration via. streaming working with Apollo Client?

There is no extra hydration beyond what I showed here - hydration would be a side effect of the `queryRef` parsing, only hydrating that small part of the store - it would all be hidden inside `transform.parse`.

These old patterns of "dehydrate the full store" and "hydrate the full store" don't work in `renderToStream` scenarios anymore, since the store is already created (and interactive) in the browser long before it will be filled on the server - so instead we have to selectively hydrate pieces as they make it over, in this case in the form of a `QueryRef` object.

We have alternative transport mechanisms for the suspenseful `useSuspenseQuery`/`useBackgroundQuery`/`useReadQuery` hooks, but these are [quite complex to set up if not directly supported by a framework](https://deploy-preview-11807--apollo-client-docs.netlify.app/ssr-and-rsc/custom-streaming-ssr), so in a first step I'd try to concentrate on the `loader` scenario, before we get into that with TanStack Start :)

phryneas:
Oh, this can be closed regarding newer PRs.

mirague:
> Oh, this can be closed regarding newer PRs.

Can you clarify what you mean? Was there another PR addressing this?

phryneas:
@mirague I have all the changes I need in #2698 and after we get that merged, we can cat this released: https://github.com/apollographql/apollo-client-nextjs/blob/fd8d289791d62f3d644fa6cbb28f827dd31a370d/packages/tanstack-start/README.md

PR #1696, opened 2024-06-02
feat: merge unions for `search` and `params` for `Link` when `from` and `to` are loose

chorobin:
`Link`, `useNavigate` and `redirect` now merge unions for `search` and `params` under the following circumstances:
- When `from` is not used, previous `search` and `params` is a merged type
- When `from` and `to` is not used `search` and `params` is a merged type
- When `from` is not used and `to` is `.` or `..`
- Introduced a few optimizations for merging unions
- Introduced more correct optional/required behaviour for `params` and `search`

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -8,9 +8,10 @@ import type { AnyRouter, ParsedLocation } from '.'
 import type { HistoryState } from '@tanstack/history'
 import type { AnyRoute, RootSearchSchema } from './route'
 import type {
+  AllParams,
   CatchAllPaths,
-  ParseRoute,
-  ParseRouteWithoutBranches,
+  FullSearchSchema,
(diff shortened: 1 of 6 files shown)


packages/react-router/src/link.tsx
```diff
@@ -275,17 +272,19 @@ type PostProcessParams<
 type ResolveFromParamType<TParamVariant extends ParamVariant> =
   TParamVariant extends 'PATH' ? 'allParams' : 'fullSearchSchema'
 
+type ResolveFromAllParams<
```
schiller-manuel, line 275: is this the same as `ResolveAllToParams`?

if yes I would not duplicate this

chorobin: They're different because of `fullSearchSchemaInput/fullSearchSchema`

Issue #1665, opened 2024-05-24
`Route.fullPath` is typed incorrectly when using path params


Tiedye:
### Describe the bug

The type of `Route.fullPath` is more or less the path that is passed to the create<xd>Route function, but during runtime it is the current path of the page when `Route,fullPath` references a path with a paramater, which included the sub path, this makes it useless as a `from` parameter when used in a layout route. 

### Your Example Website or App

https://stackblitz.com/edit/github-jhyjxh?file=src%2Froutes%2F%24subsection.tsx

### Steps to Reproduce the Bug or Issue

Click on sub, click on the links and see every navigation is relative to the current url instead of the expected Route.fullPath

### Expected behavior

Not that, the same as for static routes, should behave like this:
https://stackblitz.com/edit/github-jhyjxh-4xvfsf?file=src%2Froutes%2Fsubsection.tsx

### Platform

- OS: Windows
- Browser: Firefox
- Version: 128a

schiller-manuel:
The issue at hand is not that `Route.fullPath` is incorrectly typed, instead the following happens:

At runtime, router does **not** respect `from`, instead it always uses the current location when building relative links.

We had discussed this before, we even had a PR merged for this (see https://github.com/TanStack/router/pull/1595) but reverted since we were not confident about possible side effects.

cc @TkDodo

Tiedye:
Wow, I wouldn't've guessed that

Tiedye:
How is the behaviour of the demo I posted explained?  If I remove the path param it works fine

schiller-manuel:
which path params? can you please post a separate working example?

chorobin:
> I think the issue is simply that if you are on `/$subsection/section-a` and you make a link to `./section-b`, you will wind up on `/$subsection/section-a/section-b`, no matter if you provide `from` or not. TypeScript will allow it because `from` is defined and points to `$subsection`, so we have a discrepancy between type-level and runtime behaviour.
> 
> This should be the same no matter if `$subsection` is used or something static. If that's not the case, please show a separate reproduction.
> 
> So I see a couple of ways forward:
> 
> 1. make runtime behave like type-level and take `from` into account when building navigations. This is similar to what I've tried before but we reverted the merge.
[...]

PR #1703, opened 2024-06-03
fix: `from` was being ignored when the pathname included params or a pathless route

chorobin:
When `from` included a `path` which had `params` it was being ignored when the location was being built. Same with a layout route.

### DIFF
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -1102,6 +1102,8 @@ export class Router<
     ): ParsedLocation => {
       let fromPath = this.latestLocation.pathname
       let fromSearch = dest.fromSearch || this.latestLocation.search
+      const fromRoute =
+        this.routesByPath[dest.from as keyof typeof this.routesByPath]
 
       const fromMatches = this.matchRoutes(
         this.latestLocation.pathname,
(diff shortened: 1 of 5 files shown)


packages/react-router/tests/link.test.tsx
```diff
@@ -1090,6 +1090,138 @@ describe('Link', () => {
...
+      getParentRoute: () => postsRoute,
+      path: '$postId',
+      component: PostComponent,
+    })
+
+    const DetailsComponent = () => {
+      return (
+        <>
+          <h1>Details!</h1>
+          <Link from="/posts/$postId" to="./info">
```
TkDodo, line 1161: this is exactly what I wanted to get to work, awesome 🚀

can you maybe add two more cases:

- `<Link from="/posts/$postId" to="/posts/$postId/info">`
  - this is what we are using right now, just want to make sure it doesn't break
- `<Link from="/posts/$postId/details" to="../info">`
  - just to see if `../` is also working

packages/react-router/src/router.ts
```diff
@@ -825,6 +825,10 @@ export class Router<
     return this.routesById as Record<string, AnyRoute>
   }
 
+  get looseRoutesByPath() {
+    return this.routesByPath as Record<string, AnyRoute>
+  }
```
TkDodo, line 830: this becomes a public function exposed to users. I don't think we'd want that. Since there's only one usage, why not just inline it?

TkDodo (approved):
wow, the test coverage is amazing 👏

packages/react-router/src/router.ts
chorobin: Inlined

chorobin:
fixes #1665

levrik:
Why was this reverted? I can't find any information about that.

TkDodo:
it broke stuff. we're working on it

Issue #1670, opened 2024-05-27
Undefined context after redirect


jakst:
### Describe the bug

Seems like the issue with undefined context is not fully resolved yet. When redirecting I'm seeing undefined context.

Some criteria to trigger the issue
* `defaultPreload: 'intent'` has to be set in the router
* The `beforeLoad` function in the layout that provides the context return a promise
* The redirect is triggered in a loader fn after a delay. 500ms seems to always trigger the issue, 100ms does not trigger the issue. I assume it has to be long enough so you click the link before the preload has finished.

### Your Example Website or App

https://stackblitz.com/edit/github-xjvg3f?file=src%2Froutes%2Fabout.tsx

### Steps to Reproduce the Bug or Issue

1. Open repro project in stackblitz
2. Click the `Click me to crash` link
3. Observe crash caused by missing context

### Expected behavior

I expect the context to be present after redirect

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

### Additional context

These issues are related, but I'm not certain if they are exactly the same so I created a new issue. https://github.com/TanStack/router/issues/1656 https://github.com/TanStack/router/issues/1650

pruett:
Definitely experiencing a similar issue.

It shows up when I `redirect` from a parent (layout) route's `beforeLoad` function. I moved the `redirect` to the child's `beforeLoad` and `context` is undefined in *that* same child's `loader`

pruett:
Upon further inspection, it looks to be directly related to using `router.invalidate()` to re-trigger the route updates. Otherwise, it looks like the `throw redirect({...})` in the `beforeLoad` is working as expected. Only when it's triggered via `router.invalidate()` does it seem to not pass `context` correctly

TkDodo:
I have a very simple reproduction for this issue as well:

https://stackblitz.com/edit/tanstack-router-b9sx1q?file=src%2Froutes%2Fabout.tsx

- `/about` route redirects with `throw redirect` to `/about-new`
- `/about-new` wants to read the context in `loader`, but it receives `undefined`

The only difference from the stock examples is that I have preloading turned off. If I turn on `defaultPreload: 'intent'`, the issue disappers.

schiller-manuel:
> things seem to work fine for me if I use `navigate` instead of `throw redirect`. Is there a meaningful difference between the two? `navigate` is passed as a prop into `beforeLoad` ...

[...]

Issue #1656, opened 2024-05-23
Context is undefined when search validation fails


geoff-harper:
### Describe the bug

I've discovered that the missing context bug isn't fully fixed in `1.33.7`. Also sorry for sort of spamming the other closed issue, I thought I'd open a new one.

I'm not completely sure of what's triggering it or if there's more cases but these are the steps in my own app

1. open a route with no search validation (not sure this part is relevant)
2. directly enter a route into the address bar that has search validation, but don't include any search params in the address
3. navigate to the route, context should be undefined

My schema catches and fills values like the docs suggest. In step 3 when the error component is fired, it does still put the caught values into the url. I broke the `validateSearch` method into pieces instead of directly returning a schema, and it specifically fails when anything is returned in the method.

### Your Example Website or App

https://stackblitz.com/edit/github-spzf56-us3fam?file=src%2Froutes%2F_layout.index.tsx

### Steps to Reproduce the Bug or Issue

1. Go to the about route
2. Go to index route through address bar, omit search params

### Expected behavior

Context should be defined in all cases

### Platform

- Browser: all
- Version: 1.33.7

mirague:
I found this in `v1.34.9` as well, causing `context` to become `undefined` if the route that throws the redirect has search params that are being validated through `validateSearch`, and the validation fails/throws (e.g. with `zod`). 

Here's a complete repro:
https://stackblitz.com/edit/tanstack-router-qbczsv?file=src%2Froutes%2Findex.tsx

schiller-manuel:
this will be fixed by #1907 

reproducer with the pre-release package: https://stackblitz.com/edit/tanstack-router-93gdrp?file=src%2Froutes%2Findex.tsx

Issue #1650, opened 2024-05-22
Undefined context after redirect with wrapInSuspense at root route (introduced in 1.33.6)


Zebeqo:
### Describe the bug

Undefined context after redirect with wrapInSuspense at root route.

It used to work fine in version 1.33.4.

### Your Example Website or App

https://stackblitz.com/edit/github-jr3nax-t75gmj?file=src%2Froutes%2F__root.tsx

### Steps to Reproduce the Bug or Issue

1. See the log: `context: undefined`
2. Remove `wrapInSuspense: true` and refresh
3. See the log: `context: {"test":"abc"}`

### Expected behavior

See the log: `context: {"test":"abc"}` at step 1

### Platform

- OS: macOS
- Browser: Chrome
- Version: 125.0

Zebeqo:
> can you please check which version introduced this? 1.33.5 most likely did not since it affected @tanstack/start and not the router.

It's likely that the issue was introduced in version 1.33.6. I can confirm that version 1.33.4 was working fine, but the problem appears in version 1.33.6. As for version 1.33.5, it doesn't appear to be available on npm.

1.33.4: https://stackblitz.com/edit/github-jr3nax-xxb2ka?file=package.json
1.33.6: https://stackblitz.com/edit/github-jr3nax-uphkrn?file=package.json

schiller-manuel:
then it is caused by this PR

https://github.com/TanStack/router/pull/1611

@SeanCassiere 
@freshgiammi

freshgiammi:
Yeah, looks like this is working when my change is reverted, but I'm not sure reverting the change is the right fix. 
My patch merely fixes support for `wrapInSuspense` which got lost when some SSR changes were brought in (where the root route got forced to not suspend), so maybe this was broken before too. Seems we just keep bumping into ctx issues.

Reverting this change restores context after the redirect https://github.com/TanStack/router/commit/e19f7e8703bdc8d6ccbaf3399c82527aabdc56c2#diff-2ac234f610fa0e0c2146252c10a5266af9c5b6d0b63659a0fa4c9f2c4472596a

I wonder if the root cause lays elsewhere 🤔

PR #1708, opened 2024-06-05
test(react-router): check the route context at `beforeLoad`, `loader`, and `component`

SeanCassiere:
This is just the start to making sure that route context is correctly tested.

Route context should be checked at 3 distinct points:
* In the `beforeLoad` callback function.
* In the `loader` callback function.
* In the `component` rendered on the screen.

The idea here is that each scenario context scenario we come up with should be checked at all three points to make sure the route context experience is consistent at all levels.

Currently, these tests check for the following:
* Context available at the root route.
* Context available at the index route.
* Context modified at the root route, is correctly available at the index route.

This can be used as a base for building the tests to fix the context issues we are currently facing.

### DIFF
--- a/packages/react-router/tests/routeContext.test.tsx
+++ b/packages/react-router/tests/routeContext.test.tsx
@@ -1,0 +1,301 @@
+import React from 'react'
+import '@testing-library/jest-dom/vitest'
+import { afterEach, describe, expect, test, vi } from 'vitest'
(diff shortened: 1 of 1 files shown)


SeanCassiere:
These changes should be a good base for building the tests *and* eventual fixes for #1650, #1656, and #1670.

PR #1711, opened 2024-06-05
fix: ensure that trailingSlashes don't break path matching

TkDodo:


packages/react-router/src/link.tsx
```diff
@@ -615,8 +620,10 @@ export function useLinkProps<
   const isActive = useRouterState({
     select: (s) => {
       // Compare path/hash for matches
-      const currentPathSplit = s.location.pathname.split('/')
-      const nextPathSplit = next.pathname.split('/')
+      const currentPathSplit = removeTrailingSlash(s.location.pathname).split(
+        '/',
+      )
+      const nextPathSplit = removeTrailingSlash(next.pathname).split('/')
```
TkDodo, line 626: this is the fix: removing trailing slashes before splitting because a trailing slash results in an extra `''` entry

packages/react-router/tests/link.test.tsx
```diff
@@ -1635,6 +1635,157 @@ describe('Link', () => {
...
+    fireEvent.click(await screen.findByRole('link', { name: 'To first post' }))
+
+    const postsLink = await screen.findByRole('link', { name: 'Posts' })
+
+    expect(router.state.location.pathname).toBe('/posts/id1')
+    expect(postsLink).toHaveAttribute('data-status', 'active')
+    expect(postsLink).toHaveAttribute('aria-current', 'page')
+  })
+
+  test('when on /posts/$postId/, Links to /posts should be active (trailingSlash: always)', async () => {
```
TkDodo, line 1714: this test fails without my fix

TkDodo:
@hansyulian this slightly relates to your PR:
- https://github.com/TanStack/router/pull/1372

because `fixTrailingSlash` wasn't taking the router's `basepath` into account. We could probably have more tests with different basepaths

packages/react-router/src/utils.ts
```diff
@@ -309,8 +309,8 @@ export function escapeJSON(jsonString: string) {
     .replace(/"/g, '\\"') // Escape double quotes
 }
 
-export function removeTrailingSlash(value: string): string {
-  if (value.endsWith('/') && value !== '/') {
+export function removeTrailingSlash(value: string, basepath: string): string {
+  if (value.endsWith('/') && value !== '/' && value !== `${basepath}/`) {
```
schiller-manuel, line 313: `value !== '/'` should probably be removed since this is the case where `basepath` is empty (default)

TkDodo: I thought so too, but basepath is never empty - it's set to `/` if not provided by users. But if provided, it can be without trailing slash. A bit confusing 😅

TkDodo: so it seems that resolving a basepath works with a replace:

https://github.com/TanStack/router/blob/4327ed75d937ff4ccdf3bfa5cc4fdc7df46ef099/packages/react-router/src/path.ts#L69-L70

so whatever the basepath is will be replaced with a `/`, and since it's `/` per default, this doesn't do anything in those cases. We could use the same thing here?

PR #1721, opened 2024-06-06
fix(react-router): take `from` into account when navigating with `params` or layouts

chorobin:
- Use flatRoutes so we can select branches even if they have index routes
- Do not override `from` redirect preloads. Users should specify this manually
- Ignore `from` in declarative route masks when building the location
- Add tests to cover bugs encountered

### DIFF
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -1103,13 +1103,25 @@ export class Router<
       let fromPath = this.latestLocation.pathname
       let fromSearch = dest.fromSearch || this.latestLocation.search
 
+      const fromRoute =
+        dest.from !== undefined
+          ? this.flatRoutes.find((route) => route.fullPath === dest.from)
+          : undefined
+
       const fromMatches = this.matchRoutes(
         this.latestLocation.pathname,
         fromSearch,
       )
 
-      fromPath =
-        fromMatches.find((d) => d.id === dest.from)?.pathname || fromPath
+      const fromMatch = fromMatches.find((d) => d.routeId === fromRoute?.id)
+
+      fromPath = fromMatch?.pathname || fromPath
+
+      invariant(
+        dest.from == null || fromMatch != null,
+        'Could not find match for from: ' + dest.from,
+      )
+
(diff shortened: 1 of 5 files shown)


packages/react-router/src/router.ts
```diff
@@ -1773,8 +1786,7 @@ export class Router<
                     preload: !!preload,
                     context: parentContext,
                     location,
-                    navigate: (opts: any) =>
-                      this.navigate({ ...opts, from: match.pathname }),
+                    navigate: (opts: any) => this.navigate({ ...opts }),
```
SeanCassiere, line 1789: Could avoid the spread here yea?

`navigate: (opts: any) => this.navigate(opts as any)`

packages/react-router/src/router.ts
```diff
@@ -2183,8 +2194,6 @@ export class Router<
     } catch (err) {
       if (isRedirect(err)) {
         return await this.preloadRoute({
-          fromSearch: next.search,
```
SeanCassiere, line 2186: What's the purpose of removing `from` in the route preload?

Or is this automatically being resolved in `buildLocation`?

packages/react-router/src/useNavigate.tsx
```diff
@@ -29,7 +29,6 @@ export function useNavigate<
     (options: NavigateOptions) => {
       return router.navigate({
         ...options,
-        from: options.to ? router.state.resolvedLocation.pathname : undefined,
```
SeanCassiere, line 32: For `useNavigate`, would `from` just serve to provide type-completion then?

packages/react-router/src/router.ts
chorobin: It seems like router was using the preloaded location for `from`. Problem with this is it might not match the current location and can cause the invariant to trigger. We decided its better if the user explicitly passes `from` if they want relative pathing

packages/react-router/src/useNavigate.tsx
chorobin: It's used at runtime as well. Its used to build the location. As long as it partially matches the current location then it can be used

PR #1727, opened 2024-06-09
ci: add typescript 5.5 to the test matrix

schiller-manuel:

PR #1731, opened 2024-06-11
ci: add typescript@5.50-beta to the test matrix

SeanCassiere:


SeanCassiere:
Looks like, even setting it to  `~5.50-beta` also doesn't seem to do it.

Ref #1727

PR #1736, opened 2024-06-12
feat(router-generator): automatically convert from lazy to non-lazy file route (and vice versa) upon file rename

schiller-manuel:


SeanCassiere (approved):
Should we be adding tests for this?

Admittedly it's a small change with a well-defined surface area. Upto you.

SeanCassiere (approved):
LGTM

PR #1738, opened 2024-06-12
feat(react-router): allow `useMatch` to not throw if match was not found

schiller-manuel:
in this case, `matchRoute` will return undefined

### DIFF
--- a/docs/framework/react/api/router/useMatchHook.md
+++ b/docs/framework/react/api/router/useMatchHook.md
@@ -20,7 +20,8 @@ The `useMatch` hook accepts a single argument, an `options` object.
 ### `opts.strict` option
 
 - Type: `boolean`
-- Optional - `default: true`
+- Optional
+- `default: true`
 - If `false`, the `opts.from` must not be set and types will be loosened to `Partial<RouteMatch>` to reflect the shared types of all matches.
 
 ### `opts.select` option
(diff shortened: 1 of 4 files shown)


chorobin:
To me this makes sense, I think `matchRoute` is not the same because it doesn't provide the same stuff a `useMatch` and its definitely possible users might want to conditionally render based on a match or not

packages/react-router/src/useMatch.tsx
```diff
@@ -13,8 +13,10 @@ export type UseMatchOptions<
   TStrict extends boolean,
   TRouteMatch,
   TSelected,
+  TThrow extends boolean,
 > = StrictOrFrom<TFrom, TStrict> & {
   select?: (match: TRouteMatch) => TSelected
+  shouldThrow?: TThrow
```
SeanCassiere, line 19: I'd rather this be `throw` to match that API surface with what we are already doing with `redirect()`.

docs/framework/react/api/router/useMatchHook.md
```diff
@@ -62,3 +70,17 @@ function Component() {
   // ...
 }
 ```
+
+### checking if a specific route is currently rendered
```
SeanCassiere, line 74: Capitalizing the heading.

packages/react-router/tests/useMatch.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+  RouterProvider,
+  createMemoryHistory,
+  createRootRoute,
+  createRoute,
+  createRouter,
+  useMatch,
+} from '../src'
+import type { RouteComponent, RouterHistory } from '../src'
+
+describe('useMatch', () => {
```
SeanCassiere, line 17: We also probably have a type test in a `useMatch.test-d.tsx`, to check for the return type when the throw flag is set to `false` vs what it is when `true`.

chorobin: Yes, for sure we need the type test aswell

chorobin (changes requested):
Add type tests and I think its all good

chorobin (approved):

PR #1746, opened 2024-06-13
fix(react-router): redirect with relative path

schiller-manuel:
tests needed

### DIFF
--- a/packages/react-router/tests/link.test.tsx
+++ b/packages/react-router/tests/link.test.tsx
@@ -2584,7 +2584,7 @@ describe('Link', () => {
       loader: () => {
         throw redirect({
           from: postRoute.fullPath,
-          to: '/../../login',
+          to: '../../login',
           search,
         })
       },
@@ -2683,7 +2683,7 @@ describe('Link', () => {
       beforeLoad: () => {
         throw redirect({
           from: postRoute.fullPath,
-          to: '/../../login',
+          to: '../../login',
           search,
         })
       },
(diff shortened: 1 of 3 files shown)


packages/react-router/tests/link.test.tsx
```diff
@@ -2527,6 +2527,204 @@ describe('Link', () => {
...
+
+    const postRoute = createRoute({
+      getParentRoute: () => postsRoute,
+      path: '$postId',
+      component: PostComponent,
+      validateSearch: () => ({ postPage: 0 }),
+      loader: () => {
+        throw redirect({
+          from: postRoute.fullPath,
+          to: '/../../login',
```
schiller-manuel, line 2587: why is there a leading slash?

packages/react-router/tests/link.test.tsx
```diff
@@ -2527,6 +2527,204 @@ describe('Link', () => {
...
+
+    const postRoute = createRoute({
+      getParentRoute: () => postsRoute,
+      path: '$postId',
+      component: PostComponent,
+      validateSearch: () => ({ postPage: 0 }),
+      beforeLoad: () => {
+        throw redirect({
+          from: postRoute.fullPath,
+          to: '/../../login',
```
schiller-manuel, line 2686: same here

packages/react-router/tests/link.test.tsx
chorobin: True! weird it works with it

SeanCassiere (commented):
I think adding into a relative navigation to another route in the same route level wouldn't hurt similar to the example @schiller-manuel posted on Discord.

Something along the lines of:

```
from: /wherever/
to: ./somewhere
```

PR #1767, opened 2024-06-17
deprecate(react-router): mark `navigate` as deprecated in the `BeforeLoadFn` type definition

0xturner:
After upgrading to a newer version of tanstack router, I noticed that calling `navigate` inside of the `beforeLoad` callback was no longer working as it was previously.

The docs seems to imply `throw redirect` is the new preferred pattern, so marking as deprecated to hopefully make the issue more obvious for others when upgrading.


SeanCassiere (approved):
Tests are passing, it's the CI pipeline being a bit flaky.

https://cloud.nx.app/cipes/667278fa27cd164ad3f2a4ec

PR #1781, opened 2024-06-18
docs: add TS performance guidance with prefetching

chorobin:
When using TanStack Query or other client side caches, you might want to use `loader` to only prefetch data. In this case it is more performance for TS to never infer the complex type returned from `ensureQueryData`. Just to be clear, we're talking about this:

```tsx
  loader: (opts) => opts.context.queryClient.ensureQueryData(searchQueryOptions),
```

vs this

```tsx
  loader: (opts) => {
    opts.context.queryClient.ensureQueryData(searchQueryOptions)
  },
```

The second is more performant because the complex type is never inferred

I've profiled this with the large file based example and the difference can be quite large. 3.1s to type check the route tree vs 2.1s

### DIFF
--- a/docs/framework/react/guide/type-safety.md
+++ b/docs/framework/react/guide/type-safety.md
@@ -117,13 +117,12 @@ As your application scales, TypeScript check times will naturally increase. Ther
 
 ### Only infer types you need
 
-A great pattern with client side data caches (TanStack Query etc.) is to prefetch data. For example with TanStack Query you might have a route which calls `ensureQueryData` in a `loader`.
+A great pattern with client side data caches (TanStack Query, etc.) is to prefetch data. For example with TanStack Query you might have a route which calls `queryClient.ensureQueryData` in a `loader`.
 
 ```tsx
 export const Route = createFileRoute('/posts/$postId/deep')({
   loader: ({ context: { queryClient }, params: { postId } }) =>
     queryClient.ensureQueryData(postQueryOptions(postId)),
-  errorComponent: PostErrorComponent as any,
   component: PostDeepComponent,
 })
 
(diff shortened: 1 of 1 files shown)


docs/framework/react/guide/type-safety.md
```diff
@@ -115,6 +115,47 @@ const router = createRouter({
 
 As your application scales, TypeScript check times will naturally increase. There are a few things to keep in mind when your application scales to keep your TS check times down.
 
+### Only infer types you need
+
+A great pattern with client side data caches (TanStack Query etc.) is to prefetch data. For example with TanStack Query you might have a route which calls `ensureQueryData` in a `loader`.
```
SeanCassiere, line 120: - Missing comma (`,`)  in "(TanStack Query, etc.) ".
- ... which calls `queryClient.ensureQueryData`

docs/framework/react/guide/type-safety.md
```diff
@@ -115,6 +115,47 @@ const router = createRouter({
...
+
+```tsx
+export const Route = createFileRoute('/posts/$postId/deep')({
+  loader: ({ context: { queryClient }, params: { postId } }) =>
+    queryClient.ensureQueryData(postQueryOptions(postId)),
+  errorComponent: PostErrorComponent as any,
```
SeanCassiere, line 126: Let's remove the `errorComponent` here since it doesn't really serve a purpose for the example.

docs/framework/react/guide/type-safety.md
```diff
@@ -115,6 +115,47 @@ const router = createRouter({
...
+
+function PostDeepComponent() {
+  const params = Route.useParams()
+  const data = useSuspenseQuery(postQueryOptions(params.postId))
+
+  return <></>
+}
+```
+
+This may look fine and for small route trees you may not notice any TS performance issues. However in this case TS has to infer the loader return type despite it never beng used in your route. If the loader data is a complex type and many routes prefetch in this manner it can slow down editor performance. In this case the change is simple and you can let `Promise<void>` be inferred
```
SeanCassiere, line 138: - ... infer the loader's return type, despite ...
- ... data is a complex type with many routes ...
- In this case, the change is quite simple and let's type of `Promise<void>` be inferred.

docs/framework/react/guide/type-safety.md
```diff
@@ -115,6 +115,47 @@ const router = createRouter({
...
+
+```tsx
+export const Route = createFileRoute('/posts/$postId/deep')({
+  loader: async ({ context: { queryClient }, params: { postId } }) => {
+    await queryClient.ensureQueryData(postQueryOptions(postId))
+  },
+  errorComponent: PostErrorComponent as any,
```
SeanCassiere, line 145: Same thing about the `errorComponent`.

SeanCassiere (changes requested):

SeanCassiere (approved):
- 🚀

Issue #1762, opened 2024-06-17
Params containing slashes break the `from` prop


jakst:
### Describe the bug

When the value of `$var` contains a slash, the `$var.tsx` route will break if it contains a `from="/$var"` prop.

```tsx
// index.tsx
<Link to="/$var" params={{ var: 'some/thing' }}>...</Link>

// $var.tsx
<Link from="/$var" to="/$var">...</Link>
```

The issue was introduced in `v1.35.6` of the router

### Your Example Website or App

https://stackblitz.com/edit/github-d3fgdr?file=src%2Froutes%2F%24var.tsx

### Steps to Reproduce the Bug or Issue

1. Go to Stackblitz
2. Click "Go to problem"
3. Observer crash with message
```
Invariant failed: Could not find match for from: /$var
```

### Expected behavior

I expect URLs with params containing encoded slashes to not break routes.

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

mirague:
Was this also fixed by #1799 ?

schiller-manuel:
yes

PR #1784, opened 2024-06-19
fix: linking with from prop and a param containing slash

jakst:
A failing test case for #1762 that sets up a page on /$postId with a `<Link from="/$postId" to="/$postId">Link to self with from prop set</Link>`. When this page is visited and the $postId params contains a slash, the router crashes with `Invariant failed: Could not find match for from: /$postId`

### DIFF
--- a/packages/react-router/tests/link.test.tsx
+++ b/packages/react-router/tests/link.test.tsx
@@ -3031,7 +3031,7 @@ describe('Link', () => {
     expect(ErrorComponent).not.toHaveBeenCalled()
   })
 
-  test.only('when linking to self with from prop set and param containing a slash', async () => {
+  test('when linking to self with from prop set and param containing a slash', async () => {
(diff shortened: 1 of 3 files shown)


packages/react-router/tests/link.test.tsx
```diff
@@ -3030,6 +3030,47 @@ describe('Link', () => {
 
     expect(ErrorComponent).not.toHaveBeenCalled()
   })
+
+  test.only('when linking to self with from prop set and param containing a slash', async () => {
```
schiller-manuel, line 3034: please remove the `.only`

jakst: Oh, definitely didn't mean to include that in the PR. Good catch!

schiller-manuel:
closed in favor of https://github.com/TanStack/router/pull/1799

Issue #1779, opened 2024-06-18
`basepath` breaks routing in case its being found later within path


xobotyi:
### Describe the bug

When route path contains `basepath` - routing for these paths breaks.

Example can be found within provided stackblitz. 

### Your Example Website or App

https://stackblitz.com/edit/github-hqacy5?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

1. Set `basepath` parameter for router
2. Have a route with segment equal to `basepath`
3. Routing broken for all paths containing `basepath`

### Expected behavior

`basepath` works as a prefix to all routes of a router

### Platform

- OS: Windows
- Browser: Chrome
- Version: 126

PR #1787, opened 2024-06-19
fix: proper basename removal handling

xobotyi:
Previous implementation did not cover cases of partial match and cases with basepath matched later within the path.

Reworked version removes basepath only from the beginning of pathname and only in case it fully matches path segment.

Giving the basename is `/app`
| input | old | new |
|------:|:----|:-----|
| `/application` | `lication` | `/application` |
| `/app/project/applications` | `/projectlications` | `/project/applications` |
| `/app/app/app` | `` | `/app/app` |

fix: #1779

### DIFF
--- a/packages/react-router/src/path.ts
+++ b/packages/react-router/src/path.ts
@@ -82,6 +82,7 @@ interface ResolvePathOptions {
   to: string
   trailingSlash?: 'always' | 'never' | 'preserve'
 }
+
 export function resolvePath({
   basepath,
   base,
@@ -196,6 +197,7 @@ interface InterpolatePathOptions {
   leaveWildcards?: boolean
   leaveParams?: boolean
 }
+
 export function interpolatePath({
   path,
   params,
@@ -241,7 +243,36 @@ export function matchPathname(
 }
 
 export function removeBasepath(basepath: string, pathname: string) {
-  return basepath != '/' ? pathname.replace(basepath, '') : pathname
+  switch (true) {
+    // default behaviour is to serve app from the root - pathname
+    // left untouched
+    case basepath === '/':
+      return pathname
+
+    // shortcut for removing the basepath from the equal pathname
+    case pathname === basepath:
+      return ''
+
+    // in case pathname is shorter than basepath - there is
+    // nothing to remove
+    case pathname.length < basepath.length:
+      return pathname
+
+    // avoid matching partial segments - strict equality handled
+    // earlier, otherwise, basepath separated from pathname with
+    // separator, therefore lack of seprator means partial segment
+    // match (`/app` should not match `/application`)
+    case pathname[basepath.length] !== '/':
+      return pathname
+
+    // remove the basepath from the pathname in case there is any
+    case pathname.startsWith(basepath):
+      return pathname.slice(basepath.length)
+
+    // otherwise, return the pathname as is
+    default:
+      return pathname
+  }
 }
 
 export function matchByPath(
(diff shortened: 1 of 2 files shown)


packages/react-router/src/path.ts
```diff
@@ -241,7 +243,36 @@ export function matchPathname(
 }
 
 export function removeBasepath(basepath: string, pathname: string) {
-  return basepath != '/' ? pathname.replace(basepath, '') : pathname
+  switch (true) {
```
schiller-manuel, line 246: wouldn't it be much simpler to replace using

```tsx
new RegExp(`^${basepath}`)
```

notice the `^` to ensure it only replaces the basepath if it occurs at the start of the string

packages/react-router/tests/path.test.ts
```diff
@@ -1,5 +1,50 @@
...
+      pathname: '/application',
+      expected: '/application',
+    },
+    {
+      name: 'should remove multisegment basepath only in case it matches segments completely',
+      basepath: '/app/new',
+      pathname: '/app/new-application',
+      expected: '/app/new-application',
+    },
+  ].forEach(({ name, basepath, pathname, expected }) => {
```
schiller-manuel, line 42: please use the `describe.each` syntax: https://vitest.dev/api/#describe-each

packages/react-router/src/path.ts
xobotyi: 1) your solution does not ensure full-segment match `/app` -> `/application` will still occur.
2) such interpolation with no escaping (and there is no built-in regex-ascaping in JS AFAIK) is prone to errors, as URI's path syntax allows symbols that has meaning within regexes `/[some-segment]` obviously different outcome than expected.
3) i have no proof fo that but my gut feeling tells compillation of regex for each invocation will be suboptimal compared to bunch of `if`s

packages/react-router/tests/path.test.ts
xobotyi: According vitest's docs it will be a misuse of `each` since it is designed for `more than one test that depends on the same data` which is not true in this case since written is simply a table-test.
But okay.

packages/react-router/src/path.ts
schiller-manuel: 1. true, did not think about that
2. hmm I'd say not super likely, but possible
3. I wasn't optimizing for runtime performance, just wanted to keep the code complexity down if possible

thanks for the detailed response

packages/react-router/tests/path.test.ts
schiller-manuel: then `test.each`? https://vitest.dev/api/#test-each

packages/react-router/src/path.ts
```diff
@@ -241,7 +243,36 @@ export function matchPathname(
...
+      return ''
+
+    // in case pathname is shorter than basepath - there is
+    // nothing to remove
+    case pathname.length < basepath.length:
+      return pathname
+
+    // avoid matching partial segments - strict equality handled
+    // earlier, otherwise, basepath separated from pathname with
+    // separator, therefore lack of seprator means partial segment
```
schiller-manuel, line 263: typo in `seprator` => `separator`

packages/react-router/tests/path.test.ts
xobotyi: This will do)

packages/react-router/src/path.ts
xobotyi: fixed

schiller-manuel:
LGTM!
we need to sort out a CI issue, as soon as this is up and running again we'll merge this

PR #1790, opened 2024-06-19
feat(react-router): update `useBlocker` hook to provide current and next location when conditionally blocking

tomrehnstrom:
Currently, when using the `useBlocker` hook there is no way to ascertain where the navigation being blocked is headed.

For my usage of blocking, this is necessary to allow navigation within a selected bunch of nested routes, but not allowing the user to exit the base route. Thought this could be a useful feature to have.


schiller-manuel:
we wanted to follow react-router's API for this:

https://reactrouter.com/en/main/hooks/use-blocker

can you adjust your PR to match it?

also, the location should be typesafe  instead of just a `string`.

additionally, we would also need tests for `useBlocker`

tomrehnstrom:
> we wanted to follow react-router's API for this:
> 
> https://reactrouter.com/en/main/hooks/use-blocker
> 
> can you adjust your PR to match it?
> 
> also, the location should be typesafe instead of just a `string`.
> 
> additionally, we would also need tests for `useBlocker`

Yes sounds good, I will update the PR as soon as possible

tomrehnstrom:
This implementation should now be providing the `blockerFn` in`useBlocker` with a similar API to the hook in react-router. This also makes the blocker catch backwards navigation using the browser back buttons, which is very beneficial for our use case.

Things left to do:
- [ ] Tests
- [ ] Update docs/examples?

I am not quite sure how to do with the old `blockerFn` since it now behaves in a completely different way now. So any hints from someone on this would be appreciated.

Would also like to give a quick pitch for a different interface that could be used for `useBlocker` which I have implemented in our codebase with these changes:

```tsx
import React from 'react'
import type { BlockerFn } from '../../lib/router/qudak-router'
import { BlockerFn, useRouter } from '@tanstack/react-router'

type BlockerOpts = {
  blockerFn: BlockerFn
  disabled?: boolean
}

export function useBlocker({ blockerFn, disabled = false }: BlockerOpts) {
  const { history } = useRouter()

  React.useEffect(() => {
    return disabled ? undefined : history.block(blockerFn as BlockerFn)
  }, [blockerFn, disabled, history])
}
```

Which gives very clean implementations for custom UIs using hooks and could look something like this:

```tsx
  useBlocker({
    blockerFn: () => {
      if (blockerDisabled && blockerDisabled.current) {
        blockerDisabled.current = false
        return false
      }
      if (!blockerDisabled && _blockerDisabled.current) {
        _blockerDisabled.current = false
        return false
      }
      const shouldBlock = blockerFn()
      if (!shouldBlock) return false

      const promise = new Promise<boolean>((resolve) => {
        modals.open({
          title: 'Are you sure you want to leave?',
          children: (
            <SaveBlocker
              onSave={onSave}
              onDiscard={onDiscard}
              close={() => {
                modals.closeAll()
                resolve(false)
              }}
              confirm={() => {
                modals.closeAll()
                resolve(true)
              }}
              reject={() => {
                modals.closeAll()
                resolve(true)
              }}
            />
          ),
          onClose: () => resolve(false)
        })
      })

      return promise
    },
    disabled
  })
```

Any thoughts on this?

schiller-manuel:
This would be a breaking change to `useBlocker`, right? then we cannot do it (now, only possible in v2).

chorobin:
While this might be out of scope for this PR. I think we can likely do much better from a type safety point of view for `useBlocker`. Correct me if I'm wrong but `useBlocker` blocks a navigation and we usually describe navigations with `from` and `to`.

I was imagining a more type safe hook.

```tsx
useBlocker({ from: Route.fullPath, to: '..', blockerFn: ({ fromMatch, toMatch }) => // have I saved? })
```

- `from` would describe a route matching the current location and would narrow the `fromMatch` passed to `blockerFn` for the matching route. This will be optional `fromMatch` will be loose if not provided
- `to` would describe a route matching the destination location and would narrow the `toMatch` passed to `blockerFn` for the matching route. This will be optional `toMatch` will be loose if not provided
- `blockerFn` determines if the navigation should continue or block.
- `fromMatch/toMatch` is a match like from `useMatch` which will provide the usual `params`, `search`, `context`, etc

When `from` or `to` is provided the `blockerFn` is only called when navigating to the location described by these two properties. If they are not provided it will always call `blockerFn` for any navigation

schiller-manuel:
we should make this as typesafe as possible, as per @chorobin's comment.
@tomrehnstrom can you tackle this? if you need support let us know

ThomasStock:
@tomrehnstrom, are you planning to continue working on this PR?

Does anyone know a quick and dirty workaround I can apply in the meanwhile?

tomrehnstrom:
> @tomrehnstrom, are you planning to continue working on this PR?
> 
> Does anyone know a quick and dirty workaround I can apply in the meanwhile?

Yes, but have been quite busy elsewhere recently. The way i have solved this quickly in our codebase is just to copy the history package index.ts file and make the changes required to make it work, and then use it as a custom history:

```tsx
const history = createQudakBrowserHistory()

  routeTree,
  defaultPendingComponent: FullWidthLoader,
  defaultErrorComponent: RouterSuspenseErrorFallback,
  defaultNotFoundComponent: Error404,
  context: {
    auth: undefined!,
    queryClient
  },
  defaultPreload: 'intent',
  // Since we're using React Query, we don't want loader calls to ever be stale
  // This will ensure that the loader is always called when the route is preloaded or visited
  defaultPreloadStaleTime: 0,
  history
})
```

With the changes in my first commit in this pr

Then written a custom blocker hook like this:
```tsx
type BlockerOpts = {
  blockerFn: CustomBlockerFn
  disabled?: boolean
}

export function useCustomBlocker({ blockerFn, disabled = false }: BlockerOpts) {
  const { history } = useRouter()

  React.useEffect(() => {
    return disabled ? undefined : history.block(blockerFn as BlockerFn)
  }, [blockerFn, disabled, history])
}
```

ThomasStock:
Thanks @tomrehnstrom, I needed this urgently and managed to make do with the suggested temporary solution for now.

tomrehnstrom:
Whoops, i broke the PR

tomrehnstrom:
Hello @schiller-manuel, I have finally got this PR into a pretty nice shape. However, I am not sure about the way I implemented route from/to matching in the useBlocker hook, any suggestions there would be appreciated.

tomrehnstrom:
For anyone stumbling across the PR and really wants this now, use my fork: https://github.com/tomrehnstrom/router/tree/better-blocker-router
Until this can get merged :)

schiller-manuel:
I need to have a closer look, what's the current status of this PR? anything obvious missing ?

tomrehnstrom:
The type safety and type narrowing described by @chorobin is not really implemented, but functionally of it should be complete, have been using this method in prod for quite a while and haven't had any issues so far. Maybe the old interface for `useBlocker` should still be supported and marked as deprecated though.

schiller-manuel:
is your PR breaking the blocker API or is it backwards compatible?

packages/react-router/src/usePromiseBlocker.tsx
```diff
@@ -0,0 +1,63 @@
+import * as React from 'react'
+import { useRouter } from './useRouter'
+import { matchPathname } from './path'
+import type { BlockerFnArgs } from '@tanstack/history'
+import type { UseBlockerOpts } from './useBlocker'
+
+export function usePromiseBlocker({
```
schiller-manuel, line 7: why do we have this? isn't `useBlocker` enough?

schiller-manuel:
let's work on merging this.
[] docs need updating

packages/history/src/index.ts
```diff
@@ -117,52 +151,69 @@ export function createHistory(opts: {
...
-      blockers.push(blocker)
-
-      if (blockers.length === 1) {
-        addEventListener(beforeUnloadEvent, beforeUnloadListener, {
-          capture: true,
-        })
-      }
+      if (!opts.win) return () => {}
+      if (!opts.win.blockers) opts.win.blockers = []
+      opts.win.blockers.push(blocker)
```
schiller-manuel, line 212: why do we need to store the blockers in the window?

tomrehnstrom: We dont :)

schiller-manuel:
@tomrehnstrom thanks for the update
still not sure if the API is backwards compatible.
can you comment on that please? if it is not backwards compatible can you please fix this so we can release this in a non breaking manner

tomrehnstrom:
> @tomrehnstrom thanks for the update still not sure if the API is backwards compatible. can you comment on that please? if it is not backwards compatible can you please fix this so we can release this in a non breaking manner

It should be backwards compatible now 👍

sagi-webiks:
@Tanner-Scadden  I truly appreciate your help here. Thank you so much!

ferretwithaberet:
Could this be modified to supply the navigation type to the `notify` function? I will try to make an addon for the history API after this PR merges which will keep track of the history state. The addon will help us hide back buttons when there is only 1 entry in the navigation state and, maybe, for implementing breadcrumbs. Also, can you explain why `back`, `forward` and `go` are of type `"POP"`? It would help me if they had their own navigation type. My rough ideea for the addon right now is that `PUSH`, `POP`, `REPLACE` would do their respective action (`.push`, `.pop` or `.replace`) on an array and increment/decrement a `cursor` variable, while `BACK`, `FORWARD` and `GO` would only increment/decrement the `cursor` variable.

tomrehnstrom:
@ferretwithaberet Sorry for the long wait, would the changes in my latest commit accomplished this?

ferretwithaberet:
> @ferretwithaberet Sorry for the long wait, would the changes in my latest commit accomplished this?

@tomrehnstrom I do not mind the wait, take your time, do not stress yourself. As to the implementation, I would need to get the `index` for the `GO` action, which actually is , [as MDN calls it](https://developer.mozilla.org/en-US/docs/Web/API/History/go#delta), a delta. I need it so I can know how much I would have to increase/decrease the `cursor` of the history state. Otherwise it looks good to me, great job.

ferretwithaberet:
I tried to test the code manually. When using the browser back/forward buttons, it seems that the `popstate` event gets triggered for both buttons, which in turn triggers a `POP` event regardless of navigation type.
Also, I could not find nothing on the `pushstate` window event, does it even exist or am I missing something?
I will not be able to look much on this until monday night EEST, I will try to look for a solution then.

docs/framework/react/api/router/useBlockerHook.md
```diff
@@ -7,19 +7,57 @@ The `useBlocker` method is a hook that [blocks navigation](../../guide/navigatio
 
 ## useBlocker options
 
-The `useBlocker` hook accepts a single _optional_ argument, an option object:
+The `useBlocker` hook accepts a single \_required argument, an option object:
```
schiller-manuel, line 10:
```suggestion
The `useBlocker` hook accepts a single _required_ argument, an option object:
```

docs/framework/react/api/router/useBlockerHook.md
```diff
@@ -7,19 +7,57 @@ The `useBlocker` method is a hook that [blocks navigation](../../guide/navigatio
...
 
-### `options.blockerFn` option
+### `options.shouldBlockFn` option
 
-- Optional
+- Required
```
schiller-manuel, line 14: if this required now, that's an API break. please either update the docs or make it backwards compatible

docs/framework/react/api/router/useBlockerHook.md
```diff
@@ -7,19 +7,57 @@ The `useBlocker` method is a hook that [blocks navigation](../../guide/navigatio
...
+- Think of this function as telling the router if it should block the navigation, so returning `true` mean that it should block the navgation and `false` that it should be allowed
+
+### `options.disabled` option
+
+- Optional - defaults to `false`
+- Type: `boolean`
+- Specifies if the blocker should be entirely disabled or not
 
-### `options.condition` option
+### `options.enableBeforeUnload` option
```
schiller-manuel, line 26: same here. you can add a new property, mark the old one as deprecated but still accept and use it if is presented

docs/framework/react/api/router/useBlockerHook.md
```diff
@@ -7,19 +7,57 @@ The `useBlocker` method is a hook that [blocks navigation](../../guide/navigatio
...
 - Optional - defaults to `true`
+- Type: `boolean | (() => boolean)`
+- Tell the blocker to sometimes or always block the browser `beforeUnload` event or not
+
+### `options.withResolver` option
+
+- Optional - defaults to `false`
 - Type: `boolean`
-- A navigation attempt is blocked when this condition is `true`.
+- Specify if your the resolver that the hook returns should be used or whether the information in your `shouldBlockFn` is enough the determine blocking
```
schiller-manuel, line 36: your => you are?

schiller-manuel, line 36:
```suggestion
- Specify if you are the resolver that the hook returns should be used or whether the information in your `shouldBlockFn` is enough to determine blocking
```

docs/framework/react/guide/navigation-blocking.md
```diff
@@ -101,7 +109,7 @@ function MyComponent() {
   const [formIsDirty, setFormIsDirty] = useState(false)
 
   const { proceed, reset, status } = useBlocker({
-    condition: formIsDirty,
+    shouldBlockFn: () => formIsDirty,
```
schiller-manuel, line 112: name of prop, see above

docs/framework/react/guide/navigation-blocking.md
```diff
@@ -131,7 +139,7 @@ function MyComponent() {
   const [formIsDirty, setFormIsDirty] = useState(false)
 
   return (
-    <Block condition={formIsDirty}>
+    <Block shouldBlockFn={() => formIsDirty}>
```
schiller-manuel, line 142: name of prop, see above

docs/framework/react/guide/navigation-blocking.md
```diff
@@ -66,17 +66,25 @@ function MyComponent() {
...
     />
   )
 
   // OR
 
   return (
     <Block
-      blocker={() => window.confirm('Are you sure you want to leave?')}
-      condition={formIsDirty}
+      shouldBlockFn={() => {
```
schiller-manuel, line 82: name of prop, see above

docs/framework/react/api/router/useBlockerHook.md
tomrehnstrom: It should not be an API break since we still allow the old shape of input object
```
export type UseBlockerOpts = {
  shouldBlockFn: BlockerFn

  enableBeforeUnload?: boolean | (() => boolean)
  disabled?: boolean
  withResolver?: boolean
} & { .... }

type LegacyBlockerFn = () => Promise<any> | any
type LegacyBlockerOpts = {
  blockerFn?: LegacyBlockerFn
  condition?: boolean | any
}

 opts?: UseBlockerOpts | LegacyBlockerOpts | LegacyBlockerFn,
```
But when not using the deprecated signature, the function is required. This should be fine no?

tomrehnstrom: I think I had to rename the function property to maintain backward compatibility here, but if there is a way to do it I missed, I can fix it.

packages/history/src/index.ts
```diff
@@ -5,21 +5,36 @@
 export interface NavigateOptions {
   ignoreBlocker?: boolean
 }
+
+type SubscriberHistoryAction =
+  | {
+      action: HistoryAction | 'ROLLBACK'
```
ferretwithaberet, line 11: Can this property be renamed to `type`, having to do `action.action` is odd

schiller-manuel: done

schiller-manuel:
@sagi-webiks 
> After this PR, I will be able to block based on the condition: currentUrl != nextUrl ? Thanks

yes!

schiller-manuel:
I am working on making the API typesafe, stay tuned

tomrehnstrom:
Cool! I think using setup will be much more useful than having `from`and `to` in the hook args

tomrehnstrom:
Anything here I can help with here @schiller-manuel?

schiller-manuel:
@tomrehnstrom 

we need

- type tests for useBlocker
- more unit tests in a full router setup to check that action, current and next are correctly passed into shouldBlockFn
- e2e tests for the legacy API and the new API

I have also added a TODO whether we should return current and next in case the resolver is used. I thought about this when editing the example since we only know the status then but don't know current/next which might be necessary for UI?

tomrehnstrom:
Yes, I will jump on updating docs and supplying the current blocking state to the resolver.

schiller-manuel:
@tomrehnstrom can you ping me on discord please?

https://tlinz.com/discord

docs/framework/react/guide/navigation-blocking.md
```diff
@@ -120,6 +122,45 @@ function MyComponent() {
...
 
+### Hook/logical-based custom UI without resolver
+
+```tsx
+import { useBlocker } from '@tanstack/react-router'
+
+function MyComponent() {
+  const [formIsDirty, setFormIsDirty] = useState(false)
+
+  const { proceed, reset, status } = useBlocker({
```
schiller-manuel, line 133:
```suggestion
  useBlocker({
```

since `withResolver` is not `true`, this won't return the (unused) props

packages/react-router/tests/useBlocker.test.tsx
```diff
@@ -1,42 +1,425 @@
+import React from 'react'
```
schiller-manuel, line 1: we also need some tests that use the `<Block>` component

PR #1792, opened 2024-06-19
fix(start): adding support for streaming using `renderToReadableStream`

lithdew:
react-dom/server does not export renderToPipeableStream for JavaScript runtimes such as Bun.


lithdew:
CI appears to be failing due to some package exports that aren't touched in this PR + some unconfigured CI secrets.

packages/start/src/server/defaultStreamHandler.tsx
```diff
@@ -16,30 +16,54 @@ export const defaultStreamHandler: StartHandler<AnyRouter> = async ({
...
-  })
-
-  return new Response(passthrough as any, {
-    status: router.state.statusCode,
-    statusText: `${router.state.statusCode}`.startsWith('5')
-      ? 'Internal Server Error'
-      : 'OK',
-    headers: responseHeaders,
-  })
+  if (typeof ReactDOMServer.renderToReadableStream === 'function') {
```
schiller-manuel, line 19: `renderToReadableStream` now has precedence over `renderToPipeableStream`, do we want this?

lithdew: For environments where `renderToReadableStream` is available (environment supporting Web Streams API i.e. the browser, Bun, etc.), `renderToPipeableStream` is not available.

See:

https://github.com/facebook/react/blob/main/packages/react-dom/package.json#L60-L70
https://github.com/facebook/react/blob/main/packages/react-dom/server.browser.js
https://github.com/facebook/react/blob/main/packages/react-dom/server.node.js
https://github.com/facebook/react/blob/main/packages/react-dom/server.edge.js
https://github.com/facebook/react/blob/main/packages/react-dom/server.bun.js

schiller-manuel: ah, it's either - or. I assumed that an implementation could have both.

PR #1803, opened 2024-06-21
chore(examples): cleanup the config for the "quickstart-rspack-file-based" example

SeanCassiere:


SeanCassiere:
Need to eventually fix this flaky playright test.

PR #1805, opened 2024-06-21
Adding tests

mta-trackunit:
Added test to prove rerender is not retriggering state changes


schiller-manuel:
is this test still relevant?

PR #1817, opened 2024-06-22
fix(router-generator): format the route template before writing to disk

matinrco:
- add `tabWidth` config to `router-generator`
- apply same formatting to route code generator
- update docs

### DIFF
--- a/packages/router-generator/src/config.ts
+++ b/packages/router-generator/src/config.ts
@@ -10,6 +10,7 @@ export const configSchema = z.object({
   generatedRouteTree: z.string().optional().default('./src/routeTree.gen.ts'),
   quoteStyle: z.enum(['single', 'double']).optional().default('single'),
   semicolons: z.boolean().optional().default(false),
+  tabWidth: z.number().optional().default(2),
   disableTypes: z.boolean().optional().default(false),
   addExtensions: z.boolean().optional().default(false),
   disableLogging: z.boolean().optional().default(false),
--- a/packages/router-generator/src/generator.ts
+++ b/packages/router-generator/src/generator.ts
@@ -306,7 +306,13 @@ export async function generator(config: Config) {
 
       if (replaced !== routeCode) {
         logger.log(`🟡 Updating ${node.fullPath}`)
-        await fsp.writeFile(node.fullPath, replaced)
+        const formattedReplaced = await prettier.format(replaced, {
+          semi: config.semicolons,
+          singleQuote: config.quoteStyle === 'single',
+          tabWidth: config.tabWidth,
+          parser: 'typescript',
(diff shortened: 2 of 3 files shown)


packages/router-generator/src/config.ts
```diff
@@ -10,6 +10,7 @@ export const configSchema = z.object({
   generatedRouteTree: z.string().optional().default('./src/routeTree.gen.ts'),
   quoteStyle: z.enum(['single', 'double']).optional().default('single'),
   semicolons: z.boolean().optional().default(false),
+  tabWidth: z.number().optional().default(2),
```
SeanCassiere, line 13: Please remove the `tabWidth` configuration as we've decided to stop adding in code-styling-related (prettier) configurations.

I haven't gotten around to it yet, but we need to update the documentation to explicitly ask the user to *ignore* the generated route-tree file from both their linter (eslint, biome, etc) and their formatted (prettier,...).

The rationale behind this is quite simple. This file is something controlled by the router-generator and shouldn't be touched by the linters or formatters.

packages/router-generator/src/generator.ts
```diff
@@ -306,7 +306,13 @@ export async function generator(config: Config) {
 
       if (replaced !== routeCode) {
         logger.log(`🟡 Updating ${node.fullPath}`)
-        await fsp.writeFile(node.fullPath, replaced)
+        const formattedReplaced = await prettier.format(replaced, {
```
SeanCassiere, line 309: 👍🏼

SeanCassiere (changes requested):

packages/router-generator/src/generator.ts
schiller-manuel: do we really want to format the route files? why?

packages/router-generator/src/config.ts
matinrco: got it 🙏
I'm already ignored the route tree file in my formatter.

packages/router-generator/src/generator.ts
matinrco: @schiller-manuel 

I prefer to format the route files because they are part of my code and affect my code style.
To have full control over formatting, I think we need to separate the config, considering Sean Cassiere's opinion in the previous comment. Otherwise, the whole pull request is pointless. ty

schiller-manuel: I don't think router-generator should format route files. This are implementation files that the user has full control over when it comes to code formatting, so I'd say it is the user's duty to setup the desired codeformatting independently from the generator.

SeanCassiere:
Closing this, since we are in agreement that this is not really needed.

PR #1822, opened 2024-06-24
fix(react-router): `ResolvedSuspense` to always use `React.Suspense` never the `SafeFragment`

mta-trackunit:
If you create any hook that relies on a tanstack router hook - it becomes hard to test since the rerender recreates the elements again, I traced it down to this part in matches.tsx 
`  const ResolvedSuspense = !router.state.matches.length
    ? React.Suspense
    : SafeFragment
`
It makes state changes with only the React.Suspense it works, dont really know if its intended or not? if so please explain how we could get around it for testing.


packages/react-router/tests/link.test.tsx
```diff
@@ -34,6 +36,72 @@ afterEach(() => {
...
+    await waitFor(() => expect(screen.getByTestId('testId')).toBeVisible());
+    expect(result.current).toBeTruthy();
+  
+    const original = result.current;
+
+    rerender();
+    await waitFor(() => expect(screen.getByTestId('testId')).toBeVisible());
+    const updated  = result.current;
+    
+    expect(original).toBe(updated);
```
SeanCassiere, line 101: This uses `Object.is` which will never return true since React 😅. Could you push a change with `expect(original).toEqual(updated)` please.

packages/react-router/tests/link.test.tsx
```diff
@@ -34,6 +36,72 @@ afterEach(() => {
 })
 
 describe('Link', () => {
+
+  test('when using renderHook it returns a hook with same content to prove rerender works', async () => {
```
SeanCassiere, line 40: Also, please use prettier 😅. Our prettier configuration is available at the root.

packages/react-router/tests/link.test.tsx
mta-trackunit: yep I will do that, thanks for replying

packages/react-router/tests/link.test.tsx
mta-trackunit: I will change that :)

packages/react-router/tests/link.test.tsx
mta-trackunit: 'toBe' is actually what I expect to be true here - If I have a hook and memorise a value I expect it not to change on a rerender - if I change to toEqual it will work here - but not verify that I did not get a stateChange. Testing that way will ensure the base of the hook is stabile and not causing state changes.. what do you think ?

SeanCassiere:
🚨 Tested this out locally using the `kitchen-sink-file-based` example.

This change breaks that example where, when on`/dashboard`, you click "Invoices" and navigate to `/dashboard/invoices`, and then try clicking on "Users" to navigate to `/dashboard/users`, the navigation never resolves.

mta-trackunit:
So navigating from http://localhost:5173/dashboard/invoices
to http://localhost:5173/dashboard/users
It navigates to 
http://localhost:5173/dashboard/invoices?usersView=%7B%7D 

So it seems like tanstack router is behind 1 "tick" ... any idea how to debug that? 

In examples/react/kitchen-sink-file-based/src/routes/dashboard.users.tsx 
```ts 

  React.useEffect(() => {
    navigate({
      search: (old) => {
        return {
          ...old,
          usersView: {
            ...old?.usersView,
            filterBy: filterDraft || undefined,
          },
        }
      },
      replace: true,
    })
  }, [filterDraft])
  ```
  
That seems to navigate - so it feels like tanstack router is behind and redirects

File: packages/react-router/src/router.ts [hunk 3]
```diff
@@ -1117,16 +1117,7 @@ export class Router<
...
     ): ParsedLocation => {
-      // if the router is loading the previous location is what
-      // we should use because the old matches are still around
-      // and the latest location has already been updated.
-      // If the router is not loading we should always use the
-      // latest location because resolvedLocation can lag behind
-      // and the new matches could already render
-      const currentLocation = this.state.isLoading
-        ? this.state.resolvedLocation
-        : this.latestLocation
```
mta-trackunit, line 1128: This change made the kitchensink app work - but dont know what other consequence it will have? maybe my change is not the right solution - but it solves the kitchensink app navigate from invoices to users

mta-trackunit:
I gave up on fixing the this.state.resolvedLocation - I just added the "to" in the useEffect that fires on first render of the page. Without the "to" it will navigate back because this.state.resolvedLocation is still the "old" value - that is clearly a bug - but this is the workaround.

mta-trackunit:
@SeanCassiere The bug you found is also on main branch right now... I even hit it in our production site... adding "to": will fix it ... but it seems unrelated to this PR then... do you want me to revert the "to" fix ? or do we want this to be fixed on main this way

examples/react/kitchen-sink-file-based/src/routes/dashboard.users.tsx
```diff
@@ -68,7 +68,8 @@ function UsersComponent() {
 
   React.useEffect(() => {
     navigate({
-      search: (old) => {
+       to: "/dashboard/users",
```
SeanCassiere, line 71: The problem is that a navigate call should be possible without the `to` field.

mta-trackunit: I agree but that is also a problem on main branch already, so it has nothing to do with this fix

mta-trackunit: do you want me to revert the 2 fixes then?

SeanCassiere:
@mta-trackunit FYI, I'm currently holding this ATM till #1907 goes through.

SeanCassiere:
The issues with the kitchen sink example should be resolved with #1968

mta-trackunit:
@SeanCassiere how come you removed the 
`history: createMemoryHistory({ initialEntries: ['/'] }),`

SeanCassiere:
> @SeanCassiere how come you removed the `history: createMemoryHistory({ initialEntries: ['/'] }),`

Mostly so we emulate the other tests and use the browser history.

SeanCassiere (approved):
As mentioned in [this discord thread](https://discord.com/channels/719702312431386674/1015406599738761307/1255217684153368767), I've tested this locally with the `basic`, `basic-file-based`, `kitchen-sink-file-based`, and `start-basic` examples, and I have observed no visible regressions.

@schiller-manuel @tannerlinsley anything blocking this from your side?

mta-trackunit:
> As mentioned in [this discord thread](https://discord.com/channels/719702312431386674/1015406599738761307/1255217684153368767), I've tested this locally with the `basic`, `basic-file-based`, `kitchen-sink-file-based`, and `start-basic` examples, and I have observed no visible regressions.
> 
> @schiller-manuel @tannerlinsley anything blocking this from your side?

you are linking to a private channel - so I cannot follow along :) - any update?

PR #1824, opened 2024-06-24
feat: New Serialization Paradigm

tannerlinsley:
- This reworks the entire serialization model including deferred streaming of data.
- Instead of using stream injection, it uses suspense for promises and recursive suspense for streams.
- `defer(promise)` is no longer needed, since we are walking the loader return and extracting promises and streams now.
- `Await` is still required in React <19 to consume deferred promises, but in 19, `use` will work just fine.
- The SSR and start examples now use the new <Html>, <Head>, and <Body> tags, which are also required for non React 19 full document hydration
- It begins (but doesn't fully flesh out), a universal server entry via `createStartHandler`. More work is needed for non-start SSR envs to use it, which is where `createRequestHandler` needs some more polish.
- Provides **experimental** support for RSCs. This does not include using client components yet (which will require a bit more work around RSC manifests for the client. @devagrawal09?)
- RSC requires that you override the react and react-dom version at the mono repo level. There are `temp-` entries there for what I know works
- Rendering RSCs requires the new `renderRsc` function, exported from `@tanstack/start`, which uses the new `react-server-dom/client` to isomorphically create React elements from either Readable's on the server or ReadableStreams/Fetches on the client. One of the reasons for this `renderRsc()` step is so that RSCs **remain primitive streams until they can be handled by the router for SSR and caching**, after which they are released back to the user via useLoaderData for consumption. Diving deeper, the primary reason for this is so that we can gain access to the original binary chunks of RSC syntax that are being streamed across the wire. Without access to those chunks, we cannot serialize and recreate the RSC across any serialization boundary.
- Some components and utilities have been further extracted into their own files to avoid circular deps, types and ultimately, mixing of client and server imports during RSC. **There will be a lot more work to do here when we enable client components, as we will need to make sure server-capable router imports do not bring in `useState`, etc. I still have strong feelings that these imports shouldn't be invariants during bundling and should only warn at runtime... but I digress**.
- The <ScriptOnce> component got serious upgrades along with an undocumented feature of using the `.tsr-once` class to ensure than an element is completely removed before hydration is attempted.
-


packages/react-router/src/Matches.tsx
```diff
@@ -42,6 +38,7 @@ export interface RouteMatch<
 > {
   id: string
   routeId: TRouteId
+  index: number
```
schiller-manuel, line 41: where is this used?

packages/react-router/src/utils.ts
```diff
@@ -205,7 +205,7 @@ function hasObjectPrototype(o: any) {
   return Object.prototype.toString.call(o) === '[object Object]'
 }
 
-export function isPlainArray(value: unknown) {
+export function isPlainArray(value: unknown): value is Array<unknown> {
```
schiller-manuel, line 208: do we even need this with TS5.5 [Inferred Type Predicates](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates)?

=> worth trying to upgrade to TS5.5

tannerlinsley:
I'm not completely sure how to get things passing without forcing the entire monorepo into React 19...

packages/react-router/src/utils.ts
SeanCassiere: We can't use TS 5.5 since we've currently stated that the minimum [requirement](https://tanstack.com/router/latest/docs/framework/react/installation#requirements) for the TS version is 5.0 on the website.

We can change to TS 5.5 ofc, but I believe this is something that'd normally be a part of a major release.

examples/react/start-basic-rsc/app/utils/posts.tsx
```diff
@@ -8,26 +8,28 @@ export type PostType = {
...
+  'use server'
   console.log(`Fetching post with id ${postId}...`)
   await new Promise((r) => setTimeout(r, 500))
   const post = await axios
     .get<PostType>(`https://jsonplaceholder.typicode.com/posts/${postId}`)
     .then((r) => r.data)
     .catch((err) => {
       if (err.status === 404) {
-        throw notFound()
+        // throw notFound()
```
SeanCassiere, line 20: Is `notFound` not working with streaming?

examples/react/basic-ssr-streaming-file-based/src/routes/index.tsx
```diff
@@ -20,11 +18,9 @@ function IndexComponent() {
     <div className="p-2">
       <h3>Welcome Home!</h3>
       <p>Data: {data.date.getDate()}</p>
-      <React.Suspense fallback={'Loading...'}>
-        <Await promise={data.deferred}>
-          {(data) => <p>Deferred: {data.date.getDate()}</p>}
-        </Await>
-      </React.Suspense>
+      <Await promise={data.deferred} fallback={'Loading...'}>
```
SeanCassiere, line 21: Would `Await` also work here if the promise was the sole return (both functionality and on a type-level)?

```tsx
  loader: () => new Promise<{ date: Date }>((r) => {
    setTimeout(() => r({ date: new Date() }), 1000)
  })
})

function IndexComponent() {
  // ...
      <Await promise={useLoaderDataPromise} fallback={'Loading...'}>
        {(data) => <p>Deferred: {data..getDate()}</p>}
      </Await>
  </div>
}

```

examples/react/start-basic-rsc/package.json
```diff
@@ -12,31 +12,38 @@
...
     "@tanstack/react-router": "^1.40.0",
     "@tanstack/router-devtools": "^1.40.0",
     "@tanstack/router-vite-plugin": "^1.39.13",
     "@tanstack/start": "^1.40.1",
-    "redaxios": "^0.5.0",
+    "@vinxi/react-server-dom": "^0.0.3",
     "isbot": "^3.7.0",
-    "react": "^18.3.1",
-    "react-dom": "^18.3.1",
+    "react": "0.0.0-experimental-035a41c4e-20230704",
```
SeanCassiere, line 23: For anyone looking at these examples, whats the difference in us using the `0.0.0-experimental-...` version instead of `19.0.0-rc-...`?

packages/react-router/src/link.tsx
```diff
@@ -1,3 +1,5 @@
+'use client'
```
SeanCassiere, line 1: Does this currently work with RSCs? or is it user-land `"use client"` components that are wonky?

packages/start/src/config/index.ts
```diff
@@ -16,64 +16,55 @@ import { serverFunctions } from '@vinxi/server-functions/plugin'
...
-      .object({
-        vite: viteSchema,
-      })
-      .optional()
-      .default({}),
-  })
-  .optional()
-  .default({})
+const viteSchema = z.object({
+  plugins: z.function().returns(z.array(z.custom<vite.Plugin>())).optional(),
```
SeanCassiere, line 24: I believe the `vite.Plugin` type is currently being inherited from the monorepo and not the actual `start` package so you type mismatches like in the `start-basic` example.

Also, do we need to be `zod` validating this? Maybe this could just be a type we pass into the schemas down the line and call `.default({})` on em?

packages/react-router/package.json
```diff
@@ -74,7 +74,7 @@
     "zod": "^3.22.4"
   },
   "peerDependencies": {
-    "react": ">=16.8",
-    "react-dom": ">=16.8"
+    "react": ">=18",
```
SeanCassiere, line 77: Sticky note, that the website will need to be updated for this.

packages/start/src/server-handler/index.tsx
```diff
@@ -1,12 +1,191 @@
...
 
 export default eventHandler(handleServerAction) as any
 
-export async function handleServerAction(event: any) {
+async function handleServerAction(event: H3Event) {
   const request = toWebRequest(event)
-  return await handleRequest(request)
+  const method = request.method
+  const url = new URL(request.url, 'http://localhost:3000')
+  const search = Object.fromEntries(
```
SeanCassiere, line 19: Does this localhost URL matter here?

packages/start/tsconfig.json
```diff
@@ -4,5 +4,5 @@
     "jsx": "react",
     "module": "esnext"
   },
-  "include": ["src", "vite.config.ts"]
+  "include": ["src", "vite.config.ts", "src/client/DehydrateRouter.tsx"]
```
SeanCassiere, line 7: ?

SeanCassiere:
https://github.com/TanStack/router/blob/5e1cb46e31077e2659d97754b25690f31d0439d7/pnpm-lock.yaml#L1

@tannerlinsley, your pnpm version may be out of date, since the quest into the rsc branch began. 

https://github.com/TanStack/router/pull/1757 moved the monorepo to pnpm v9 (9.3.0) to be specific.

SeanCassiere:
Regarding the e2e test. It is failing the navigation to `/posts` in `start-basic`. This can be seen in dev as well. It is possible that it should be using the experimental react version.

> Error when evaluating SSR module /@fs/Users/seanc/code/tanstack/router/packages/start/dist/esm/server-handler/server-handler/index.js: failed to import "react"
|- Error: This entry point is not yet supported outside of experimental channels
    at /Users/seanc/code/tanstack/router/node_modules/.pnpm/react@18.3.1/node_modules/react/cjs/react.shared-subset.development.js:18:7
    at Object.<anonymous> (/Users/seanc/code/tanstack/router/node_modules/.pnpm/react@18.3.1/node_modules/react/cjs/react.shared-subset.development.js:19:5)

packages/start/tsconfig.json
SeanCassiere: Nvm... the test matrix breaks without it.

examples/react/start-basic-rsc/app/utils/posts.tsx
tannerlinsley: This will be more complex to solve with RSCs, but we're punting on that for now.

examples/react/start-basic-rsc/package.json
tannerlinsley: One works with Vinxi, the other one doesn't. This will change soon.

packages/start/src/config/index.ts
tannerlinsley: It's not really validating the runtime value inside of the array, just providing the type information through the schema. Should be fine.

packages/start/src/server-handler/index.tsx
tannerlinsley: Not really. It's just a placeholder to make it a valid URL.

packages/router-plugin/src/ast.ts
```diff
@@ -1,4 +1,19 @@
 import * as babel from '@babel/core'
+// @ts-expect-error
+import _babelPluginJsx from '@babel/plugin-syntax-jsx'
```
SeanCassiere, line 3: Are the changes here actually needed? What was broken before?

examples/react/start-basic-react-query/app/routes/posts.index.tsx
```diff
@@ -0,0 +1,10 @@
+import { createFileRoute } from '@tanstack/react-router'
+
+// @ts-ignore
```
SeanCassiere, line 3: can we nuke that ts-ignore?

packages/router-plugin/tests/code-splitter/test-files/function-as-parameter.tsx
```diff
@@ -8,22 +8,22 @@ const ReactUseMemoCall1 = React.useMemo(function performAction() {
   return 'true'
 }, [])
 
-console.log(ReactUseMemoCall1)
+console.info(ReactUseMemoCall1)
```
SeanCassiere, line 11: Why are these tests being changed? No real big deal, just wondering?

examples/react/basic-ssr-streaming-file-based/src/routes/index.tsx
tannerlinsley: Yes it would

packages/router-plugin/src/ast.ts
tannerlinsley: Yes, babels plugin resolver was not working here.

examples/react/start-basic-react-query/app/routes/posts.index.tsx
tannerlinsley, line 3:
```suggestion
```

packages/router-plugin/tests/code-splitter/test-files/function-as-parameter.tsx
tannerlinsley: I changed to info so I could more easily find lingering console logs from development across the code base.

PR #1830, opened 2024-06-26
Update parallel-routes.md (TODO page) with reference to "Route Masking"

colelawrence:
Add something for people who stumble upon this to continue figuring out what options they have for parallel routes in TanStack Router.

### DIFF
--- a/docs/framework/react/guide/parallel-routes.md
+++ b/docs/framework/react/guide/parallel-routes.md
@@ -3,3 +3,5 @@ title: Parallel Routes
 ---
 
 TODO: Write this guide
+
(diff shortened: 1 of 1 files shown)


tannerlinsley:
I don't think this is the right description for parallel routes though. They are related, but very separate. We don't currently support parallel routes (which is essentially having multiple matching branches of a route tree at the same time).

It's on our radar though!

mta-trackunit:
@tannerlinsley do you know when it is going to be implemented?

SeanCassiere:
Closing this for now, since we are currently pushing on Start grind.

https://x.com/tannerlinsley/status/1813747203686666434

When we begin parallel routes implementation, we'll be sure to document its usage.

PR #1835, opened 2024-06-26
fix(router-generator): add virtual parent for routes in addition to layouts

d-a-s:
Currently the file-based route generator has a bug with `route.tsx` files, which causes the "parent" path to be lost. Creating a virtual parent for `route.tsx` files (similarly to `_layout` files) fixes this problem so that the path is retained. Example:
```
post/route.tsx
post/index.tsx
post/$id.tsx
```
Currently the presence of the `route.tsx` file causes `post/index.tsx` to be mapped to url path `/`, instead of being properly mapped to `/post/`. If you attempt to use `/post/`, it actually gets evaluated by `post/$id.tsx`, and `$id = post`.

The documentation [file-based-routing#file-naming-conventions](https://tanstack.com/router/latest/docs/framework/react/guide/file-based-routing#file-naming-conventions) seems to indicate that this is a bug:
> When using directories to organize your routes, the `route` suffix can be used to create a route file at the directory's path. For example, `blog.post.route.tsx` or `blog/post/route.tsx` can be used at the route file for the `/blog/post` route.

This description doesn't match current behavior; this code change should fix it.


schiller-manuel:

schiller-manuel:
looks like the existing tests fail with this change

d-a-s:
Sorry about that, it turns out it only seems to be a problem if you had a route inside of a layout directory (e.g. `_layout-a/route.tsx`). I've altered the code, added some tests for that scenario, and updated the relevant snapshot.

PR #1843, opened 2024-06-27
ci: fix the workflow for running the e2e test on the "start-basic" example

callingmedic911:
`build:example` wasn't building right dep. This PR builds all of the packages before starting the tests.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -100,7 +100,7 @@ jobs:
         with:
           main-branch-name: 'main'
       - name: Build dependecies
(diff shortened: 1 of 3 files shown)


SeanCassiere:
@callingmedic911 let me know when you want something merged in.

callingmedic911:
I couldn't replicate initially too. But then cleaned up locally (`git clean -dfx`) and tests failed with the same error as CI after doing just `build:examples`.

@SeanCassiere this is ready for review.

SeanCassiere:
Thank you! @callingmedic911

PR #1857, opened 2024-06-29
docs: add quickstart guide for TanStack Start

crutchcorn:
This is an early quick-start guide for TanStack Start. I haven't added it to the routes configuration guide yet, as I'm not sure how public we want this guide to be (not against, just wanting to follow @tannerlinsley's guidance).

> This PR is related to and followed up by: https://github.com/TanStack/router/pull/1858

PR #1858, opened 2024-06-29
docs: add initial documentation for createServerFn

crutchcorn:
> This PR builds on top of #1857, we might want to merge that one first.

This PR documents usage of `createServerFn` from TanStack Start. Like the other PR, this one is intentionally not added to the sidebar, as it's unclear how public we want these draft documents.

### DIFF
--- a/docs/framework/react/guide/server-functions.md
+++ b/docs/framework/react/guide/server-functions.md
@@ -3,8 +3,21 @@ id: server-functions
 title: Server Functions
 ---
 
-As [TanStack Start](./tanstack-start.md) is a full-stack framework, we need a way to write code that executes on the
-server and can be called from the client.
+As [TanStack Start](./tanstack-start.md) is a full-stack framework, we need a way to call a function on the client which then executes some code on the server.
+
+This is where Server Functions come in.
+
+## What are Server Functions?
+
+> 🧠 You can think of Server Functions like mini-portals that let your users trigger a pre-defined action on the server (like a built-in RPC solution 🤯).
+
+Server Functions are actions that can be executed on the server from the client or from other server functions.
+
+When a server function is called, it will execute the server-side code and return the result to the caller (regardless of whether the caller is a client or another server function).
+
+Server Functions as more than just a way to send and receive data. They're like a bridge that lets you connect your backend code (the behind-the-scenes magic) directly to your frontend code (what the user sees), making it easier to build a seamless user experience.
+
+## How to use Server Functions in TanStack Start
 
 While some frameworks use special pragmas to denote a function that will execute on the server, Start uses a utility to
 create an instance of a server function.
@@ -58,7 +71,7 @@ async function someServerFunction() {
 When this function is called from the client, it will make a request to the server, execute the server-side code, and
 return the serialized result to the client.
 
-# Server Function Arguments
+## Server Function Arguments
 
 Like any function, server functions can take arguments:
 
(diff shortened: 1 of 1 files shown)


docs/framework/react/guide/server-functions.md
```diff
@@ -0,0 +1,245 @@
...
+  // `val` will be '123'
+  const val = formData.get('val');
+  // ...
+})
+
+function Component() {
+  return (
+    //  We need to tell the server that our data type is `multipart/form-data` by setting the `encType` attribute on the form.
+    <form action={yourFn.url} method="POST" encType="multipart/form-data">
+      <input type="hidden" name="val" value="123" />
```
SeanCassiere, line 163: I'd add a callout on how client-side fields like this should not be "trusted". Purely for security due-dilligence.

docs/framework/react/guide/server-functions.md
```diff
@@ -0,0 +1,245 @@
+---
+id: server-functions
+title: Server Functions
+---
+
+As [TanStack Start](./tanstack-start.md) is a full-stack framework, we need a way to write code that executes on the
```
SeanCassiere, line 6: I am not sure if it works with the rest of the document flow, but would likening it to dynamic-RPC could help readers understand server functions better.

Purely adding this in here, because of the mental model that people need to keep in mind, since these aren't just API-endpoints per-se. Rather, little portals through which server-side code can be executed through an API-surface the user controls.

Since currently our server functions are also "server actions", I just feel like we either need to explain a bit more from the get-go or possibly link to external resources.

docs/framework/react/guide/server-functions.md
crutchcorn: Oh! I didn't even realize `type="hidden"` was there. Removing!

docs/framework/react/guide/server-functions.md
crutchcorn: > I just feel like we either need to explain a bit more from the get-go

Let's do that - explain a bit more from the get-go.

Could you possibly work on adding in a section that explains what the differences are and the mental model folks need to be in when using these?

SeanCassiere: @crutchcorn I think what I've added gets the message across. Feel free to cut out any bits that feel a bit repetitive.

docs/framework/react/guide/server-functions.md
```diff
@@ -0,0 +1,258 @@
...
+
+## What are Server Functions?
+
+> 🧠 You can think of Server Functions like mini-portals that let your users trigger a pre-defined action on the server (like a built-in RPC solution 🤯).
+
+Server Functions are actions that can be executed on the server from the client or from other server functions.
+
+When a server function is called, it will execute the server-side code and return the result to the caller (regardless of whether the caller is a client or another server function).
+
+Server Functions as more than just a way to send and receive data. They're like a bridge that lets you connect your backend code (the behind-the-scenes magic) directly to your frontend code (what the user sees), making it easier to build a seamless user experience.
```
schiller-manuel, line 18: "as more" -> "are more"

SeanCassiere (approved):

schiller-manuel:
should this have been merged into `main` instead?

crutchcorn:
@schiller-manuel I think the idea is that I wanted us to merge the Start quickstart docs first, since this PR references that one.

PR #1870, opened 2024-07-01
docs: remove the unnecessary encode and decode `URIComponent` calls in the zipson example of the "custom search param serialization" guide

Engid:
Howdy! 

When using Zipson per this original example I had unexpected results. The URL produced was actually larger than the base-64 encoding example. I believe it was because `decodeURIComponent` and `encodeURIComponent` were being ran twice (ie: once in the lambda passed to `parseSearchWith` and once inside `decodeFromBinary`). Removing the duplicate function call seemed to do the trick. 

If anyone wants to try to reproduce it, copy-pasting from this change should be enough. 

Thanks!


Engid:
Thank you!

Issue #1638, opened 2024-05-20
Route component not showing


Maquinours:
### Describe the bug

After clicking on a link with has its preload as intent and the route does not have the time to load, the new route component will not show.

### Your Example Website or App

https://stackblitz.com/edit/vitejs-vite-sb9dg1?file=src%2Frouter.tsx

### Steps to Reproduce the Bug or Issue

Download the project and run it. Open it and restrict your connection to "slow 3g" then click on "enterprises" link. If you click instantly without leaving the time to the preload to end, then it will show you nothing, as if there wasn't a component associated to the link.

### Expected behavior

It should show the loader component and then, when the loader has finished, show the route component.

### Screenshots or Videos

https://gyazo.com/ada043a794530f4d4c044512523b18f1

### Platform

- OS: Windows
- Browser: Chrome/Firefox
- Version: 1.32.12

### Additional context

In this case, we need to use a restricted connection so the preload does not have the time to load before we click. In routes that has bigger components, the issue happens even with a good connection cause it couldn't load huge routes before we click.

schiller-manuel:
duplicate of #1534

Issue #1534, opened 2024-04-29
redirects thrown in the __root beforeLoad aren't showing the correct route on first-load


maksymskuibida:
### Describe the bug

I am facing an issue on admin panel I am developing now. Sometimes, I cannot find exact steps to reproduce it just returns nothing in Outlet. I checked all my code tree and there is no chance it is bug in my code. Then I reload the page and then I am facing the next issue. beforeLoad of my $lang route is called before __root.tsx beforeLoad what causes other bugs

### Your Example Website or App

admin.7loc.com

### Steps to Reproduce the Bug or Issue

If you want, you can 
1) register here,
2) create profile in form 
3) Navigate between pages on phone or with efficiency slowed down x6 + mobile mode in devtools

### Expected behavior

1) When I am on some route, and there is some content, Outlet have to return it always
2) __root.tsx beforeLoad should be called before other routes beforeLoad as well, as any parent beforeLoad should be called before beforeLoad of any children

### Platform

- OS: It happens more often on phone or if I slow dows processor x6 in browser devtools
- Browser: all
- Version: 1.30.1—1.31.2

tannerlinsley:
Please try with the latest version

Tanner Linsley

maksymskuibida:
1.31.2 is last available version. I am currently on it and it also happens. 
As __root beforeLoad is called before my $lang's beforeLoad when I try to use variable from context, which I return from __root beforeLoad, it is undefined. I use this variable to change lang in params, if it was changed, for example, in user object. So, I set undefined to params, what causes infinite pending state. I reload page, undefined is replacing to actual lang, but I don't know why it also falls into infinite pending and only if I reload the page for third time, it starts working. I fixed it a little bit by checking for if (variable) in $lang beforeLoad, so now when Outlet returns nothing, I reload the page and it works from first reload, but the issue with calling other routes beforeLoad before __root beforeLoad still exists. And issue with Outlet returns nothing still exists. 

Anyway thank you for your hard work on this project. You are doing a lot and present fixes so quick. I beleive issues will be fixed soon, because without this bugs it is perfect router

BourqueCharles:
@maksymskuibida By any chance, are you using redirects in your beforeLoad? I had an issue similar to your's and changing the redirect function to using the navigate function "fixed" those issues.

maksymskuibida:
[...]

PR #1876, opened 2024-07-02
Correctly load matched routes when navigated before preload is finished

freshgiammi:
Ref: https://github.com/TanStack/router/issues/1534
Ref: https://github.com/TanStack/router/issues/1638

### DIFF
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -1938,20 +1938,11 @@ export class Router<
                     if (match.isFetching === 'beforeLoad') {
                       // If the user doesn't want the route to reload, just
                       // resolve with the existing loader data
+                      // Otherwise, load the route
 
                       // if (match.fetchCount && match.status === 'success') {
                       //   resolve()
                       // }
-
-                      // Otherwise, load the route
-                      matches[index] = match = updateMatch(
-                        match.id,
-                        (prev) => ({
-                          ...prev,
-                          isFetching: 'loader',
-                          fetchCount: match.fetchCount + 1,
-                        }),
-                      )
 
                       lazyPromise =
                         route.lazyFn?.().then((lazyRoute) => {
@@ -1977,6 +1968,15 @@ export class Router<
                       // so we need to wait for it to resolve before
                       // we can use the options
                       await lazyPromise
+
+                      matches[index] = match = updateMatch(
+                        match.id,
+                        (prev) => ({
+                          ...prev,
+                          isFetching: 'loader',
+                          fetchCount: match.fetchCount + 1,
+                        }),
+                      )
 
                       checkLatest()
 
--- a/packages/react-router/tests/components/mockHeavyDependenciesRoute.tsx
+++ b/packages/react-router/tests/components/mockHeavyDependenciesRoute.tsx
@@ -1,0 +1,6 @@
+// This mimicks the waiting of heavy dependencies, which need to be streamed in before the component is available.
+await new Promise((resolve) => setTimeout(resolve, 5000))
+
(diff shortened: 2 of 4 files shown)


freshgiammi:
There's still an issue here when with context when it gets merged with beforeLoad.

```
export const Route = createFileRoute('/about')({
  beforeLoad: async () => {   
    await new Promise((resolve) => setTimeout(resolve, 5000));
    return { someData: 'hello' };
  },
   loader: async ({context}) => {
    await new Promise((resolve) => setTimeout(resolve, 2000));
    console.log(context.someData)
  },
})
```

Supposing `/about` is the lazy route that takes a long time to resolve, if beforeLoad isn't finished before the navigation event happens, the loader gets invoked with an `undefined` context, breaking the route.

This possibly happens because [execution flow is skipped](https://github.com/TanStack/router/blob/main/packages/react-router/src/router.ts#L1772) when a route already in `beforeLoad` fetching state is re-loaded (because the loader gets invoked regardless of the beforeLoad state, for which we have no pending information on reload)

schiller-manuel:
are you able to reproduce this in a test somehow? this would be awesom.

packages/react-router/src/link.tsx
```diff
@@ -545,6 +545,7 @@ export function useLinkProps<
   options: UseLinkPropsOptions<TRouter, TFrom, TTo, TMaskFrom, TMaskTo>,
 ): React.AnchorHTMLAttributes<HTMLAnchorElement> {
   const router = useRouter()
+  const isClicking = React.useRef(false)
```
schiller-manuel, line 548: why do we need to track if a link was clicked?
what happens if we have two link instances pointing to the same target, this then won't work, right?

i think the solution must be independent from click tracking etc, and also must work when programatically preloading / navigating without clicking on a link

schiller-manuel: this most likely needs to be fixed inside of `router.ts`

freshgiammi: This is a different issue than the references (so we could have it in a separate PR), but essentially `onFocus` is also invoked when a button is clicked, which causes the route to get preloaded twice (`onMouseEnter` + `onFocus`).

When focusing a link via keyboard navigation, we obviously skip `onMouseEnter` so it gets correctly preloaded once.
Given that there's no way to distinguish between mouse and keyboard origin for focus events, we can "trap" it by skipping the preload call when a focus event is detected between `onMouseDown` and `onMouseUp`: this makes preload only happen once and ensures that the behaviour is the same for both mouse and keyboard navigation.

freshgiammi:
> are you able to reproduce this in a test somehow? this would be awesom.

The issue I'm still facing in https://github.com/TanStack/router/pull/1876#issuecomment-2202756943 should be easily reproducible.

A test case for this preload issue when lazy components haven't finished fetching yet might be more complicated, but I haven't worked on either yet. Still, it could be something close to what I've done in the [reproduction template](https://stackblitz.com/edit/github-fsj9kg)

packages/react-router/src/link.tsx
schiller-manuel: ok then please put this in a separate PR then

freshgiammi: https://github.com/TanStack/router/pull/1885

freshgiammi:
Split https://github.com/TanStack/router/pull/1876#issuecomment-2202756943 into a separate issue: https://github.com/TanStack/router/issues/1886

schiller-manuel:
closing in favor of https://github.com/TanStack/router/pull/1907
I copied the test out of this PR, so thanks for this!

Issue #1886, opened 2024-07-03
preload: undefined context in loader if navigation happens before beforeLoad is done


freshgiammi:
### Describe the bug

There's still an issue here when with context when it gets merged with beforeLoad.

```
export const Route = createFileRoute('/about')({
  beforeLoad: async () => {   
    await new Promise((resolve) => setTimeout(resolve, 5000));
    return { someData: 'hello' };
  },
   loader: async ({context}) => {
    await new Promise((resolve) => setTimeout(resolve, 2000));
    console.log(context.someData)
  },
})
```

Supposing `/about` is the lazy route that takes a long time to resolve, if beforeLoad isn't finished before the navigation event happens, the loader gets invoked with an `undefined` context, breaking the route.

This possibly happens because [execution flow is skipped](https://github.com/TanStack/router/blob/main/packages/react-router/src/router.ts#L1772) when a route already in `beforeLoad` fetching state is re-loaded (because the loader gets invoked regardless of the beforeLoad state, for which we have no pending information on reload)

### Your Example Website or App

https://stackblitz.com/edit/github-nkiyw7

### Steps to Reproduce the Bug or Issue

1. Hover `About` link
2. Immediately click to navigate before the 5000ms beforeLoad is done executing
3. Undefined context in loader 🤯 

### Expected behavior

As a user, I expect context to not be undefined

### Platform

- OS: MacOS
- Browser: Chrome
- Version: 1.43.6

schiller-manuel:
this will be fixed by #1907 

reproducer with pre-release package: https://stackblitz.com/edit/github-nkiyw7-ojt8yk?file=src%2Froutes%2Fabout.tsx

freshgiammi:
> this will be fixed by #1907 
> 
> reproducer with pre-release package: https://stackblitz.com/edit/github-nkiyw7-ojt8yk?file=src%2Froutes%2Fabout.tsx

I guess I can close https://github.com/TanStack/router/pull/1891 then?

schiller-manuel:
yes we'll close #1891, but I will take over the test from there, so thanks for implementing that!

PR #1891, opened 2024-07-04
fix: ensure context is consistent on concurrent match loads

freshgiammi:
If a match has multiple load routines running (such as navigation happening while preload isn't finished yet), we might start the `loader` function too early when the context isn't ready for consumption yet.

This happens because [execution flow is skipped](https://github.com/TanStack/router/blob/main/packages/react-router/src/router.ts#L1772) when a route already in beforeLoad fetching state is re-loaded.

This patch adds beforeLoad tracking to matched routes, so that we can eventually await any pending beforeLoad function and ensure we have up to date context to feed into the `loader` function.

Fixes https://github.com/TanStack/router/issues/1886

### DIFF
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -1082,6 +1082,7 @@ export class Router<
           isFetching: false,
           error: undefined,
           paramsError: parseErrors[index],
+          beforeLoadPromise: createControlledPromise<void>(),
           loaderPromise: Promise.resolve(),
           loadPromise,
           routeContext: undefined!,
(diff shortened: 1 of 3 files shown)


packages/react-router/src/router.ts
```diff
@@ -1982,7 +1996,8 @@ export class Router<
                       checkLatest()
 
                       // Kick off the loader!
-                      loaderPromise = route.options.loader?.(loaderContext)
```
freshgiammi, line 1985: Weirdly enough, doing
```
matches[index] = match = updateMatch(
  match.id,
  (prev) => ({
    ...prev,
    loaderPromise: route.options.loader?.(loaderContext),
   }),
)
```

breaks `navigate` from working 🤔

schiller-manuel:
closing in favor of https://github.com/TanStack/router/pull/1907
I copied the test out of this PR, so thanks for this!

PR #1898, opened 2024-07-05
ci: Only preview on approved PRs

lachlancollins:


.github/workflows/preview.yml
```diff
@@ -14,6 +14,7 @@ env:
 jobs:
   preview:
     name: Preview
+    if: github.event.review.state == 'APPROVED'
```
SeanCassiere, line 17: If a member opens a PR, this means that another member would have to go in and approve the PR correct?

lachlancollins: Correct - however after looking at how other projects have used it, it's probably safe enough to not need this check and to just run on PR. New contributors already need to have workflows approved.

This is an early setup, and I'll probably tweak it over the coming days for all the TanStack repos (especially once this PR gets merged: https://github.com/stackblitz-labs/pkg.pr.new/pull/134

schiller-manuel: yeah I think this check is not necessary, would be great to remove it

PR #1713, opened 2024-06-05
fix(react-router): route-context being available and loaders being called after a redirection

SeanCassiere:
This PR initially was just to add tests for the route-context checks after navigation, but it became a bit more.

- It checks route context being available after a navigation.
- It checks route context being available after a redirection.
- It checks that the destination route loaders are being fired when redirected to.
- The tests introduced in #1710 are also being amended to better simulate a "real-world experience". No more relying on calls to `router.navigate()`, instead a button is clicked which calls `navigate` from `useNavigate()`.
- The tests in `packages/react-router/tests/redirects.test.tsx` related to route-context have been __REMOVED__ in favor of the ones in `packages/react-router/tests/routeContext.test.tsx` as they simulate real-world usage.

I've manually tested these changes against the reproductions in the following issues and can confirm that this PR fixes the following.

- #1730
- #1670 
- #1650 
- This [comment](https://github.com/TanStack/router/issues/1656#issuecomment-2145989592) in #1656

Context-related issues still remain broken in this though.

- The original reproduction in #1656

Thanks to @freshgiammi [pointing out](https://github.com/TanStack/router/pull/1662) the unnecessary `this.load()`.

PR #1907, opened 2024-07-06
fix(router): context issues

### LINKED ISSUE #1534: redirects thrown in the __root beforeLoad aren't showing the correct route on first-load

### LINKED ISSUE #1551: `notFoundComponent` doesn't render when `notFound` is thrown in the loader since 1.28.2
### Describe the bug

A route wrapped in `Suspense` either by setting the prop `wrapInSuspense` to `true` or defining a `pendingComponent` doesn't render the `notFoundComponent` or the `defaultNotFoundComopnent`  when the `notFound` function is thrown in the loader.

https://codesandbox.io/p/devbox/beautiful-archimedes-7vwmwj

To reproduce :
1. Go to index route
2. `Loading` is rendered instead of `Not found`

Things that doesn't trigger the bug :
- There is no Promise in the loader or the Promise is faster
- The prop `wrapInSuspense` is set to `false`
- The prop `pendingComponent` is not set

### Expected behavior

The prop `notFoundComponent` should be rendered when the `notFound` function is thrown in the loader even if the route is wrapped in `Suspense` and the Promise is taking longer to resolve.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Chrome
- Version: 124

### Additional context

_No response_

### LINKED ISSUE #1646: defaultPendingMinMs blocks the page on load
### Describe the bug

when hard-loading a page for the first time, it will show the loader (or an empty page) for the `pendingMinMs` time even if the route has no loader at all or the loader resolves before `pendingMs` has expired (default 1 second).

This is _not_ the case when navigating to a route

https://stackblitz.com/edit/tanstack-router-hzdror?file=src%2Fmain.tsx&preset=node

- reload the sandbox on the `Home` tab (this route has no loader)
- you should see an empty page for 5 seconds.
  - if you comment in the `defaultPendingComponent`, you should see `loading...` for 5 seconds.
- now go to the `about` tab (which has a 500ms loader, so shorter than the default 1s `pendingMs`) and reload it.
  - you should, again see the `loading...` (or an empty screen) for 5 seconds
- now navigate to `Home` via the link
  - home loads instantly, as expected, because it has no loader.

### Expected behavior

I'd expect `pendingMinMs` to not influence the hard reload times, just like it doesn't block navigations either.

### Screenshots or Videos

_No response_

### Platform

any

### Additional context

This behaviour is observable even with the default settings, but because `defaultPendingMinMs` is "just" 500ms, it's harder to spot.

### LINKED ISSUE #1656: Context is undefined when search validation fails

### LINKED ISSUE #1751: Context is undefined in weird case.
### Describe the bug

For some unknown reason context is undefined in loader, but it was present in beforeLoad

https://github.com/Arilas/tanter-bug

1. Clone
2. `yarn`
3. `yarn dev`
4. Open Host
5. Obserse error in `src/routes/_secured.tsx`
6. Refresh the page
7. Now it's working

### Expected behavior

It should not clear the context

### Screenshots or Videos

image

### Platform

- OS: macOS, iOS, Android
- Browser: Arc, Safari
- Version: latest

### Additional context

I've over-simplified our case to the reproducible state. In our case we have a ServiceLocator with multiple factories, dependencies, events.

`index.lazy.tsx` can be replaced with an route with `beforeLoad` which will have redirects, but it was an workaround to fix problem previously.

### LINKED ISSUE #1752: beforeLoad doesn't get called
### Describe the bug

I expect that every every time I enter the route beforeLoad will be called

https://stackblitz.com/edit/tanstack-router-94bxv9?file=src%2Froutes%2F_auth.tsx

1. Click "Go to the auth-only invoices page".
2. Enter any login and access the authorized part of the site.
3. Quickly click "Logout".
4. Quickly click "Go to the auth-only invoices page" again.
5. You are on the "Invoices" page without any authentication.

(You may need a few tries.)

### Expected behavior

On Step 4, loading happens again, and I'm redirected to the login page.

### Screenshots or Videos

_No response_

### Platform

```json
    "@tanstack/react-router": "^1.36.3",
    "@tanstack/router-devtools": "^1.36.3",
    "@tanstack/router-vite-plugin": "^1.38.0",
```

### Additional context

If it's expected behavior, how can I avoid it?

### LINKED ISSUE #1774: when `trailingSlash: always` loaders do not run if the trailing-slash is not present
### Describe the bug

From @TkDodo 

found a small issue with route loaders and `trailingSlash: always`:

the loaders don't run if I navigate to a route without a trailing slash, like `/about`. This can happen if the user types in the url manually. What happens is:

- the router appends a slash `/about/`
- it renders the right component
- but the loader of that component never runs.

reproduction: https://stackblitz.com/edit/tanstack-router-lhw7fj?file=src%2Fmain.tsx,src%2Froutes%2F__root.tsx,src%2Froutes%2Fabout.tsx&preset=node

if you navigate with the link to `About`, it will work, but if you open the sandbox preview in a new tab and go to `/about`, it doesn't work (I hope this link works): https://tanstackrouterlhw7fj-yhec--3001--9e2d28a3.local-credentialless.webcontainer.io/about (mising trailing slash on purpose)

it works fine with `trailingSlash: 'preserve'`

https://stackblitz.com/edit/tanstack-router-lhw7fj?file=src/main.tsx,src/routes/__root.tsx,src/routes/about.tsx&preset=node

❗ Navigating via `Links` works correctly, so you need to directly visit the URL.

- To directly to `/about` without the ending `/`.
- Observe loaders not running

### Expected behavior

Loaders should run 👍🏼

### Screenshots or Videos

_No response_

### Platform

- TSR Version: 1.38.1

### Additional context

_No response_

### LINKED ISSUE #1777: loader doesn't run after redirects
### Describe the bug

using the latest version, the loader with loaderDeps doesn't run initially.

the page needs to be reloaded for it to run.

Reproduction on version ^1.20.4 (✅ works fine)
[https://stackblitz.com/edit/github-p2w83v-xh9id6](https://stackblitz.com/edit/github-p2w83v-xh9id6)

Reproduction on the latest version (^1.38.1) (❌ doesn't run initially)
[https://stackblitz.com/edit/github-p2w83v-pz6euc](https://stackblitz.com/edit/github-p2w83v-pz6euc)

https://stackblitz.com/edit/github-p2w83v-pz6euc

update to the latest version

### Expected behavior

run initially, without having to reload the page.

### Screenshots or Videos

_No response_

### Platform

- Version: 1.38.1

### Additional context

_No response_

### LINKED ISSUE #1795: Throwing `notFound()` in loader causes infinite "pending" state
### Describe the bug

If you `throw notFound()` in a loader after a short delay (e.g. when awaiting a request), the parent routes stay forever stuck in "pending" state.

This behaviour does not happen if the loader promise resolves very quick, e.g. within 100ms.

Simplest reproduction:
```tsx
export const Route = createFileRoute('/posts/$postId')({
  loader: async ({ params }) => {
    try {
      // Simulate a slow request. Resolving this within 100ms works fine, but if it takes longer
      // the route state will be forever stuck in pending.
      await new Promise((resolve, reject) => {
        setTimeout(() => {
          reject();
        }, 1000);
      });
    } catch (err) {
      // Never shows
      throw notFound();
    }
  },

  notFoundComponent: () => <div>Post not found</div>,
});
```

https://stackblitz.com/edit/tanstack-router-jc3pyp?file=src%2Froutes%2Findex.tsx

1. Go to `/posts/whatever` in Stackblitz (`posts.$postId.tsx`)

### Expected behavior

A "Not found message"

### Actual behavior

The matched root route is stuck in a pending state according to the dev tools, the final route correctly has "not found" state.

This causes the page to never load.

### Platform

- OS: macOS
- Browser: Arc (Chrome)
- Version: 1.47

### Additional context

_No response_

### LINKED ISSUE #1796: Throwing `notFound()` in `beforeLoad()` already shows the error during preload
### Describe the bug

I don't even need to click the link, as soon as I hover over it, I get the "Not found" error message?

https://stackblitz.com/edit/tanstack-router-xxntjv?file=src%2Froutes%2Fbrowse.%24.tsx

1. Go to `/browse`
2. Hover over "Browse 404"

### Expected behavior

No error, just a preloading in the background. Only when I click the link will I see the not found error.

### Actual behavior
The 404 error is shown as soon as I hover over the link as a result of preloading that page.

### Platform

- OS: macOS + Arc 1.47

### Additional context

_No response_

### LINKED ISSUE #1814: When `validateSearch` overrides search param, lazy route won't load
### Describe the bug

When a `validateSearch` function gets executed and overrides a search param, the lazy loaded components for that route won't render.

With the following code:

/test/route.tsx
```code
export const Route = createFileRoute('/test')({
  validateSearch: z.object({
    status: z.enum(['one', 'two', 'three']).catch('one'),
  }),
});
```

/test/route.lazy.tsx
```code
import { createLazyFileRoute } from '@tanstack/react-router';

export const Route = createLazyFileRoute('/test')({
  component: TestComponent,
});

function TestComponent() {
  const { status } = Route.useSearch();

  return <div>Hello from test with {status}</div>;
}
```

If I try to manually update the `status` query param for a value that is not within the enum set. E.g: `?status=four`, `validateSearch` will catch the error and set it to 'one' but the route.lazy.tsx won't be rendered.

I created the following snippet. https://stackblitz.com/edit/tanstack-router-4bpt3z?file=src%2Froutes%2Ftest%2Froute.lazy.tsx

https://tanstackrouter4bpt3z-j5cp--3001--dc4d7514.local-credentialless.webcontainer.io/test?status=one

1. Go to https://tanstackrouter4bpt3z-j5cp--3001--dc4d7514.local-credentialless.webcontainer.io/test?status=one
2. Change `status` query param from one to four
3. the content of the /test route won't show up.

### Expected behavior

Expected behaviour would be to render the route.lazy.tsx component.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- @tanstack/react-router: 1.38.1

### Additional context

_No response_

### LINKED ISSUE #1827: Component being rendered before `beforeLoad()` has completed
### Describe the bug

After setting up a simple app with authentication, when I log in, the `beforeLoad()` function does not complete before my route is rendered. This causes an error as the context does not contain values I expect.

https://stackblitz.com/edit/tanstack-router-ep72yj?file=src%2Froutes%2F_auth.dashboard.tsx

1. Make sure you are in a logged out state
2. Click on the second link 'Go to the auth-only dashboard page'
3. Enter a username and click 'Login'

After an expected delay an error occurs. Since the `beforeLoad()` has not completed, the `value` property has not been added to the context.

### Expected behavior

I would expect the `beforeLoad()` to finish running before the `loader()` runs or the component is rendered.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Edge
- Version: 126.0.2592.68 (Official build) (64-bit)

### Additional context

This issue seems to happen in part due to the fact that when the user logs in, the `InnerApp` (in main.tsx) rerenders due to a state update in the AuthContext. This results in the RouterProvider being re-rendered. This re-render happens in addition to the call to `router.navigate()` from the login.

### LINKED ISSUE #1882: BeforeLoad causing infinite pending state when redirecting to current nested route
### Describe the bug

I've noticed that when you are on a route, and then you hover over another link with preload: 'intent' enabled, and said route throws a redirect to the current route you are on, an infinite pending state is set on the router. For example

User is on `/my-services/hosting` and hovers over a link for `my-services`, and the route for my-services has a before load something like this:
```
if (location.pathname === '/my-services') {
    throw redirect({ to: '/my-services/hosting', replace: true });
}
```

When this happens, the router goes into a pending state and never leaves it. This also means that when no default pending component is specified, react does the default behavior of setting all direct children to display: none, so troubleshooting this becomes even harder because the dev tools are hidden 😭

While I do acknowledge that realistically, there should not be a button on `/my-services/hosting` that takes you to a route that is intended to redirect, I also encountered this while in the process of refactoring our application so ideally this case can be handled. 

I believe that the route has to be a nested route for this to happen. It also does not happen for any nested route, only if you are on the nested route that is being redirected in the prefetch. 

I've attached a stackblitz below that should encounter the error when you are on the `about` route and hover the `/` route link. Clicking any route refreshes the pending state and resolves the issue

Note: I also noticed this older thread which is very similar:
https://github.com/TanStack/router/issues/1480

**Edit**:
Obviously this can be worked around for now using the following check if necessary for anyone else having this issue
```
if (location.pathname === '/my-services' && cause !== 'preload') {
    throw redirect({ to: '/my-services/hosting', replace: true });
}
```

https://stackblitz.com/edit/tanstack-router-b6npj2?file=src%2Fmain.tsx

1. Create Route
2. Create Child Route
3. In the parent route, in the before load, redirect the user to the child route
4. Visit the child route
5. Hover a link to the Parent route

### Expected behavior

As a user, I expect the the application to continue rendering the page but the page enters a permanent pending state.

### Screenshots or Videos

https://stackblitz.com/edit/tanstack-router-b6npj2?file=src%2Fmain.tsx

### Platform

- OS: Windows
- Browser: Chrome@latest & Arc@latest (For windows)

### Additional context

_No response_

### LINKED ISSUE #1886: preload: undefined context in loader if navigation happens before beforeLoad is done

### LINKED ISSUE #1892: `useRouteContext` returns undefined when navigating with `<Navigate />`
### Describe the bug

Context is undefined when using `<Navigate />` to redirect.
The bug does not occur if you remove React.StrictMode
The bug does not occur if you remove the empty beforeLoad from postsRoute on line 46

https://stackblitz.com/edit/tanstack-router-fuvvzs

1. run the app
2. wait 1 second for the setTimeout to trigger
3. check logs in console, context is undefined

### Expected behavior

context to be an object

### Screenshots or Videos

_No response_

### Platform

- OS: mac
- Browser: Chrome
- Version: any

### Additional context

Could be related or a duplicate of:
https://github.com/TanStack/router/issues/1886
https://github.com/TanStack/router/issues/1751
https://github.com/TanStack/router/issues/1740
but non of those mentions `<Navigate />`

### LINKED ISSUE #1899: `defaultPendingComponent` always rendered once on load
### Describe the bug

defaultPendingComponent always rendered once on load, even if there are no loaders

https://stackblitz.com/edit/tanstack-router-ho2fk8?file=src%2Fmain.tsx&preset=node

1. Run the app
2. Default pending component is rendered
3. Home route is rendered

### Expected behavior

1. Run the app
2. Home route is rendered

### Screenshots or Videos

_No response_

### Platform

- OS: mac
- Browser: Chrome
- Version: any

### Additional context
related to: https://github.com/TanStack/router/issues/1349 but this issue is for initial load only

_No response_

### LINKED ISSUE #1928: If `validateSearch` rewrites URL, router nav works but initial load fails (with strange behavior)
### Describe the bug

If the validation function not only validates but also returns something other than what it got (e.g. get empty input, return object with defaults), then the URL gets rewritten and page load fails. When using router navigation the URL is not rewritten and page load is successful.

What I mean by "page load fails": Not sure exactly what is going wrong, but in my repro the manifestation is that `throw notFound()` gets ignored and the route's match status is erroneously "success".

The repro uses a `validateSearch` function from the docs so I assume it's supposed to work.

https://stackblitz.com/edit/tanstack-router-uyzm48?file=src%2Froutes%2Fposts.%24postId.tsx

1. Use one of the links to get to a `/posts/<something>` route (WORKS)
1. Refresh (DOESN'T WORK)
    - See that it was automatically rewritten to `/posts/<something>?page=1&filter=&sort=newest` (I think this is the thing that isn't handled well)
1. Refresh again (WORKS)
    - It wasn't rewritten this time because it was already filled

### Expected behavior

Full page load and router navigation should share the same URL rewriting (or non-rewriting) behavior. They should result in the same page behavior as well. A validateSearch function that returns extra defaults or something else not strictly the same as its input should not cause a `throw notFound()` error to be ignored or any other strange behavior.

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Chrome
- Version: 126.0.6478.127

### Additional context

_No response_

### LINKED ISSUE #1929: Slow initial load
### Describe the bug

- At the time of writing, `latest` version is v1.44.0
- The initial load of a skeleton React app using the router is perceivably slow (~500ms from React profiler)
- [v1.28.1](https://github.com/TanStack/router/releases/tag/v1.28.1) is the latest good version (good as in not perceivably slow) which suggests that [v1.28.2](https://github.com/TanStack/router/releases/tag/v1.28.2) introduced a regression

https://stackblitz.com/edit/tanstack-router-vjwuwc?file=index.html,package.json,package-lock.json,src%2Fmain.tsx,vite.config.js

- Bad: https://stackblitz.com/edit/tanstack-router-vjwuwc?file=index.html,package.json,package-lock.json,src%2Fmain.tsx,vite.config.js
- Good: https://stackblitz.com/edit/tanstack-router-u6raco?file=index.html,package.json,package-lock.json,src%2Fmain.tsx,vite.config.js

The only difference between Good and Bad is the router version.

1. Refresh the page on the StackBlitz browser
2. Notice that `This sentence is from React` loads instantly in the Good version, almost at the same time as `This sentence is from HTML`. 
3. Notice that `This sentence is from React` loads after a delay in the Bad version, ~500ms after `This sentence is from HTML`

### Expected behavior

As a user, I expected no perceivable delays on initial app load

### Screenshots or Videos

_No response_

### Platform

- macOS
- Chrome Version 126.0.6478.127 (Official Build) (arm64)

### Additional context

_No response_

tannerlinsley:
It looks big, yes. But it boils down to:
- Better encapsulation for the different states of the match loading process. That’s really just moving some code into functions like `runBeforeLoad`, etc
- A change to use `getMatch()` and `updateMatch` exclusively, instead of reading and writing to a mix of local and state variables. This has already gotten rid of a few areas where we were definitely reading from stale data (but likely not getting any bugs reported)
- Tracking and deduping promises for beforeLoad. This is a very important piece to the fix, since prior to this, we were skipping the before load stage if it was in progress and immediately running the loader. This tracking is what enables each load to dedupe and share in-flight promises for both beforeLoad, component loading, lazy loading, and ultimately loader calls.

fixes #1892 
fixes #1886 
fixes #1752
fixes #1752
fixes #1827
fixes #1534
fixes #1777 
fixes #1814
fixes #1656
fixes #1796
fixes #1774 
fixes #1751
fixes #1882
fixes #1795
fixes #1551
fixes #1928
fixes #1899
fixes #1646
fixes #1929


packages/react-router/src/router.ts
```diff
@@ -1545,6 +1543,7 @@ export class Router<
           location: next,
           checkLatest: () => this.checkLatest(promise),
           onReady: async () => {
+            // eslint-disable-next-line ts/require-await
```
schiller-manuel, line 1546: since we are awaiting this, shouldn't `this.startViewTransition` then return a promise?

tannerlinsley: eslint complains that there is no await 🤷

packages/react-router/src/router.ts
```diff
@@ -1797,91 +1779,146 @@ export class Router<
...
                   abortController: new AbortController(),
                 }))
               }
 
-              if (match.paramsError) {
-                handleSerialError(match.paramsError, 'PARSE_PARAMS')
-              }
+              const runBeforeLoad = async () => {
+                if (getMatch().beforeLoadPromise) {
+                  return await getMatch().beforeLoadPromise!
```
schiller-manuel, line 1802: remove `!`

packages/react-router/src/router.ts
schiller-manuel: if I remove the `await` and this disable comment, eslint does not complain

packages/react-router/src/router.ts
tannerlinsley: TypeScript still thinks it's necessary though, since it *could* have changed somehow during the second `getMatch` call. I hate this about TS, and it's one of the annoyances with using TS with Solid.

schiller-manuel: I just noticed this when testing the PR locally... well then we'll keep it

packages/react-router/tests/routeContext.test.tsx
```diff
@@ -381,6 +381,42 @@ describe('beforeLoad in the route definition', () => {
...
+    await router.load()
+
+    // Don't await, simulate user clicking before preload is done
+    router.preloadRoute(aboutRoute)
+
+    await router.navigate(aboutRoute)
+    await router.invalidate()
+
+    // Expect double call: once from preload, once from navigate
+    expect(mock).toHaveBeenCalledTimes(2)
```
schiller-manuel, line 417: @tannerlinsley is this expectation correct?

packages/react-router/tests/routeContext.test.tsx
```diff
@@ -381,6 +381,42 @@ describe('beforeLoad in the route definition', () => {
...
+      },
+    })
+
+    const routeTree = rootRoute.addChildren([aboutRoute, indexRoute])
+    const router = createRouter({ routeTree, context: { foo: 'bar' } })
+
+    await router.load()
+
+    // Don't await, simulate user clicking before preload is done
+    router.preloadRoute(aboutRoute)
```
SeanCassiere, line 411: In addition to this test, I think we should also be testing the user hover event from testing-library.

Setting `defaultPreload: 'intent'` and doing `element.focus()` would also simulate a "real" interaction.

packages/react-router/src/index.tsx
```diff
@@ -183,6 +183,7 @@ export {
   PathParamError,
   getInitialRouterState,
   defaultSerializeError,
+  getRouteMatch,
```
SeanCassiere, line 186: We avoid exporting this to the public, yes?

packages/react-router/src/router.ts
```diff
@@ -1797,91 +1778,146 @@ export class Router<
...
+                  isFetching: 'beforeLoad',
+                  beforeLoadPromise,
+                  loadPromise,
                   routeContext: replaceEqualDeep(
-                    match.routeContext,
+                    prev.routeContext,
                     parentContext,
                   ),
-                  context: replaceEqualDeep(match.context, parentContext),
+                  context: replaceEqualDeep(prev.context, parentContext),
```
SeanCassiere, line 1821: @tannerlinsley what is the difference between `context` and `routeContext` here? As in, what's purpose of having both?
I never fully understood the reasoning for both existing.

packages/react-router/src/router.ts
```diff
@@ -1797,91 +1778,146 @@ export class Router<
...
+
+              if (paramsError) {
+                handleSerialError(paramsError, 'PARSE_PARAMS')
+              }
+
+              if (searchError) {
+                handleSerialError(searchError, 'VALIDATE_SEARCH')
+              }
+
+              try {
```
SeanCassiere, line 1915: I'd decorate this with a comment like "actually fire off the `beforeLoad` callback", purely for readability for others in the future.

packages/react-router/src/router.ts
```diff
@@ -1890,235 +1926,251 @@ export class Router<
...
+                  const parentMatchPromise = matchPromises[index - 1]
+                  const route = this.looseRoutesById[routeId]!
+
+                  const {
+                    params,
+                    loaderDeps,
+                    abortController,
+                    context,
+                    cause,
+                  } = getMatch()
```
SeanCassiere, line 1943: Should this `getMatch()` call be moved so it gets invoked inside the `getLoaderContext()` function?
Functionally does it make a difference?

packages/react-router/src/router.ts
```diff
@@ -1890,235 +1926,251 @@ export class Router<
...
-                  abortController: match.abortController,
-                  context: match.context,
-                  location,
-                  navigate: (opts) =>
-                    this.navigate({ ...opts, _fromLocation: location }),
-                  cause: preload ? 'preload' : match.cause,
-                  route,
-                }
+            validResolvedMatches.forEach(
+              ({ id: matchId, routeId, params: p }, index) => {
```
SeanCassiere, line 1930: I think we can remove this destructuring of `params: p` since we don't use `p` anywhere downstream.

packages/react-router/src/router.ts
```diff
@@ -1890,235 +1926,251 @@ export class Router<
...
+                  // If the route is successful and still fresh, just resolve
+                  const { status, invalid } = getMatch()
+
+                  if (
+                    status === 'success' &&
+                    (invalid || (shouldReload ?? age > staleAge))
+                  ) {
+                    ;(async () => {
+                      try {
+                        await fetchWithRedirectAndNotFound()
```
SeanCassiere, line 2158: Does this still need to be within a try-catch block considering `fetchWithRedirectAndNotFound` already uses one within the function definition?
Asking since we don't use a try-catch block below at `if (status !== 'success') {`.

packages/react-router/tests/routeContext.test.tsx
SeanCassiere: I think possibly not, since the `beforeLoad` is being deduped here: https://github.com/TanStack/router/pull/1907/files#diff-2ac234f610fa0e0c2146252c10a5266af9c5b6d0b63659a0fa4c9f2c4472596aR1801

SeanCassiere:
@tannerlinsley This in the `Transitioner` was previously changed from `useLayoutEffect` to `useEffect` in an attempt to fix the context issues. Here: https://github.com/TanStack/router/pull/1713/files#diff-8f39ee7f0cd19573624baf67df23085405045c8e446263697011394594e56e17R32

For the function of subscribing to the current location, should this go back to being a `useLayoutEffect`? or remain as a `useEffect`?

examples/react/quickstart-file-based/src/routes/about.tsx
```diff
@@ -3,6 +3,14 @@ import { createFileRoute } from '@tanstack/react-router'
 
 export const Route = createFileRoute('/about')({
   component: AboutComponent,
+  beforeLoad: async () => {
```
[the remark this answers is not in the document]
SeanCassiere: Resolved in https://github.com/TanStack/router/pull/1908

packages/react-router/tests/createLazyRoute.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    router.preloadRoute({ to: '/heavy' })
+    await router.navigate({ to: '/heavy' })
+
+    await router.invalidate()
+
+    expect(router.state.location.pathname).toBe('/heavy')
+
+    const lazyRoute = router.routesByPath['/heavy']
+
+    expect(lazyRoute.options.component).toBeDefined()
```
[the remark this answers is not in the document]
SeanCassiere: Resolved in https://github.com/TanStack/router/pull/1908

packages/react-router/tests/routeContext.test.tsx
SeanCassiere: Resolved in https://github.com/TanStack/router/pull/1908

packages/react-router/src/router.ts
SeanCassiere: Resolved in https://github.com/TanStack/router/pull/1908

packages/react-router/src/router.ts
SeanCassiere: Resolved in https://github.com/TanStack/router/pull/1908

packages/react-router/src/router.ts
tannerlinsley: Let’s do that then

mirague:
Great to see this! 🌟

Does this also aim to address #1795?

packages/react-router/tests/routeContext.test.tsx
schiller-manuel: i changed the test to expect a single call

packages/react-router/tests/routeContext.test.tsx
```diff
@@ -381,6 +381,42 @@ describe('beforeLoad in the route definition', () => {
...
+    const rootRoute = createRootRoute()
+    const indexRoute = createRoute({
+      getParentRoute: () => rootRoute,
+      path: '/',
+    })
+    const aboutRoute = createRoute({
+      getParentRoute: () => rootRoute,
+      path: '/about',
+      beforeLoad: async () => {
+        await sleep(WAIT_TIME) // Use a longer delay here
```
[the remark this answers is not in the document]
schiller-manuel: not necessary,I removed that comment

packages/react-router/src/index.tsx
schiller-manuel: I removed the export

packages/react-router/src/router.ts
tannerlinsley: ... My gut says yes, because there can be synchronous errors that we need to catch, not just async ones

packages/react-router/src/router.ts
tannerlinsley: Done

tannerlinsley:
> @tannerlinsley This in the `Transitioner` was previously changed from `useLayoutEffect` to `useEffect` in an attempt to fix the context issues. Here: https://github.com/TanStack/router/pull/1713/files#diff-8f39ee7f0cd19573624baf67df23085405045c8e446263697011394594e56e17R32
> 
> For the function of subscribing to the current location, should this go back to being a `useLayoutEffect`? or remain as a `useEffect`?

I think we should subscribe as soon as possible. So even if we need to split them up, let's do that.

mrlubos:
This looks pretty good! One thing I am still struggling to understand is how does `notFound()` work @tannerlinsley? I have this `route.tsx` file which isn't an actual path, it's a scaffolding for other matched routes.

```tsx
import { createFileRoute, notFound } from '@tanstack/react-router'

export const Route = createFileRoute('/_loggedin/_layout/_adminuser')({
  beforeLoad: async ({ context }) => {
    const canShow = await context.checkIfCanShow()

    if (!canShow) {
      throw notFound()
    }
  },
  component: () => <div>A</div>,
  errorComponent: () => <div>B</div>,
  notFoundComponent: () => <div>C</div>,
  pendingComponent: () => <div>D</div>,
})
```

I would expect/like to render "C" when `notFound()` is thrown, but this doesn't work. In the older versions (1.33.x), actually nothing would get rendered, the whole component tree was gone (😳). In this pull request, I see the default Not Found text and the warning

>A notFoundError was encountered on the route with ID "__root__", but a notFoundComponent option was not configured, nor was a router level defaultNotFoundComponent configured

If I comment out the `throw notFound()` part, I see "A" rendered as expected. Any hints?

schiller-manuel:
@mrlubos if you throw a `notFound` in a `beforeLoad`, only a single not found component can be shown.
Currently, this PR will always invoke the root not found component and **not** the route's one.

@tannerlinsley told me we could change this to show the route's not found component. Could we also make it bubble up so it chooses the nearest not found component in case the route, whose `beforeLoad` had thrown the `notFound`, does not have a not found component?

mrlubos:
I'd like that @schiller-manuel. I will admit that I still have knowledge gaps around how layout routes are matched as my understanding is they're not a part of the same tree as regular routes and I thought that might be causing the problem here. Ultimately my goal is to be able to throw and handle from within the same component (I tried using the routeId option in notFound() but it made no difference). I have a layout component I'm using to check permissions for certain user type and if the user doesn't have sufficient permissions, they should be shown the 404 page without revealing the existence of specific routes.

packages/react-router/src/router.ts
chorobin: I thought one is the current route context (`routeContext`) where the other is merged from the parent. At least, this is how the types work

chorobin (approved):

mrlubos:
This pull request can't be released fast enough. I discovered another issue introduced in 1.36.3 where a simple `<Link to="/path" />` keeps reverting me back to the current path. Unfortunately, it's not obvious why or how to debug, but this problem didn't exist prior to 1.36.3, is present in the latest 1.44.4, and I can confirm is resolved in this pull request 🎉

flipvh:
> This pull request can't be released fast enough. I discovered another issue introduced in 1.36.3 where a simple `<Link to="/path" />` keeps reverting me back to the current path. Unfortunately, it's not obvious why or how to debug, but this problem didn't exist prior to 1.36.3, is present in the latest 1.44.4, and I can confirm is resolved in this pull request 🎉

thats sounds very similar to problems I am experiencing too! Will do a one time bonus donation if this big PR gets done 😃 ! thnks!

packages/history/src/index.ts
```diff
@@ -7,6 +7,7 @@ export interface NavigateOptions {
 }
 export interface RouterHistory {
   location: HistoryLocation
+  subscribers: Set<() => void>
```
SeanCassiere, line 10: Should we straight away use the function-getter pattern?

We always end up refactoring down the line to use this pattern anyways 😅 

```suggestion
  getSubscribers () => Set<() => void>
```

tannerlinsley: I think this one is fine as is. It *should* be very low traffic for a long time.

schiller-manuel:
@tannerlinsley there is currently a bug that causes routes with loaders to sometimes (?) keep pending

I forked the `basic-file-based` example and installed the package of this PR:

https://stackblitz.com/edit/tanstack-router-2zgjft

1. click on posts
2. click on various posts
3. some load and display, others keep the previous post displayed or just show a blank page

https://github.com/user-attachments/assets/8cef0772-568d-4c56-8d11-2a58b9d63424

mrlubos:
@schiller-manuel i also noticed my app became way less performant, but as always hard to say why, I didn't dig into the internals

sepirak-nick:
> i also noticed my app became way less performant, but as always hard to say why, I didn't dig into the internals

I've noticed that with intent preloading enabled, my application hangs very noticeably during the preload (hover over a link).

ruisaraiva19:
> > i also noticed my app became way less performant, but as always hard to say why, I didn't dig into the internals
> 
> I've noticed that with intent preloading enabled, my application hangs very noticeably during the preload (hover over a link). 

I have the same degraded performance outcome with preload enabled.

schiller-manuel:
@ruisaraiva19 can you please provide a minimal complete example by forking one of the existing examples on Stackblitz that shows this performance degradation?

ruisaraiva19:
> @ruisaraiva19 can you please provide a minimal complete example by forking one of the existing examples on Stackblitz that shows this performance degradation?

Yes, I'll do it later today. Just to be clear, the performance issue happens specifically when hovering links (triggering the preloading)

mrlubos:
Nice find, I believe it's the same for me, I use the same preload setting

martylukyy:
I can confirm performance degradation after running a snapshot of this PR too.
Will wait for @ruisaraiva19 example to provide an example for this.
In the meanwhile here are some flame graphs to showcase how much the degradation in our case:

oldandfast

newandslow

(Ignore the entries generated by browser addons - it doesn't look much better without them.)

tannerlinsley:
> I can confirm performance degradation after running a snapshot of this PR too. Will wait for @ruisaraiva19 example to provide an example for this. In the meanwhile here are some flame graphs to showcase how much the degradation is in our case:
> 
> Old and fast: oldandfast
> 
> New and slow: newandslow
> 
> (Ignore the entries generated by browser addons - it doesn't look much better without them.)

I’m seeing some function calls that I recently removed in the latest version of this PR. Is this up to date?

ruisaraiva19:
@schiller-manuel @tannerlinsley [Here](https://stackblitz.com/edit/neie9x?file=src%2Froutes%2Fdashboard.index.tsx) is a bare minimum demo based on the `kitchen-sink-react-query-file-based` example. The Dashboard index route is the one with the code changes.

A couple of notes regarding the demo:

- At first, I couldn't recreate the performance issue with a bare `kitchen-sink-react-query-file-based` example (maybe too simple of an app?).
- After debugging with a couple of console logs + the React Profile flame graphs, I noticed that route components are rerendered multiple times when hovering links.
	- With `defaultPreload: false`, route components would render **5** times on the first hover event and **3** times on the following hover events.
	- With `defaultPreload: 'intent'`, route components would render **7** times on the first hover event and **5** times on the following hover events.
	- So with `defaultPreload: 'intent'`, we get a few extra rerenders.
	- This can be confirmed by looking at the console logs when inside the Dashboard index route.
- My experience with the performance issue happens on a web app with complex dashboards.
	- To confirm this with a minimal demo, I created a `SlowComponent` component that takes at least 5ms to render to simulate a complex component.
	- After rendering a few component instances, I reproduced the same performance issue when hovering over links like on the web app mentioned above. You can check that when hovering over a link, the color change (from blue to red) is not responsive.
- Then, I wrapped the `DashboardIndexComponent` component in a `React.memo` and the rerender performance issue is gone since the route component does not rerender on the hover events anymore.
	- I tried the `React.memo` approach on the web app mentioned above and it fixed the issue as well which uses version `1.43.10` of the router.

To summarize, it seems that when hovering over links the router is rerendering pretty much all `Outlet` components, hence the `React.memo` approach to fix the issue.

Please let me know if any of my conclusions aren't correct and feel free to ask any questions!

tannerlinsley:
Good find. I can take a look at these perf regressions later tonight.

martylukyy:
> I’m seeing some function calls that I recently removed in the latest version of this PR. Is this up to date?

I just remembered i pinned a certain commit.
Will redo this without pinning a certain commit but pinning the PR today or tomorrow.

martylukyy:
Made another comparison with our app now without the addons.
Result is pretty much the same in terms of the factor of the performance degradation. The addons just add overall.
It's still a performance degradation compared to the old version from last months npm update cycle we did,
even with the newer commit.

1.34.9:
oldandfast2

commit [64a8bdd](https://github.com/TanStack/router/pull/1907/commits/64a8bddb8ad384f276409334dea8946598aa568c):
newandslow2

tannerlinsley:
Sweet. I’ll take a look. I already have a good idea of where to start.

Tanner Linsley
TanStack LLC

tannerlinsley:
I just pushed a few fixes to the MatchInner reactivity. Can you wait for that PR version to release and give it a shot?

martylukyy:
> I just pushed a few fixes to the MatchInner reactivity. Can you wait for that PR version to release and give it a shot?

Looks a lot better to me now!
Performance measurement is done without addons again to get a better read on it.
Here is another flame graph:

newandfast

tannerlinsley:
> > I just pushed a few fixes to the MatchInner reactivity. Can you wait for that PR version to release and give it a shot?
> 
> Looks a lot better to me now! Performance measurement is done without addons again to get a better read on it. Here is another flame graph:
> 
> newandfast

Better or worse than original?

martylukyy:
Let me build 2 versions real quick so that devtools won't interfere that much in the measurement
and get back to you in a few minutes!

martylukyy:
It's a worse by a one digit millisecond margin compared to 1.34.9.
Didn't attach a flame graph of the production versions since there isn't much to read out of there.
If you want numbers we are talking about 155ms +- 5ms scripting time measured between 1.34.9 and 0f34fbe.

Chance are, this difference comes from other changes made since 1.34.9 and it's on par with what it used to be.

Given that it solves so many bugs it's better than the original. 😄

tannerlinsley:
Perfect! Just need to get tests passing and we’re good. I don’t have anything else to push.

ruisaraiva19:
> * After debugging with a couple of console logs + the React Profile flame graphs, I noticed that route components are rerendered multiple times when hovering links.

> To summarize, it seems that when hovering over links the router is rerendering pretty much all Outlet components, hence the React.memo approach to fix the issue.

@tannerlinsley regarding the above mentioned. Is my assumption correct? If so, is this behavior expected (by design) or not really?

mrlubos:
@tannerlinsley is there any checklist on what's remaining until this gets merged? Do you think it will be released by the next week?

schiller-manuel:
this needs to be fixed 
https://github.com/TanStack/router/pull/1907#issuecomment-2226991746

and of course test need to pass

schiller-manuel:
@tannerlinsley this was probably a mistake so I added that `any` back:

https://github.com/TanStack/router/pull/1907/commits/2a85a587be896576dc221ceb8a89ec206a823823

schiller-manuel (approved):

mrlubos:
Thank you 🕺

PR #1910, opened 2024-07-07
ci: Add local TS version tests

lachlancollins:
Uses the strategy implemented in TanStack Form, Store, and Query to take full advantage of Nx.

This strategy can be tested and debugged locally rather than needing to run in CI (e.g. the matrix approach). It also all runs in a single GitHub Action job, making it easier to work out which task has failed (all logs are in a unified Nx cloud report).


packages/react-router/tsconfig.legacy.json
```diff
@@ -0,0 +1,4 @@
+{
```
SeanCassiere, line 1: What's the purpose of this file?

lachlancollins: It only runs type tests on the src files for TS 5.0-5.4, while TS 5.5 tests src and tests. This cuts down the task's time dramatically.

SeanCassiere (approved):

PR #1911, opened 2024-07-07
ci: Use distributed Nx agents for e2e tests

lachlancollins:
This changes the `"test:e2e"` tasks to run on the distributed Nx agents, rather than on its own GitHub job. This means the tasks are cached and parallelised optimally.

Unfortunately, this isn't compatible with the `"@replayio/playwright"` setup (it uses the standard `"@playwright/test"` setup), but I'm not sure how much this was being used - if this is a necessary part of the workflow, then this change won't work.


SeanCassiere (commented):
Currently, it's a good sanity check to make sure the Start examples (`start-basic` and `start-basic-react-query`), still work correctly after changes have been made to any of our packages. _In the future, we should probably expand it to crawl our `kitchen-sink` or `quickstart` examples as well._

The replay.io is quite useful for actually watching the playback of what happened during the tests (especially in the failed cases).

@lachlancollins from what I can gather, this [run](https://github.com/TanStack/router/actions/runs/9823513309/job/27121489942) only failed because an API key was missing. Could the job, not inherit the API key from the workflow?

Also, @jasonLaster @callingmedic911, could you please reconfirm that having this particular API key being publicly viewable is OK? And do we need different API keys for each example project? Currently, on the replay dashboard I'm only seeing the tests being recorded for `start-basic` but not the ones for `start-basic-react-query`.
image
image

SeanCassiere:
Closing for now

jasonLaster:
@SeanCassiere Yes - it's fine for the API key to be publicly available

PR #1883, opened 2024-07-03
discussion: plugging in a convex database

thomasballinger:
This PR is here because code diffs are a nice walk to discuss things (and I accidentally opened it here instead of on a fork). Really excited about the new SSR suspense stuff so wanted to try it out. Wondering if this API seems reasonable: a `convexQueryOptions()` function and requiring settings a queryFn and queryKeyHashFn globally.

This example uses https://github.com/thomasballinger/convex-react-query-draft for live updates, but that isn't apparent in this example since there are no mutations.

PR #1962, opened 2024-07-15
examples: Use ensureQuery in start-basic-react-query

thomasballinger:
Use ensureQuery for loaders so that if the data is not stale it won't need to be refetched.

As suggested in https://github.com/TanStack/router/pull/1883/files#r1667101376


thomasballinger:
If I'm understanding the difference right then this should be used everywhere in loaders, yeah?

PR #1964, opened 2024-07-15
fix(react-router): URI path segments segments should match on equality regardless of their encoding state

### LINKED ISSUE #1959: NotFound route when refreshing or manual url to a route with special characters
### Describe the bug

Using the file based routing with file names/routes that have special characters, in my case it is accented characters such as â or é etc., causes not found issues. (Running using vite and the file based plugin)

Manually routing to a page, i.e. typing in a url and hitting enter, or refreshing or opening in a new tab, throws notfound and consequentally will render the 404 or notfound component. 

I can easily fix it by renaming my files to use english characters only but thought I would flag it anyway.

https://stackblitz.com/edit/github-sxumx6?file=src%2Froutes%2Fex%C3%A2mple.tsx,src%2Froutes%2F__root.tsx

Link that actually throws not found instead of routing:
https://wmaibnrpmngithub-zclp--3001--70dbe416.local-corp.webcontainer.io/exâmple

1. Create a new route / file with special characters like é
2. Let routetree.gen generate
3. Route to page using app to check it is working
4. Then refresh, open in new tab or manually retype url to reproduce issue (will throw a 404 instead of routing)

### Expected behavior

It should route to the page but instead 404s the exisiting route

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 91.1]

### Additional context

_No response_

SeanCassiere:
Closes #1959

### DIFF
--- a/packages/react-router/src/path.ts
+++ b/packages/react-router/src/path.ts
@@ -175,7 +175,7 @@ export function parsePathname(pathname?: string): Array<Segment> {
 
       return {
         type: 'pathname',
-        value: part,
+        value: decodeURIComponent(part),
       }
     }),
   )
--- a/packages/react-router/tests/router.test.tsx
+++ b/packages/react-router/tests/router.test.tsx
@@ -50,39 +50,6 @@ function createTestRouter(initialHistory?: RouterHistory) {
   }
 }
 
-function createTestRouterWithObjects(initialHistory?: RouterHistory) {
-  const history =
-    initialHistory ?? createMemoryHistory({ initialEntries: ['/'] })
-
-  const rootRoute = createRootRoute({})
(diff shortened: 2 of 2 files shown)


braineo:
@SeanCassiere it seems this fix breaks other encoded url see https://github.com/TanStack/router/issues/2046

Suppose there is a route defined as `/somePath/$someParam`

now manually entering `/somePath/sub%2Ffield` in the address bar and enter will automatically transit to `/somePath/sub/field`, which breaks the navigation

braineo:
turns out to be a trivial fix. I opened a pr https://github.com/TanStack/router/pull/2261

@SeanCassiere could you take a look at it?

SeanCassiere:
@braineo sorry I didn't see this till now. Notifications on the closed issues often get a bit buried in the github notifications 😅

PR #1968, opened 2024-07-15
fix(react-router): use current matches for `fromMatches` and only build new matches when location overridden

### LINKED ISSUE #1852: Dynamic invoice route in kitchen sink examples break (`could not find match...`)
### Describe the bug

When you open one of the kitchen sink examples, the route breaks when you navigate to the invoices. Given the fact that I experience issues with dynamic routes in another codebase (that I haven't been able to succesfully reproduce in a small reproduction) this may indicate there's an underlying bug.

https://stackblitz.com/github/tanstack/router/tree/main/examples/react/kitchen-sink-file-based?embed=1&theme=light&preset=node&file=src/main.tsx

1. Open the [official kitchen-sink example](https://stackblitz.com/github/tanstack/router/tree/main/examples/react/kitchen-sink-file-based?embed=1&theme=light&preset=node&file=src/main.tsx).
2. Click on Dashboard › Invoices › select an invoice.
3. Note that the invoice is loaded and the url is updated properly.
4. Click on 'Home'.
5. Click on Dashboard › Invoices › select an invoice.
6. Note that the invoice is NOT loaded and the URL is NOT updated properly (flips back to /invoices).
7. The error displayed in the console says: "Could not find match for from: /dashboard/invoices/$invoiceId"

### Expected behavior

As a user I expect the route to be properly loaded the second time also.

### Screenshots or Videos

https://github.com/TanStack/router/assets/487182/60e99303-744c-4811-9a12-9396a265cc65

### Platform

- OS: macOS
- Browser: All

### Additional context

_No response_

chorobin:
Due to the nature of `useLayoutEffect` and `resolvedLocation` lagging behind, it seems possible that a match can render before the `resolvedLocation` has been updated. Threfore I've decided just not to rely on it at all and to use the current matches for `fromMatches` and only build new matches if the location has been overridden.

Fixes #1852


chorobin:
I need to add a test to replicate this. Might be a bit hard but definitely not merge yet

chorobin:
Fixes #1852

SeanCassiere (commented):
Are there any specific scenarios in which you've run into this problem? and have you checked this against the examples?

P.S. You might want to run a `pnpm install` and re-run prettier again, those formatting changes below shouldn't be happening.

chorobin:
> Are there any specific scenarios in which you've run into this problem? and have you checked this against the examples?
> 
> P.S. You also, might want to run a `pnpm install` and re-run prettier again, those formatting changes below shouldn't be changing.

It was an issue in the kitchen sink example

SeanCassiere:
> It was an issue in the kitchen sink example

What are the reproduction steps? I'd like to wrap my head around the issue at play here.

chorobin:
> > It was an issue in the kitchen sink example
> 
> What are the reproduction steps? I'd like to wrap my head around the issue at play here.

See this https://github.com/TanStack/router/issues/1852

chorobin:
Also not sure whats going on with prettier as I've nuked my `node_modules`. Will fix it tomorrow

SeanCassiere:
> See this #1852

Thanks! Yup, can see the issue clear as day.

SeanCassiere:
> Also not sure whats going on with prettier as I've nuked my `node_modules`. Will fix it tomorrow

👍🏼 I've run prettier and pushed up the changes.

SeanCassiere (approved):

schiller-manuel:
@SeanCassiere we need a test first before we merge this. would be awesom if you can contribute this.

SeanCassiere:
> @SeanCassiere we need a test first before we merge this. would be awesom if you can contribute this.

A bit easier said than done 😅

I'm unable to reliably reproduce the error even in between runs.

chorobin:
> > @SeanCassiere we need a test first before we merge this. would be awesom if you can contribute this.
> 
> A bit easier said than done 😅
> 
> I'm unable to reliably reproduce the error even in between runs even on `main`.
> 
> @schiller-manuel

I could reliable produce it but now theres been a larger PR merged, it only happens sometimes. It must be to do with `resolvedLocation` being updated after though if its inconsistent

chorobin:
Ok. To me this looks like it happens actually when a preload is successful and its successfully cached a match. Likely because it flips to the match immediately, there is no 'pending' state and `resolvedLocation` is not correct

SeanCassiere:
@chorobin I'm able to sometimes get the issue on main, but it never comes up when testing with these changes.

Perhaps update this branch with main and give it a shot. If all is good, then I'd push this through.

tannerlinsley (approved):

schiller-manuel (approved):

PR #1969, opened 2024-07-16
test(react-router): active state for `<Link>` is correct when a search parameter value is set to `undefined`

### LINKED ISSUE #1920: Active link based on undefined search param
### Describe the bug

If you have a `Link` with `search={{ foo: undefined }}`, it will NOT be "active" when the query param `foo` is not present in the URL.

In my app, I have a navigation bar which controls a query param (`foo`). The first link unsets this query param, while subsequent links set it to different values. I want these links to have active status based just on this query param (disregarding other query params).

The default matching works great for the links which have a value for the query param. For the "home" link, I could use `activeOptions={{ exact: true }}` so that the link isn't active if the `foo` query param has a value. However, it also won't be active when other query params have a value.

Ideally I could set `search={{ foo: undefined }}` on the link. However, this does not actually match when the `foo` query parameter isn't present in the URL.

https://stackblitz.com/edit/github-7rehud-pn5er4?file=src%2Froutes%2F__root.tsx

1. Click the `Foo = a` and `Foo = b` links. Observe that they become active.
2. Click the `Bar = a` link. Observe that it becomes active, and the `Foo = a|b` link remains active.
3. Click the `Exact` link, then the `Bar = a` link. Observe that `Bar = a` becomes active, but `Exact` is no longer active.
4. Click the `Foo = undefined` link. Observe that it does NOT become active.

### Expected behavior

The `Foo = undefined` link should be active when the `foo` query parameter is not present in the URL.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Safari, Chrome
- Version: all

### Additional context

_No response_

SeanCassiere:
Fixes #1920


packages/react-router/src/utils.ts
```diff
@@ -209,6 +209,32 @@ export function isPlainArray(value: unknown): value is Array<unknown> {
   return Array.isArray(value) && value.length === Object.keys(value).length
 }
 
+/**
+ * This function will remove all undefined values from an object.
+ * It will also remove any undefined values from nested objects.
+ * It will not remove undefined values from arrays.
+ */
+export function deepRemoveUndefinedFromObject<T>(obj: T): T {
```
schiller-manuel, line 217: can we, instead of this function, enhance `deepEqual` to ignore properties that are explicitly set to `undefined`?

SeanCassiere: I assume you object properties yes? Not to strip out `undefined` as an array value, only as an object value, when comparing `[undefined, 'foo', { a: undefined }]` with `['bar', undefined, { a: undefined }]`.

I was hesitant to make such changes to the `deepEqual` function itself since that'd have an impact on the `router.matchRoute` and `router.commitLocation` methods, especially without understanding the conditions in which `undefined` would get either pass or fail the function.

It's why I introduced that function since it explicitly does a single job and has a narrow impact area (only the link component).

SeanCassiere: @schiller-manuel thoughts on having the `deepEqual` function internally call the `deepRemoveUndefinedFromObject` function on both its inputs `a` and `b`?

SeanCassiere: Yup, that's what'll work best here. I'll beef up those tests for the `deepRemoveUndefinedFromObject` function and push this.

schiller-manuel:
using `replaceEqualDeep` to implement `deepEqual` would be much easier, see this PR 

https://github.com/TanStack/query/pull/999/files

(even though they removed it since they did not need it in query, the implementation is still visible)

code-elf:
Since this change, is there an easy way of checking whether a specific search parameter is explicitly _not_ set (without `exact` which also rules out unrelated params)? The only way I've found is something like this:

```tsx
...
<Link
  ...
  activeOptions={{ exact: true }}
  search={{ ...location.search, param: undefined }}
>
```
...which is obviously not super readable (plus `exact` has other side effects). Thanks in advance!

schiller-manuel:
@code-elf  can you please create a new discussion entry for this question?

PR #1984, opened 2024-07-18
fix(react-router): deepEqual correctly handles undefined properties

schiller-manuel:


SeanCassiere (commented):
That's much simpler! I'd probably just switch to using Object.values to avoid indexing into the object.
Will make that change.

SeanCassiere (approved):
Nope, this approach works best after reading the source.

PR #1987, opened 2024-07-19
ci: use distributed nx agents for e2e tests

SeanCassiere:
From #1911 

> This changes the `"test:e2e"` tasks to run on the distributed Nx agents, rather than on its own GitHub job. This means the tasks are cached and parallelised optimally.
> 
> Unfortunately, this isn't compatible with the `"@replayio/playwright"` setup (it uses the standard `"@playwright/test"` setup), but I'm not sure how much this was being used - if this is a necessary part of the workflow, then this change won't work.


SeanCassiere:
Pause a moment, the tests on `start-basic-react-query` needs to be fixed.

lachlancollins (changes requested):
Leaving a "request changes" because GitHub Actions seem to be having a moment (the last 5 workflows haven't launched for some reason).

I also suspect this won't work right now, as Nx specifically looks at the root package.json to see if @playwright/test is present to decide if it should install chromium: https://github.com/nrwl/nx-cloud-workflows/blob/main/workflow-steps/install-browsers/main.js

SeanCassiere:
All good now!

lachlancollins (approved):

PR #1988, opened 2024-07-19
Add Convex DB example

thomasballinger:
Use the (still alpha) `@convex-dev/react-query` library to demonstrate live updates and hydration via TanStack Query client.


thomasballinger:
https://github.com/user-attachments/assets/d504c046-dc0e-4cfa-839d-6eb24b1efc5f

SeanCassiere:
If we include a Convex example, it should probably be separate from the existing Trellaux example. We should keep the existing example "as-is" in terms of data storage, since we want people to be able to view the example on the documentation website without having to add in an API key 😅.

Could you copy over this example under `examples/react/start-trellaux-with-convex`?

SeanCassiere (changes requested):
If we include a Convex example, it should probably be separate from the existing Trellaux example. We should keep the existing example "as-is" in terms of data storage, since we want people to be able to view the example on the documentation website without having to add in an API key 😅.

Could you copy over this example under `examples/react/start-trellaux-with-convex`?

thomasballinger:
Makes sense! Thanks @SeanCassiere

thomasballinger:
Other examples say

> It's deployed automagically with Vercel!

Is this set up?

I'm thinking the default setup can be to talk to an existing backend with creating new cards disabled so all folks can do it reorder to there's no need to moderate it, similar to the algolia example in this repo. Then when run locally both options are available, you can create your own backend by signing up for Convex or talk to the public one.

SeanCassiere:
> Is this set up?

Tanner cloned the tanstack.com repo and stripped it down into the first "start-basic" example. The examples that followed all just copied the start-basic example. So these READMEs were inherited from there.

> the algolia example in this repo?

Do we have an algolia example 🤔?

> I'm thinking the default setup can be to talk to an existing backend with creating new cards disabled so all folks...

This PR looks good to merge as it is.

Perhaps deployment is something to discuss with @tannerlinsley. Maybe a solution here would be to host this example and have a corn (or something) set-up that resets the DB every 30 minutes or so to a known "good" state.

thomasballinger:
Cleaned up the READMEs, that makes sense.

Sorry about that, no the [Algolia example](https://github.com/TanStack/query/tree/main/examples/react/algolia) I'm thinking of is in TanStack/query!

For now there's nothing special about deployment, so actually running this requires entering an email address during the `pnpm dev` command will require signing up for a Convex account. If we figure out hosting I can make a version that resets with a [corn](https://docs.convex.dev/scheduling/cron-jobs) or similar, and when running locally that can either be a default or we'll get local dev set up. Right now running Convex locally requires downloading a binary manually, I'd like to make it npm-installable in the next month or two so there's never a signup step required.

SeanCassiere (approved):
I'll release this as a patch, since the `@tanstack/react-router-with-query` package has some minor changes going through.

thomasballinger:
Give me a sec to make sure those are intended changes

thomasballinger:
Ah the hashKey stuff, yeah I think those are changes we want. But happy to pull those out, either way.

SeanCassiere:
> Ah the hashKey stuff, yeah I think those are changes we want. But happy to pull those out, either way.

No it's fine. It's a minor change, that has a fallback.

PR #2000, opened 2024-07-21
feat(react-router): support input types for `search` validation schemas

chorobin:
### Why

Previously we have always inferred the `Output` type for `zod` and friends. This causes TS headaches when using `Link` when a route definition has transformations in a schema.

For example, given a route definition like this:

```tsx
export const Route = createFileRoute('/search/searchPlaceholder')({
  component: SearchComponent,
  validateSearch: z.object({
    searchPlaceholder: z.literal('searchPlaceholder'),
    page: z.number().optional().default(0),
    offset: z.number(),
    search: z.string(),
  }),
})
```

When navigating to this route, the `page` `search` param is always required, despite the fact it is an optional **input**.

```tsx
     <Link
        to="/search/searchPlaceholder"
        className="block py-1 text-blue-800 hover:text-blue-600"
        search={{
          searchPlaceholder: 'searchPlaceholder',
          page: 0, // page is required here :(
          offset: 10,
          search: 'search',
          rootSearch: 0,
        }}
      >
        Search
      </Link>
```

This was because we never attempted to infer the input type and we had a workaround with a tag type of `SearchSchemaInput`.

With this PR `Link` and other navigation now uses  the `zod` `Input` type and `useSearch`  uses the `zod` `Output` type.

That means `page` in the above example is now optional.

~~The approach taken is we take advantage of TS being structurally typed, so we can infer `_input` and `_output` from the `zod` schema without having a dependency on `zod`~~

### Update

This PR introduces search validation adapters to separate validation libraries from the core of tanstack router while keeping the ability to infer `input` and `output` types. This change is made with no breaking changes to the types and we're just extending the types to also allow adapters.

A route definition can use an adapter by wrapping the validation schema. For example, with `zod` usage:

```typescript
import { zodSearchValidator } from '@tanstack/react-router-with-zod'

export const Route = createFileRoute('/search/searchPlaceholder')({
  component: SearchComponent,
  validateSearch: zodSearchValidator(
    z.object({
      searchPlaceholder: z.literal('searchPlaceholder'),
      page: z.number().optional().default(0),
      offset: z.number(),
      search: z.string(),
    }),
  ),
  loader: (opts) =>
    opts.context.queryClient.ensureQueryData(searchQueryOptions),
})

```

Using `zodSearchValidator` will infer the `input` type for navigation like `Link` and infer the `output` type for `useSearch`

```tsx
      <Link
        to="/search/searchPlaceholder"
        className="block py-1 text-blue-800 hover:text-blue-600"
        search={{
          searchPlaceholder: 'searchPlaceholder',
          page: 0, // this is optional
          offset: 10,
          search: 'search',
          rootSearch: 0,
        }}
      >
        Search
      </Link>
```

`search` here is the `zod` `input` type which means `page` is optional

```tsx
     <Link
        from={Route.fullPath}
        to="/search/searchPlaceholder"
        className="block py-1 text-blue-800 hover:text-blue-600"
        search={(prev) => prev}
      >
        Search
      </Link>
```
`prev` here is the `zod` `output` type while the return type is `input` type

```typescript
 const search = Route.useSearch()
```

`search` here is the `output` type.

### Solving the `catch` problem

A design decision by `zod` means that when a `catch` is used in validation schema, the `input` type becomes `unknown`

```typescript
export const Route = createFileRoute('/search/searchPlaceholder')({
  component: SearchComponent,
  validateSearch: zodSearchValidator(
    z.object({
      searchPlaceholder: z.literal('searchPlaceholder'),
      page: z.number().optional().default(0).catch(0),
      offset: z.number(),
      search: z.string(),
    }),
  ),
  loader: (opts) =>
    opts.context.queryClient.ensureQueryData(searchQueryOptions),
})
```
```tsx
      <Link
        to="/search/searchPlaceholder"
        className="block py-1 text-blue-800 hover:text-blue-600"
        search={{
          searchPlaceholder: 'searchPlaceholder',
          page: 0, // this is now unknown :(
          offset: 10,
          search: 'search',
          rootSearch: 0,
        }}
      >
        Search
      </Link>
```

The `page` type is lost to `unknown`. Here I'm proposing a few solutions to this problem:

1. Accept this is a limitation of `zod` and maybe consider an alternative if keeping the type is important to you. `valibot` does not have the same issue with their `fallback` variation
2. Use a custom schema for a `fallback/catch` that keeps the type information
3. Change the `input` type to be the `output` type for navigation

### Use a custom schema for a `fallback/catch` that keeps the type information

This problem can be solved using a custom schema like so

```typescript
export const fallback = <TSchema extends z.ZodTypeAny>(
  schema: TSchema,
  fallback: TSchema['_input'],
) => {
  return z.custom<TSchema['_input']>().pipe(schema.catch(fallback))
}

export const Route = createFileRoute('/search/searchPlaceholder')({
  component: SearchComponent,
  validateSearch: zodSearchValidator(
    z.object({
      searchPlaceholder: z.literal('searchPlaceholder'),
      page: fallback(z.number(), 0).optional().default(0),
      offset: z.number(),
      search: z.string(),
    }),
  ),
  loader: (opts) =>
    opts.context.queryClient.ensureQueryData(searchQueryOptions),
})
```

Here we're creating a custom `fallback` schema which forces the `input` type to remain and pipes a dummy `custom` schema to the actual `schema` which uses `catch`.

The result is the type we expect for `input`

The `fallback` custom schema can be provided by the adapter package?

```tsx
   <Link
        to="/search/searchPlaceholder"
        className="block py-1 text-blue-800 hover:text-blue-600"
        search={{
          searchPlaceholder: 'searchPlaceholder',
          page: 0, // page is optional and a number (yay!) :)
          offset: 10,
          search: 'search',
          rootSearch: 0,
        }}
      >
        Search
      </Link>
```

### Change the `input` type to be the `output` type for navigation

The `zod` adapter is configurable so it possible to use the `output` type for `navigation` by passing a options object to `zodSearchValidator`

```typescript
export const Route = createFileRoute('/search/searchPlaceholder')({
  component: SearchComponent,
  validateSearch: zodSearchValidator({
    schema: z.object({
      searchPlaceholder: z.literal('searchPlaceholder'),
      page: fallback(z.number(), 0).optional().default(0),
      offset: z.number(),
      search: z.string(),
    }),
    input: 'output',
  }),
  loader: (opts) =>
    opts.context.queryClient.ensureQueryData(searchQueryOptions),
})
```

The `input` type has now been set to the `zod` output type which means for navigation (`Link`, `useNavigate`), `page` is required again as it was previously and keeps the `number` type

```tsx
    <Link
        to="/search/searchPlaceholder"
        className="block py-1 text-blue-800 hover:text-blue-600"
        search={{
          searchPlaceholder: 'searchPlaceholder',
          page: 0, // page is now the output type and is required again
          offset: 10,
          search: 'search',
          rootSearch: 0,
        }}
      >
        Search
      </Link>
```
### TS Performance

The performance of tanstack router has been improved by resolving the full search schema to when we need it, rather than it being a defaulted type parameter. TS will instantiate all defaulted type parameters

Before
```
> tsc --extendedDiagnostics

Files:                         648
Lines of Library:            38680
Lines of Definitions:        83589
Lines of TypeScript:         21604
Identifiers:                135323
Symbols:                    309502
Types:                      104016
Instantiations:             740250
Memory used:               333716K
Assignability cache size:    48247
Identity cache size:          8856
Subtype cache size:           2426
Strict subtype cache size:     413
I/O Read time:               0.09s
Parse time:                  0.59s
Program time:                0.94s
Bind time:                   0.26s
Check time:                  5.27s
Total time:                  6.47s
```

After
```
> tsc --extendedDiagnostics

Files:                         648
Lines of Library:            38680
Lines of Definitions:        83583
Lines of TypeScript:         21604
Identifiers:                135323
Symbols:                    309441
Types:                      101158
Instantiations:             705433
Memory used:               331772K
Assignability cache size:    48334
Identity cache size:          8654
Subtype cache size:           2426
Strict subtype cache size:     413
Parse time:                  0.56s
ResolveModule time:          0.13s
Program time:                0.88s
Bind time:                   0.25s
Check time:                  5.08s
Total time:                  6.21s
```


chorobin:
Left to do
[x]  - Zod input types
[  ] - Valibot input type
[  ] - Arktype input type
[  ] - Type tests
[  ] - Runtime tests
[  ] - Optimise types

packages/react-router/src/route.ts
```diff
@@ -291,18 +284,16 @@ type LdJsonValue = LdJsonPrimitive | LdJsonObject | LdJsonArray
...
 
-// The parse type here allows a zod schema to be passed directly to the validator
-export type SearchSchemaValidator<TInput, TReturn> =
-  | SearchSchemaValidatorObj<TInput, TReturn>
-  | SearchSchemaValidatorFn<TInput, TReturn>
-
-export type SearchSchemaValidatorObj<TInput, TReturn> = {
-  parse?: SearchSchemaValidatorFn<TInput, TReturn>
+export interface ZodSearchValidator {
+  _input: any
```
chorobin, line 288: This is the trick. We can infer any schema which has an `_input` property

schiller-manuel:
I don't think putting zod specific (and other library specific stuff) into the core library. I would make this an optional package that a user of that specific library can install.

packages/react-router/src/route.ts
schiller-manuel: how stable is this? what if zod renames this at some point?

chorobin:
> I don't think putting zod specific (and other library specific) stuff into the router core library. I would make this an optional package that a user of that specific library can install.

What is `zod` specific here? `ZodSearchValidator` is our own type which is 'zod like' but is not type from `zod`? Much like how we accepted an object with `parse` method. We did this to support `zod`, I am assuming.

packages/react-router/src/route.ts
schiller-manuel: this won't work for schemas using `catch`: https://www.typescriptlang.org/play/?ssl=14&ssc=1&pln=1&pc=1#code/JYWwDg9gTgLgBAbwF4F84DMoRHA5EiAE1wG4AoMgYwgDsBneOygCwFMQBDAdWBmYBFW6DgFcANvAC8cJADoIAIwBWrSjAAUCdBAgAuGbIZRgNAObqAlLMJDRE9bgUcouCygvkyMAJ5hWcAGUARjhpHz8IdDgmNk4ePkFhcRgvX38ASRDpYIBtXAB9EzARGFwAXTIAekq4WoA9AH4KanpGFnZuXmYAYQ4YFlCDRRU1TW09AyMTcytKPpYHJxc3Dwpw-wCAJkH1yOj2uK7e-uZUvzh07ezNvMKaYtKK6tq4RrIgA

schiller-manuel:
You are right about the `parse` method, but it wasn't a good design IMO.

> What is zod specific here?

The usage of zod internal structure that might or might not change in the future.
Let's assume `_input` is renamed, would we update router? if yes, this would mean users who update router would also need to update zod. or we try to support old versions, but this will get complicated.

> I'm not sure about the extra package. How would this look? And how would you infer the types from it?

I haven't thought about this in detail. Can we use declaration merging to bring in "type logic" from a separate package into router?

One more thought:
I assume if we add support for various validation libraries, this would slow down type checking even though a user would likely only use a single one.

packages/react-router/src/route.ts
chorobin: This seems by design of `zod` :(

schiller-manuel: I know, but what I mean by this: it maybe is not as simple as it seems. I use `catch` a lot, so I could not use this (directly).
Maybe we need much more complicated type logic to fully support zod's "quirks".

chorobin:
> You are right about the `parse` method, but it wasn't a good design IMO.
> 
> > What is zod specific here?
> 
> The usage of zod internal structure that might or might not change in the future. Let's assume `_input` is renamed, would we update router? if yes, this would mean users who update router would also need to update zod. or we try to support old versions, but this will get complicated.
> 
> > I'm not sure about the extra package. How would this look? And how would you infer the types from it?
> 
> I haven't thought about this in detail. Can we use declaration merging to bring in "type logic" from a separate package into router?
> 
> One more thought: I assume if we add support for various validation libraries, this would slow down type checking even though a user would likely only use a single one.



> You are right about the `parse` method, but it wasn't a good design IMO.
> 
> > What is zod specific here?
> 
> The usage of zod internal structure that might or might not change in the future. Let's assume `_input` is renamed, would we update router? if yes, this would mean users who update router would also need to update zod. or we try to support old versions, but this will get complicated.
> 
> > I'm not sure about the extra package. How would this look? And how would you infer the types from it?
> 
> I haven't thought about this in detail. Can we use declaration merging to bring in "type logic" from a separate package into router?
> 
> One more thought: I assume if we add support for various validation libraries, this would slow down type checking even though a user would likely only use a single one.

Maybe there is a way with an extra package but this is definitely all tradeoffs, we either accept the validation schema directly or wrap it like this?

```tsx

import {zod} from '@tanstack/react-router-zod`

export const Route = createFileRoute('/search/searchPlaceholder')({
  component: SearchComponent,
  validateSearch: zod(z.object({
    searchPlaceholder: z.literal('searchPlaceholder'),
    page: z.number().optional().default(0),
    offset: z.number(),
    search: z.string(),
  })),
})
```

packages/react-router/src/route.ts
chorobin: That might get ugly! D:

schiller-manuel: see e.g. https://tanstack.com/router/latest/docs/framework/react/guide/search-params

> In the above example, we used Zod's .catch() modifier instead of .default() to avoid showing an error to the user because we firmly believe that if a search parameter is malformed, you probably don't want to halt the user's experience through the app to show a big fat error message. That said, there may be times that you do want to show an error message. In that case, you can use .default() instead of .catch().

chorobin: This is annoying because `valibot` actually behaves how we expect here for `Input` with their `fallback` variation. `Input` is the correct thing to use here but its not how we expect it to work

chorobin:
> You are right about the `parse` method, but it wasn't a good design IMO.
> 
> > What is zod specific here?
> 
> The usage of zod internal structure that might or might not change in the future. Let's assume `_input` is renamed, would we update router? if yes, this would mean users who update router would also need to update zod. or we try to support old versions, but this will get complicated.
> 
> > I'm not sure about the extra package. How would this look? And how would you infer the types from it?
> 
> I haven't thought about this in detail. Can we use declaration merging to bring in "type logic" from a separate package into router?
> 
> One more thought: I assume if we add support for various validation libraries, this would slow down type checking even though a user would likely only use a single one.

In terms of performance theres. This is currently worse but is likely to get better if we move the resolving of `input` and `output` to later, not in a defaulted type parameter. 

Likely with the seperate package approach, TS still need to look inside an interface of a base type likely.

```
// validators can implement this base interface 
interface SearchValidator<TInput, TOutput> {
}

type ResolveSearchSchemaInput<TSearchValidator extends SearchValidator> = TSearchValidator extends SearchValidator<infer TInput, any> ? TInput : never
```

packages/react-router-with-zod/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+export type ZodSearchValidatorAdapter<
+  T extends ZodTypeLike,
+  TInput extends InputOutputOption,
+  TOutput extends InputOutputOption,
+> = SearchValidatorAdapter<
+  'input' extends TInput ? T['_input'] : T['_output'],
+  'output' extends TOutput ? T['_output'] : T['_input']
+>
+
+export const zodSearchValidator = <
```
chorobin, line 31: What do people think about this naming?

packages/react-router-with-zod/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  return {
+    types: {
+      input: input === 'output' ? schema._output : schema._input,
+      output: output === 'input' ? schema._input : schema._output,
+    },
+    parse: (input) => schema.parse(input),
+  }
+}
+
+export const fallback = <TSchema extends z.ZodTypeAny>(
```
chorobin, line 50: Should we have this custom schema in the adapter or should be maintained in user land?

packages/react-router-with-zod/src/index.ts
SeanCassiere: Cool with it.

packages/react-router-with-zod/tests/index.test-d.ts
```diff
@@ -0,0 +1,98 @@
+import {
```
SeanCassiere, line 1: We should add some runtime unit tests (in addition to the type tests here) to confirm that the validator helper is functioning as intended.

SeanCassiere (commented):
Can we confirm that this doesn't break the type inference for anyone already `validateSearch` in its current form?

```ts
{
  validateSearch: (inp) => searchSchema.parse(inp),
  validateSearch: searchSchema,
  validateSearch: (inp: Record<string, unknown>): SearchOutputType => {
  // ...
  return {} satisfies SearchOutputType
  },  
}
```

SeanCassiere:
It's expected that the preview workflow will fail since the new package hasn't been published on npm yet.

chorobin:
> Can we confirm that this doesn't break the type inference for anyone already `validateSearch` in its current form?
> 
> ```ts
> {
>   validateSearch: (inp) => searchSchema.parse(inp),
>   validateSearch: searchSchema,
>   validateSearch: (inp: Record<string, unknown>): SearchOutputType => {
>   // ...
>   return {} satisfies SearchOutputType
>   },  
> }
> ```

I do have type tests that are passing. I will check if I've covered all of it

package.json
```diff
@@ -72,6 +72,7 @@
       "@tanstack/router-plugin": "workspace:*",
       "@tanstack/router-vite-plugin": "workspace:*",
       "@tanstack/react-router-with-query": "workspace:*",
+      "@tanstack/react-router-with-zod": "workspace:*",
```
schiller-manuel, line 75: quick discussion of naming of the library: 

form uses this pattern: @tanstack/zod-form-adapter

should we use a similar one?

schiller-manuel:
a full blown example in router-examples would be nice

schiller-manuel:
docs should also be added

package.json
SeanCassiere: I'm not a fan of the adapter name being before the project name tbh. 

I'd be more inclined towards `router-zod-adapter`.

packages/react-router/src/route.ts
tannerlinsley: @colinhacks, what would your advice be?

package.json
chorobin: My thought process is we already named -with-query like this

schiller-manuel: I prefer the adapter one

packages/router-arktype-adapter/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "node": ">=12"
+  },
+  "devDependencies": {
+    "@testing-library/jest-dom": "^6.4.6",
+    "@testing-library/react": "^16.0.0",
+    "@tanstack/react-router": "workspace:*",
+    "arktype": "^2.0.0-beta.2"
+  },
+  "peerDependencies": {
+    "@tanstack/react-router": ">=1.43.2"
```
schiller-manuel, line 73: should we also add `arktype` as peer dependency to express compatibility with some version?
e.g. what happens if the internal structure of arktype changes as of some version?

packages/router-arktype-adapter/tests/index.test-d.ts
```diff
@@ -0,0 +1,95 @@
...
+    .parameter(0)
+    .toHaveProperty('search')
+    .parameter(0)
+    .toEqualTypeOf<{ page: number } | { page: number; indexPage: number }>()
+
+  expectTypeOf(rootRoute.useSearch()).toEqualTypeOf<{ page: number }>()
+  expectTypeOf(rootRoute.useSearch()).toEqualTypeOf<{ page: number }>
+})
+
+test('when using valibot schema with function', () => {
```
schiller-manuel, line 55: valibot => arktype

packages/router-arktype-adapter/tests/index.test.tsx
```diff
@@ -0,0 +1,80 @@
+import * as v from 'valibot'
```
schiller-manuel, line 1: remove

packages/router-arktype-adapter/tests/index.test-d.ts
```diff
@@ -0,0 +1,95 @@
...
+  Link,
+  createRootRoute,
+  createRoute,
+  createRouter,
+} from '@tanstack/react-router'
+import { expectTypeOf, test } from 'vitest'
+import { type } from 'arktype'
+import { arkTypeSearchValidator } from '../src'
+
+test('when creating a route with valibot validation', () => {
```
schiller-manuel, line 11: valibot => arktype

packages/router-arktype-adapter/package.json
chorobin: I think `infer` and `inferIn` should be a part of the public api for arktype

schiller-manuel: so what happens if a major version of arktype changes this? currently our package does not express which arktype version it is compatible to.

SeanCassiere: Yea, something like `>=2 <=3` should suffice for the time being.

schiller-manuel (approved):

docs/framework/react/guide/search-params.md
```diff
@@ -172,6 +172,147 @@ In the above example, we used Zod's `.catch()` modifier instead of `.default()`
...
+})
+
+export const Route = createFileRoute('/shop/products/')({
+  validateSearch: valibotSearchValidator(productSearchSchema),
+})
+```
+
+### Arktype Adapter
+
+When using [ArkType](https://arktype.io/) we recommend using the adapter. This ensures the correct `input` and `output` types are used for navigation and reading search params
```
SeanCassiere, line 297: I'd add a gfm alert here to let the user know that the package expects the arktype beta.


```
> [!WARNING]
> This adapter...
```

packages/router-zod-adapter/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+    types: {
+      input: input === 'output' ? _output : _input,
+      output: output === 'input' ? _input : _output,
+    },
+    parse: (input) =>
+      'schema' in options ? options.schema.parse(input) : options.parse(input),
+  }
+}
+
+export const fallback = <TSchema extends z.ZodTypeAny>(
```
SeanCassiere, line 64: This is for when you want the search param to appear in the URL yes?
Maybe its name should reflect it like `uriFallback`, `urlFallback``, or something to that effect.

docs/framework/react/guide/search-params.md
```diff
@@ -172,6 +172,147 @@ In the above example, we used Zod's `.catch()` modifier instead of `.default()`
 
 The underlying mechanics why this works relies on the `validateSearch` function throwing an error. If an error is thrown, the route's `onError` option will be triggered (and `error.routerCode` will be set to `VALIDATE_SEARCH` and the `errorComponent` will be rendered instead of the route's `component` where you can handle the search param error however you'd like.
 
+#### Adapters
+
+When using a library like [Zod](https://zod.dev/) to validate search params you might want to `transform` search params before committing the search params to the URL. A common `zod` `transform` is `default` for example.
+
+```typescript
```
SeanCassiere, line 179: Can we decorate these markdown code blocks as just `tsx` like the rest of the page? so we don't introduce much divergence in the docs.

SeanCassiere (changes requested):

packages/router-zod-adapter/src/index.ts
chorobin: This is when the input does not validate. Like `catch` but it keeps the `input` type where `catch` does not

SeanCassiere (approved):

PR #2007, opened 2024-07-22
fix(router-with-query): context type too narrow

juliusmarminge:
type currently expects exact match on context type:

CleanShot 2024-07-22 at 15 29 35@2x


packages/react-router-with-query/tests/index.test.tsx
schiller-manuel, line 1: can you please rename to `index.test-d.tsx`

packages/react-router-with-query/tests/index.test.tsx
```diff
@@ -1,5 +1,64 @@
...
+  createRootRouteWithContext,
+  createRouter,
+} from '@tanstack/react-router'
+
+import { expect, expectTypeOf, test } from 'vitest'
+
+import { routerWithQueryClient } from '../src'
 
 test('it has a test', async () => {
   expect(true).toBe(true)
```
schiller-manuel, line 12: delete this dummy test

packages/react-router-with-query/src/index.tsx
```diff
@@ -4,19 +4,23 @@ import {
...
->(router: T, queryClient: QueryClient): T {
+export function routerWithQueryClient<TRouter extends AnyRouter>(
+  _router: TRouter extends AnyRouterWithContext<infer Ctx>
+    ? Ctx extends { queryClient: QueryClient }
+      ? TRouter
+      : 'Context type mismatch. Expected queryClient to be in context type.'
+    : 'Context type mismatch. Expected queryClient to be in context type.',
+  queryClient: QueryClient,
+): TRouter {
+  const router = _router as unknown as AnyRouter
```
schiller-manuel, line 23:
```suggestion
export function routerWithQueryClient<TRouter extends AnyRouter>(
  router: TRouter extends AnyRouterWithContext<infer Ctx>
    ? Ctx extends { queryClient: QueryClient }
      ? TRouter
      : never
    : never,
  queryClient: QueryClient,
): TRouter {
```

schiller-manuel: when using `never`, you don't have to do any type assertions

packages/react-router-with-query/src/index.tsx
```diff
@@ -123,5 +127,5 @@ export function routerWithQueryClient<
     },
   }
 
-  return router
+  return router as TRouter
```
schiller-manuel, line 130: not necessary when using `never` above

schiller-manuel (changes requested):

packages/react-router-with-query/src/index.tsx
juliusmarminge: unfortunately it appears I have to in order to support all the ts versions:
CleanShot 2024-07-23 at 10 04 29

https://github.com/TanStack/router/pull/2007#discussion_r1686586556

schiller-manuel: @chorobin can you please have a look at this?

tannerlinsley: I feel like I’ve run into this before too. Very odd.

chorobin: Fairly sure this is due to invariance. If a type parameter is in a `in` and an `out` position then a router with a different context type is not assignable to `AnyRouterWithContext`

chorobin: Why not
```typescript
export function routerWithQueryClient<TRouter extends AnyRouter>(
  router: TRouter['options']['context'] extends { queryClient: QueryClient }
    ? TRouter
    : AnyRouterWithContext<{ queryClient: QueryClient }>,
  queryClient: QueryClient,
): TRouter
```

tannerlinsley: Well, what do we do? We need this to move forward.

chorobin: I would say my above suggestion works. What do you think @schiller-manuel

schiller-manuel: yes please go for https://github.com/TanStack/router/pull/2007#discussion_r1691905804

juliusmarminge: Will see if it works for all versions

juliusmarminge: CleanShot 2024-07-26 at 23 38 33
 Nope...

PR #2017, opened 2024-07-24
Warn against DOM-rendering elements in Wrap and InnerWrap

thomasballinger:


thomasballinger:
Is this correct? I'm not sure if this belongs here since I see these hydration errors in Start.

schiller-manuel:
please also update the docs

https://github.com/TanStack/router/blob/f7c74bc2d9fb8a587d300482e1d658a021792439/docs/framework/react/api/router/RouterOptionsType.md?plain=1#L192

https://github.com/TanStack/router/blob/f7c74bc2d9fb8a587d300482e1d658a021792439/docs/framework/react/api/router/RouterOptionsType.md?plain=1#L173

thomasballinger:
ah thanks, I wrongly assumed they were autogenerated

thomasballinger:
Are these autogenerated? I just made the change manually, I couldn't find how to generate docs if they are.

schiller-manuel:
not autogenerated (yet 🙃)

PR #1946, opened 2024-07-11
fix(react-router): when children are added to the route tree, the inference of RouterContext is lost

chorobin:
When `addChildren` was called, we lost the type of `RootRoute` and there could no longer infer `RouterContext`.

Fixes #1943 
Fixes #1926

PR #2018, opened 2024-07-24
Specify context with routerWithQueryClient

thomasballinger:
https://github.com/TanStack/router/pull/1946 broke these types as they're used, for now just specify the query client explicitly.


thomasballinger:
Oops this is fixed by https://github.com/TanStack/router/pull/2007

PR #2027, opened 2024-07-26
fix(react-router): retain path parameters that are falsy but still valid 

### LINKED ISSUE #2024: Path parameter with value 0 is ignored
### Describe the bug

`<Link
          to={'/employee/$id'}
          params={{
            id: 0,
          }}
        >
          Employee with id 0
        </Link>`

With route:

`export const Route = createFileRoute('/employee/$id')({
  component: AboutComponent,
});`

Since a recent update (1.38.1 -> 1.45.10), this link generates: `/employee` while it used to generate `/employee/0`

`<Link
          to={'/employee/$id'}
          params={{
            id: 5,
          }}
        >
          Employee with id 5
        </Link>`
Still correctly generates  `/employee/5`

https://stackblitz.com/edit/tanstack-router-jnrhtc?file=src%2Froutes%2F__root.tsx

- Click on 'About with id 0'
- Notice it renders "Not found" because the link does not include the parameter

### Expected behavior

Path parameters should support the number 0 as we use index based routing in a List->Detail view for entities that don't have a uuid.

### Screenshots or Videos

_No response_

### Platform

- Browser: Chrome

### Additional context

Why I need this:
I use react-hook-form's fieldarray to let the user modifiy an array of 'things'. When the user clicks the 'add thing'
 button I add a 'thing' and navigate to /mypage/things/0 to let the user enter the thing's details.

SeanCassiere:
Fixes #2024

### DIFF
--- a/packages/react-router/src/path.ts
+++ b/packages/react-router/src/path.ts
@@ -8,7 +8,13 @@ export interface Segment {
 }
 
 export function joinPaths(paths: Array<string | undefined>) {
-  return cleanPath(paths.filter(Boolean).join('/'))
+  return cleanPath(
+    paths
+      .filter((val) => {
+        return val !== undefined
+      })
+      .join('/'),
+  )
 }
 
 export function cleanPath(path: string) {
@@ -193,7 +199,7 @@ export function parsePathname(pathname?: string): Array<Segment> {
 
 interface InterpolatePathOptions {
   path?: string
-  params: any
+  params: Record<string, unknown>
   leaveWildcards?: boolean
   leaveParams?: boolean
 }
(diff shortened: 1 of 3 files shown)


packages/react-router/src/path.ts
```diff
@@ -8,7 +8,13 @@ export interface Segment {
 }
 
 export function joinPaths(paths: Array<string | undefined>) {
-  return cleanPath(paths.filter(Boolean).join('/'))
+  return cleanPath(
+    paths
+      .filter((val) => {
```
SeanCassiere, line 13: This was `Boolean` filtering was causing falsey number values (like `0`, `-1`, etc...) to be excluded out of the joined path.

This ripples out into the router logic, where all the right routes don't get matched since  the pathname it's using for checking will be incomplete or in some cases even outright incorrect.

Issue #2030, opened 2024-07-26
Vite plugin regression causes plugin to break when having a root configured


reihwald:
### Describe the bug

I noticed that my routes are no longer getting generated after updating my vite plugin. This is most noticeable by the fact that creating a new file in the routes folder no longer generates the small createFileRoute snippet. 

It is possible to trigger the plugin to update the gen.ts file by stopping the relaunching the vite dev command but I am still not getting the correct type inference in my editor when using search params. Everything works fine when doing a production build.

This is caused by the root setting in vite. Removing it and configuring vite to run without it makes the problem disappear.

### Your Example Website or App

https://stackblitz.com/edit/github-snnf6b

### Steps to Reproduce the Bug or Issue

1. Run `npm run dev`
2. Create a new file under `someFolder/anotherFolder/routes` with a name like test.tsx
3. No `createFileRoute` snippet is being generated

Please note that the vite example does not show anything when running the dev command. We run vite in a customized environment but this should not be of concern for this bug. 

### Expected behavior

Please take a look at this example: https://stackblitz.com/edit/github-snnf6b-aobdnw

1. This is the exact same vite setup as the first example but using version `1.34.8` of `@tanstack/router-vite-plugin`.
2. The problem is not reproducible in this version. Doing the same steps as above creates a `createFileRoute` snippet

I'd expect that the vite plugin still work when using the root setting.

### Platform

- OS: Windows
- Browser: Chrome
- Version: 127

schiller-manuel:
is this the same as https://github.com/TanStack/router/issues/1550?

reihwald:
@schiller-manuel It certainly looks very similar. This issue is however still happening on the latest version even after that merged PR.

schiller-manuel:
could be a regression caused by moving over to unplugin

cc @SeanCassiere

SeanCassiere:
> could be a regression caused by moving over to unplugin
> 
> cc @SeanCassiere

Very unlikely tbh. Their implementation for vite is [dead-simple](https://github.com/unjs/unplugin/blob/30b323996919cc6ea8bf0fe53e7cdc5485461302/src/vite/index.ts).

vbifonixor:
Found a tiny crutch to work around it - used `vite-plugin-restart` to restart vite every time route files change. I think it should be definitely a bad one for large projects (mine was just started two days ago). Hope the issue is fixed soon!

Issue #2026, opened 2024-07-25
Outlet in root route prevents suspense from bubbling up


lukasgeiter:
### Describe the bug

I would like to use my own loading component at the root level. To do that I'm wrapping the `<Outlet>` in the root component with a `<Suspense>` block. Unfortunately, `Outlet` wraps the children in a `<Suspense>` block of its own, preventing mine from being used. Since I have no `defaultPendingComponent` configured, it renders nothing.

I have traced the issue to this part of the code: https://github.com/TanStack/router/blob/f7c74bc2d9fb8a587d300482e1d658a021792439/packages/react-router/src/Match.tsx#L294-L302

### Your Example Website or App

https://stackblitz.com/edit/github-s38jue?file=src%2Froutes%2F__root.tsx

### Steps to Reproduce the Bug or Issue

1. Run the Stackblitz

### Expected behavior

As a user, I expect to see the fallback content `Suspense fallback - I expect this to show up`, but I'm seeing nothing (besides the navigation)

### Platform

- OS: macOS
- Browser: Arc (Chrome)
- Version: 1.52.0 (51895)

### Additional context

Related issue: #1049

schiller-manuel:
Would it help if you could configure the root route to not wrap in suspense?

schiller-manuel:
does https://github.com/TanStack/router/pull/2035 solve your issue?

lukasgeiter:
@schiller-manuel that would certainly be an improvement 👍 

Although, I'm still of the opinion that if no pending components are configured, the router shouldn't wrap anything in a suspense at all. I don't think rendering a suspense with `null` as fallback is a good default.

schiller-manuel:
like this? https://github.com/TanStack/router/pull/2035/commits/3518ac37e1178b1738ff6288f6ebaa5bed1257a3

I am not sure which implications this has, so I need reviews.

lukasgeiter:
Yeah, that's what I had in mind. Not sure if the failing tests mean that this is a bad idea, or if they should be adjusted 😅

TkDodo:
> Although, I'm still of the opinion that if no pending components are configured, the router shouldn't wrap anything in a suspense at all

I’m not sure about that. Usually, we try to determine if wrapping in a suspense boundary is necessary, by checking for the presence of a `loader` or `beforeLoad`. This can be force-overwritten with `wrapInSuspense: true`.

[...]

PR #2035, opened 2024-07-26
fix(react-router): do not wrap children in outlet in suspense unless (default)pendingComponent is configured

schiller-manuel:


SeanCassiere:
This would resolve https://github.com/TanStack/router/issues/2030 yes? Since the user would now be able to add their own `pendingComponent` on the root route.

schiller-manuel:
it maybe solves https://github.com/TanStack/router/issues/2026

SeanCassiere:
> it maybe solves #2026

Whoops, my bad, linked the wrong issue.

schiller-manuel:
@tannerlinsley please have a look if omitting this suspense is ok, check #2026  for the motivation.

one thing I dislike is that there is no way to opt out of suspense here if `defaultPendingComponent` is set.

SeanCassiere (commented):
> one thing I dislike is that there is no way to opt out of suspense here if defaultPendingComponent is set.

This could be changed so that we don't use the `defaultPendingComponent` for the `rootRoute` and only rely on the `rootRoute.pendingComponent`. So if `PendingComponent===undefined` then just return `nextMatch`.

The only issue here is that this could be a breaking change for anyone currently relying on that default behaviour of not requiring a Suspense boundary above the RouterProvider.
We can ofc, get around this by introducing a _future_ flag and stuff this new behaviour under that flag (and remove the old behaviour in v2).

lukasgeiter:
@SeanCassiere @schiller-manuel @tannerlinsley Is there any chance of getting at least the [first commit](https://github.com/TanStack/router/pull/2035/commits/16a42fd3f7426172e5b68f354d3dcf270ae5d945) merged?
That would allow me to work around this bug without having to patch the package itself.

dy0gu:
Any updates on getting this merged? We would really appreciate it, some workarounds are getting messy!

LadyBluenotes:
Closing this since the issue was addressed in #5495

PR #2037, opened 2024-07-27
build(react-router-with-query): add `@tanstack/react-router-with-query` to the publish list

SeanCassiere:
For whatever reason, https://github.com/TanStack/router/commit/2c94fa0ee8f54f7039649f5e7e2c3e74087df5f1 did not trigger a package to be published.

Looking at our `publish` script, we don't seem to have the `@tanstack/react-router-with-query` as a package to be published.

### DIFF
--- a/scripts/publish.js
+++ b/scripts/publish.js
@@ -17,6 +17,10 @@ await publish({
       packageDir: 'packages/react-router',
     },
     {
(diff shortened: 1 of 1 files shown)


thomasballinger:
Thank you! I've been waiting for this, should have looked into the publish script.

PR #2056, opened 2024-07-29
fix(react-router): memoize Match, MatchInner as well as route components to prevent unnecessary re-renders

### LINKED ISSUE #1462: I'm reaching out to report a performance issue.
I'm reaching out to report a performance issue.

The problem involves performance in two scenarios:

1. When updating the search parameters using `useNavigation`, the entire page, including all layout components and child components, is re-rendered.
2. When there is a layout component with multiple child pages, navigating between these child pages triggers a re-render of the entire page, including all layout components.

In the classic `react-router` package, this problem does not exist, there it works as expected, parts are rerendered as expected.

Example Link: https://stackblitz.com/edit/vitejs-vite-ifxs98

_Originally posted by @AndrejNemec in https://github.com/TanStack/router/discussions/1460_

schiller-manuel:
fixes #1462

### DIFF
--- a/packages/react-router/src/Match.tsx
+++ b/packages/react-router/src/Match.tsx
@@ -16,7 +16,11 @@ import { renderRouteNotFound } from './renderRouteNotFound'
 import { rootRouteId } from './root'
 import type { AnyRoute } from './route'
 
-export function Match({ matchId }: { matchId: string }) {
+export const Match = React.memo(function Match({
+  matchId,
+}: {
+  matchId: string
+}) {
   const router = useRouter()
   const routeId = useRouterState({
     select: (s) => s.matches.find((d) => d.id === matchId)?.routeId as string,
@@ -99,8 +103,13 @@ export function Match({ matchId }: { matchId: string }) {
       </ResolvedSuspenseBoundary>
     </matchContext.Provider>
   )
-}
-function MatchInner({ matchId }: { matchId: string }): any {
+})
+
+export const MatchInner = React.memo(function MatchInner({
+  matchId,
+}: {
+  matchId: string
+}): any {
   const router = useRouter()
   const routeId = useRouterState({
     select: (s) => s.matches.find((d) => d.id === matchId)?.routeId as string,
@@ -126,6 +135,10 @@ function MatchInner({ matchId }: { matchId: string }): any {
       ])
     },
   })
+
+  const Comp = route.options.component ?? router.options.defaultComponent
+
+  const out = React.useMemo(() => (Comp ? <Comp /> : <Outlet />), [Comp])
 
   // function useChangedDiff(value: any) {
   //   const ref = React.useRef(value)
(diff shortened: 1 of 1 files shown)


packages/react-router/src/Match.tsx
```diff
@@ -127,6 +136,10 @@ function MatchInner({ matchId }: { matchId: string }): any {
     },
   })
 
+  const Comp = route.options.component ?? router.options.defaultComponent
+
+  const out = React.useMemo(() => (Comp ? <Comp /> : <Outlet />), [Comp])
```
SeanCassiere, line 141: It may be worth checking out the benefit of using the `matchId` as the key on the `<Comp/>`.

Ultimately may not be what we are looking for if stuff like `useSearch` doesn't trigger a re-render, but might be worth checking out.

packages/react-router/src/Match.tsx
```diff
@@ -127,6 +136,10 @@ function MatchInner({ matchId }: { matchId: string }): any {
     },
   })
 
+  const Comp = route.options.component ?? router.options.defaultComponent
```
SeanCassiere, line 139: Could this also be memoed? Or maybe pulled into `useMemo` below?

Since these values don't change very often, there is no need to recompute this.

packages/react-router/src/Match.tsx
```diff
@@ -16,7 +16,11 @@ import { renderRouteNotFound } from './renderRouteNotFound'
 import { rootRouteId } from './root'
 import type { AnyRoute } from './route'
 
-export function Match({ matchId }: { matchId: string }) {
+export const Match = React.memo(function Match({
```
SeanCassiere, line 19: Could we name the inner `Match` as `MatchImpl`?

Same for `MatchInner` and `Outlet`.

packages/react-router/src/Match.tsx
schiller-manuel: done

packages/react-router/src/Match.tsx
schiller-manuel: done

packages/react-router/src/Match.tsx
schiller-manuel: like this?

```tsx
<Comp key={matchId}/> 
```

which effect would you expect from that?

SeanCassiere: Yup. 

It's since keyed components don't re-render unless the key changes, it may help.

And we shouldn't have to worry about the render being stale, since the `matchId` changes if actual match info has changed.

schiller-manuel: tried it with kitchen sink example, using the `matchId` as key causes the `InvoiceComponent` to be rendered only once instead of twice without it.
great find!

SeanCassiere: 🚀

Yeah it's a simple hack I tend to use to either trigger a child component's re-render or to prevent unneeded renders.

SeanCassiere (approved):

PR #2058, opened 2024-07-29
feat(react-router): allow narrowing of matches with `useMatches`

chorobin:
### useMatches returns a union of all matches

Currently `useMatches` returns single match type with different properties which are a union of everything. `loaderData` is a union of all `loaderData`, `routeId` is a union of all `routeId`, `search` is a union of all `search`. The problem with this approach is its impossible for TS to narrow a match based on the `routeId` of the `match`.

For example this code doesn't type check in TS 5.5:

```typescript
  const match = useMatches({
    select: (matches) =>
      matches.find((match) => match.routeId === '/posts/$postId'),
  })

  const postId = match?.params.postId // the type is not narrowed :(
```

Another example that would not type check

```tsx
export const PageTitle = () => {
  const matches = useMatches()
  return (
    <h1>
      {matches.map((match) => {
        switch (match.routeId) {
          case '/dashboard/invoices/$invoiceId':
            return match.loaderData?.title // the match should be narrowed here to specific `loaderData`
          case '/dashboard/users/user':
            return match.loaderData?.username
        }
      })}
    </h1>
  )
}
```

But now it does with this PR

This is because TS 5.5 can narrow a `match` based on the `routeId` of a `match`. This is made possible by creating a `match` type for each route which is a part of a union

### isMatch type predicate (idea?)

Narrowing a union based on `routeId` isn't always feasible. Sometimes you want to narrow to matches which contain a certain shape or data. For example, a match might contain data necessary to build breadcrumbs and we want to find matches with the necessary data. 

This can be achieved with the `isMatch` type predicate

```tsx
export const Breadcrumbs = () => {
  const matches = useMatches({
    select: (matches) =>
      matches.filter((match) => isMatch(match, 'loaderData.crumb')),
  })

  return (
    <nav>
      <ul>
        {matches.map((match, i) => (
          <li>
            <Link from={match.fullPath}>{match.loaderData?.crumb}</Link>
          </li>
        ))}
      </ul>
    </nav>
  )
}
```

`isMatch` will narrow a the type of matches to only `matches` which contain `loaderData.crumb`. TypeScript is happy because the match types are narrowed


packages/react-router/src/Matches.tsx
```diff
@@ -27,8 +27,75 @@ import type {
...
+  TPath extends string,
+> = TPath extends SuggestMatchesByPath<TPath, TMatch>['suggestions']
+  ? TPath
+  : SuggestMatchesByPath<TPath, TMatch>['suggestions']
+
+export const isMatch = <TMatch extends AnyRouteMatch, TPath extends string>(
+  match: TMatch,
+  path: ValidateMatchSuggestions<TMatch, TPath>,
+): match is SuggestMatchesByPath<TPath, TMatch>['match'] => {
+  const parts = path.split('.')
```
chorobin, line 85: This runtime logic is really basic. I'm going to be adding tests so I assume it might get better

packages/react-router/src/Matches.tsx
```diff
@@ -27,8 +27,75 @@ import type {
...
+    : never
+  : never
+
+export type SuggestMatchByPath<
+  TPath,
+  TMatch,
+  TValue = TMatch,
+  TParentPath extends string = '',
+> = TPath extends `${string}.${string}`
+  ? TPath extends `${infer TFirst}.${infer TRest}`
```
chorobin, line 57: Optimised the types a bit but I have some ideas to go further if we like it

packages/react-router/src/router.ts
```diff
@@ -1140,6 +1140,7 @@ export class Router<
           scripts: route.options.scripts?.(),
           staticData: route.options.staticData || {},
           loadPromise: createControlledPromise(),
+          fullPath: route.fullPath,
```
chorobin, line 1143: I think it would be good to have `fullPath` on a match to enable `<Link from={match.fullPath} />`

examples/react/kitchen-sink-file-based/src/components/Breadcrumbs.tsx
```diff
@@ -0,0 +1,28 @@
+import { isMatch, Link, useMatches } from '@tanstack/react-router'
+
+export const Breadcrumbs = () => {
+  const matches = useMatches()
+
+  if (matches.some((match) => match.status === 'pending')) return null
+
+  const matchesWithCrumbs = matches.filter((match) =>
+    isMatch(match, 'loaderData.crumb'),
+  )
```
TkDodo, line 10: looking at [the sandbox](https://stackblitz.com/edit/dj9kzo?file=src%2Fcomponents%2FBreadcrumbs.tsx), `matchesWithCrumbs` is always `RouteMatch<any, any, any, any, any, any, any, any>[]`:

Screenshot 2024-08-02 at 17 01 52

Is that because the sandbox is using the wrong TS version?

TkDodo: also - why couldn't we just check for `match.loaderData?.crumb !== undefined` ?

chorobin: I noticed this. It's weird. Works well in vscode and sometimes in the sandbox

chorobin: > also - why couldn't we just check for `match.loaderData?.crumb !== undefined` ?

The issue is TS is not able to narrow this. If you have loader data in matches which do not have `crumb`, you can't get to the property without using `in`. Furthermore TS is not able to narrow the parent type of a property if it's not a literal type. This is why `routeId` works but `crumb` does not. Discriminated unions work but other properties do not

chorobin: > I noticed this. It's weird. Works well in vscode and sometimes it works in the sandbox but not all the time

Maybe the types just need optimising more. I think there's still too much instantiations happening in the recursive type due to it distributing over the recursive type. Maybe stackblitz is more limited in resources or uses an older version of TS which kicks the bucket with more complex types?

PR #2062, opened 2024-07-31
feat(react-router): support for `preload='viewport'`

SeanCassiere:
This adds support for `preload='viewport'`.

Similar to how NextJS does preloading on anchor elements that enter the viewport, this change would make this possible for the `useLinkProps` hook, `Link` component, and the `createLink` utility function.

Testing required the mocking of the IntersectionObserver API (can't be helped 🤷🏼).

Tasklist:
- [x] Implementation
- [x] Testing
- [x] Docs on the preloading guide
- [x] Docs on the API reference


docs/framework/react/api/router/RouterOptionsType.md
```diff
@@ -37,11 +37,12 @@ The `RouterOptions` type accepts an object with the following properties and met
 
 ### `defaultPreload` property
 
-- Type: `undefined | false | 'intent'`
+- Type: `undefined | false | 'intent' | 'viewport`
```
SeanCassiere, line 40:
```suggestion
- Type: `undefined | false | 'intent' | 'viewport'`
```

PR #2066, opened 2024-07-31
fix(react-router): when navigating from a index the types should resolve to a valid route

chorobin:
There is an issue in the types where `from` is an index route but there is no `to` or the resolve path is the same as `from`. It means that

`<Link from='/posts/' />` always requires a `search` param because the resolved route is incorrect and returns `never`

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -258,7 +258,7 @@ export type ResolveRoute<
   TTo,
   TPath = ResolveRelativePath<TFrom, TTo>,
 > = TPath extends string
-  ? string extends TTo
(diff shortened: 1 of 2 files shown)


packages/react-router/src/link.tsx
```diff
@@ -258,7 +258,7 @@ export type ResolveRoute<
   TTo,
   TPath = ResolveRelativePath<TFrom, TTo>,
 > = TPath extends string
-  ? string extends TTo
+  ? TFrom extends TPath
```
chorobin, line 261: If the resolved path is the same as `TFrom` then we know we should use `RouteByPath`

PR #2091, opened 2024-08-06
docs: remove the invalid `context` and `location` arguments from being shown as available in the `loaderDeps`

SeanCassiere:
We haven't had the `context` and `location` arguments present as `loaderDeps` options, so this corrects this error in Router's documentation.

The reason we don't have `location` as an argument of `loaderDeps` is quite simple. If the `location` changes, the loader would anyways have to be refired (or retrieved from a cache using something like Query). The scenario in which this wouldn't be the case, is if you wanted to prevent this from happening when search params change. This is why only `search` is provided as its own argument.

The reason we don't have `context` as an argument of `loaderDeps` is because, when the root context has been changed, the user is expected to call `router.invalidate()`. This `invalidate` call discards all cached data and re-evaluates the entire route-tree. As such, having `context` as an argument in `loaderDeps`, makes no sense, since it wouldn't actually contribute anything towards the actual `loaderDeps` functionality.

### DIFF
--- a/docs/framework/react/api/router/RouteOptionsType.md
+++ b/docs/framework/react/api/router/RouteOptionsType.md
@@ -143,18 +143,13 @@ type loader = (
 - Type:
 
 ```tsx
-type loaderDeps = (opts: {
-  search: TFullSearchSchema
-  location: ParsedLocation
-  context: TAllContext
-}) => Record<string, any>
+type loaderDeps = (opts: { search: TFullSearchSchema }) => Record<string, any>
 ```
 
 - Optional
-- [`ParsedLocation`](./ParsedLocationType.md)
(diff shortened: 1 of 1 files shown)


waweber:
@SeanCassiere is the loader intended to be fired when _any_ property of `location` changes? (I assume the whole object is replaced on navigate.)

I noticed the route loader did not appear to fire when only `location.hash` was changed, so I'm not sure if I missed something or if that's a bug.

SeanCassiere:
> @SeanCassiere is the loader intended to be fired when _any_ property of `location` changes? (I assume the whole object is replaced on navigate.)
> 
> I noticed the route loader did not appear to fire when only `location.hash` was changed, so I'm not sure if I missed something or if that's a bug.

Please open a new issue for this with a reproduction. This is sounding like a bug, but I can't be sure till someone investigates.

PR #2099, opened 2024-08-07
docs: add command for all package manager

sagardwivedi:


SeanCassiere (changes requested):
If we are looking to make changes here, we should support all the package managers.

See the installation pages for Query and Form:
- https://tanstack.com/query/latest/docs/framework/react/installation
- https://tanstack.com/form/latest/docs/installation

SeanCassiere (approved):

PR #2100, opened 2024-08-07
feat(router-generate): add option to skip generating route manifest

honzahruby:
Add an option to the generator config to skip generating route manifest, as it's currently only being used in tanstack start. I added it as a non-breaking change so it stays enabled by default.

It's just a comment, but it can grow to thousands of lines of code that need to be processed and removed when we can just not generate it.


packages/router-generator/src/config.ts
```diff
@@ -13,6 +13,7 @@ export const configSchema = z.object({
   disableTypes: z.boolean().optional().default(false),
   addExtensions: z.boolean().optional().default(false),
   disableLogging: z.boolean().optional().default(false),
+  generateManifest: z.boolean().optional().default(true),
```
SeanCassiere, line 16: I believe this should follow the naming above and should be `disableManifestGeneration` or something to that effect.

SeanCassiere (changes requested):
@schiller-manuel Since we only use manifest for Start, I'd vote for this going in. Any blockers you see over here?

@honzahruby This'd also have to be updated in the docs, under the `docs/framework/react/guide/file-based-routing.md`.

SeanCassiere (approved):

honzahruby:
Thanks a lot @SeanCassiere 🫶

PR #2104, opened 2024-08-07
feat(react-router): add `context` option to route

chorobin:
This PR adds `context` to each route


chorobin:
- [ ] Type tests
- [ ] Optimize & refactor types

chorobin:
Lets not force push this one if multiple of us are working on it. Merge only!

packages/react-router/src/route.ts
```diff
@@ -122,39 +122,64 @@ export interface FullSearchSchemaOption<TFullSearchSchema> {
...
     | ((
-        match: LoaderFnContext<TAllParams, TFullSearchSchema, TAllContext>,
+        match: LoaderFnContext<
+          TAllParams,
+          TFullSearchSchema,
+          ResolveAllContext<TParentAllContext, TRouteContextFn, TBeforeLoadFn>
+        >,
       ) => any)
+
+  routeContext?: Constrain<
```
chorobin, line 156: Limitations of default type parameters and constraints leads me to use this pattern which I've made into a utility type. Long story short, it easy to get stale types which are the default types/constraints and not the inferred types. 

Instead `Constrain` is a conditional type which checks if a type is assignable to another type, if it is then all good, we can infer the type, otherwise we use the target type to force a type check error

packages/react-router/src/route.ts
```diff
@@ -165,11 +190,11 @@ export type BaseRouteOptions<
   TFullSearchSchema = TSearchSchema,
   TParams = {},
   TAllParams = {},
-  TRouteContextReturn = RouteContext,
-  TParentAllContext = AnyContext,
-  TAllContext = AnyContext,
   TLoaderDeps extends Record<string, any> = {},
   TLoaderDataReturn = {},
+  TParentAllContext = {},
+  TRouteContextFn = AnyContext,
```
chorobin, line 196: I'm trying out patterns to remove unnecessary type parameters. Usually we require 2 type parameters per function because we need to check the return type is `never`. If we infer the whole function, we don't need two type paramters

packages/react-router/src/route.ts
```diff
@@ -426,31 +488,12 @@ export interface AnyRoute
...
-  any,
-  any,
-  TContext,
-  any,
-  any,
-  any,
-  any
->
+export type AnyRouteWithContext<TContext> = AnyRoute & {
+  types: { allContext: TContext }
```
SeanCassiere, line 495: This is only a type-level thing yes? No something that's available at runtime.

packages/react-router/src/route.ts
```diff
@@ -178,16 +203,16 @@ export type BaseRouteOptions<
...
+    TLoaderDataReturn,
+    TParentAllContext,
+    TRouteContextFn,
+    TBeforeLoadFn
   > & {
     getParentRoute: () => TParentRoute
   }
 
-export interface BeforeLoadContext<
+export interface ContextOptions<
```
SeanCassiere, line 215: Just to keep a note here, we should probably type [this beforeLoad](https://github.com/TanStack/router/blob/08b85242f875d619ed285f0ff3dd06db00b30807/packages/react-router/src/router.ts#L1961-L1972) parameter object with a type like we do so for [`the loader`](https://github.com/TanStack/router/blob/08b85242f875d619ed285f0ff3dd06db00b30807/packages/react-router/src/router.ts#L2032-L2054).

tannerlinsley:
@chorobin I hope this doesn't screw anything up, but I removed `routeContext` from `RouteMatch`. `Route` remains untouched. There just wasn't any use for it in the match and in fact, since we're doing immutable updates to matches, it never made much sense to have a separate `routeContext` and `context` for the match.

Now it's just `context`. This okay?

tannerlinsley:
Fudge. We might need to keep it, but only as a runtime thing. I don't want people relying on it as a public API, but I do need it to build up merges correctly...

packages/react-router/package.json
```diff
@@ -77,6 +77,7 @@
     "jsdom": "^24.1.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "vitest": "^1.6.0",
```
SeanCassiere, line 80: This shouldn't be needed, since Vitest gets inherited from the monorepo. Plus we don't have vitest as deps or devDeps in any of the other packages.

Its probably whats causing the lockfile failure in CI.

tannerlinsley: Hmm. It was failing to run the tests locally without it.

chorobin:
> @chorobin I hope this doesn't screw anything up, but I removed `routeContext` from `RouteMatch`. `Route` remains untouched. There just wasn't any use for it in the match and in fact, since we're doing immutable updates to matches, it never made much sense to have a separate `routeContext` and `context` for the match.
> 
> Now it's just `context`. This okay?

No problem from my side on this. It probably would be confusing to keep it

packages/react-router/package.json
SeanCassiere: Oh, I see!!!

This tends to happen for me locally as well when I haven't kept up to date with the **lockfile** on _main_ or if my pnpm version isn't in sync with `packageManager` field in the `package.json` file.

Some common indicators of happening which I've noted are:
- local tests suddenly thinks vitest is missing
  - for which "fixing" it causes CI installs to fail
- local formatting using prettier causes unrelated files or areas of affected code to change their code-style.
- local builds failings with nx being unable to build one or more packages.

To fix this, I often switch to _main_ and pull down all the latest changes. Then make sure the pnpm version is correct and run `pnpm install`. Then for good measure, I run `pnpm dedupe`.

chorobin:
Will fix the conflicts with the types

tannerlinsley (approved):

tannerlinsley:
@SeanCassiere @chorobin @schiller-manuel we good here?

packages/react-router/src/router.ts
```diff
@@ -1180,6 +1203,38 @@ export class Router<
...
+        : (parentMatch.context ?? this.options.context ?? {})
+
+      match.context = {
+        ...parentContext,
+        ...match.__routeContext,
+        ...match.__beforeLoadContext,
+      }
+
+      // Update the match's context
+      const contextFnContext = {
```
SeanCassiere, line 1219: Typing this object with `RouteContextOptions`.

packages/react-router/src/router.ts
```diff
@@ -1946,41 +1997,55 @@ export class Router<
...
+                      ...prev.__routeContext,
+                      ...prev.__beforeLoadContext,
+                    },
                   }))
 
-                  const { search, params, routeContext, cause } =
+                  const { search, params, context, cause } =
                     this.getMatch(matchId)!
 
                   const beforeLoadFnContext = {
```
SeanCassiere, line 2021: Typing this with `BeforeLoadContextOptions`.

SeanCassiere (commented):
@tannerlinsley could the `defaultTransformer` issue from https://github.com/TanStack/router/issues/2113 also be pushed with this? It's just a matter of moving the transformer from Start into Router.

Other than that, it's just the typing of the arguments of the `routeContext` and `beforeLoad` functions. I can make these changes later. Just need the go-ahead.

tannerlinsley:
“Go, Go, Go”

Tanner Linsley
TanStack LLC

packages/react-router/src/router.ts
```diff
@@ -1180,6 +1199,41 @@ export class Router<
...
+      // Update the match's context
+      const contextFnContext: RouteContextOptions<any, any, any, any> = {
+        search: match.search,
+        params: match.params,
+        context: match.context,
+        location: next,
+        navigate: (opts: any) =>
+          this.navigate({ ...opts, _fromLocation: next }),
+        buildLocation: this.buildLocation,
+        cause: match.cause,
```
SeanCassiere, line 1223: @tannerlinsley not sure if this was intentionally omitted or not, but `cause`, `abortController`, and `preload`, were not present on this object before but were _typed_ on the `RouteContextOptions` type.

SeanCassiere (approved):

packages/react-router/src/router.ts
chorobin: Yeah, I made it the same as `beforeLoad` in this regard. If the types are wrong here they can be removed

schiller-manuel: could we take this chance to add previous and next location (in parsed form) as params to all of those functions?

SeanCassiere: @schiller-manuel I'd push for that kind of change to go separate from this PR.

A change of this nature would require more sweeping changes to make sure those values are present in the `loadMatches` function, whilst not really contributing to the task of adding this `context` function to the route definition and wiring it up.

Once this goes in, if we are adding in the _previous_ and _next_ location options into the shared definition used for `context`, `beforeLoad`, and the `loader`, it'd certainly be more straight-forward.

shfrmn:
@tannerlinsley Thank you for the work on the library. I thought this would be an appropriate place to ask some questions, since you just closed this PR very recently.

1. What is the difference between `context` and `beforeLoad`? For now documentation is not updated to explain the `context` option. I assume `context` does not block component rendering? Or does it?
2. What is the reason the context type is not restricted by the root route? Meaning, `beforeLoad` and `context` can set a context of type that's different from the defined root context type or not return any context at all. I expected to get TypeScript errors in those cases to prevent incorrect context states. For example, breadcrumbs feature requires every page to have a title, but there is apparently no way to enforce this on the type level.
3. Since context is inherited from the parent component, there's an undesired side effect when using it for features like breadcrumbs. Let's say we have a route `/user/$userId`. While user object is being loaded, the breadcrumb data will end up being duplicated, something like `[{pageTitle: "Users"}, {pageTitle: "Users"}]` (because the second match initially inherits the context identical to the first). It would be helpful to have some way to prevent context inheritance. Perhaps `context` + `beforeLoad` actually allow to achieve that, but I can't be sure. If not perhaps some boolean flag could do. Does it make any sense?

Thank you and have a good day!

SeanCassiere:
`1` will be documented, we just haven't gotten around to it yet.

If anything is breaking, please create an issue for it. These closed PRs don't get much visibility in our inboxes 😅

PR #2117, opened 2024-08-10
fix(start): using TextEncoder to encode the chunks when using `renderToReadableStream`

SeanCassiere:
Using PR to test out the preview packages on the cf-deployment.

This led to CF deployments just showing a blank screen with this error in the dashboard.

image

<details><summary>"This ReadableStream did not return bytes."</summary>
<pre>
{
  "truncated": false,
  "outcome": "exception",
  "scriptName": "pages-worker--3280425-production",
  "diagnosticsChannelEvents": [],
  "exceptions": [
    {
      "name": "TypeError",
      "message": "This ReadableStream did not return bytes.",
      "timestamp": 1723290907130
    }
  ],
  "logs": [],
  "eventTimestamp": 1723290907070,
  "event": {
    "request": {
      ...
    },
    "response": {
      "status": 200
    }
  },
  "id": 0
}
</pre>
</details> 

CF deployment: https://try-out-the-textencoder-chan.start-basic-cf-pages.pages.dev/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ artifacts
 .rpt2_cache
 coverage
 *.tgz
+.wrangler
 
 # tests
 packages/router-generator/tests/**/*.gen.ts
--- a/packages/react-router/src/Matches.tsx
+++ b/packages/react-router/src/Matches.tsx
@@ -109,15 +109,11 @@ export function Matches() {
     <router.options.defaultPendingComponent />
   ) : null
 
-  const ResolvedSuspense = !router.state.matches.length
-    ? React.Suspense
-    : SafeFragment
-
   const inner = (
-    <ResolvedSuspense fallback={pendingElement}>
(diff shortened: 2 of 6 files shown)


SeanCassiere:
Definitely on the right track here:

Now the invariant error UI is being displayed, with the correct HTML being sent down.

image

Likely something to do with the hydration or reading from the `__TSR__` globals is causing this invariant failure.

SeanCassiere:
Currently on "cf-pages" deployments, two known errors do exist.

- On the client, hydration fails because of the invariant failure. 
  - We know that React is working though since the invariant error is being displayed using the default TSR `<ErrorComponent>` with the button successfully toggling the `React.useState` boolean for the "open" state.
- Loading the `/posts` route fails, because the routeTree has an error listed in the output logs.
  - `"Error: Context is not available"`

PR #2118, opened 2024-08-11
fix(start): remove debug log that causes dehydration errors

lithdew:
The debug 'Injected from Server: ' logs are not properly escaped, causing dehydration to fail. These logs have been removed.

`<ScriptOnce/>` debug logs are removed, and the 'tsr-once' class name set on <ScriptOnce/> elements has an extraneous space suffixed to it which has been removed as well.

image

### DIFF
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -2558,15 +2558,10 @@ export class Router<
     )
 
     this.streamedKeys.add(key)
-    const children = `__TSR__.streamedValues['${key}'] = { value: ${this.serializer?.(this.options.transformer.stringify(value))}}`
+    const children = `__TSR__.streamedValues['${key}'] = { value: ${this.serializer?.(this.options.transformer.stringify(value))} }`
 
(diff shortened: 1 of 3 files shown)


packages/react-router/src/router.ts
```diff
@@ -2558,15 +2558,10 @@ export class Router<
...
+    const children = `__TSR__.streamedValues['${key}'] = { value: ${this.serializer?.(this.options.transformer.stringify(value))} }`
 
     this.injectHtml(
-      `<script class='tsr-once'>${children}${
-        process.env.NODE_ENV === 'development'
-          ? `; console.info(\`Injected From Server:
-        ${children}\`)`
-          : ''
-      }; __TSR__.cleanScripts()</script>`,
+      `<script class='tsr-once'>${children}; __TSR__.cleanScripts()</script>`,
```
SeanCassiere, line 2564: Do we need to remove the logging for this? Couldn't we just correctly escape the `console.log` statements instead?

PR #2120, opened 2024-08-12
fix(react-router,start): ensure document is only used when available

lo1tuma:
I’m trying to use `react-test-renderer` to render a component that uses router hooks or components like `Meta`. It seems like `react-test-renderer` is always executing `useLayoutEffect` and `useEffect`. Since I don’t have a global `window` or `document` object during my test run, some components fail when using `document` without checking if it exist.

See also discussion https://github.com/TanStack/router/discussions/2109.

### DIFF
--- a/packages/react-router/src/Transitioner.tsx
+++ b/packages/react-router/src/Transitioner.tsx
@@ -111,13 +111,11 @@ export function Transitioner() {
         resolvedLocation: s.location,
       }))
 
-      if (!router.isServer) {
-        if (tyepof document !== 'undefined' && (document as any).querySelector) {
-          if (router.state.location.hash !== '') {
-            const el = document.getElementById(router.state.location.hash)
-            if (el) {
-              el.scrollIntoView()
-            }
+      if (typeof document !== 'undefined' && (document as any).querySelector) {
+        if (router.state.location.hash !== '') {
+          const el = document.getElementById(router.state.location.hash)
+          if (el) {
+            el.scrollIntoView()
           }
         }
       }
--- a/packages/react-router/src/Transitioner.tsx
+++ b/packages/react-router/src/Transitioner.tsx
@@ -56,7 +56,7 @@ export function Transitioner() {
   // Try to load the initial location
   useLayoutEffect(() => {
     if (
-      window.__TSR__?.dehydrated ||
+      (typeof window !== 'undefined' && window.__TSR__?.dehydrated) ||
       (mountLoadForRouter.current.router === router &&
         mountLoadForRouter.current.mounted)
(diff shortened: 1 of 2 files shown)


packages/react-router/src/Transitioner.tsx
```diff
@@ -55,6 +55,9 @@ export function Transitioner() {
 
   // Try to load the initial location
   useLayoutEffect(() => {
+    if (router.isServer) {
```
SeanCassiere, line 58: I'd get rid of the `route.isServer` exit condition and just check below for
```
(typeof window !== 'undefined' && window.__TSR__?.dehydrated) ||
(mountLoadForRouter.current.router === router && mountLoadForRouter.current.mounted)
```

packages/start/src/client/Meta.tsx
```diff
@@ -173,6 +173,10 @@ export const Meta = ({ children }: { children?: React.ReactNode }) => {
   React[
     typeof document !== 'undefined' ? 'useLayoutEffect' : 'useEffect'
   ](() => {
+    if (router.isServer) {
```
SeanCassiere, line 176: Actually, check for the document.

```suggestion
    if (typeof document !== 'undefined') {
```

packages/start/src/client/Meta.tsx
```diff
@@ -190,7 +194,7 @@ export const Meta = ({ children }: { children?: React.ReactNode }) => {
     end?.remove()
 
     setMounted(true)
-  }, [])
+  }, [router.isServer])
```
SeanCassiere, line 197: No longer needed.

```suggestion
  }, [])
```

SeanCassiere (changes requested):

lo1tuma:
@SeanCassiere thanks for the review. I’ve addressed all comments so far.

packages/react-router/src/Transitioner.tsx
```diff
@@ -111,11 +111,13 @@ export function Transitioner() {
         resolvedLocation: s.location,
       }))
 
-      if ((document as any).querySelector) {
-        if (router.state.location.hash !== '') {
-          const el = document.getElementById(router.state.location.hash)
-          if (el) {
-            el.scrollIntoView()
+      if (!router.isServer) {
+        if ((document as any).querySelector) {
```
SeanCassiere, line 115:
```suggestion
        if (tyepof document !== 'undefined' && (document as any).querySelector) {
```
Here as well.

SeanCassiere (changes requested):

SeanCassiere (approved):

PR #2161, opened 2024-08-20
perf(react-router): reduce type parameters of route definitions

chorobin:
This just optimizes the types a bit after the `context` and other merges. Mainly less type parameters continuing on the tread of reducing them


packages/react-router/src/fileRoute.ts
```diff
@@ -6,23 +6,17 @@ import { useLoaderData } from './useLoaderData'
 import { useSearch } from './useSearch'
 import { useParams } from './useParams'
 import { useNavigate } from './useNavigate'
-import type { NoInfer } from '@tanstack/react-store'
+import { type Constrain } from './utils'
```
SeanCassiere, line 9:
```suggestion
import type { Constrain } from './utils'
```
Just for consistency.

PR #2166, opened 2024-08-22
fix(start): load meta only on successful hydration

lithdew:
When a route is matched to be unsuccessful such that no loader data is available, the meta function still gets called with undefined loader data. Given that the type signature of the meta function assumes that loader data is always to be defined, a change was made to only call the meta function when a match is successful.

### DIFF
--- a/packages/start/src/client/serialization.tsx
+++ b/packages/start/src/client/serialization.tsx
@@ -110,13 +110,18 @@ export function afterHydrate({ router }: { router: AnyRouter }) {
       }
     }
 
+    const meta =
+      match.status === 'success'
+        ? route.options.meta?.({
+            matches: router.state.matches,
+            match,
+            params: match.params,
+            loaderData: match.loaderData,
+          })
+        : undefined
+
     Object.assign(match, {
-      meta: route.options.meta?.({
-        matches: router.state.matches,
-        match,
-        params: match.params,
-        loaderData: match.loaderData,
-      }),
+      meta,
       links: route.options.links?.(),
       scripts: route.options.scripts?.(),
     })


packages/react-router/src/router.ts
```diff
@@ -2675,7 +2674,7 @@ export class Router<
     )
 
     this.streamedKeys.add(key)
-    const children = `__TSR__.streamedValues['${key}'] = { value: ${this.serializer?.(this.options.transformer.stringify(value))}}`
+    const children = `__TSR__.streamedValues['${key}'] = ${jsesc({ value: this.serializer?.(this.options.transformer.stringify(value)) }, { isScriptContext: true, wrap: true, json: true })}`
```
SeanCassiere, line 2677: What's the purpose of this change?

lithdew: Depending on the transformer/serializer used, the streamed value may still contain illegal characters that need to be escaped.

SeanCassiere: Are you running into this issue?

I'm asking since the `defaultTransformer` handles `undefined` and `Date` values already.

If you aren't running this issue, I'd rather not add this package into core, that'd have a downstream effect on SPA users as well.

lithdew: Yep, I tried to use @tanstack/react-router-with-query - I am using `devalue` as a transformer b/c there are some values i.e. `Decimal` that cannot be serialized with the default transformer. Using `devalue` as a transformer causes illegal values to be injected into the <script/> tag which causes hydration to fail.

tannerlinsley:
The meta change makes sense I think. jsesc should probably instead be injected by the server utilities, same as router.serializer

lithdew:
Removed the stream log changes for now.

SeanCassiere (approved):

PR #2172, opened 2024-08-23
fix(start): deferred `createServerFn` calls fails on the client

SeanCassiere:
When wrapping a `createServerFn` call that returns something more complex than a number or string (ie: an Object or an Array), it fails on the client since the `transformer` uses `JSON.parse` which expects a string input (but can also gracefully handle a number input as well).

This fails on the client without the changes being applied in this PR.
```tsx
import { Await, createFileRoute, defer } from '@tanstack/react-router'
import { createServerFn } from '@tanstack/start'
import { Suspense, useState } from 'react'

const serverObject = createServerFn('GET', async () => {
  await new Promise((r) => setTimeout(r, 1000))
  return {
    foo: 'bar',
  }
})

export const Route = createFileRoute('/deferred')({
  loader: () => {
    return {
      deferredObject: defer(serverObject()),
    }
  },
  component: Deferred,
})

function Deferred() {
  const [count, setCount] = useState(0)
  const { deferredObject } = Route.useLoaderData()

  return (
    <div className="p-2">
      <Suspense fallback="Loading...">
        <Await
          promise={deferredObject}
          children={(data) => (
            <h3>
              {data.foo} // uh oh error happens here
            </h3>
          )}
        />
      </Suspense>
      <div>Count: {count}</div>
      <div>
        <button onClick={() => setCount(count + 1)}>Increment</button>
      </div>
    </div>
  )
}

```

### DIFF
--- a/packages/start/src/client/serialization.tsx
+++ b/packages/start/src/client/serialization.tsx
@@ -101,9 +101,9 @@ export function afterHydrate({ router }: { router: AnyRouter }) {
         Object.entries(extracted).forEach(([_, ex]: any) => {
           if (ex.value instanceof Promise) {
             const og = ex.value
-            ex.value = og.then((data: any) =>
-              router.options.transformer.parse(data),
-            )
+            ex.value = og.then((data: any) => {
+              return data
+            })
           }
           deepMutableSetByPath(match, ['loaderData', ...ex.path], ex.value)
         })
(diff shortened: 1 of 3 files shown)


packages/start/src/client/serialization.tsx
```diff
@@ -101,9 +101,9 @@ export function afterHydrate({ router }: { router: AnyRouter }) {
         Object.entries(extracted).forEach(([_, ex]: any) => {
           if (ex.value instanceof Promise) {
             const og = ex.value
-            ex.value = og.then((data: any) =>
-              router.options.transformer.parse(data),
-            )
+            ex.value = og.then((data: any) => {
+              return data
```
SeanCassiere, line 105: When it gets here, it's already in Javascript form (ie: `Object`, `Array`, ...) and not just a string.

PR #2193, opened 2024-08-27
Update getting-started.md

sorrycc:
Remove trailing commas that will cause this problem.

✘ [ERROR] JSON does not support trailing commas

    package.json:26:26:
      26 │     "start": "vinxi start",
         ╵


SeanCassiere (approved):

SeanCassiere (approved):
We are using `jsonc` for that block, which is why that trailing comma is added.

This is low-surface enough that the `jsonc` comment for showing the user that they should spread the rest of their `package.json` configuration is worth it.

PR #2194, opened 2024-08-27
fix(react-router): accept valid ReactNode as route component

honzahruby:
In `route.ts`, there's a custom alias for `ReactNode` set to `any`. Based on the commit history alone, it looks like it was introduced for backward compatibility with typescript <5.1. This PR conditionally uses the real type from react if typescript >5.0 is available.

Using `any` allowed to pass any function as a route component, even when it wasn't a valid react component, causing a runtime crash.

### DIFF
--- a/packages/react-router/src/route.ts
+++ b/packages/react-router/src/route.ts
@@ -1499,8 +1499,8 @@ export type NotFoundRouteProps = {
 export type ReactNode = any
 
 export type SyncRouteComponent<TProps> =
-  | ((props: TProps) => ReactNode)
-  | React.LazyExoticComponent<(props: TProps) => ReactNode>
+  | ((props: TProps) => React.ReactNode)
+  | React.LazyExoticComponent<(props: TProps) => React.ReactNode>
 
 export type AsyncRouteComponent<TProps> = SyncRouteComponent<TProps> & {
   preload?: () => Promise<void>
--- a/packages/react-router/src/route.ts
+++ b/packages/react-router/src/route.ts
@@ -1495,7 +1495,14 @@ export type NotFoundRouteProps = {
   data: unknown
 }
 
-// workaround for typescript 5.0, can be replaced with React.ReactNode once 5.0 is not supported
+/**
+ * Up until typescript 5.0, react components could only return a JSX element or null.
+ * Typescript 5.1 changed the signature for React.JSXElementConstructor,
+ * which allows components to return any valid ReactNode, e.g. string or undefined.
+ *
+ * Since we still support typescript 5.0, we need to use this workaround to support both versions.
+ * Once typescript 5.0 is not supported, this type can be removed and replaced by React.ReactNode.
+ */
 export type ReactNode =
   (() => 'x') extends React.JSXElementConstructor<any>
     ? React.ReactNode
--- a/packages/react-router/src/route.ts
+++ b/packages/react-router/src/route.ts
@@ -1496,7 +1496,10 @@ export type NotFoundRouteProps = {
 }
 
 // workaround for typescript 5.0, can be replaced with React.ReactNode once 5.0 is not supported
-export type ReactNode = React.ReactElement | null
+export type ReactNode =
+  (() => 'x') extends React.JSXElementConstructor<any>
+    ? React.ReactNode
+    : React.ReactElement | null
 
 export type SyncRouteComponent<TProps> =
   | ((props: TProps) => ReactNode)
--- a/packages/react-router/src/route.ts
+++ b/packages/react-router/src/route.ts
@@ -1494,13 +1494,13 @@ export type NotFoundRouteProps = {
   // TODO: Make sure this is `| null | undefined` (this is for global not-founds)
   data: unknown
 }
-//
 
-export type ReactNode = any
+// workaround for typescript 5.0, can be replaced with React.ReactNode once 5.0 is not supported
+export type ReactNode = React.ReactElement | null
 
 export type SyncRouteComponent<TProps> =
-  | ((props: TProps) => React.ReactNode)
-  | React.LazyExoticComponent<(props: TProps) => React.ReactNode>
+  | ((props: TProps) => ReactNode)
+  | React.LazyExoticComponent<(props: TProps) => ReactNode>
 
 export type AsyncRouteComponent<TProps> = SyncRouteComponent<TProps> & {
   preload?: () => Promise<void>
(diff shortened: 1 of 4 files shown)


honzahruby:
@SeanCassiere Would you have a look? 🙏

packages/react-router/src/route.ts
```diff
@@ -1461,9 +1461,12 @@ export type NotFoundRouteProps = {
   // TODO: Make sure this is `| null | undefined` (this is for global not-founds)
   data: unknown
 }
-//
 
-export type ReactNode = any
+// workaround for typescript 5.0, can be replaced with React.ReactNode once 5.0 is not supported
+export type ReactNode =
```
schiller-manuel, line 1466: what's that workaround doing exactly? why does TS 5.0 not like `React.ReactNode`?

SeanCassiere: Good question. I feel like `React.ReactNode` has been around for quite a while. Even before TS `5.0` I believe.

@honzahruby is there any reason why we can't just default it to `React.ReactNode`?

honzahruby: @schiller-manuel @SeanCassiere

Sure, happy to elaborate 🙌 

Typescript 5.1 added support for more constructable JSX types, other than `JSX.Element`, see the [release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-1.html#decoupled-type-checking-between-jsx-elements-and-jsx-tag-types) for more details.

React itself (not considering types) allows components to return other types than just JSX elements, e.g. strings. On the type level, this is reflected in a way that @types/react now comes with two versions - one for TS 5.0 and one for the rest.

The main difference boils down to this line of code in what react considers to be contructable JSX: [5.0](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d9b1cf511408316932aff73e4bcc8749613c871d/types/react/ts5.0/index.d.ts#L124) vs [>5.0](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d9b1cf511408316932aff73e4bcc8749613c871d/types/react/index.d.ts#L124)

When using `React.ReactNode` directly, the type check for TS 5.0 results in these errors:

```sh
│ src/CatchBoundary.tsx(19,38): error TS2769: No overload matches this call.
│   The last overload gave the following error.
│     Argument of type '(((props: ErrorComponentProps) => ReactNode) & { preload?: (() => Promise<void>) | undefined; }) | (({ error }: { error: any; }) => Element)' is not assignable to parameter of type 'string | FunctionComponent<ErrorComponentProps> | ComponentClas…
│       Type '((props: ErrorComponentProps) => ReactNode) & { preload?: (() => Promise<void>) | undefined; }' is not assignable to type 'string | FunctionComponent<ErrorComponentProps> | ComponentClass<ErrorComponentProps, any>'.
│         Type '((props: ErrorComponentProps) => ReactNode) & { preload?: (() => Promise<void>) | undefined; }' is not assignable to type 'FunctionComponent<ErrorComponentProps>'.
│           Type 'ReactNode' is not assignable to type 'ReactElement<any, any> | null'.
│ src/Match.tsx(39,46): error TS2786: 'PendingComponent' cannot be used as a JSX component.
│   Its return type 'ReactNode' is not a valid JSX element.
│     Type 'undefined' is not assignable to type 'Element | null'.
│ src/Match.tsx(97,42): error TS2769: No overload matches this call.
│   The last overload gave the following error.
│     Argument of type '(props: NotFoundRouteProps) => ReactNode' is not assignable to parameter of type 'string | FunctionComponent<NotFoundRouteProps> | ComponentClass<NotFoundRouteProps, any>'.
│       Type '(props: NotFoundRouteProps) => ReactNode' is not assignable to type 'FunctionComponent<NotFoundRouteProps>'.
│         Type 'ReactNode' is not assignable to type 'ReactElement<any, any> | null'.
│           Type 'undefined' is not assignable to type 'ReactElement<any, any> | null'.
│ src/Match.tsx(295,6): error TS2786: 'router.options.defaultPendingComponent' cannot be used as a JSX component.
│   Its return type 'string | number | boolean | ReactElement<any, string | JSXElementConstructor<any>> | Iterable<ReactNode> | null | undefined' is not a valid JSX element.
│ src/Matches.tsx(209,6): error TS2786: 'router.options.defaultPendingComponent' cannot be used as a JSX component.
│   Its return type 'string | number | boolean | ReactElement<any, string | JSXElementConstructor<any>> | Iterable<ReactNode> | null | undefined' is not a valid JSX element.
│ src/renderRouteNotFound.tsx(14,15): error TS2786: 'router.options.defaultNotFoundComponent' cannot be used as a JSX component.
│   Its return type 'string | number | boolean | ReactElement<any, string | JSXElementConstructor<any>> | Iterable<ReactNode> | null | undefined' is not a valid JSX element.
│     Type 'undefined' is not assignable to type 'Element | null'.
│ src/renderRouteNotFound.tsx(27,11): error TS2786: 'route.options.notFoundComponent' cannot be used as a JSX component.
│   Its return type 'string | number | boolean | ReactElement<any, string | JSXElementConstructor<any>> | Iterable<ReactNode> | null | undefined' is not a valid JSX element.
│ src/route.ts(1473,31): error TS2344: Type '(props: TProps) => ReactNode' does not satisfy the constraint 'ComponentType<any>'.
│   Type '(props: TProps) => ReactNode' is not assignable to type 'FunctionComponent<any>'.
│     Type 'ReactNode' is not assignable to type 'ReactElement<any, any> | null'.
```

I can update the comment in code to explain more of this, feel free to do it yourself if you can put it in better words 🙂

schiller-manuel: @chorobin can you have a look please ?

chorobin: Yep, this is correct but I'm not sure if changing the types to `React.ReactElement` for 5.0 is correct because from a public interface point of view, we should accept `string` etc in 5.0 aswell

honzahruby: React itself doesn’t accept a string if you’re running on TS 5.0, this just makes it consistent 🤔

honzahruby: @SeanCassiere @chorobin I've updated the comment, could you have another look, please? 🙏

SeanCassiere (approved):
As long as we don't have a type-level regression, I'm fine with it.

packages/react-router/src/route.ts
schiller-manuel: @chorobin will check again once he finished up a big PR he is currently working on

SeanCassiere (commented):
Now that we've dropped TS `5.0` in favor of `5.1` as the minimum version, can we just set this to `React.ReactNode`?

SeanCassiere:
Closing this PR, as the minimum typescript version has been moved up to `5.2` as listed on the website.

@honzahruby thank you for contributing towards this! we do very much appreciate it!

PR #2196, opened 2024-08-27
docs(start): fix arguments for `createServerFn`

yuheiy:
This PR fixes a problem where the `context` object that should be the second argument was the first argument.


SeanCassiere:
Thank you for contributing this correction, but it looks like Tanner's also corrected this with https://github.com/TanStack/router/commit/610fd101c7d5ec04b9749abf519ff920211c9aee 😅.

Closing this, since the correction is already in.

PR #2203, opened 2024-08-27
feat(router-generator): allow to customize `index` and `route` token

schiller-manuel:


SeanCassiere (commented):
Should we have rules about setting tokens starting with an underscore (`_`)? since that could clash with a pathless/layout route.

For example, what would happen if the `index` or `route` tokens are changed to `_index` or `_route`.
So, what'd happen when then this route is created `src/routes/dashboard/_index.tsx`? Would this be an `index` route, a pathless/layout route, or both?

Perhaps just disallow, `_index` but not `__index`.

schiller-manuel:
I need to test but I think underscore prefix does not cause an issue since index/route tokens are replaced before we parse the route name

SeanCassiere:
> I need to test but I think underscore prefix does not cause an issue since index/route tokens are replaced before we parse the route name

Definitely something worth adding a test for since the `routePath` replacement currently happens after the layout route status has been determined.

tannerlinsley:
@arackaf

SeanCassiere (approved):

Issue #2174, opened 2024-08-23
Routes with lazy and those with loader async do not set view to elementet id after refresh


JayCodeGitHub:
### Describe the bug

Routes with lazy and those with loader async do not set view to elementet id after refresh. When it is on the path to a specific element with id

files `.lazy`, `loader: async () => {},`

### Your Example Website or App

https://stackblitz.com/edit/vitejs-vite-rz7exv

### Steps to Reproduce the Bug or Issue

1. Go to Form (lazy) or Post (async loader)
2. Click on Test link which leads to an element with id
3. Refresh the site (The side inside the sandbox, not the entire sandbox)

### Expected behavior

I would expect the viewer's view to be set on the element the link points to as it is on paths without lazy

### Screenshots or Videos

https://github.com/user-attachments/assets/38969708-9927-4872-83ca-1141fd71975b

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari]
- Version: [e.g. 91.1]

schiller-manuel:
seems to be a duplicate of https://github.com/TanStack/router/issues/2014, but #2014 does not mention the loader issue

SeanCassiere:
Completed with https://github.com/TanStack/router/releases/tag/v1.98.1.

Issue #2014, opened 2024-07-23
Hashed link not scrolling with lazy file routes


patatoes1235:
### Describe the bug

Urls with hashes not scrolling to correct section when using lazy file routes. This occurs any time the page is fully rerendered (e.g. on reload and when accessing a link with a hash in a new tab). Works as expected with normal (non-lazy) file routing. 

### Your Example Website or App

https://stackblitz.com/edit/github-4rybtq

### Steps to Reproduce the Bug or Issue

1. Run the stackblitz example
2. Paste [https://stackblitz.com/edit/github-4rybtq/#section](https://stackblitz.com/edit/github-4rybtq/#section) into a new tab
3. Click "connect" if needed
4. Notice how even though the #section tag is appended to the url, the section is not scrolled to. 

### Expected behavior

Regardless of whether the route is lazy loaded or not, I would expect that when accessing a url with a section, the DOM element with the corresponding `id` is scrolled to. 

### Platform

- OS: Windows
- Browser: Chrome

SeanCassiere:
This also doesn't seem to work every with the `<ScrollRestoration />` component.

patatoes1235:
My best guess is that the browser is trying to scroll down before the lazily loaded component is fully rendered.

SeanCassiere:
Completed with https://github.com/TanStack/router/releases/tag/v1.98.1

SeanCassiere:
Closing once again, after confirming that this issue is indeed fixed.

There was an issue with the original reproduction, whereby the `div` used a `my-[100px]` which in turn moved the start of the element much higher than where the actual content was placed.

PR #2205, opened 2024-08-27
test(react-router): e2e scroll restoration reproducers

SeanCassiere:
Reproductions for
- #2014
- #2174


SeanCassiere:
Looks like this PR is no longer needed since everything is already available on "main" with the merging of https://github.com/TanStack/router/pull/3191.

PR #2213, opened 2024-08-29
refactor(react-router): support the `exactOptionalPropertyTypes` ts option in `<Link>`

kshramt:


SeanCassiere:
https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes

kshramt:
CI failures addressed and changes force pushed.

schiller-manuel:
do we have runtime  tests and type tests for all those situations where you can pass in an explicit undefined?

kshramt:
To clarify, this PR doesn't alter the type definition, but rather makes the existing behavior more explicit (see https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes ).

schiller-manuel:
thanks for clarification.
can you please add the `exactOptionalPropertyTypes`flag to the root tsconfig.json so we catch this with our type tests?

kshramt:
Setting `"exactOptionalPropertyTypes": true` in the root tsconfig.json will result in many type-check errors like this:

```
src/CatchBoundary.tsx:14:6 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(props: { getResetKey: () => string | number; children: (props: { error: Error | null; reset: () => void; }) => ReactNode; onCatch?: (error: Error, errorInfo: ErrorInfo) => void; }): CatchBoundaryImpl', gave the following error.
    Type '{ getResetKey: () => string | number; onCatch: ((error: Error, errorInfo: ErrorInfo) => void) | undefined; children: ({ error, reset }: { error: Error | null; reset: () => void; }) => string | ... 6 more ... | undefined; }' is not assignable to type 'Readonly<{ getResetKey: () => string | number; children: (props: { error: Error | null; reset: () => void; }) => ReactNode; onCatch?: (error: Error, errorInfo: ErrorInfo) => void; }>' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
      Types of property 'onCatch' are incompatible.
        Type '((error: Error, errorInfo: ErrorInfo) => void) | undefined' is not assignable to type '(error: Error, errorInfo: ErrorInfo) => void'.
          Type 'undefined' is not assignable to type '(error: Error, errorInfo: ErrorInfo) => void'.
  Overload 2 of 2, '(props: { getResetKey: () => string | number; children: (props: { error: Error | null; reset: () => void; }) => ReactNode; onCatch?: (error: Error, errorInfo: ErrorInfo) => void; }, context: any): CatchBoundaryImpl', gave the following error.
    Type '{ getResetKey: () => string | number; onCatch: ((error: Error, errorInfo: ErrorInfo) => void) | undefined; children: ({ error, reset }: { error: Error | null; reset: () => void; }) => string | ... 6 more ... | undefined; }' is not assignable to type 'Readonly<{ getResetKey: () => string | number; children: (props: { error: Error | null; reset: () => void; }) => ReactNode; onCatch?: (error: Error, errorInfo: ErrorInfo) => void; }>' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
      Types of property 'onCatch' are incompatible.
        Type '((error: Error, errorInfo: ErrorInfo) => void) | undefined' is not assignable to type '(error: Error, errorInfo: ErrorInfo) => void'.
          Type 'undefined' is not assignable to type '(error: Error, errorInfo: ErrorInfo) => void'.

14     <CatchBoundaryImpl
```

There are many places outside of `link.tsx` that implicitly assume `x?: T` allows `undefined`, making them incompatible with this flag.

schiller-manuel:
would it make sense to fix all of those issues?

kshramt:
WIP: I am trying to add the flag to `packages/react-router/tsconfig.json` (changing the root `tsconfig.json` seems a bit involved).
It seems that https://github.com/TanStack/router/blob/f36f92b459bf0aabbc4d49c1e72c8b2ad6894622/packages/react-router/src/router.ts#L707-L710 may not work well with explicit `undefined`s, as `{...{transformer: 1}, ...{transformer: undefined}}` results in `{transformer: undefined}`.

kshramt:
Sorry, as this turned out to be more complex than I expected, I have rolled back the branch to https://github.com/TanStack/router/pull/2213/commits/7d892ac8ef96a46995b4fff31481f1cc2b6e927a . Since I've already implemented a workaround without this PR, please feel free to close it, although I believe the changes, while not comprehensive, could still be useful in some cases.

SeanCassiere:
Closing this as there's no movement in the requirement for `exactOptionalPropertyTypes`.

PR #2229, opened 2024-08-30
adds `onStreamedValue` router event, `Router.streamedKeys` return a value in the browser

phryneas:
Just a suggestion for some additional functionality - do you think going this way would be feasible?


phryneas:
closing in favor of #2698

PR #2243, opened 2024-09-01
perf(react-router): constant editor performance regardless how large a file based route tree is

chorobin:
# What?

Editor performance for file based routing will now remain constant for suggestions no matter how large a route tree is.

This is visible when tracing our large file based example.

Before this change, it looked like this:

tsr-current

The large wall was when TS hit a `Link` for the first time and had to type check the whole route tree. Unfortunately this is what our editors were doing when we were interacting with `Link` for example.

After this change, the trace is very different

tsr-healthy

This means that TS no longer needs to type check the whole route tree when it hits `Link` for the first time. **This translates to a very large performance improvement for our editors** but type checking time remains unaffected because the work is now split between each `Link` that references the particular route.

This is achieved by generating types which are easier on the compiler however, **the DX will remain the same**. **Everything in route definitions remains totally inferred**

# How?

It comes down to generating interfaces to help the compiler resolve a route on the type level, TS no longer needs to walk down the whole route tree, it can quickly resolve a route by interfaces which are generated. This means code based and file based type level logic needs to diverge and there are essentially short circuit cases for file based routing


examples/react/large-file-based/src/routeTree.gen.ts
```diff
@@ -127,18 +127,97 @@ declare module '@tanstack/react-router' {
...
-    ParamsParamsPlaceholderRoute,
-  }),
-  SearchRouteRoute: SearchRouteRoute.addChildren({
-    SearchSearchPlaceholderRoute,
-  }),
-  AbsoluteRoute,
-  LinkPropsRoute,
-  RelativeRoute,
-})
+const IndexRouteWithChildren = IndexRoute
```
chorobin, line 130: I'll get rid of the these redundant const's.

examples/react/large-file-based/src/routeTree.gen.ts
```diff
@@ -127,18 +127,97 @@ declare module '@tanstack/react-router' {
...
+const ParamsParamsPlaceholderRouteWithChildren = ParamsParamsPlaceholderRoute
+interface ParamsRouteRouteChildren {
```
chorobin, line 132: We now declare children as an interface. This means basically just delays the inference until you index into the interface

examples/react/large-file-based/src/routeTree.gen.ts
```diff
@@ -127,18 +127,97 @@ declare module '@tanstack/react-router' {
...
+  SearchSearchPlaceholderRoute: SearchSearchPlaceholderRouteWithChildren,
+}
+const SearchRouteRouteWithChildren = SearchRouteRoute.addChildren(
+  SearchRouteRouteChildren,
+)
+const AbsoluteRouteWithChildren = AbsoluteRoute
+const LinkPropsRouteWithChildren = LinkPropsRoute
+const RelativeRouteWithChildren = RelativeRoute
+
+interface FileRoutesByFullPath {
```
chorobin, line 155: This allows us to resolve routes without parsing the route tree

examples/react/large-file-based/src/routeTree.gen.ts
```diff
@@ -127,18 +127,97 @@ declare module '@tanstack/react-router' {
...
+  '/params': typeof ParamsRouteRouteWithChildren
+  '/search': typeof SearchRouteRouteWithChildren
+  '/absolute': typeof AbsoluteRouteWithChildren
+  '/linkProps': typeof LinkPropsRouteWithChildren
+  '/relative': typeof RelativeRouteWithChildren
+  '/params/$paramsPlaceholder': typeof ParamsParamsPlaceholderRouteWithChildren
+  '/search/searchPlaceholder': typeof SearchSearchPlaceholderRouteWithChildren
+}
+
+interface FileRoutesByTo {
```
chorobin, line 166: This allows us to resolve routes without parsing the route tree for `to`

examples/react/large-file-based/src/routeTree.gen.ts
```diff
@@ -127,18 +127,97 @@ declare module '@tanstack/react-router' {
...
+  '/params': typeof ParamsRouteRouteWithChildren
+  '/search': typeof SearchRouteRouteWithChildren
+  '/absolute': typeof AbsoluteRouteWithChildren
+  '/linkProps': typeof LinkPropsRouteWithChildren
+  '/relative': typeof RelativeRouteWithChildren
+  '/params/$paramsPlaceholder': typeof ParamsParamsPlaceholderRouteWithChildren
+  '/search/searchPlaceholder': typeof SearchSearchPlaceholderRouteWithChildren
+}
+
+interface FileRouteTypes {
```
chorobin, line 177: Note. Need to add `id` aswell

examples/react/large-file-based/src/routeTree.gen.ts
```diff
@@ -127,18 +127,97 @@ declare module '@tanstack/react-router' {
...
+  ParamsRouteRoute: ParamsRouteRouteWithChildren,
+  SearchRouteRoute: SearchRouteRouteWithChildren,
+  AbsoluteRoute: AbsoluteRouteWithChildren,
+  LinkPropsRoute: LinkPropsRouteWithChildren,
+  RelativeRoute: RelativeRouteWithChildren,
+}
+
+export const routeTree = rootRoute
+  .addChildren(rootRouteChildren)
+  ._addFileTypes<FileRouteTypes>()
```
chorobin, line 220: I've taken this approach so I can write type tests for it. Effectively I'm attaching additional types to the route tree which will short circuit types for later

packages/react-router/src/route.ts
```diff
@@ -1056,11 +1056,7 @@ export class Route<
     this._to = fullPath as TrimPathRight<TFullPath>
   }
 
-  addChildren<
```
chorobin, line 1059: I need to undo this change and possible add a new internal method to add children only for file based

packages/react-router/src/routeInfo.ts
```diff
@@ -53,22 +55,34 @@ export type CurrentPath<TOption> = 'always' extends TOption
...
-  | CurrentPath<TOption>
-  | ParentPath<TOption>
-  | ''
+export type CatchAllPaths<TOption> = CurrentPath<TOption> | ParentPath<TOption>
 
-export type RoutesByPath<TRouteTree extends AnyRoute> = {
-  [K in ParseRoute<TRouteTree> as K['fullPath']]: K
-}
+export type CodeRoutesByPath<TRouteTree extends AnyRoute> =
+  ParseRoute<TRouteTree> extends infer TRoutes extends AnyRoute
```
chorobin, line 61: You'll see this trick used a lot. It basically makes `ParseRoutes` lazily instantiated so that when we go down the file based routing path, the route tree is not parsed at all

examples/react/large-file-based/src/routeTree.gen.ts
schiller-manuel: any downsides of adding those types just for the tests? e.g. does it have performance implications ? 
we could only add them conditionally when building / running the tests

packages/router-generator/tests/generator/custom-tokens/routeTree.snapshot.ts
```diff
@@ -127,15 +127,93 @@ declare module '@tanstack/react-router' {
...
-export const routeTree = rootRoute.addChildren({
-  R1nd3xRoute,
-  BlogR0ut3Route: BlogR0ut3Route.addChildren({ BlogSlugRoute, Blog1nd3xRoute }),
-  PostsR0ut3Route: PostsR0ut3Route.addChildren({
-    Posts1nd3xRoute,
-    PostsPostIdDeepRoute,
-    PostsPostId1nd3xRoute,
-  }),
-})
+const R1nd3xRouteWithChildren = R1nd3xRoute
```
schiller-manuel, line 130: should we maybe not generate that line for leave routes?

chorobin: Yeah, I'll remove these from the generator

examples/react/large-file-based/src/routeTree.gen.ts
chorobin: Nah, no performance implications from what I can tell

toteto:
This is AMAZING! 🥳 
Today I celebrate!
You probably stopped Apple from receiving my next paycheck for laptop with M4.

stevenlyd:
Is it possible to make this optional(can be turn on or off in the TSR config file)? Currently I'm having a hybrid setup for my project, majority of the routes are file based and part of the routs are code based. I wrote an utility function to help me inject the code based routes into a file based routes as children, and make sure the types flow into the route tree correctly. After this change, it no longer works, since types are generated and exported from the `routeTree.gen.ts`, as long as the plugin is used. I appreciate the performance improvement, but if we can choose between performance or compatibility that would be great! And as long as code based routing will be supported, there is no extra maintenance needed I guess.

chorobin:
> Is it possible to make this optional(can be turn on or off in the TSR config file)? Currently I'm having a hybrid setup for my project, majority of the routes are file based and part of the routs are code based. I wrote an utility function to help me inject the code based routes into a file based routes as children, and make sure the types flow into the route tree correctly. After this change, it no longer works, since types are generated and exported from the `routeTree.gen.ts`, as long as the plugin is used. I appreciate the performance improvement, but if we can choose between performance or compatibility that would be great! And as long as code based routing will be supported, there is no extra maintenance needed I guess.

Hey, Open a question/thread of discord so we can discuss and understand your specific use case

stevenlyd:
> > Is it possible to make this optional(can be turn on or off in the TSR config file)? Currently I'm having a hybrid setup for my project, majority of the routes are file based and part of the routs are code based. I wrote an utility function to help me inject the code based routes into a file based routes as children, and make sure the types flow into the route tree correctly. After this change, it no longer works, since types are generated and exported from the `routeTree.gen.ts`, as long as the plugin is used. I appreciate the performance improvement, but if we can choose between performance or compatibility that would be great! And as long as code based routing will be supported, there is no extra maintenance needed I guess.
> 
> Hey, Open a question/thread on discord so we can discuss and understand your specific use case

No problem, I'll do it later today, thanks for the prompt reply!

PR #2244, opened 2024-09-01
fix(react-router): reset error boundary after the new matches are ready

### LINKED ISSUE #2043: Error propagated from child-route to parent
### Describe the bug

I have 2 routes:
- `/$slug` which displays the detail of an entity
- `/$slug/report` which displays the report for the entity

Bot have `loader` and `errorComponent` on the routes as they have to load some data to display and handle errors. If an exception is thrown on `/$slug/report` route it displays the error from `errorComponent` which is to be expected. But if I navigate (via `Link`) to the detail `/$slug` it also shows the `errorComponent` with the same error as was thrown in the report route.

https://stackblitz.com/edit/github-3mpcwb?file=src%2Fmain.tsx

1. Click on "Report"
2. See error thrown on Report route
3. Click on "Detail"
4. See error thrown on Detail route

### Expected behavior

I would expect the "Detail" route to not thrown an error as none was thrown in it.

### Screenshots or Videos

https://github.com/user-attachments/assets/acf445a9-cd93-4c04-9686-d5d0498f727e

### Platform

- OS: macOS
- Browser: Chrome
- Version: 127.0.6533.72

### Additional context

Hopefully I didn't miss some obvious

### LINKED ISSUE #2162: Failed to reset error component when defaultErrorComponent is set
### Describe the bug

I can't reset error component through navigation when defaultErrorComponent is set

https://stackblitz.com/edit/vitejs-vite-7thcsy

1. Click on the link `About`
2. Error will show up on `About` page
3. Click on the link `Home`
4. Error persist on `Home` page
5. Click on the link `Home` again - error disappears

### Expected behavior

1. Click on the link `About`
2. Error will show up on `About` page
3. Click on the link `Home`
4. Error disappears and `Home` page content should be displayed

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 127.0

### Additional context

Try to remove or comment out defaultErrorComponent - now it's working as expected

cristi-lng:
In the current version of the router, the error boundary is reset at the beginning of the loading phase of the new matching routes. This is a problem because the currently rendered routes can re-render while the new ones ar being in the loading process. If this is the case, the re-render will trigger the error boundary again and will void the reset.

I believe the reset should happen only after the new routes have finished loading. In this way, the reset will be in sync with the matched routes.

I don't know if this fix might have some side effects, but on our project it seems to work. If you think that this fix is not ok or you have a better solution, I would be happy to try it out.

Here is the issue I am referring to: https://github.com/TanStack/router/issues/2162

fixes #2162
fixes #2043

### DIFF
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -1643,11 +1643,6 @@ export class Router<
   load = async (): Promise<void> => {
     this.latestLocation = this.parseLocation(this.latestLocation)
 
-    this.__store.setState((s) => ({
-      ...s,
-      loadedAt: Date.now(),
-    }))
-
     let redirect: ResolvedRedirect | undefined
     let notFound: NotFoundError | undefined
 
@@ -1738,6 +1733,7 @@ export class Router<
                     return {
                       ...s,
                       isLoading: false,
+                      loadedAt: Date.now(),
                       matches: newMatches,
                       pendingMatches: undefined,
                       cachedMatches: [
--- a/packages/react-router/tests/link.test.tsx
+++ b/packages/react-router/tests/link.test.tsx
@@ -760,6 +760,77 @@ describe('Link', () => {
     expect(onError).toHaveBeenCalledOnce()
   })
 
+  test('when navigating away from a route with a loader that errors', async () => {
+    const postsOnError = vi.fn()
+    const indexOnError = vi.fn()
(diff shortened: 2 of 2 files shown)


schiller-manuel:
can you please add a unit test that reproduces the issue?

Issue #2008, opened 2024-07-22
Intent-based loading prevents pendingComponent from showing.


ArmandRedgate:
### Describe the bug

Since #1534 was fixed with #1907, I've found that the pendingComponent will not show if intent-based loading is enabled. The age blocks indefinitely until the dependencies are loaded.

This is better than the behaviour from before the fix, but still leaves something to be desired, and is not the behaviour I'd expect.

### Your Example Website or App

https://stackblitz.com/edit/github-gh73by

In the above, see navigation to the About page, which does not show the specified pending component unless intent-based loading is off.

### Steps to Reproduce the Bug or Issue

Make sure your project has [intent-based loading](https://tanstack.com/router/latest/docs/framework/react/guide/preloading#supported-preloading-strategies) enabled.

1) Load into some part of your UI that contains a link with lazy-loaded code behind it.
2) Slow down your connection speed using browser throttling tools.
3) Hover and immediately click on the link

You will see that you will remain on the current page and no `pendingComponent` will render, even after the default 1 second wait time. This could be several seconds for slower connections and results in a poor user experience as they wonder "did I really click that link?"

### Expected behavior

I expect the pendingComponent to show after the default 1 second wait time, as it does when preloading is disabled.

### Platform

- OS: Windows 11
- Browser: Chrome 
- Version: 126

jude98:
Similar issue here.
I have updated to version `^1.47.1`.
Now when I navigate to a page, it takes few seconds to mount even though the route gets updated, but its not showing any kind of loading states.
My router setup is like:

```javascript
const router = createRouter({
  context: {
    queryClient,
  },
  defaultNotFoundComponent: () => (
    <div>
      <p>Oops, not found!</p>
    </div>
  ),
  defaultPendingComponent: () => <Text>Loading...</Text>,
  // defaultPendingMs: 0,
  defaultPreload: 'intent',
  // defaultPreloadDelay: 0,
  routeTree,
});
```
I have nested routes

`src/routes/~pool/~swap/~$tokenAddress.lazy.tsx` file:
```javascript
export const Route = createLazyFileRoute('/pool/swap/$tokenAddress')({
  component: () => (
      <Trade />
  ),
});
```

`src/routes/~pool/~swap/~$tokenAddress.tsx` file:
```javascript
const Route = createFileRoute('/pool/swap/$tokenAddress')({
  validateSearch: (search: Record<string, unknown>): SwapSearchParams => ({
    in: (search.in as string) || '',
[...]

PR #2247, opened 2024-09-02
fix(react-router): `pendingComponent` not rendering with preload='intent'

MaciejGarncarski:
Hi, I tried fixing this issue #2008. If route had preload: 'intent' enabled, the pendingComponent would not show, UI just waited for promise to end, and then it would switch route. I found out, that triggerOnReady() in router.ts weren't used if route had beforeLoadPromise or loaderPromise to await for.

https://github.com/user-attachments/assets/960609ad-1e4f-4cf9-a376-ad32f16e3e2a

https://github.com/user-attachments/assets/7701b0ba-4def-42a9-b543-89f26e9003bb

### DIFF
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -1997,12 +1997,8 @@ export class Router<
                 existingMatch.beforeLoadPromise ||
                 existingMatch.loaderPromise
               ) {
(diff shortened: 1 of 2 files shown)


schiller-manuel:
can you please add a unit test that reproduces this issue?

SeanCassiere (approved):

PR #2258, opened 2024-09-03
feat: pass `matches` into `context()` and `beforeLoad()`

schiller-manuel:
this will allow typesafe access to the matches so something like this can be done:

```tsx
  beforeLoad: (c) => {
    const leafMatch = c.matches[c.matches.length - 1];
    if (leafMatch.fullPath === '/dashboard/invoices/$invoiceId') {
      if (leafMatch.params.invoiceId === '1') {
         // do something
      }
    }
```

### DIFF
--- a/packages/react-router/src/Matches.tsx
+++ b/packages/react-router/src/Matches.tsx
@@ -352,8 +352,8 @@ export function MatchRoute<
   return params ? props.children : null
 }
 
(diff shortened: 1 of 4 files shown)


lithdew:
Should the `matches` type be updated as well for i.e. the `meta` fn?

Issue #2255, opened 2024-09-03
Root route component is rendered for non-matching-route path but not for missing resource


lo1tuma:
### Which project does this relate to?

Router

### Describe the bug

The component defined in the `rootRoute` is rendered when a non-matching path riggers the rendering of the `notFoundComponent` but it is not rendered when a child route trows a `notFound()` value in its loader. 

### Your Example Website or App

https://stackblitz.com/edit/github-yk78pg-lfpyxu?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

* define a root route with `component` and `notFoundComponent`
* navigate to a non-matching path
* -> the `component` as well as the `notFoundComponent` gets rendered

### Expected behavior

I would expect the behavior to be the same in both scenarios.

### Platform

- Version: 1.52.5

SeanCassiere:
Please attach a valid minimal reproduction showing off the issue in question.

lo1tuma:
I’ve used the `quickstart-code-based` as a template. Here is the adjusted version that demonstrates the issue https://stackblitz.com/edit/github-bfat87?file=src%2Fmain.tsx

Steps to reproduce:
* In the browser when you are on the home page
* click the "About" link
* -> you see the not found page **without menu**
* click on "back to home"
* click on "Non-Existing Page" link in the menu
* -> you will see the not-found page but this time **with menu**

SeanCassiere:
Throwing a `notFound` error in a route's loader, outputs the following route-tree.

```
<root.NotFoundComponent />
```

Navigating to a route that **DOES NOT EXISTS** in the route-tree, results in this route-tree.

```
<root.Component>
  <root.NotFoundComponent />
</root.Component>
```

Note that this configuration does not have a `defaultNotFoundComponent`.

By default the Router has the setting of `notFoundMode: 'fuzzy'`, where the router traverses up the tree until it finds the nearest `notFoundComponent` (which in this case would be the one configured on the root route).

lo1tuma:
Thanks for the further investigation and writing a test-case @SeanCassiere. I haven’t thought about the `non-async` case.

My expected behavior would be that the behavior is consistent across all scenarios, so either the `NotFoundComponent` should be always wrapped or never.

pupuking723:
I can check whether this still reproduces on current main. If it does, I’ll start with a focused regression around the two root notFound paths and keep the fix scoped to making the layout behavior consistent, rather than changing fuzzy matching more broadly.

[...]

PR #2262, opened 2024-09-04
test(react-router): reproducer starter for #2255

SeanCassiere:
> [!IMPORTANT]
> These reproductions need to be corrected to match the expected route tree and the content to be displayed on the screen.

Starting for the reproductions for #2255

### DIFF
--- a/packages/react-router/tests/notFoundComponent.test.tsx
+++ b/packages/react-router/tests/notFoundComponent.test.tsx
@@ -20,17 +20,19 @@ afterEach(() => {
 
 const getStartComponent = (href: string) => {
   const startTextLinkName = 'Link to test'
-  function RouteComponent() {
+  function StartComponent() {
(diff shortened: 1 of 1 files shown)


SeanCassiere:
Currently, we've identified 3 scenarios that need resolution.

1. `/about` route with an `async` loader where this is thrown `notFound()`
```
<root.NotFoundComponent />
```

2. `/about` route with a non-async loader where this is thrown `notFound()`
```
<root.Component>
  <@tanstack/react-router->DefaultGlobalNotFound />
</root.Component>
```

3. `/route-that-does-not-exist`
```
<root.Component>
  <root.NotFoundComponent />
</root.Component>
```

`2` definitely shouldn't be happening since it should be traversing up the route-tree to select the `notFoundComponent`.

SeanCassiere:
Closing this out since it was never truly complete to begin with.

PR #2277, opened 2024-09-06
Example: Add example with Supabase Auth

ridwansameer:
### What does it do

Adds an example of using TanStack start with Suapabase auth.

### How to run?

Simply add your Supabase Keys to your environment variable and it should work out of the box.

If there's any feedback, let me know. First time hacking something together using Supabase, and first time using TanStack router/start as well!


examples/react/start-supabase-basic/app/utils/supabase.ts
```diff
@@ -0,0 +1,25 @@
+import { parseCookies, setCookie } from 'vinxi/http'
+import { createServerClient } from '@supabase/ssr'
+
+export function useSupabase() {
```
SeanCassiere, line 4: Please rename this util so it doesn't clash with the naming of React hooks. Maybe `getSupabaseServer`?

SeanCassiere (changes requested):

examples/react/start-supabase-basic/app/utils/supabase.ts
ridwansameer: Oops! Was trying to follow the previous example but that was indeed returning a hook. Changed!

SeanCassiere (approved):
Tested the example locally with my own Supabase credentials and the example worked. It could use some better error-handling, but that's more of a user-side implementation detail.

Pushing this through.

ridwansameer:
Thanks! I'll try and take a look and fix up the error-handling later. I think I need to remove the "Sign up instead?" link since supabase doesn't directly specify it's not an existing  account.

PR #2278, opened 2024-09-06
fix: make router-generator a peer dep of react-router

schiller-manuel:
type tests of generated code must be implemented in a separate package to break up this circular dependency


chorobin (approved):

levrik:
Could `@tanstack/router-generator` be made an optional peer dep? I just wanted to upgrade to latest TanStack Router but stopped after I saw that it pulls in tons of ESBuild dependencies through `tsx` which is a direct dep of `@tanstack/router-generator`. I do not use the file-based routing so I would like to be able not pulling in big dependencies like ESBuild.

PR #2280, opened 2024-09-07
examples(stark-clerk-basic): fix duplicated dependencies

irvile:


SeanCassiere (changes requested):
Following the installation guide on the website, please remove the dep from the `devDependencies`. Also, please run a `pnpm install` after removing the dep from the dev-deps just to make sure the root lockfile is kept up to date.

PR #2292, opened 2024-09-08
Replacing <StartClient /> with <RouterProvider /> fixes HMR

artimath:
Maybe this is obvious, but I wanted to document it just in case this is helpful with isolating the issue.


SeanCassiere:
Closing this, since we've solved pretty much all the HMR issues we were running into earlier in #2316.

@artimath thank you for looking into this issue, it definitely spurred the exploration we did into looking to solve this HMR issue for Start.

PR #2299, opened 2024-09-09
chore: support TS 5.6 and drop support for 5.0

chorobin:


packages/react-router/src/router.ts
```diff
@@ -1816,12 +1816,16 @@ export class Router<
...
     delete this.shouldViewTransition
     // Attempt to start a view transition (or just apply the changes if we can't)
-    ;(shouldViewTransition && typeof document !== 'undefined'
-      ? document
-      : undefined
-    )
-      // @ts-expect-error
-      ?.startViewTransition?.(fn) || fn()
+    if (
+      shouldViewTransition &&
```
chorobin, line 1820: Unfortunately had to rework the code a bit here to pass both 5.6 and the rest :(

SeanCassiere (commented):
The docs need to be updated as well.

SeanCassiere (approved):

PR #2300, opened 2024-09-09
feat(create-router): added project generator to scaffold router projects

schiller-manuel:


SeanCassiere (commented):
My suggestion would be to not include the other bundlers as options and only scaffold out a **Vite** project. Since Start is a migration path we want to support and encourage down the road, just leave it with Vite so it isn't a crazy jump to migrate over.

PR #2321, opened 2024-09-12
docs(react): fix typo in search params docs

3fuyang:
Replaced `default()` with `catch()`, which aligns with the "...use of `catch` here overrides the types..." from [L228](https://github.com/TanStack/router/blob/main/docs/framework/react/guide/search-params.md?plain=1#L228).


schiller-manuel:
the docs are correct as is.
if you need a `catch()`-like behavior, you should use `fallback` as described

PR #2334, opened 2024-09-14
fix(router-plugin): do not split if known ident is exported

SeanCassiere:
From https://discord.com/channels/719702312431386674/1283438765817331712

When using `autoCodeSplitting`, we need to take into account the scenario where the user is exporting their `component` or `loader` from the route file.  Consider this route before code-splitting:

```tsx
import { createFileRoute, Outlet } from '@tanstack/react-router'
import {
  importedComponent as ImportedComponent,
  importedLoader,
} from '../shared/imported'

export const loaderFn = () => {
  return importedLoader()
}

const Layout = () => {
  return (
    <main>
      <header style={{ height: HEADER_HEIGHT }}>
        <nav>
          <ul>
            <li>
              <a href="/">Home</a>
            </li>
          </ul>
        </nav>
      </header>
      <ImportedComponent />
      <Outlet />
    </main>
  )
}

export const Route = createFileRoute('/_layout')({
  component: Layout,
  loader: loaderFn,
})

const HEADER_HEIGHT = '63px'
export const SIDEBAR_WIDTH = '150px'
export const SIDEBAR_MINI_WIDTH = '80px'
const ASIDE_WIDTH = '250px'

export default Layout
```

An assumption can be made here, that if the user is exporting the `component` or `loader` from this file, then they expect to import it into another file.

As such, the current strategy of nuking the `component` or `loader`, once it has been detected for use in the `createFileRoute` function will not work, as a user wouldn't then be able to make the import from another file since it would no longer exist.

Therefore, if the `component` or `loader` is being exported, then code removal shouldn't happen and it should retain the item in the same *compiled* file. As such, since the item remains in the same file, we can skip performing an import.

A warning is also going to be added into the `console`, so the user knows that this function is not being code-split.


packages/router-plugin/src/core/code-splitter/compilers.ts
```diff
@@ -449,6 +493,28 @@ export function compileCodeSplitVirtualRoute(opts: ParseAstOptions) {
...
+  // to the file to let the user know that the exported identifiers
+  // will not in the split file but in the original file, therefore
+  // increasing the bundle size
+  if (knownExportedIdents.size > 0) {
+    const list = Array.from(knownExportedIdents).reduce((str, ident) => {
+      str += `\n- ${ident}`
+      return str
+    }, '')
+
+    const warningMessage = `These exports from "${opts.filename.replace('?' + splitPrefix, '')}" are not being code-split and will increase your bundle size: ${list}\nThese should either have their export statements removed or be imported from another file that is not a route.`
```
schiller-manuel, line 506: does this cause a problem with Start?
will Start properly work with some non-code-split routes?

it Start will not work, then we should throw a fatal error in case we are building for Start

SeanCassiere: This is only capturing if the `component` or `loader` is being exported out of the route file. It'll detect it both for  Start and Router since this happens during dev and at build time (not at runtime).

This is the example that are checking for:
```ts
// this export ident here
export const loadUser = async () => {
	return { user: 'john' }
}

export const Route = createFileRoute('/login')({
	loader: loadUser
})
```

In this case, we can't code-split the `loadUser` fn, since other external files may be imported from this source file.

So it needs to remain as-is.

schiller-manuel: understood, but is skipping a route from code splitting posing any problems in TanStack Start? i.e. could we just disable automatic code splitting for start and it would still work?

SeanCassiere: > could we just disable automatic code splitting for start and it would still work?

We *could* choose to turn it off automatic code-splitting on our side, but currently, we force it on in the Start config. https://github.com/TanStack/router/blob/8fde2ef13d887f6657a0534c5f2e9e6216329121/packages/start/src/config/index.ts#L292

PR #2359, opened 2024-09-18
feat(react-router): improve performance of Link

honzahruby:
Addresses https://github.com/TanStack/router/discussions/2011 to improve performance of `Link` and `router.buildLocation`.

`matchRoutesInternal` traverses all routes in `flatRoutes`, which degrades the performance of Link with the number of routes registered. This PR adds an optimization to look up the route in `routesByPath` in constant time for absolute paths in `to`.

Rendering times of 10000 Links as described in the original discussion thread https://github.com/TanStack/router/discussions/2011#discussioncomment-10668917
| | 2 routes | 400+ routes | 400+ routes optimized
|-|-|-|-|
| hardcoded href | 150ms | 160ms | 150ms |
| interpolatePath | 180ms | 190ms | 180ms |
| buildLocation | 360ms | 1760ms ⚠️  | 320ms |
| Link component | 660ms | 2160ms ⚠️  | 590ms |

### DIFF
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -956,7 +956,8 @@ export class Router<
   ```ts
   matchRoutes (
     next: ParsedLocation,
-    opts?: { preload?: boolean; throwOnError?: boolean },
+    opts?: MatchRoutesOpts,
+    routeFullPath?: string,
   ): Array<AnyRouteMatch>;
   ```
 */
@@ -964,18 +965,19 @@ export class Router<
     pathname: string,
     locationSearch: AnySearchSchema,
     opts?: MatchRoutesOpts,
+    routeFullPath?: string,
   ): Array<AnyRouteMatch>
   public matchRoutes(
     next: ParsedLocation,
     opts?: MatchRoutesOpts,
+    routeFullPath?: string,
   ): Array<AnyRouteMatch>
 
   public matchRoutes(
     pathnameOrNext: string | ParsedLocation,
-    locationSearchOrOpts?:
-      | AnySearchSchema
-      | { preload?: boolean; throwOnError?: boolean },
-    opts?: { preload?: boolean; throwOnError?: boolean },
+    locationSearchOrOpts?: AnySearchSchema | MatchRoutesOpts,
+    optsOrRouteFullPath?: MatchRoutesOpts | string,
+    routeFullPath?: string,
   ) {
     if (typeof pathnameOrNext === 'string') {
       return this.matchRoutesInternal(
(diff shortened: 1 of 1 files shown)


schiller-manuel:
how did you measure the time?
would make sense to add this benchmark to the repo

packages/react-router/src/router.ts
```diff
@@ -956,65 +956,94 @@ export class Router<
...
     next: ParsedLocation,
     opts?: { preload?: boolean; throwOnError?: boolean },
+    routeFullPath?: string,
   ): Array<AnyRouteMatch> {
     let routeParams: Record<string, string> = {}
 
-    const foundRoute = this.flatRoutes.find((route) => {
+    let foundRoute: AnyRoute | undefined = routeFullPath
+      ? this.routesByPath[routeFullPath]
+      : undefined
```
schiller-manuel, line 1009: does this mean if `to` is not specified, the link will still be slow?

in this case, can we use the current location of router instead?

honzahruby: Is it a valid use case to use a Link without `to`? If yes, we could provide the current match's `fullPath` as a fallback. 

Currently, it only speeds up the resolve process for links with absolute paths, because they can be easily looked up in the `routeById` map. Is there any util to quickly resolve relative paths?

honzahruby:
> how did you measure the time? would make sense to add this benchmark to the repo

I used [this code](https://gist.github.com/honzahruby/8e189b6f9743ba78962577a8bf346c6a) in the _large-file-based_ example and measured the time to render the Home component in react dev tools. It renders 10000 links with generated IDs used as a route param.

Is there any performance benchmark setup in this repo? How about https://www.npmjs.com/package/reassure?

schiller-manuel:
a performance setup as you proposed would be very good. can you set something up in a separate PR so we can compare different solutions to this perf issue?

we should also look into whether we can cache those expensive things

honzahruby:
Sure I can try.

As for this problem, caching won’t help when first generating the links, you have to traverse all the routes for each link. I was playing with caching various steps for a good week before I came up with this PR 🙂

packages/react-router/src/router.ts
schiller-manuel: > Is it a valid use case to use a Link without to?
yes, see e.g. https://tanstack.com/router/latest/docs/framework/react/guide/search-params#link-search-

in short: if you omit `to` but specify `from`, you will get typesafe links that point to the current location and can be used for e.g. updating search params.

schiller-manuel: > Is there any util to quickly resolve relative paths?

we have https://github.com/TanStack/router/blob/main/packages/react-router/src/path.ts#L92
whether that's fast should be benchmarked

honzahruby:
@schiller-manuel I set up a basic performance testing in #2398. Interestingly, the use case without `to` turned out to be the slowest one at the moment (11x slower than absolute path). With fixes applied, it's on par with the normal one. I'll update this PR when we move forward with the testing stuff.

schiller-manuel:
ok let's continue here

schiller-manuel:
I applied the idea in #2516
thanks for this!
if you have more perf ideas (or anything else), PRs are always welcome!

honzahruby:
@schiller-manuel Thank you so much for taking over, I had no time to get back to this 🫶

PR #2398, opened 2024-09-23
test(react-router): set up performance testing for Link

honzahruby:
Related to #2359, I set up a basic performance testing of the Link component and its simplified versions, using vitest `bench` functionality.

link perf

### DIFF
--- a/packages/react-router/tests/link.bench.tsx
+++ b/packages/react-router/tests/link.bench.tsx
@@ -50,17 +50,23 @@ type LinkProps = React.PropsWithChildren<{
 
 describe.each([
   {
-    name: 'small router (1 route with params)',
+    name: 'small router',
     numberOfRoutes: 1,
     matchedParamId: 0, // range from 0 to numberOfRoutes-1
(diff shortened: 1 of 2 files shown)


schiller-manuel:
great work. we should consider running the tests in CI at some point just so we make they do not break. but for now, we won't do this.

packages/react-router/tests/link.bench.tsx
```diff
@@ -0,0 +1,182 @@
...
+  (routesCount: number) => (children: React.ReactNode) => {
+    const rootRoute = createRootRoute()
+    const indexRoute = createRoute({
+      getParentRoute: () => rootRoute,
+      path: '/',
+      component: () => children,
+    })
+    const paramRoutes = Array.from({ length: routesCount }).map((_, i) =>
+      createRoute({
+        getParentRoute: () => indexRoute,
```
SeanCassiere, line 25: The parent route here is incorrect, it should be `rootRoute`.

Issue #2400, opened 2024-09-23
Wrong route id for preloaded routes after rebuild route tree


stevenlyd:
### Which project does this relate to?

Router

### Describe the bug

After a route (it needs to have a lazy route) `/_protected/test` got preloaded, then we call `router.buildRouteTree()` to rebuild the route tree again, that route will have a wrong id `/_protected/_protected/test`.

### Your Example Website or App

https://stackblitz.com/edit/vitejs-vite-s8qswn?file=src%2FApp.tsx

### Steps to Reproduce the Bug or Issue

1. Open my StackBlitz example.
2. It should automatically run the project. If not, just `npm run start`
3. Open browser devtool's console.
4. Click the first button to print `router.routesById`, in the console you can confirm all routes' ids are fine.
5. Click the second button, it preloads the `/_protected/test` route.
6. Click the third button to run `router.buildRouteTree()`
7. Click the fourth button to print `router.routesById` again, in the console you can see the preloaded route `/test` now has a wrong id: `/_protected/_protected/test`

### Expected behavior

No matter how many times I rebuild the route tree, the result should be consistent. 
I'm doing dynamic route injection, so I need to rebuild the route tree after there is a new route got injected.

### Screenshots or Videos

https://github.com/user-attachments/assets/d7db270f-4d76-4f0d-a1bc-5bc433e8c53b

### Platform

- OS: [Windows]
- Browser: [Chrome]
- Version: [1.58.3]

### Additional context

**How the issue happened**
1. This is currently how we generate id for a route instance in its `init()` method:
```typescript
    const customId = options?.id || path // Uses path as id's fallback

    // Strip the parentId prefix from the first level of children
    let id = isRoot
      ? rootRouteId
      : joinPaths([
          this.parentRoute.id === rootRouteId ? '' : this.parentRoute.id,
          customId,
        ])
``` 
Related code: https://github.com/TanStack/router/blob/main/packages/react-router/src/route.ts#L1032

2. During the first initialization of the router, everything is fine because it fall back to the path, and the path doesn't contain a prefix from parent.

3. When we preload the route, the lazy route's options will be assign to the primary route object, so that the primary route object now has an `id` in its `options` property. Related code: https://github.com/TanStack/router/blob/main/packages/react-router/src/router.ts#L2247

4. Now we call `router.buildRouteTree()` to manually rebuild it, now the `customId` will be `/_protected/test`, and we are joining it with parent's id again, so we get `/_protected/_protected/test`.

[...]

PR #2401, opened 2024-09-23
Add test cases for route id consistency

stevenlyd:
#2400


schiller-manuel:

stevenlyd:
> can you please add a test for this?

No problem, I'll do it later today. Actually I found this fix won't work for lazy pathless routes, since they don't have path, will always use `options.id` as fallback. I guess the key is to make sure the `option.id` we use when generating `customId` doesn't contain any parent prefix. I'm thinking about adding a util function to trim the parent prefix, what do you think?

stevenlyd:
@schiller-manuel 
Good afternoon! I've updated my fix and added test cases.
The root issue is we need to trim parent prefix when generating lazy routes (using either `createLazyRoute` or `createLazyFileRoute`). 
So that  it won't mess things up after the `id` in the lazy route object got assigned back to the primary route object,

packages/react-router/tests/router.test.tsx
```diff
@@ -580,3 +618,38 @@ describe('router matches URLs to route definitions', () => {
...
+    const rebuiltRouteIds = Object.keys(router.routesById)
+
+    console.log(originalRouteIds, rebuiltRouteIds)
+    originalRouteIds.forEach((id) => {
+      expect(rebuiltRouteIds).toContain(id)
+    })
+
+    // rebuiltRouteIds.forEach((id) => {
+    //   expect(originalRouteIds).toContain(id)
+    // })
```
schiller-manuel: remove?

packages/react-router/tests/router.test.tsx
```diff
@@ -580,3 +618,38 @@ describe('router matches URLs to route definitions', () => {
...
+      router.navigate({
+        to: '/fileBasedTest',
+      }),
+    )
+
+    router.buildRouteTree()
+
+    const rebuiltRouteIds = Object.keys(router.routesById)
+
+    console.log(originalRouteIds, rebuiltRouteIds)
```
schiller-manuel: remove

packages/react-router/tests/path.test.ts
```diff
@@ -276,3 +277,88 @@ describe('interpolatePath', () => {
...
+      path: '/',
+      expected: undefined,
+    },
+    {
+      name: 'should handle path with multiple trailing slashes',
+      path: '/multiple/trailing/slashes///',
+      expected: undefined,
+    },
+  ].forEach(({ name, path, expected }) => {
+    it(name, () => {
```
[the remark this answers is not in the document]
stevenlyd: done!

packages/react-router/src/fileRoute.ts
```diff
@@ -276,5 +280,8 @@ export function createLazyFileRoute<
   TRoute extends FileRoutesByPath[TFilePath]['preLoaderRoute'],
 >(id: TFilePath) {
   return (opts: LazyRouteOptions) =>
-    new LazyRoute<TRoute>({ id: removeGroups(id), ...opts })
+    new LazyRoute<TRoute>({
+      id: getLastPathSegment(removeGroups(id)) as any,
```
schiller-manuel, line 284: i am wondering about why the last path segment should be used as `id`.
this might work, but aren't there cases where this is not unique?

e.g. consider the following routes:
```
/foo/bar
/hello/bar
```

both would end up with the same `id`, right?

stevenlyd: Yes, you're right that both `/foo/bar` and `/hello/bar` have the same last path segment bar. However, this only affects the `options.id` of the lazy route object. From what I understand, the `options.id` in the lazy route object is mainly used during the preloading of a route. When a route is preloaded, these options are assigned back to the primary route object like this:
```typescript
// ./packages/react-router/src/router.ts
route._lazyPromise =
  route._lazyPromise ||
  (route.lazyFn
    ? route.lazyFn().then((lazyRoute) => {
        Object.assign(route.options, lazyRoute.options)
      })
    : Promise.resolve())
``` 

I believe that the `options.id` should represent the route's own ID, not the full ID that includes prefixes from parent routes. This seems to be consistent with how IDs and paths are assigned in `routeTree.gen.ts`. For example:

```typescript
const ProtectedRouteRoute = ProtectedRouteImport.update({
  id: '/_protected',
} as any)

const ProtectedLayoutRouteRoute = ProtectedLayoutRouteImport.update({
  id: '/_layout', // Full id should be /_protected/_layout
  getParentRoute: () => ProtectedRouteRoute,
} as any)

const ProtectedLayoutParentRouteRoute = ProtectedLayoutParentRouteImport.update(
  {
    path: '/parent',
    getParentRoute: () => ProtectedLayoutRouteRoute,
  } as any,
)

const ProtectedLayoutParentTestRouteRoute =
  ProtectedLayoutParentTestRouteImport.update({
    path: '/test', // Full path should be /parent/test
    getParentRoute: () => ProtectedLayoutParentRouteRoute,
  } as any).lazy(() =>
    import('./routes/_protected/_layout/parent/test/route.lazy').then(
      (d) => d.Route,
    ),
  )
``` 
I guess the reason why we assign the full ID at the beginning but update it to a trimmed version later is that the code generator needs to consume that full path or full ID information when generating the route tree.

From my understanding, `route.options.id` represents the route's own ID, whereas `route.id` is the full ID that exists only in the primary route object and is generated when `route.init` is called. This process combines the `customId` with the parent's full ID as a prefix. Additionally, the `customId` is derived from either the `path` (not `fullPath`) or `options.id`, depending on whether the route is pathless.
As a result, the route's ID ultimately become `/foo/bar` and `/hello/bar`, maintaining uniqueness even if the last segments are the same.

schiller-manuel: can you add a test for this as well please?

stevenlyd: Sure, you mean a test that checks if the lazy route object getting the trimmed version of id?

schiller-manuel: and that IDs are still unique then

stevenlyd: > and that IDs are still unique then

I've added tests for route ID uniqueness, and also added routes ending with same last segments to that test router.

stevenlyd: > and that IDs are still unique then

And also added tests that check if the lazy route's `options.id` is trimmed.

stevenlyd:
@schiller-manuel I've extracted test cases for route id consistency. All other changes have been reverted, since a simpler fix is already been implemented.

stevenlyd:
@schiller-manuel for some reason the automated unit test failed on a file I didn't touch: `loaders.test.tsx`. I ran unit test locally and it passed, weird. Do you have any idea why? https://cloud.nx.app/runs/lk9M3B5Ckg/task/%40tanstack%2Freact-router%3Atest%3Aunit

schiller-manuel:
thanks for your contribution!

test retriggered and passed

PR #2404, opened 2024-09-23
fix(start): rename `deployment` config to `server`

schiller-manuel:


SeanCassiere (approved):
Looks good, can good call on removing that static boolean since we can just get it from the types now.

PR #2406, opened 2024-09-24
fix(router-plugin): ignore initial files

honzahruby:
The chokidar watcher in `router-plugin` is supposed to be looking only for newly created files in webpack/rspack apps. By default, it considers all files as "new" when starting the dev server, causing the generator to run multiple times, and slowing down the actual startup of the app.

There's a question if we want to always run the generator at least once when starting the dev server 🤔 

Screenshot 2024-09-24 at 12 22 23


schiller-manuel:
> There's a question if we want to always run the generator at least once when starting the dev server 

definitely. imagine if a file had been created / moved before the dev server is started, so the code generator must re-run and the file must be updated.

PR #2409, opened 2024-09-24
fix(start): extend `defineConfig` server options with VinxiAppOptions.server

rizesql:
Ok so this PR follows some discussions from the Tanstack discord. For deploying Tanstack start to cloudflare, one must add some additional configs, which were not available in the config.

Current implementation copies `HTTPSOptions` from `@vinxi/listhen` because the type is not portable and the build fails. Ideally the implementation shouldn't do such things, so I'm open to suggestions.

With this change, the Tanstack start config is more aligned with the Solid start one:))

(P.S. this is my first OSS contribution, hello to everyone)

### DIFF
--- a/packages/start/src/config/index.ts
+++ b/packages/start/src/config/index.ts
@@ -114,24 +114,41 @@ function checkDeploymentPresetInput(preset: string): DeploymentPreset {
   return preset
 }
 
-const serverSchema = z.object({
-  routeRules: z.custom<NitroOptions['routeRules']>().optional(),
-  preset: z.custom<DeploymentPreset>().optional(),
-  static: z.boolean().optional(),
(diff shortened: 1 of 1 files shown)


schiller-manuel:
> because the type is not portable

what does this mean exactly ?

rizesql:
> > because the type is not portable
> 
> what does this mean exactly ?

Ok so if I just use `VinxiAppOptions['server']` instead of `ServerOptions`, it doesn't compile and I get the following error:

> error TS2742: The inferred type of 'inlineConfigSchema' cannot be named without a reference to '.pnpm/@vinxi+listhen@1.5.6/node_modules/@vinxi/listhen'. This is likely not portable. A type annotation is necessary.

PR #2427, opened 2024-09-27
refactor(start): supply `links()` with same context signature as `meta()`

### LINKED ISSUE #2426: Similar to how `route.options.meta()` already behaves it would be nice to pass the `loaderData` and `match` to the `links()` function as well.
Similar to how `route.options.meta()` already behaves it would be nice to pass the `loaderData` and `match` to the `links()` function as well.

For my specific use-case I want to generate a `canonical` link element that is rendered via the `Meta` component inside the head. To do this I need some data from the current `match` and `loaderData`. Those are available inside `meta()` but not inside `links()`.

_Originally posted by @lo1tuma in https://github.com/TanStack/router/discussions/2419_

SeanCassiere:
Closes #2426

### DIFF
--- a/examples/react/start-basic/app/routes/index.tsx
+++ b/examples/react/start-basic/app/routes/index.tsx
@@ -2,6 +2,16 @@ import { createFileRoute } from '@tanstack/react-router'
 
 export const Route = createFileRoute('/')({
   component: Home,
+  loader: () => Promise.resolve({ num: 20 }),
+  links: () => {
+    // console.log({ loaderData })
+    return [
+      {
+        rel: 'stylesheet',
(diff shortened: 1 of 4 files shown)


SeanCassiere (commented):
Two things open for discussion:
1. whether we want to replicate this behaviour for `scripts`.
2. whether we want to include `matches` and `match` in the function's context like we do for `meta`.

lo1tuma:
For `scripts()` I would appreciate to get the same information.

SeanCassiere:
Converted back into a draft, pending a decision on the API change suggested by @rehos.

SeanCassiere:
Closing this in favor of the new API suggested. Will push that through separately.

PR #2437, opened 2024-09-29
fix(start): use client base config value to get client vite manifest path

rehos:
Currently the build fails when setting the client base in app.config.ts (routers > client > base). This PR replaces the hardcoded `_build` with the value from the configuration.

Tested by changing the `app.config.ts` file of the start-basic example to:

```
import { defineConfig } from '@tanstack/start/config'
import tsConfigPaths from 'vite-tsconfig-paths'

export default defineConfig({
  vite: {
    plugins: () => [
      tsConfigPaths({
        projects: ['./tsconfig.json'],
      }),
    ],
  },
  routers: {
    client: {
      base: 'custom-base'
    }
  }
})
```


rehos:
I can't figure out why this change causes the tests to fail. When I run them locally in the failing packages the tests succeed.

PR #2442, opened 2024-09-29
fix(history): correctly handle history state in `createMemoryHistory`

matthiasschwarz:
Fixed the behaviour of the history state in createMemoryHistory to be the same than of the browser. The current implementation doesn't handle the state correctly when traversing the history.
Also added a state test.


SeanCassiere (approved):

SeanCassiere:
@matthiasschwarz thank you for this contribution!

PR #2443, opened 2024-09-30
fix(react-router): fix matchPathname with basepath

czizzy:
In micro-frontend applications, host app will inject `basepath` to micro app. When the host app is navigating to its own route, the mounted micro apps will monitor and trigger redirect as well without respecting the `basepath`.

### DIFF
--- a/packages/react-router/src/path.ts
+++ b/packages/react-router/src/path.ts
@@ -298,6 +298,10 @@ export function matchByPath(
   from: string,
   matchLocation: Pick<MatchLocation, 'to' | 'caseSensitive' | 'fuzzy'>,
 ): Record<string, string> | undefined {
+  // check basepath first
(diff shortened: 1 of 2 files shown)


czizzy:
@SeanCassiere could you please take a look at this? We're encountering a bug in our product.

braineo:
@SeanCassiere we actually also encountered this problem. we have app1 and app2 running in the same pager and both of them use tanstank router.

app1 does not have `basepath` while app2 has `/app2` as `basepath` and a redirect on root router

when i navigated to href `/` (which is app1's route), it triggers the app2's redirect on `/` (which it shouldn't and it should only care about the href staring with `/app2`)

SeanCassiere:
@czizzy I'm away till mid October, as such I've asked @schiller-manuel to review this PR.

PR #2463, opened 2024-10-03
feat: allow to exclude routes from SSR

schiller-manuel:


packages/react-router/src/lazyRouteComponent.tsx
```diff
@@ -77,10 +101,21 @@ export function lazyRouteComponent<
       throw error
     }
 
-    if (!comp) {
-      throw load()
+    if (typeof window === 'undefined' && ssr?.() === false) {
```
tannerlinsley, line 104: Should we use document here?

schiller-manuel: done

SeanCassiere (commented):
Since the usage for this looks to be setting `ssr` to `false` on the `Route.options`, should we have a `defaultSsr` flag on the `RouterOptions` as well?

So that'd allow both opting out certain routes from SSR as well as the choice of being able to opt-in for SSR on certain routes.

Or conversely, should this maybe be something named like `renderMode`? where maybe in the future `RSC` would be slotted in. - `"csr" | "ssr" | "rsc"`

schiller-manuel:
the case against a default value is that we don't have access to the router instance here. so I would not add this right now

PR #2464, opened 2024-10-03
chore: move Prettier to devDependencies

jlucaso1:
This reduce node_modules size on projects using tanstack_router


SeanCassiere (commented):
Prettier isn't a devDep in this instance since it is actively being used during runtime. It can't be a peerDep either, since we can't guarantee that prettier is the formatter in use.

PR #2467, opened 2024-10-04
docs(start): enable strictNullChecks

Sebring:
Must be enabled for app/router.tsx->createRouter()

_app/router.tsx line: 7_
```
Argument of type 
'{ routeTree: RootRoute<undefined, {}, AnyContext, AnyContext, {}, undefined, RootRouteChildren, FileRouteTypes>; }'
 is not assignable to parameter of type '"strictNullChecks must be enabled in tsconfig.json"'.ts(2345)
```


schiller-manuel:
good catch
we should advocate for `strict: true` though.

also, did you trip over the error message now? I added this error just yesterday :)

SeanCassiere (approved):

PR #2470, opened 2024-10-04
feat(router-valibot-adapter): upgrade valibot to v1

fabian-hiller:
I've upgraded the TanStack Router Valibot adapter to v1. We've changed the type signature, so this change is necessary. All other changes in our v1 beta version are only internal.

> I recommend marking the `types` property of the `SearchValidatorAdapter` interface as optional, as there is no need to actually include it at runtime.

---

An union of schema library authors (Zod, Valibot, ArkType, ...) collaborating on a standard interface for schema libraries called [Standard Schema](https://github.com/standard-schema/standard-schema). This simplifies implementation and prevents vendor lock-in. So far Valibot v1 supports this spec and Zod will follow soon. In the long run, the explicit implementation of a Valibot and Zod adapter may no longer be necessary.


SeanCassiere:
The unrelated start-basic-auth e2e test looks to be flaky here. Since it is not part of this change, I'm pushing this through.

@fabian-hiller thanks for the heads up on Standard Schema. It's definitely something worth keeping an eye.

SeanCassiere (approved):

PR #2480, opened 2024-10-05
fix(react-router): do not unnecessarily render notFoundComponent

### LINKED ISSUE #2478: Unexpected render of notFoundComponent when generic error is thrown

Router

### Describe the bug

I have both `notFoundComponent` and `errorComponent` defined on __root. When there is an error in loader or in child routes components, there is unexpected render of `notFound` component

https://stackblitz.com/edit/github-shvffc?file=src%2Froutes%2F__root.tsx

1. define a root route with `notFoundComponent` and `errorComponent`
2. throw error in `loader` or in `component` of root or any child component

### Expected behavior

only root's `errorComponent` will be rendered

### Screenshots or Videos

`notFoundComponent` is rendered

Image

### Platform

- OS: Windows
- Browser Chrome
- Version: 129.0.6668.90 (Official Build) (64-bit)

### Additional context

_No response_

lislon:
fixes #2478

### DIFF
--- a/packages/react-router/tests/link.test.tsx
+++ b/packages/react-router/tests/link.test.tsx
@@ -1107,7 +1107,9 @@ describe('Link', () => {
       getParentRoute: () => rootRoute,
       path: '/',
       component: () => {
-        throw new Error('Error from component should not render notFoundComponent')
+        throw new Error(
+          'Error from component should not render notFoundComponent',
+        )
       },
     })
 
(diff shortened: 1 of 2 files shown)


packages/react-router/tests/link.test.tsx
```diff
@@ -1095,6 +1095,33 @@ describe('Link', () => {
     expect(errorText).toBeInTheDocument()
   })
 
+  test('when navigating to the root with an error in component', async () => {
+    const notFoundComponent = vi.fn()
+
+    const rootRoute = createRootRoute({
+      errorComponent: () => <span>Oops! Something went wrong!</span>,
```
SeanCassiere, line 1102: Please use different error messages both in the `errorComponent` and the `component` so that anyone reading the test knows exactly what the expected outcome is.

Something like:
- `errorComponent` - "Rendering errorComp message"
- `component` - "Throwing from route component"

Feel free to change the exact verbiage as you please. Just please make them very distinctly unique.

SeanCassiere (changes requested):

PR #2484, opened 2024-10-06
fix:  mistake in search-params.md about `catch` with `fallback` usage

jdavidferreira:
I think this is a mistake. You are talking about the use of zod `catch` with `fallback`, but the example uses `default` instead of `catch`.


schiller-manuel:
no this is not a mistake. 
zod's `catch()` erases the type information, so we cannot infer it anymore. instead, we provide `fallback`.

PR #2486, opened 2024-10-06
feat: add `linkOptions` for re-usable props

chorobin:
`linkOptions` is a new function which is similar to TanStack Query's `queryOptions` in naming. Similar to `queryOptions` it does not change the input in anyway but it provides type safety of object literals intended to be passed to `Link`, `navigate` or `redirect`.

For example, consider the following from one of our examples

```tsx

function DashboardComponent() {
  return (
    <>
      <div className="flex items-center border-b">
        <h2 className="text-xl p-2">Dashboard</h2>
      </div>

      <div className="flex flex-wrap divide-x">
        {(
          [
            ['/dashboard', 'Summary', true],
            ['/dashboard/invoices', 'Invoices'],
            ['/dashboard/users', 'Users'],
          ] as const
        ).map(([to, label, exact]) => {
          return (
            <Link
              key={to}
              to={to}
              activeOptions={{ exact }}
              activeProps={{ className: `font-bold` }}
              className="p-2"
            >
              {label}
            </Link>
          )
        })}
      </div>
      <hr />

      <Outlet />
    </>
  )
}
```

The array here allows us to loop over and render a `Link` per item. But the items in the array are not type checked and only when they are passed to `Link` TypeScript complains if something is wrong.

Instead we can now use `linkOptions` to get type checking (and all the autocompletion goodness) of options as they are defined.

```tsx
const options = [
  linkOptions({
    to: '/dashboard',
    label: 'Summary',
    activeOptions: { exact: true },
  }),
  linkOptions({
    to: '/dashboard/invoices',
    label: 'Invoices',
  }),
  linkOptions({
    to: '/dashboard/users',
    label: 'Users',
  }),
]

function DashboardComponent() {
  return (
    <>
      <div className="flex items-center border-b">
        <h2 className="text-xl p-2">Dashboard</h2>
      </div>

      <div className="flex flex-wrap divide-x">
        {options.map((option) => {
          return (
            <Link
              key={option.to}
              {...option}
              activeProps={{ className: `font-bold` }}
              className="p-2"
            >
              {option.label}
            </Link>
          )
        })}
      </div>
      <hr />

      <Outlet />
    </>
  )
}

```

`linkOptions` works very similar to `satisfies` in TypeScript. It type checks the input of `linkOptions`, providing all the great type safety of `Link`. **But it returns the type of whatever you input as is** meaning any additional props are also inferred. As you can see with `label` here

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -105,6 +105,10 @@
               "to": "framework/react/guide/search-params"
             },
             {
+              "label": "Link Options",
+              "to": "framework/react/guide/link-options"
+            },
+            {
               "label": "Custom Links",
               "to": "framework/react/guide/custom-link"
             },
--- a/docs/framework/react/api/router.md
+++ b/docs/framework/react/api/router.md
@@ -17,6 +17,7 @@ title: Router API
   - [`isNotFound`](./router/isNotFoundFunction.md)
   - [`isRedirect`](./router/isRedirectFunction.md)
   - [`lazyRouteComponent`](./router/lazyRouteComponentFunction.md)
+  - [`linkOptions`](./router/linkOptions.md)
   - [`notFound`](./router/notFoundFunction.md)
   - [`redirect`](./router/redirectFunction.md)
 - Components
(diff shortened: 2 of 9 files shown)


SeanCassiere (commented):
love this!

Can we also document this on the "Reference" page, under "functions".

Issue #2495, opened 2024-10-09
Custom Links not getting correct href value in the DOM


andyswanepoel:
### Which project does this relate to?

Router

### Describe the bug

I'm using I followed the docs for creating `@chakra-ui/react` and was following the docs for creating a custom link.

The routing all works as expected, however when hovering over the links in the browser, I noticed the URL was that of the path for the route I was on.

I did also quickly test is with the exact set up with Material UI as shown in the docs with the same issue.

### Your Example Website or App

https://stackblitz.com/edit/github-ygojrn?file=CustomLink.tsx

### Steps to Reproduce the Bug or Issue

1. Go to https://stackblitz.com/edit/github-ygojrn?file=CustomLink.tsx
2. Visit the home page of the App.
3. Open the dev tool and look at the link to the About page
4. See the href is `/`

### Expected behavior

I expect that the URL in the browser would reflect the destination path/route.

### Screenshots or Videos

https://github.com/user-attachments/assets/e7957293-017d-4d70-8057-0461fd0ae9ef

### Platform

- OS: macOS
- Browser: Chrome
- Version: 129.0.6668.90

SeanCassiere:
Confirmed this is not a Chakra-specific issue. This is happening with MUI as well.
https://stackblitz.com/edit/github-ygojrn-jjltmm?file=src%2Froutes%2Findex.tsx

christofferbergj:
I have the same issue with a fairly simple custom link component. 
Navigation and prefetching on hover intent etc works as normal. 

```tsx
interface NavLinkProps extends Omit<RACLinkProps, 'className' | 'slot' | 'children' | 'style'> {
  children?: React.ReactNode
}

export const NavLink = createLink(
  React.forwardRef<React.ElementRef<typeof Link>, NavLinkProps>(
    ({ isDisabled, children, ...rest }, forwardedRef) => {
      const ref = useObjectRef(forwardedRef)

      const { isHovered, hoverProps } = useHover({ isDisabled, ...rest })
      const { isPressed, linkProps } = useLink({ isDisabled, ...rest }, ref)

      return (
        <Link
          {...mergeProps(hoverProps, linkProps, rest)}
          ref={ref}
          data-hovered={isHovered || undefined}
          data-pressed={isPressed || undefined}
          className={navItemStyles({ isDisabled, isHovered, isPressed })}
          activeOptions={{ includeSearch: false }}
        >
          {children}
        </Link>
      )
    },
  ),
)
```

[...]

PR #2506, opened 2024-10-10
docs: add to `createLink` documentation

christofferbergj:
Revise the text and examples in the `createLink` guide for better clarity and completeness. 
Added sections for React Aria Components, Chakra UI and MUI with code examples to illustrate the usage of third-party libraries.

Reason:
[Custom Links not getting correct href value in the DOM #2495
](https://github.com/TanStack/router/issues/2495)


docs/framework/react/guide/custom-link.md
```diff
@@ -2,91 +2,160 @@
...
+### Basic example
+
+If you want to create a basic custom link component, you can do so with the following:
 
 ```tsx
-import { Link, createLink } from '@tanstack/react-router'
+import { createLink, LinkComponent } from '@tanstack/react-router'
+
+interface BasicLinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
```
nstepien, line 19: If we allow all the `<a>` props via `React.AnchorHTMLAttributes<HTMLAnchorElement>`, then the `href` prop would be allowed:
```tsx
<CreatedLinkComponent href="/path" to="/path" />
```
This would be allowed, which is probably not what we want, no?

Tangentially, I've also noticed that the `disabled` prop is passed, which is not a valid `<a>` prop.

christofferbergj: You're right, but I'm unsure how deeply it makes sense to example such things. The user can easily omit the `href` from the props in this case but it always depends on the component. 

I'm uncertain how to handle the `disabled` prop with a custom component.

TkDodo (approved):

PR #2513, opened 2024-10-11
Chained Server Fn Syntax, ServerFn Middleware

tannerlinsley:


ssijak:
Let's goooo 🚀

packages/start-vite-plugin/src/compilers.ts
```diff
@@ -156,3 +164,450 @@ export function compileCreateServerFnOutput(opts: ParseAstOptions) {
...
+      }
+    },
+  })
+}
+
+function handleCreateMiddlewareCallExpression(
+  path: babel.NodePath<t.CallExpression>,
+  opts: ParseAstOptions,
+) {
+  // const firstArg = path.node.arguments[0]
```
SeanCassiere, line 348: cleanup

packages/start/src/config/index.ts
```diff
@@ -270,56 +292,79 @@ export function defineConfig(
...
-                // }),
-              ],
-              // link: {
-              //   client: 'client',
-              // },
-            }),
-          ]
-        : []),
-      withStartPlugins(tsrConfig)({
+      // ...(apiEntryExists
```
SeanCassiere, line 305: API routes need to be turned back on.

packages/start/src/config/index.ts
```diff
@@ -360,6 +435,16 @@ export function defineConfig(
               conditions: [],
             },
           }),
+          config('debug-server', {
+            plugins: [
+              {
+                name: 'custom-hot-update-client',
+                handleHotUpdate(ctx) {
+                  console.log('server hot', ctx)
```
SeanCassiere, line 443: remove

packages/start/src/server-handler/index.tsx
```diff
@@ -55,47 +51,38 @@ export async function handleServerRequest(request: Request, _event?: H3Event) {
...
-          return [
-            method.toLowerCase() === 'get'
-              ? (() => {
-                  // TODO: This is a weird use case. Can you actually GET form data?
-                  const { _serverFnId, _serverFnName, payload } = search
-                  return payload
-                })()
-              : await request.formData(),
-          ] as const
+          // We don't support GET requests with FormData payloads... that seems impossible
```
SeanCassiere, line 59: What is blocking on this?

convo on discord ←

packages/react-router/src/router.ts
```diff
@@ -2493,6 +2493,10 @@ export class Router<
                         // to be preloaded before we resolve the match
                         await route._componentsPromise
                       } catch (err) {
+                        updateMatch(matchId, (prev) => ({
```
SeanCassiere, line 2496: Is this fixing an existing bug?

e2e/start/website/app/routes/$project.$version.docs.framework.$framework.$.tsx
```diff
@@ -7,7 +7,10 @@ import { capitalize, seo } from '~/utils/seo'
 export const Route = createFileRoute(
   '/$project/$version/docs/framework/$framework/$',
 )({
-  loader: ({ params: { _splat } }) => getDocument(_splat!),
+  loader: ({ params: { _splat } }) =>
+    getDocument({
+      data: _splat!,
```
SeanCassiere, line 12: Is there a reason why the `_splat` param needs to be asserted here as NOT undefined? 

Unless, is the splat route also expected to be matched when a not is in the `"notFound"` status? This wouldn't make sense to me, since if we hit a `"notFound"` the router should grab the nearest `notFoundComponent` by default (or the root route's `notFoundComponent` based on the router's setting).

packages/react-router/src/router.ts
tannerlinsley: Yep

e2e/start/website/app/routes/$project.$version.docs.framework.$framework.$.tsx
tannerlinsley: I think we could probably ensure it's at least a string, even if it's empty

ttsirkia:
This PR included a new Middleware Guide page for Start. It is visible as a link in Server Functions page but not in the main contents menu. Should it also be visible there?

tannerlinsley:
Fixed.

thebuilder:
This PR silently changes the package name of adapters between two patch versions (`1.18.5` and `1.18.6`).

```diff
-import { zodSearchValidator } from '@tanstack/router-zod-adapter';
+import { zodValidator } from '@tanstack/zod-adapter';
```

After bumping dependencies, the types for `searchParams` becomes `any`, until the adapter package is changed to the new one.

Why isn't something like this marked as a breaking change, but just thrown into a huge PR with no description?
Everyone using the adapters will run this issue once they start bumping dependencies, and have to do the same detective work to find out why it breaks.

The damage is done, but could you at least:
- Mark the old packages as deprecated, to give a hint to the developer that they should replace them
- Update the release note with a mention of the packages having changed name

PR #2518, opened 2024-10-12
fix(react-router): populate result of `validateSearch` into URL when navigating

### LINKED ISSUE #1184: Default search params from validateSearch do not populate the URL
### Describe the bug

When using `validateSearch`, if you return a default value it becomes available via the `useSearch` hook in a component, but it does not appear in the URL. 

https://stackblitz.com/edit/github-cqxxd6?file=src%2Froutes%2Fabout.tsx

1. Click on the 'about' link

### Expected behavior

The default parameter from validateSearch is available via `useSearch` but does not show in the URL search params. 

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Chrome, Edge

### Additional context

_No response_

### LINKED ISSUE #2496: Default search params do not work with tanstack start

Start

### Describe the bug

When using tanstack start and default search params when we navigate to a page where all search params have default values, but do not specify the params on the navigation and then refresh the page we get a console error.

https://stackblitz.com/edit/tanstack-router-xmbwsm?file=app%2Froutes%2Fposts.tsx

1. Go to posts route
2. There are two links for each post details, when navigating to the non working one and refreshing the page we get the error

Sometimes we also get some context not defined errors on stackblitz so to reliably reproduce this, this needs to be run locally.

### Expected behavior

We should be able to refresh the page and default values for search. params should be used. 

### Screenshots or Videos

_No response_

### Platform

- OS: all
- Browser: Chrome

### Additional context

_No response_

schiller-manuel:
fixes #2496
fixes #1184

### DIFF
--- a/packages/react-router/src/RouterProvider.tsx
+++ b/packages/react-router/src/RouterProvider.tsx
@@ -48,6 +48,7 @@ export type BuildLocationFn = <
 >(
   opts: ToOptions<TRouter, TFrom, TTo, TMaskFrom, TMaskTo> & {
     leaveParams?: boolean
+    _includeValidateSearch?: boolean
   },
 ) => ParsedLocation
 
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -757,8 +757,9 @@ export function useLinkProps<
       })
 
       // All is well? Navigate!
-      router.commitLocation({
-        ...next,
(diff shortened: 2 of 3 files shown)


SeanCassiere (commented):
Some e2e tests checking the search values on a rendered `<Link>` would be good here.

PR #2529, opened 2024-10-13
Add file extensions to generated imports

10thfloor:
Adds file extensions to auto-generated imports so user's can use Tanstack's file-based router with **Deno** (v2+)
(and other tools that enforce strict ECMAScript module compatibility.)

To test: 
- Have Deno v2 installed
- Build and run the project as outlined in `CONTRIBUTING.md`
- Navigate to an example project
- Delete `tsconfig.json` file
- Run : `deno task dev`

The example project should start without errors.


schiller-manuel:
I think we already handle this via `addExtensions`, see https://github.com/TanStack/router/discussions/2528#discussioncomment-10926617

PR #2536, opened 2024-10-13
fix(start): do not try compiling start itself

schiller-manuel:


schiller-manuel:
waiting for server function rewrite, might make this unnecessary

PR #2546, opened 2024-10-15
Update path-aliases.md

justin13888:
Just small typo fixed


justin13888:
Must be me but it seems type definitions got changed after update

PR #2558, opened 2024-10-16
Update Convex codegen

thomasballinger:
Convex now uses prettierignore directives for generated files, update this so anyone running this locally isn't confused.


thomasballinger:
Huh, it looks like these files were formatted anyway. Is there another format process running to ignore here?

SeanCassiere:
> Huh, it looks like these files were formatted anyway. Is there another format process running to ignore here?

It's possible that your pnpm version or node_modules have just fallen behind.

PR #2565, opened 2024-10-17
fix(react-router): update link href upon search change to prevent stale links

schiller-manuel:


packages/react-router/src/link.tsx
```diff
@@ -595,12 +595,7 @@ export function useLinkProps<
     activeProps = () => ({ className: 'active' }),
     inactiveProps = () => ({}),
     activeOptions,
-    hash,
-    search,
-    params,
```
lislon, line 600: @schiller-manuel Should we keep this or remove it from `rest`? Otherwise the `<a>` get polluted in `when navigating to /posts with params` test:

```
    <a
      href="/posts/id1"
      params="[object Object]"
    >
      To first post
    </a>
```

schiller-manuel: yes we already noticed this. will push a fix later.

PR #2571, opened 2024-10-18
feat(start): move `scripts`,`links`, and `meta` to the `head`

### LINKED ISSUE #2426: Similar to how `route.options.meta()` already behaves it would be nice to pass the `loaderData` and `match` to the `links()` function as well.

SeanCassiere:
Closes #2426 

As discussed in #2426, this change adds a new method named `head` onto `Route`. This function takes in the same parameters as `meta` and returns the values used for `scripts`, `links`, and `meta`. Additionally, the `loaderData` argument is made to be possibly `undefined` as this execution may happen both before and after the `loader` has resolved.

The `meta`, `links`, and `scripts` methods on `Route` as being dropped and replaced by this new `head` method.


packages/start/src/client/serialization.tsx
```diff
@@ -111,20 +111,17 @@ export function afterHydrate({ router }: { router: AnyRouter }) {
       }
     }
 
-    const meta =
-      match.status === 'success'
```
schiller-manuel, line 115: is the status not relevant anymore?

SeanCassiere: No anymore, since that was only dependent for the loaderData.
Now the loaderData is always possibly undefined.

tannerlinsley (approved):

PR #2588, opened 2024-10-19
feat(react-router): expose 'key' for search param serialization

MarkusWendorf:
Hi, 

I would like to expose the JSON key for serialization. 

I have a use-case where I want to serialize a `Date`, but I have query params (`fromDate`, `toDate`) that should be serialized to `2024-10-19` and others with a different key that should also include minutes, seconds, etc.

So the type is the same for both cases, so I have to make the distinction based on the `key`.

This PR exposes the `key` in addition to the `value`.

### DIFF
--- a/docs/framework/react/guide/custom-search-param-serialization.md
+++ b/docs/framework/react/guide/custom-search-param-serialization.md
@@ -31,8 +31,8 @@ import {
 
 const router = createRouter({
   // ...
-  parseSearch: parseSearchWith(JSON.parse),
-  stringifySearch: stringifySearchWith(JSON.stringify),
+  parseSearch: parseSearchWith((value) => JSON.parse(value)),
+  stringifySearch: stringifySearchWith((value) => JSON.stringify(value)),
 })
 ```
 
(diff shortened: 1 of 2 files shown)


SeanCassiere (commented):
This would technically be a breaking change, the current usage of `parseSearchWith` and `stringifySearchWith` have already been long defined. As such, this code will already exist in people's codebases (`parseSearchWith(JSON.parse)`).

This poses a problem, the addition of this change would likely have people running into issues regarding the parameters of the `JSON.stringify/JSON.parse` functions.

@schiller-manuel thoughts on how you'd proceed here?

schiller-manuel:
@MarkusWendorf you can define this custom serialization logic all in userland, right?

SeanCassiere:
@schiller-manuel how about we make this its own standalone feature under the Router options?

Currently, we have a key under `RouterOptions` named search. Perhaps we move this in there and deprecate the existing `stringifySearch` and `parseSearch` options?

```tsx
const router = createRouter({
	// ...
	search: {
		parse: parseSearchWithV2(...),
		stringify: parseSearchWithV2(...)
	}
})
```

schiller-manuel:
while I think moving those props into `search` should be done, I wouldn't mix it up with this change.
let's just consider this as a potential breaking change for v2 as a solution in userland is perfectly possible right now.

PR #2602, opened 2024-10-20
feat(react-router): implement standard schema into router core

chorobin:
This PR implements [Standard Schema](https://github.com/standard-schema/standard-schema) in router core. Currently only valibot implements this but the intention is for all validation libraries to implement it. Which means we don't need to support adapters


chorobin:
- [ ] Update docs
- [ ] Type tests

PR #2604, opened 2024-10-20
fix(react-router): fix explicitUndefined links never matching

code-elf:
Small follow-up to #2592, as I noticed that it doesn't quite seem to work as I think would be expected - it does correctly _exclude_ the route when the parameter exists, however it doesn't _match_ the route when it doesn't. In other words, there's currently no way for `explicitUndefined` links to ever match a route. This PR fixes that.

### DIFF
--- a/packages/react-router/src/utils.ts
+++ b/packages/react-router/src/utils.ts
@@ -248,7 +248,7 @@ export function deepEqual(
       return false
     }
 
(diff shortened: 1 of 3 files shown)


code-elf:
@schiller-manuel Thanks so much for making that change - noticed a small issue when trying to use it in my project that this should fix. Lmk if I'm missing anything!

schiller-manuel:
whoops, thanks for fixing this 👍

PR #2606, opened 2024-10-21
examples: use a shared Convex backend for StackBlitz

thomasballinger:
Custom StackBlitz dev command that uses a shared backend. This makes https://tanstack.com/router/v1/docs/framework/react/examples/start-convex-trellaux usable.

Clear this shared table every 20 minutes.


thomasballinger:
Is there a nicer way to do this, e.g. a custom StackBlitz command maybe? I need to find where StackBlitz config is done.

I see https://developer.stackblitz.com/platform/webcontainers/project-config, maybe a package.json field is cleaner here.

thomasballinger:
Updated, no more wacky environment detection, just add a .stackblitzrc

thomasballinger:
The next thing to figure out is how to make this deploy with one click on Vercel. I think this should use the same approach: if you haven't actually closed this project, use the shared backend?

PR #2607, opened 2024-10-21
docs(start): update the getting-started guide's template

pixelmord:
## Adjusts tutorial code to 
- remove non-null assertion from client entry, throw if element cannot be found
- add namespace to fs import
- add type attribute to button

just some thoughts that my linter had ;)

### DIFF
--- a/docs/framework/react/start/getting-started.md
+++ b/docs/framework/react/start/getting-started.md
@@ -179,9 +179,9 @@ import { createRouter } from './router'
 
 const router = createRouter()
 
-const root = document.getElementById("root")
+const root = document.getElementById('root')
 if (!root) {
-	throw new Error("Root element not found")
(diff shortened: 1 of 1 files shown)


docs/framework/react/start/getting-started.md
```diff
@@ -241,7 +246,7 @@ Now that we have the basic templating setup, we can write our first route. This
 
 ```tsx
 // app/routes/index.tsx
-import * as fs from 'fs'
+import * as fs from 'node:fs'
```
schiller-manuel, line 249: that's the only thing I would take over, please remove the others

SeanCassiere: better to keep this in, since Bun and Deno want the node prefix

SeanCassiere (approved):

PR #2613, opened 2024-10-21
examples: no trailing comma in stackblitzrc

thomasballinger:
https://github.com/TanStack/router/pull/2606 didn't work, maybe the problem was the trailing comma? I'll try to test this first.


thomasballinger:
Thanks @schiller-manuel, https://tanstack.com/router/v1/docs/framework/react/examples/start-convex-trellaux works now!

PR #2625, opened 2024-10-23
fix(start): Send error messages from server functions

danecando:
Removes some unnecessary logs and passes error messages from server function errors. Passes error messages directly to errors thrown from the client instead of putting a json payload in the message. 

This should align better with whats shown in the docs. It seems unsafe to send along server stack traces though. 

> Aside from special redirect and notFound errors, server functions can throw any custom error. These errors will be serialized and sent to the client as a JSON response along with a 500 status code.

```tsx

import { createServerFn } from '@tanstack/start'

export const doStuff = createServerFn('GET', async () => {
  throw new Error('Something went wrong!')
})

// Usage
function Test() {
  try {
    await doStuff()
  } catch (error) {
    console.error(error)
    // {
    //   message: "Something went wrong!",
    //   stack: "Error: Something went wrong!\n    at doStuff (file:///path/to/file.ts:3:3)"
    // }
  }
}
```

Would like to support returning other error responses without having them thrown on the client side as well. Curious how other people feel about that.

### DIFF
--- a/packages/start/src/client-runtime/fetcher.tsx
+++ b/packages/start/src/client-runtime/fetcher.tsx
@@ -136,16 +136,9 @@ async function handleResponseErrors(response: Response) {
     const message = `Request failed with status ${response.status}`
 
     if (isJson) {
-      throw new Error(
-        JSON.stringify({
-          message,
-          body,
-        }),
-      )
+      throw new Error(body.message || message)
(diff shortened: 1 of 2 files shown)


schiller-manuel:
can you please add tests to the start e2e tests?

PR #2644, opened 2024-10-26
feat(react-router): add `preload='render'` support

SeanCassiere:
This is the last preloading strategy which was left to implement.

This option makes it so that a preload will be fired off as soon as the Link component is rendered into the DOM. Once fired, it won't ever fire again unless the component is destroyed.

### DIFF
--- a/docs/framework/react/api/router/RouterOptionsType.md
+++ b/docs/framework/react/api/router/RouterOptionsType.md
@@ -52,7 +52,7 @@ The `RouterOptions` type accepts an object with the following properties and met
 - If `false`, routes will not be preloaded by default in any way.
 - If `'intent'`, routes will be preloaded by default when the user hovers over a link or a `touchstart` event is detected on a `<Link>`.
 - If `'viewport'`, routes will be preloaded by default when they are within the viewport of the browser.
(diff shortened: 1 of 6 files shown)


docs/framework/react/api/router/RouterOptionsType.md
```diff
@@ -46,12 +46,13 @@ The `RouterOptions` type accepts an object with the following properties and met
...
 ### `defaultPreload` property
 
-- Type: `undefined | false | 'intent' | 'viewport'`
+- Type: `undefined | false | 'intent' | 'viewport' | 'render'`
 - Optional
 - Defaults to `false`
 - If `false`, routes will not be preloaded by default in any way.
 - If `'intent'`, routes will be preloaded by default when the user hovers over a link or a `touchstart` event is detected on a `<Link>`.
 - If `'viewport'`, routes will be preloaded by default when they are within the viewport of the browser.
+- If `'render'`, routes will be preloaded by default when as soon as they are rendered on in the DOM.
```
schiller-manuel, line 55:
```suggestion
- If `'render'`, routes will be preloaded by default as soon as they are rendered in the DOM.
```

schiller-manuel: other than that, LGTM!

PR #2649, opened 2024-10-26
fix(react-router): correctly set `pathChanged` in router events

schiller-manuel:
fixes #2638


SeanCassiere (approved):

schiller-manuel:
needs some tests

schiller-manuel:
@merthanmerter not aware of such use cases, if they arise, we can still add it later.

PR #2652, opened 2024-10-27
docs(start): resize Convex logo for tighter SVG

thomasballinger:


thomasballinger:
current
image

after
image

PR #2667, opened 2024-10-28
fix(router-valibot-adapter): support for v1 dist-tags

fabian-hiller:
Looks like the npm install broke with the upgrade to Valibot v1. This PR should fix that.


schiller-manuel:
we don't really need the adapter anymore, but let's merge this for those who still use it

fabian-hiller:
Sounds good! But keep in mind that the Standard Schema spec is in beta and some things may change in the next few weeks before we release our RC.

PR #2677, opened 2024-10-30
feat(react-router): allow unencoded characters in dynamic path segments

alma-lp:
Currently any path params with a "@" character are URI encoded. Modern browsers handle this fine and it is a common pattern (e.g. youtube.com/@exampleChannel) to have them unencoded in a path.

### DIFF
--- a/docs/framework/react/api/router/RouterOptionsType.md
+++ b/docs/framework/react/api/router/RouterOptionsType.md
@@ -286,3 +286,9 @@ const router = createRouter({
 - Optional
 - Defaults to `never`
 - Configures how trailing slashes are treated. `'always'` will add a trailing slash if not present, `'never'` will remove the trailing slash if present and `'preserve'` will not modify the trailing slash.
+
+### `pathParamsAllowedCharacters` property
+
+- Type: `Array<';' | ':' | '@' | '&' | '=' | '+' | '$' | ','>`
+- Optional
+- Configures which URI characters are allowed in path params that would ordinarily be escaped by encodeURIComponent.
--- a/packages/react-router/src/path.ts
+++ b/packages/react-router/src/path.ts
@@ -204,6 +204,8 @@ interface InterpolatePathOptions {
   params: Record<string, unknown>
   leaveWildcards?: boolean
   leaveParams?: boolean
+  // Map of encoded chars to decoded chars (e.g. '%40' -> '@') that should remain decoded in path params
+  decodeCharMap?: Map<string, string>
 }
 
 export function interpolatePath({
@@ -211,6 +213,7 @@ export function interpolatePath({
   params,
   leaveWildcards,
   leaveParams,
+  decodeCharMap,
 }: InterpolatePathOptions) {
   const interpolatedPathSegments = parsePathname(path)
   const encodedParams: any = {}
@@ -222,7 +225,9 @@ export function interpolatePath({
       // the splat/catch-all routes shouldn't have the '/' encoded out
       encodedParams[key] = isValueString ? encodeURI(value) : value
     } else {
-      encodedParams[key] = isValueString ? encodeURIComponent(value) : value
+      encodedParams[key] = isValueString
+        ? encodePathParam(value, decodeCharMap)
+        : value
     }
   }
 
@@ -245,6 +250,16 @@ export function interpolatePath({
       return segment.value
     }),
   )
+}
+
+function encodePathParam(value: string, decodeCharMap?: Map<string, string>) {
+  let encoded = encodeURIComponent(value)
+  if (decodeCharMap) {
+    for (const [encodedChar, char] of decodeCharMap) {
+      encoded = encoded.replaceAll(encodedChar, char)
+    }
+  }
+  return encoded
 }
 
 export function matchPathname(
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -464,6 +464,15 @@ export interface RouterOptions<
      */
     strict?: boolean
   }
+  /**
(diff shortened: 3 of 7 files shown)


packages/react-router/src/path.ts
```diff
@@ -222,7 +222,10 @@ export function interpolatePath({
       // the splat/catch-all routes shouldn't have the '/' encoded out
       encodedParams[key] = isValueString ? encodeURI(value) : value
     } else {
-      encodedParams[key] = isValueString ? encodeURIComponent(value) : value
+      encodedParams[key] = isValueString
+        ? // allow '@' in path segments
+          encodeURIComponent(value).replaceAll('%40', '@')
```
SeanCassiere, line 227: This wouldn't really scale well should additional characters need to be escaped.

Perhaps this `replaceAll` call should be moved into a function, where it can be referenced against an object by its keys and replaced with its values.

Maybe something like this?

```js
function unescapeParamValues(str) {
	const obj = { '%40': '@' }
	const enc = encodeURIComponent(str)
	Object.entries(obj).forEach(([key, val]) => {
		enc.replaceAll(key, val)
	})
	return enc
}
```

SeanCassiere (changes requested):

packages/react-router/src/path.ts
alma-lp: Thanks @SeanCassiere for the feedback. See latest for changes.

SeanCassiere (approved):

schiller-manuel:
should we maybe expose this as a config option to the user? some might want different characters, some might want no characters etc?

SeanCassiere:
> should we maybe expose this as a config option to the user? some might want different characters, some might want no characters etc?

That's a good idea!

@alma-lp would this be something you'd be able to implement?

alma-lp:
> > should we maybe expose this as a config option to the user? some might want different characters, some might want no characters etc?
> 
> That's a good idea!
> 
> @alma-lp would this be something you'd be able to implement?

I should. I'll take a look.

docs/framework/react/api/router/RouterOptionsType.md
```diff
@@ -286,3 +286,9 @@ const router = createRouter({
 - Optional
 - Defaults to `never`
 - Configures how trailing slashes are treated. `'always'` will add a trailing slash if not present, `'never'` will remove the trailing slash if present and `'preserve'` will not modify the trailing slash.
+
+### `pathParamsAllowedCharacters` property
+
+- Type: `Array<';' | ':' | '@' | '&' | '=' | '+' | '$' | ','>`
```
schiller-manuel, line 292: where does this list come from?

packages/react-router/src/router.ts
```diff
@@ -768,6 +778,15 @@ export class Router<
...
     this.isServer = this.options.isServer ?? typeof document === 'undefined'
 
+    this.pathParamsDecodeCharMap = this.options.pathParamsAllowedCharacters
+      ? new Map(
+          this.options.pathParamsAllowedCharacters.map((char) => [
+            encodeURIComponent(char),
+            char,
+          ]),
+        )
+      : new Map()
```
schiller-manuel, line 788: leave `pathParamsDecodeCharMap`  `undefined` in case `pathParamsAllowedCharacters` is not configured

schiller-manuel:
please add a test that configures router with some allowed chars, navigate to a route by setting a path param to contain one of the allowed chars and then check if the url is not encoded

docs/framework/react/api/router/RouterOptionsType.md
alma-lp: > where does this list come from?

These are characters encoded by `encodeURIComponent` but not by `encodeURI` minus `/`, `?`, and `#` since those would be problematic as a path param.

CanRau:
Interesting. this is a completely different approach, but seemingly much simpler, than adding prefix/suffix support, tho if I understand it correctly would actually be sufficient for my usecase as we use /@username routes as well.

Just for reference, I added some thoughts here https://github.com/TanStack/router/issues/707#issuecomment-2433814404

schiller-manuel:
@CanRau  this does not allow prefixes. this just prevents path params containing e.g. "@" to be urlencoded. the path param will still be the full string "@foo"

CanRau:
Yea just realized that it wouldn't help with route matching 😔 which we need as we have those /@username paths at the root next to other dynamic and non-dynamic routes 😬

SeanCassiere (approved):

SeanCassiere:
> Yea just realized that it wouldn't help with route matching 😔 which we need as we have those /@username paths at the root next to other dynamic and non-dynamic routes 😬

To confirm, are we merging this 'as-is' and working on the other part as a separate issue? or is there more work happening on this PR?

jmkd3v:
It looks like this fix is incomplete. Links to routes such as `<Link to="/$handle" params={{ handle: "@hello" }} />` still reference `/%40handle`. Including the url-encoded character in the `to` prop itself, like `<Link to="/@hello" />`, directs to the correct page, but my understanding is that this is not the correct approach.

jmkd3v:
It looks like this fix is incomplete. Links to routes such as `<Link to="/$handle" params={{ handle: "@hello" }} />` still direct to `/%40handle`. Something like `<Link to="/@hello" />` where the url-encoded character is in the `to` prop itself directs to the correct page, but my understanding is that this is not the correct approach.

CanRau:
@jmkd3v have you configured allowed characters on your router?

```ts
const router = createRouter({
  ...
  pathParamsAllowedCharacters: ['@']
})
```

jmkd3v:
oh, I didn't know I had to configure it! that fixed it, thank you so much @CanRau!

PR #2679, opened 2024-10-31
feat(router-generator): remove prettier ignore comment blocks

### LINKED ISSUE #2673: Generated prettier-ignore comments do nothing

Router

### Describe the bug

By default, the router generator will add `prettier-ignore-start` and `prettier-ignore-end` comments to the generated file (functionality added in #1157):

https://github.com/TanStack/router/blob/2c4ff9162cca457fa00ee326367fbe258f212a95/packages/router-generator/src/config.ts#L24
https://github.com/TanStack/router/blob/2c4ff9162cca457fa00ee326367fbe258f212a95/packages/router-generator/src/config.ts#L32

However, these comments are not actually supported by prettier - the `start`/`end` range comments are only supported in Markdown right now - prettier/prettier#5287

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBbAhjATgSwB4AUAOlAASkCMANKQAw30nn2UNVN1uUkCUJJAegBUpAA5Y4MGDjhYAtDgDmUCBLkBnGOiwxSQgSQzZ8xMq06cOLapbMtrvQSPGTpshctVw5CACZ6BECoQCFFpaHVkUG0sCAB3AAVtBEiUdAAbOPQAT0jggCMsdDAAa0kAZXRUOAAZHCg4ZAAzDPU4AqLSitFi+sVkbABXdpA21BwBrGHguDxRWRxq2AyAFVkobRlUlvS24PU+9LgARUGIeGbWkYArdTxyw5Ozi6QdvZAAR2e4BNjRVJA6HUcgacF8YKCIGw6Bw6T6AGEIKgMMhAel0pCDlBFEcAIJSXD5QbwBKyOoNS67EYACxgqHSAHVqTh4OoemA4OUUiycAA3FnZVFgdR5EC84YASSg4Ng5TAuDCuOl5Rg2SOlPe4ggbQZRVEqJcbSwvMawXqRpgv3QihRryuwR6WCNqNV83U8pwYUh4nqMAZOF8MGpyAAHPQQBIvjgJFabegNSMtPl-YHg0gAEzBQZtFbofLbe0gOCofJg8G+GrobGDa1wABiqiM0mxqPQxIgIAAvp2gA

1. Take any generated TSR route tree file and run it through prettier
2. Notice prettier still formats the entire file

### Expected behavior

The presence of the non-functional prettier-ignore comments is confusing. They should just be removed from the default configuration since there is no way to get prettier to ignore a whole file through comment directives ([prettier/prettier#3634](https://github.com/prettier/prettier/issues/3634#issuecomment-364297344)).

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: N/A

### Additional context

_No response_

SeanCassiere:
Resolves #2673


docs/framework/react/guide/file-based-routing.md
```diff
@@ -329,6 +348,26 @@ export default defineConfig({
...
+If you are using VSCode, you can also add the following to your `.vscode/settings.json` file to exclude the generated route tree file from the editor's file watcher:
+
+```json
+{
+  "files.watcherExclude": {
+    "**/routeTree.gen.ts": true
+  }
+}
+```
+
```
ethanwu10, line 370: Will this cause problems with intellisense not picking up changes to the file made by the generator? In my quick and dirty testing it doesn't seem to be the case. But when running neovim, I need to make sure that my editor loads modifications to the generated route tree for the language server to be able to know about them.

Regardless, is there even a real benefit to ignoring it from the file watcher? I don't think VSCode would do anything itself to the file when notified of a change, and having language intelligence services be aware of changes as soon as possible is probably a good thing.

SeanCassiere: This won't have any bearing on the intellisense service provided by VSCode which bases the majority of its analysis based on what is returned by `tsc`. As such, this change would have no bearing on it.

Currently, when VSCode's background file watcher gets notified of an update to the active/inactive/closed generated route tree file (triggered by actions like adding or removing a route), it then proceeds to open the file in your active workbench in its _previous_ state, thereby interrupting the save triggered by the router-generator. This sometimes leads to the file being formatted incorrectly or having syntactical errors. This is a long-standing issue we've had and this change solves it.

Issue #2514, opened 2024-10-12
`loader` promise never resolves when error thrown in params.parse in child route


lislon:
### Which project does this relate to?

Router

### Describe the bug

When the error is inside `params.parse`, it doesn't reset parent's routes loader states and the UI is stuck in the loading state.

Configuration:
- `__root.tsx` with async loader + `pendingComponent` + `notFoundComponent` + `wrapInSuspense: true`
- `_reustarant.tsx` just a template for inner routes, no logic
- `_reustarant/$pizzaType.tsx` here I throw an error in `params.parse`, catch it in `onError` and re-throw it as `throw notFound()`

```tsx
// ..._root.tsx`
export const Route = createRootRoute({
  component: RootComponent,
  notFoundComponent: () => (
    <div>
      No pizza (this is expected)
    </div>
  ),
  pendingComponent: () => (
    <div>
      Loading...
    </div>
  ),
  loader: async () => {
    function sleep(ms: number) {
      return new Promise((resolve) => setTimeout(resolve, ms));
    }

    await sleep(3000);
    return { ok: true };
  },
  wrapInSuspense: true,
});
```

```tsx
// ..._reustarant/$pizzaType.tsx`
export const Route = createFileRoute('/_reustarant/$pizza')({
  component: Component,
  params: {
    parse: (p) => {
      if (p.pizza === 'rotten') {
        throw new Error('404 No rotten pizzas')
      }
      return {
        pizza: p.pizza,
      }
    },
    stringify: (p) => ({ pizza: p.pizza }),
  },
  onError: () => {
    throw notFound()
  },
})
```

### Your Example Website or App

https://stackblitz.com/edit/github-efx5cv-t3bmly?file=README.md

### Steps to Reproduce the Bug or Issue

1. Open example app
2. Click on "Click to go to /rotten"

Infinite `pendingComponent` will be shown 'Loading...'

### Expected behavior

I expect to see `notFoundComponent`

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/16cf3059-3b2f-4fe7-aec7-1c4138539a8e)

### Platform

- OS: Windows
- Browser: Chrome
- Version: Version 129.0.6668.101 (Official Build) (64-bit)

schiller-manuel:
could you provide a unit test as a PR?

schiller-manuel:
this has been fixed

PR #2682, opened 2024-10-31
fix(react-router): loader promise not resolving when error thrown in `params.parse`

vishwamartur:
Related to #2514

Add test to verify `notFoundComponent` is displayed when an error is thrown in `params.parse` in a child route.

* Add a new test case in `errorComponent.test.tsx` to check if the `notFoundComponent` is rendered when an error is thrown in `params.parse`.
* Create a root route with a `notFoundComponent`.
* Create an index route with a link to a child route.
* Create a child route with a `params.parse` function that throws an error for a specific condition and an `onError` function that re-throws the error as `notFound()`.
* Render the router and simulate a click on the link to the child route.
* Verify that the `notFoundComponent` is displayed.

### DIFF
--- a/packages/react-router/tests/errorComponent.test.tsx
+++ b/packages/react-router/tests/errorComponent.test.tsx
@@ -7,6 +7,7 @@ import {
   createRootRoute,
   createRoute,
   createRouter,
+  notFound,
 } from '../src'
 import type { ErrorComponentProps } from '../src'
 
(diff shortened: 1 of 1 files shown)


schiller-manuel:
is this testing the same thing as https://github.com/TanStack/router/pull/2636 ?

schiller-manuel:
closing in favor of https://github.com/TanStack/router/pull/7336

PR #2698, opened 2024-11-04
exploration for Apollo Client integration

phryneas:


packages/start/src/client/Meta.tsx
```diff
@@ -122,7 +122,17 @@ export const useMetaElements = () => {
...
+    {},
+    {
+      set(t, key, v) {
+        const res = Reflect.set(t, key, v)
+        window.__TSR__ROUTER__ &&
+          __TSR__ROUTER__.emit({ type: 'onStreamedValue', key })
+        return res
+      },
+    },
+  ),
```
phryneas, line 135: Streamed values arriving now emits a router event.

examples/react/start-basic-apollo-client/app/components/ApolloProvider.tsx
```diff
@@ -0,0 +1,98 @@
...
+  useMemo,
+  useRef,
+} from 'react'
+
+const APOLLO_EVENT_PREFIX = '@@apollo.event/'
+const APOLLO_HOOK_PREFIX = '@@apollo.hook/'
+
+const handledEvents = new WeakSet<object>()
+
+export const ApolloProvider = WrapApolloProvider((props) => {
```
phryneas, line 21: This file is essentially all that's needed for Apollo Client streaming SSR integration with fetch-as-you-render (`useSuspenseQuery` etc).

phryneas: This would end up in it's own package, I guess?

examples/react/start-basic-apollo-client/app/apollo/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  // it would be nice to do this in the long run
+  // @ts-ignore unavoidable due to the ternary in arguments
+  //   router.options.context.preloadQuery = router.isServer
+  //     ? createTransportedQueryPreloader(apolloClient)
+  //     : createQueryPreloader(apolloClient)
+
+  // @ts-ignore unavoidable due to the ternary in arguments
+  router.options.context.preloadQuery =
+    createTransportedQueryPreloader(apolloClient)
+  router.options.transformer = defaultTransformer.withTransformers([
```
phryneas, line 30: I'm gonna be honest, I would prefer to have a `router.options.transformers` option - that way we could merge user-supplied transformers with "library-supplied" ones, which is currently not possible.

packages/start/src/client/Meta.tsx
```diff
@@ -126,7 +126,17 @@ export const useMetaElements = () => {
           // children={`
           // __TSR__ = {
           //   matches: [],
-          //   streamedValues: {},
+          //   streamedValues: new Proxy(
```
schiller-manuel, line 129: is there as solution that does not need a proxy?

packages/start/src/client/Meta.tsx
```diff
@@ -195,7 +205,7 @@ export const useMetaElements = () => {
...
+          children={`__TSR__={matches:[],streamedValues:new Proxy({},{set(e,_,t){const r=Reflect.set(e,_,t);return window.__TSR__ROUTER__&&__TSR__ROUTER__.emit({type:"onStreamedValue",key:_}),r}}),queue:[],runQueue:()=>{let e=!1;__TSR__.queue=__TSR__.queue.filter((_=>!_()||(e=!0,!1))),e&&__TSR__.runQueue()},initMatch:e=>{__TSR__.queue.push((()=>(__TSR__.matches[e.index]||(__TSR__.matches[e.index]=e,Object.entries(e.extracted).forEach((([e,_])=>{if("stream"===_.type){let e;_.value=new ReadableStream({start(_){e=_}}),_.value.controller=e}else if("promise"===_.type){let e,t;_.value=new Promise(((_,r)=>{e=_,t=r})),_.resolve=e,_.reject=t}}))),!0))),__TSR__.runQueue()},resolvePromise:e=>{__TSR__.queue.push((()=>{const _=__TSR__.matches[e.matchIndex];if(_){const t=_.extracted[e.id];if(t)return t.resolve(e.value.data),!0}return!1})),__TSR__.runQueue()},cleanScripts:()=>{document.querySelectorAll(".tsr-once").forEach((e=>{e.remove()}))}};`}
```
schiller-manuel, line 208: I know this is outside of the scope of this PR but I'd really like to automate the minification during build somehow

packages/start/src/client/Meta.tsx
phryneas: Not while keeping this a "plain object", as afaik there is no other way of detecting a property is added. 

If the injected script tag would so something like `__TSR__.streamedValues.push([name, value])` on an array, or `__TSR__.streamedValues.set(name, value)` on a Map, those `.push` or `.set` properties could be overwritten to also call `__TSR__ROUTER__.emit` - but for a plain object property assignment the only way to do this is a Proxy.

For an "array-type" implementation, see https://github.com/apollographql/apollo-client-nextjs/blob/main/packages/client-react-streaming/src/ManualDataTransport/lateInitializingQueue.ts as an example - it is just created as an array and later `registerLateInitializingQueue` is called, consuming backpressure once and passing newly incoming data directly into the callback function.
Of course that's a very different approach and I didn't want to change too much, so I went with this for now.

packages/react-router/src/router.ts
```diff
@@ -3145,9 +3166,9 @@ ${jsesc(script, { quotes: 'backtick' })}\`)`
       'Key has already been streamed: ' + key,
     )
 
-    this.streamedKeys.add(key)
+    this.emit({ type: 'onStreamedValue', key })
     this.injectScript(
-      `__TSR__.streamedValues['${key}'] = { value: ${this.serializer?.(this.options.transformer.stringify(value))}}`,
+      `__TSR__.streamedValues.push([${this.serializer?.(key)}, ${this.serializer?.(this.options.transformer.stringify(value))}])`,
```
phryneas, line 3171: `key` was not serialized before, so keys containing `'` could actually cause JavaScript errors or code injection.

phryneas:
This is now deployed at  https://apolloclient-integration-tanstack-start-b3k74ebxl.vercel.app/useSuspenseQuery so you can poke it there if you want to.

phryneas:
Reworked as requested :)

phryneas:
Latest deploy: https://apolloclient-integration-tanstack-start-3he6bcptx.vercel.app/useSuspenseQuery

phryneas:
And reworked again so it works with the changed streaming :)

packages/start-server/src/tsrScript.ts
```diff
@@ -1,9 +1,17 @@
...
 import type { StartSsrGlobal } from '@tanstack/start-client'
 
+declare const __TSR_ROUTER__: AnyRouter | undefined
+
 const __TSR_SSR__: StartSsrGlobal = {
   matches: [],
   streamedValues: {},
+  setStreamedValue(key, value) {
+    __TSR_SSR__.streamedValues[key] = { value }
+    if (typeof __TSR_ROUTER__ !== 'undefined') {
```
schiller-manuel, line 11: there might a race condition where a value is streamed but router is not hydrated yet.
`streamedKeys` is initialized with all previously streamed keys, so that won't be affected. 

remind me again please, why don't we just call `__TSR_ROUTER__.clientSsr!.streamedKeys.add(key)` here? why do we need that event emitter pattern?

phryneas: Because the event is the thing I am really interested in. 

Our library is listening for these incoming events. 
It only uses `streamedKeys` on initialization to catch up with previously streamed values to consume back pressure, but it also needs everything happening in the future when it happens.

Of course we could emote the event *and* call .add here, but to me it seemed more intuitive to write to `streamedKeys` only in one place, not in two.

phryneas: https://github.com/apollographql/apollo-client-nextjs/blob/1306572238ff9d9421b16164fc039d5b61f17ce9/packages/tanstack-start/src/ApolloProvider.tsx#L79-L85

We stream events like "query started", "query received data", "query finished" etc. from SSR to the browser by calling `router.streamValue` (still have to push the change to the new call here).

https://github.com/apollographql/apollo-client-nextjs/blob/1306572238ff9d9421b16164fc039d5b61f17ce9/packages/tanstack-start/src/ApolloProvider.tsx#L57-L68

We do this as these events could happen in a waterfalled nested suspense boundary, so they wouldn't necessarily be available at the time `router.options.dehydrate` runs - even the start of a query might happen much later, and multiple values might be streamed in for a single query with a multipart response.

I'm honestly a bit surprised that (back when I looked at the code, might have changed in the meantime) the React Query implementation only collects values at the time of `router.options.dehydrate`, I would assume it would also need an approach like this - could that be a bug in the RQ integration?

schiller-manuel: it does not only collect values at time of `router.options.dehydrate`, it also collects values after a query finished.

phryneas: But does it collect new queries that start after `router.options.dehydrate` and when no query was running?

PR #2710, opened 2024-11-06
Fix hydration of server functions that return strings with backticks in them (in development)

### LINKED ISSUE #2510: Returning strings from server functions with some specific characters causes hydration errors (e.g '`')

Start

### Describe the bug

Returning strings with some specific characters causes hydration errors e.g:

```tsx
import { Outlet, createFileRoute } from "@tanstack/react-router";
import { createServerFn } from "@tanstack/start";

  loader: async () => {
    return await getSomeDataWithError();
  },
  component: LayoutComponent,
});

const getSomeDataWithError = createServerFn("GET", async () => {
  return {
    test: "test",
    error: "`", // This will cause hydration errors
  };
});

function LayoutComponent() {
  const { test } = Route.useLoaderData();
  return (
    <div className="p-2">
      <div className="border-b">{test}</div>
      <div>
        <Outlet />
      </div>
    </div>
  );
}
```

https://stackblitz.com/edit/tanstack-router-2feezb?file=app%2Froutes%2Freproduce.tsx

Check the /reproduce route

### Expected behavior

We should be able to return any kind of valid strings from our server functions. 

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 91.1]

### Additional context

_No response_

paneidos:
If a server function returns a string with a backtick in it, or an object that contains such a string, hydration will fail due to the backtick not being escaped in a console.info that only gets logged in development.

Fixes #2510

### DIFF
--- a/packages/react-router/package.json
+++ b/packages/react-router/package.json
@@ -68,10 +68,12 @@
   "dependencies": {
     "@tanstack/history": "workspace:*",
     "@tanstack/react-store": "^0.5.6",
+    "jsesc": "^3.0.2",
     "tiny-invariant": "^1.3.3",
     "tiny-warning": "^1.0.3"
   },
   "devDependencies": {
+    "@types/jsesc": "^3.0.3",
     "@testing-library/jest-dom": "^6.6.3",
     "@testing-library/react": "^16.0.1",
     "@vitejs/plugin-react": "^4.3.3",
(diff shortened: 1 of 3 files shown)


adrifer:
I'm facing a similar issue: ##2725

In my case is due to some routes having an id like `show/2/${}` which breaks as well with the `console.info(`...`)` as `${}` is an invalid character. 

I've created another PR #2726 which I think it fixes this issue and mine.

schiller-manuel:
closed in favor of #2726

PR #2711, opened 2024-11-07
idea exploration for validated env vars in Start

SeanCassiere:
Just exploring an idea for environment variables validation and accessibility using TanStack Start.

I'm currently referencing prior art of how Astro has tackled the problem to see if there's something we can offer to provide type-safe environment variables or whether we should be recommending users an external package like t3-env.

🤷🏼‍♂️ this may not even ever get merged in.


SeanCassiere:
Closing this for now, since the Nitro migration will likely open up more code-based avenues.

PR #2719, opened 2024-11-09
start: allow disabling server functions logs

kandros:
Allow disabling serverFn logs in development as requested in https://discord.com/channels/719702312431386674/1304470693970444318

I would like to also make this configurable in `app.config.ts` but I'm not sure what is the vinxi/nitro way to read configs in places like `packages/start/src/server-runtime/index.tsx`


SeanCassiere:
This has been superceded and completed.

PR #2724, opened 2024-11-10
chore: upgrade valibot dependency

fabian-hiller:
We have improved the [Standard Schema spec](https://github.com/standard-schema/standard-schema). Unfortunately, this leads to a breaking change in the Valibot's type signature. The implementation remains the same.


schiller-manuel:
closed in favor of https://github.com/TanStack/router/pull/2729

PR #2734, opened 2024-11-11
Support square brackets for advanced dynamic path segments (prefix, suffix & multiple)

CanRau:
Hey, this is my take so far on #707 inspired by #709 

Also tried my luck making it type safe, mostly ChatGPT tho and I'm actually not completely sure it works 😬 but I think it's a good enough start to open a PR, as I'd love prefixes to be officially supported 😇

I'm very curious what the team thinks about this approach, for explanation, I decided to not stick to `$` because having start `[` and end `]` properly delimited make suffixes and even multiple dynamic params easier or even possible. Also as mentioned before in https://github.com/TanStack/router/issues/707#issuecomment-2433814404 SvelteKit, Rakkas and possibly others already use square brackets, tho I'm curious if there's potential downsides?

So this PR should make the following possible:
prefix: `@[username]`
suffix: `[username]@`
both: `@[username]@`
where `@` could of course be anything
multiple: `[firstname]-[lastname]`

~What I think might be nice for multiple params, would be better matching so that it could properly distinguish/handle `[firstname]-[lastname]` and `[firstname]-[middlename]-[lastname]` which it doesn't yet, so currently the first route would also handle the second and the second would never match. 😅~ should be fixed now

Closes: #707


schiller-manuel:
As we will also add optional path params we must check whether this can be decoupled or not, especially when it comes to types. @chorobin let's discuss this

CanRau:
@schiller-manuel interesting I see 🤔 yea would of course be nice if everything would work properly together 🤞
for when do you have optional params planned?

just pushed an improvement so routes should be better sorted now to address the last part of my initial comment.

Also not sure breaking routing completely is desirable 😅 tho would it be helpful to drop `$param` for params and only use square brackets `[param]` for everything? Or find an alternative to square brackets 🤔

CanRau:
Yea totally understand just wanted to've mentioned it 😇
Currently patching this in via `pnpm patch` 🤓

CanRau:
What could I do to potentially help move this forward?
If it's about `optional params`? I guess no one is currently working on them right?

schiller-manuel:
first, we need to define the full set of features we want to support with path params (and maybe other related features that would be influenced by it such as parallel routes ) so we can solve this holistically. 

so you could open up an RFC style discussion to get this started.

CanRau:
Opened the RFC as WIP to get started https://github.com/TanStack/router/discussions/2800

schiller-manuel:
this will be supported as soon as we merge alpha in to main (you can try it out in alpha already)

PR #2761, opened 2024-11-14
Fix getting started example

unDemian:
The example presented in the **Getting Started** [section](https://tanstack.com/router/latest/docs/framework/react/start/getting-started) of Start is not working properly.

### Issues
1. There is no element with root `id` which means the app will not start properly.
SCR-20241114-upjf

2. `updateCount` requires a `{ data, ... }` object
3.  Apparently the default method for `createServerFn` is not `GET` so it should be provided.


adgang:
The updateCount method type should be POST, no?

unDemian:
> The updateCount method type should be POST, no?
Indeed, updated.

> 1. There is no element with root id which means the app will not start properly.
Has been fixed by https://github.com/TanStack/router/pull/2757

> 2. updateCount requires a { data, ... } object
Has been fixed by https://github.com/TanStack/router/pull/2775

SeanCassiere:
@unDemian thank you for your contribution here!

It looks like everything being addressed here has been closed out via other PRs. As such, I'm closing this issue.

Issue #2759, opened 2024-11-14
Start server function validator only supports the `ValidatorFn` shape (so things like Zod don't work as expected)


marbemac:
### Which project does this relate to?

Start

### Describe the bug

On the Start v1.81.7, the server function validator middleware does not actually support the options described by its typings.

It accepts `AnyValidator` (https://github.com/TanStack/router/blob/main/packages/react-router/src/validators.ts#L62-L67).

However, the actual implementation assumes the validator is the `ValidatorFn` shape (https://github.com/TanStack/router/blob/main/packages/start/src/client/createServerFn.ts#L325-L330).

Looks like `nextMiddleware.options.validator` in that `createServerFn.ts` has the `any` type - any ways to cleanup the types in there as well? If `nextMiddleware.options.validator` were set to `AnyValidator`, the issue would have been surfaced right away.

### Your Example Website or App

NA

### Steps to Reproduce the Bug or Issue

Attempt to use zod, or zod + @tanstack/zod-adapter, with a server function validator and it'll throw.

### Expected behavior

The server function validator supports the validator shapes described in [@tanstack/zod-adapter](https://github.com/TanStack/router/blob/main/packages/react-router/src/validators.ts#L62-L67).

### Platform

- OS: mac
- Browser: chrome

schiller-manuel:
do you have a minimal example that we can use to debug this?

schiller-manuel:
thanks.
yes we will probably reuse this from router.
if you want, you can also create a PR for this

marbemac:
Happy to - will open it shortly. Just wanted to confirm ya'll are OK exporting and re-using that function from router.

schiller-manuel:
exporting it does not really mean it is public API IMO. it's public when we add API docs (which we don't do here)

tannerlinsley:
I just pushed a "mostly" fix for this, but as we've been discussing the adoption of the "Standard Validation" schema internally, it might need to be massaged for backwards compatibility.

marbemac:
Getting familiar with the typings for functions / middleware's will take me longer than I have right now, but just one more pitch to somehow get things like `nextMiddleware.options.validator` typed correctly if possible - even though it's internal, would help to catch this kind of issue in the future especially as there is so much code churn happening in Start atm. I know ya'll have a lot on your plate though and it's prob low priority which is ok.

tannerlinsley:
[...]

PR #2762, opened 2024-11-14
improve internal middleware validation typings

marbemac:
Just constrains the validation generic to require it to implement AnyValidator.

This would have caught the error in https://github.com/TanStack/router/issues/2759:

Screenshot 2024-11-14 at 4 21 54 PM

### DIFF
--- a/packages/start/src/client/createServerFn.ts
+++ b/packages/start/src/client/createServerFn.ts
@@ -308,10 +308,13 @@ const applyMiddleware = (
 function execValidator(validator: AnyValidator, input: unknown): unknown {
   if (validator == null) return {}
 
-  if ('~validate' in validator) {
-    const result = validator['~validate']({ value: input })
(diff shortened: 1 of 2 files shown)


packages/start/src/client/createServerFn.ts
```diff
@@ -308,11 +308,14 @@ const applyMiddleware = (
...
-  if ('~validate' in validator) {
-    const result = validator['~validate']({ value: input })
+  if ('~standard' in validator) {
+    const result = validator['~standard'].validate(input)
 
     if ('value' in result) return result.value
 
+    if (result instanceof Promise)
+      throw new Error('Async validation not supported')
+
```
marbemac, line 318: I also fixed an existing typing error that was blocking the build process. @tannerlinsley I updated this to be consistent with the router implementation. Or was using `~validate` intentional?

PR #2781, opened 2024-11-16
fix(start): Fix `FormData` handling in server functions

np-13:
- Ensure context is included when serializing `FormData` on the client (serialized using default transformer and added to the `__TSR_CONTEXT` field in the payload)
- Add input type check on the server and, if it's `FormData`, extract context value from the `__TSR_CONTEXT` field before passing the actual payload to the middleware
- Add E2E test for `multipart/form-data` form submission

### DIFF
--- a/packages/start/src/client/createServerFn.ts
+++ b/packages/start/src/client/createServerFn.ts
@@ -1,4 +1,5 @@
 import invariant from 'tiny-invariant'
+import { defaultTransformer } from '@tanstack/react-router'
 import { mergeHeaders } from './headers'
 import type { AnyValidator, Constrain } from '@tanstack/react-router'
 import type {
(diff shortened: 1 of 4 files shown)


packages/start/src/client/createServerFn.ts
```diff
@@ -216,21 +217,49 @@ export function createServerFn<
...
+    const context = defaultTransformer.parse(serializedContext)
+    return {
+      context,
+      data: formData,
+    }
+  } catch (e) {
+    return {
+      data: formData,
+    }
+  }
```
np-13, line 260: I need others' opinions on this piece since it allows the request to be processed if the context cannot be parsed. However, I'm not sure if this is the desired behaviour, and we'd rather throw an error on context deserialisation failure.

PR #2782, opened 2024-11-16
feat(react-router): generic router history

matthiasschwarz:
Changes type of history in router to a generic type parameter to support advanced use cases like providing an own implementation of a `RouterHistory` or to enchance a specific `RouterHistory` with additional features.

### DIFF
--- a/examples/react/router-monorepo-react-query/packages/router/package.json
+++ b/examples/react/router-monorepo-react-query/packages/router/package.json
@@ -8,6 +8,7 @@
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "dependencies": {
+    "@tanstack/history": "^1.81.9",
     "@tanstack/react-query": "^5.61.3",
     "@tanstack/react-router": "^1.84.4",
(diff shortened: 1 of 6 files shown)


schiller-manuel:
thanks for the contibution!

please have a look at the build failures and fix those.
please also add a type test (e.g. in packages/react-router/tests/router.test-d.tsx)

matthiasschwarz:
> thanks for the contibution!
> 
> please have a look at the build failures and fix those. please also add a type test (e.g. in packages/react-router/tests/router.test-d.tsx)

- Fixed the build issues of the monorepo examples. This is a breaking change though since it was required to add `@tanstack/history` as dependency for the build typescript declarations of the router packages. I'm not sure if this is the correct way to handle that issue.
- The e2e tasks that are still failing are running successfully on my machine.
- Added tests for the history type of the router.

schiller-manuel:
@chorobin please have a look about the type changes

schiller-manuel:
sorry for the delay

Issue #2720, opened 2024-11-09
serverFn base url is hardcoded to localhost:3000


kandros:
### Which project does this relate to?

Start

### Describe the bug

https://github.com/tanstack/router/blob/cbc588dde615770891059186092fd030166370f0/packages/start/src/server-runtime/index.tsx#L87

This prevents using things serverFn URL property for progressive enhancement in production
https://tanstack.com/router/latest/docs/framework/react/start/server-functions#no-js-server-functions

SeanCassiere:
The issue here is that the URL attached to the server function has the base of `http://localhost:3000` hardcoded into it. This should probably be stripped out and left as an absolute navigation, just without the URI as a part of the URL.

```ts
const $testFn = createServerFn().handler(async () => {
  await doSomething();
  return { foo: 'bar' }
})

console.log($testFn.url) // "http://localhost:3000/_server/?_serverFnId=xxxxxxxxx"

// should probably just be "/_server/?_serverFnId=xxxxxxxxx"
```

PR #2787, opened 2024-11-17
test(start): checking if the bound server fn URL is hardcoded

SeanCassiere:
E2E Test for #2720


SeanCassiere:
This test is no longer needed since https://github.com/TanStack/router/commit/56fae6a0e4833b7c9414f9986c0ea03ea3e0dcb2 resolves this issue and only rendering the localhost origin on the client-side (with the correct host).

PR #2790, opened 2024-11-17
fix: router should have 'pending' when loaders in progress (draft)

lislon:
The `router` had a state `idle`, while the `load` and `beforeLoad` was not resolved. 

I noticed this in tests, when I attempted to fix the warning
```
Warning: An update to Transitioner inside a test was not wrapped in act(...).
```
by waiting for `state = idle`. 

I also noticed that `Transitionier` warning happens in `afterEach`, so I've added a unsubscribe method for history:

```ts
afterEach(() => {
  vi.resetAllMocks()
  window.history.replaceState(null, 'root', '/') <-- also causes 'Warning: An update to Transitioner inside a test was not wrapped in act(...).'
```

Should I expand this approach for other tests?

### DIFF
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -2113,7 +2113,11 @@ export class Router<
       }
     }
 
-    if (!this.isServer && !this.state.matches.length) {
+    if (
+      !this.isServer &&
+      !this.state.matches.length &&
+      !this.state.pendingMatches?.length
+    ) {
       triggerOnReady()
     }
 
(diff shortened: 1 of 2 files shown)


schiller-manuel:
this looks promising. can you resolve the merge conflict and ideally tackle those other tests as well?

LadyBluenotes:
Closing this since the router lifecycle has moved into `RouterCore` and currently it already keeps the router pending while loads are in progress.

Relevant code: [https://github.com/TanStack/router/blob/main/packages/router-core/src/router.ts](https://github.com/TanStack/router/blob/main/packages/router-core/src/router.ts) and [https://github.com/TanStack/router/blob/main/packages/react-router/src/Transitioner.tsx](https://github.com/TanStack/router/blob/main/packages/react-router/src/Transitioner.tsx)

PR #2796, opened 2024-11-18
Allow rerender test

mta-trackunit:
This changes so Tanstack router will support rerenders in tests using the renderHooks from testing library.

import { renderHook } from '@testing-library/react'

We heavily depend on the rerender tests to ensure no state changes are done in hooks, to avoid infinite rerender issues that are really hard to find and fix, so would really appriciate getting this issue fixed, I hope this solution is sufficient.


schiller-manuel (changes requested):
please find a way that does not remove the memos

mta-trackunit:
After thorough investigation, I couldn't find a viable way to support re-renders without removing the memoization. As a result, I’ve decided to close this PR and explore alternative approaches to address the issue.

While this change might introduce a small trade-off of a few additional re-renders, I believe it's a reasonable compromise to ensure compatibility with the React Testing Library and improve the framework's usability for others.

Thank you for your input and understanding as I work toward a better solution.

PR #2806, opened 2024-11-19
docs(start): explicitly set POST method `updateCount` server function examples

Balastrong:
Since the default method is GET I think it's correct to explicitly set these examples to POST

### DIFF
--- a/docs/framework/react/guide/tanstack-start.md
+++ b/docs/framework/react/guide/tanstack-start.md
@@ -258,7 +258,7 @@ const getCount = createServerFn({
   return readCount()
 })
 
-const updateCount = createServerFn()
+const updateCount = createServerFn({ method: 'POST' })
   .validator((d: number) => d)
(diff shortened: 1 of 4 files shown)


SeanCassiere (commented):
What's the goal behind this change?

From my point of view:
- The general convention is to send mutation-related actions over a "POST" request so that you are not opted-in to any unintentional browser-side caching.
- As a quickfire documentation approach, this works as an easy method showing off the ability to use POST requests.

Perhaps, it's the explicit definition of the usage of "GET" is what needs to be taken out?

Balastrong:
For sure we have to explicitly set POST otherwise it's gonna default to GET which feels wrong in the mutations, I'd keep that as is in the PR.

Speaking of the explicit GET we can remove it if we think it's redundant and we want to remove an extra step but being explicit on the method doesn't seem like a bad idea to me.

PR #2841, opened 2024-11-24
fix(start): fix server function handling of standard schemas

EskiMojo14:
Some implementations (namely Valibot) still include the value key for failure results, which is technically still correct in structural typing (since an object type is a *minimum* requirement). What can actually be relied on is that `result.issues` is undefined if the result is valid, as this is in the spec.

Currently failure results from Valibot are still treated as success - this PR flips to checking for `result.issues` instead of `result.value`.

### DIFF
--- a/packages/start/src/client/createServerFn.ts
+++ b/packages/start/src/client/createServerFn.ts
@@ -17,8 +17,6 @@ import type {
   ResolveAllValidators,
 } from './createMiddleware'
 
-//
-
 export interface JsonResponse<TData> extends Response {
   json: () => Promise<TData>
 }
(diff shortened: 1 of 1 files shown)


packages/start/src/client/createServerFn.ts
```diff
@@ -17,8 +17,6 @@ import type {
   ResolveAllValidators,
 } from './createMiddleware'
 
-//
```
EskiMojo14, line 20: i thought these were a typo added by me but apparently they were there before - happy to add them back if they're there for a reason

schiller-manuel:
thanks for the contribution. we need the same fix for search param validation in router.ts
do you want to provide a PR for this as well?

EskiMojo14:
> thanks for the contribution. we need the same fix for search param validation in router.ts do you want to provide a PR for this as well?

sure, will do

EskiMojo14:
raised #2847

PR #2843, opened 2024-11-24
docs(start): `FormData` to use `e.currentTarget` instead of `e.target`

AntanasMisiunas:
Currently [this example](https://tanstack.com/router/latest/docs/framework/react/start/server-functions#formdata-parameters) has a type error:

Screenshot 2024-11-24 at 09 29 16

because:

Screenshot 2024-11-24 at 09 29 45

At first I ended up type casting `event.target` but there's a better solution: use `event.currentTarget`.

Screenshot 2024-11-24 at 09 31 45

For context: `event.currentTarget` refers to the element the event handler is attached to, while `event.target` refers to the element that fired the event. [Src](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget)

### DIFF
--- a/docs/framework/react/start/server-functions.md
+++ b/docs/framework/react/start/server-functions.md
@@ -306,7 +306,7 @@ function Test() {
     <form
       onSubmit={async (event) => {
         event.preventDefault()
-        const formData = new FormData(event.target)
+        const formData = new FormData(event.currentTarget)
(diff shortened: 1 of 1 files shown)


SeanCassiere (approved):

SeanCassiere:
@AntanasMisiunas thank you for your contribution here in spotting and fixing this mistake!

PR #2854, opened 2024-11-25
fix(react-router): ensure links and scripts are available on routes created with `createRoute`

GregoryCollett:
We found that creating a route with `createRoute` would not expose links and scripts where as when creating a route with `createFileRoute` the link and script properties would exist as expected. We expect the behaviour to be the same across the board.

### DIFF
--- a/packages/react-router/src/router.ts
+++ b/packages/react-router/src/router.ts
@@ -2528,6 +2528,8 @@ export class Router<
                             loaderData,
                           })
                           const meta = headFnContent?.meta
+                          const links = headFnContent?.links
+                          const scripts = headFnContent?.scripts
 
                           const headers = route.options.headers?.({
                             loaderData,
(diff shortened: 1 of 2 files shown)


schiller-manuel:
a e2e test in one of the e2e start tests would be nice.

packages/react-router/src/router.ts
```diff
@@ -2528,6 +2528,8 @@ export class Router<
                             loaderData,
                           })
                           const meta = headFnContent?.meta
+                          const links = headFnContent?.links
```
schiller-manuel, line 2531: I don't understand why this worked in `createFileRoute` but in in `createRoute`. can you explain this please?

SeanCassiere (approved):

packages/react-router/src/router.ts
GregoryCollett: Sorry, I haven't had time to fully evaluate this but agree it didn't seem immediately apparent why this worked using createRootRoute and createFileRoute vs createRoute. I have to assume they go through the same path and it seems file route calls createRoute under the hood... I will explore this more but as I say I haven't yet found the time to give this an in depth diagnosis beyond the initial findings.

schiller-manuel:
thanks for checking this

Issue #2799, opened 2024-11-18
Default search becomes empty object when used in `search`


OwenVey:
### Which project does this relate to?

Router

### Describe the bug

I am using both the `stripSearchParams` and `retainSearchParams` middlewares to strip default values from the query params as well as removing any parameters not related to that route. This works as expected as shown in the reproduction URL below but the moment I use my `DefaultSearch` object in a `Link` like so:
```tsx
<Link to={Route.fullPath} search={DefaultSearch}>
  Clear Search
</Link>
```
the `DefaultSearch` object itself becomes `{}` which then causes problems with the middlewares.

My guess is there is a bug somewhere where the object being passed to `search` is being overwritten.

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-fyxlu6?file=src%2Froutes%2F__root.tsx

### Steps to Reproduce the Bug or Issue

1. Open reproduction link and click the "Open Preview in new tab" button in the top right corner (so that we can see search params in URL)
2. Notice that when switching between the "Home" and "About" tabs, the middlewares are working as expected and the `DefaultSearch` objects are being displayed as expected.
3. Go to the `__root.tsx` file and uncomment lines 20 and 29
4. Switch between the two tabs and notice that the `DefaultSearch` objects are displayed as `{}` and the `search` middlewards no longer function as expected.

### Expected behavior

I expect my `DefaultSearch` object be unmodified and log out as originally created but I am seeing it log as an empty object when using my `DefaultSearch` in a `<Link>`'s `serach`.

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/e5ce8592-f9d2-4866-9ae7-afd3fd133047)

### Platform

- OS: Windows
- Browser: Chrome
- Version: 131.0.6778.70 (Official Build) (arm64)

### Additional context

Mentioned in the TanStack Discord here: https://discord.com/channels/719702312431386674/1306270495016095847/1306442814875242548

GiulioMabiloft:
For anyone encountering this bug, spreading the default search params object—thereby creating a new reference—resolves the issue. Hope this helps

schiller-manuel:
not reproducible anymore

PR #2855, opened 2024-11-25
fix: don't overwrite searchParam values if they are provided in url with what is provided in validateSearch

### LINKED ISSUE #2830: Navigation with retainSearchParams middleware not working

Router

### Describe the bug

Navigating when using a retainSearchParams middleware seems to be broken.
The URL for the link shows the search parameter but once we navigate, the search parameter is lost.

https://stackblitz.com/edit/tanstack-router-retain-nok?file=src%2Froutes%2F__root.tsx

Navigate to the about route using the link. The `rootValue` search parameter is lost.

### Expected behavior

 `rootValue` search parameter should be preserved.

This behavior used to work. It was working on this example:
https://stackblitz.com/edit/tanstack-router-retain-ok?file=src%2Froutes%2F__root.tsx

If on the working version we delete the lock file and recreate it, it should stop working.

### Screenshots or Videos

_No response_

### Platform

- OS: Linux
- Browser: Chrome
- Version: 130

### Additional context

_No response_

### LINKED ISSUE #2845: Router searchParameter gets reset eventhough we tell it to be persisted using retainSearchParams

Router

### Describe the bug

Using retained searchParams, if we change the retained param `x` form `a` to `b` on some route (lets say route-a) and we navigate to route-b, which doesn't change param `x`, param `x` will revert back to its default value.

This is a regression because it works on using older versions.

Same code older versions:
[stackblitz](https://stackblitz.com/edit/tanstack-router-retain-ok-oemzf3?file=src%2Froutes%2F__root.tsx)
Same code latest versions:
[stackblitz](https://stackblitz.com/edit/tanstack-router-retain-ok-bwm1lv?file=src%2Froutes%2F__root.tsx,package.json,src%2Froutes%2F_auth%2Fabout.tsx,src%2Froutes%2F_auth%2Fsettings.tsx,src%2Froutes%2Findex.tsx,src%2Fmain.tsx)

-see desc

1. click on Home route (see Auth defaults to true)
2. click on Settings route (see Auth remains true)
3. click on About route (see now Auth is false)
4. click on Home route:
4.a older version: Auth remains as false (which is expected)
4.b newer version: Auth changes back to true (which is unexpected) 

### Expected behavior

after chaning the retained searchParams, it should stay unchanged after navigations, and only change if explicitly said so.
* some other weird thing here is that even without using this middle params stay on URL, kinda making retainSearchParams useless, unless im not grasping the use of it correctly. 

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 91.1]

### Additional context

_No response_

FoHoOV:
Title is actually good enough :D 
Fixes #2845
Fixes #2830


schiller-manuel:
can you please also have a a look at https://github.com/TanStack/router/issues/2830 and https://github.com/TanStack/router/issues/2799 as they seem related?

packages/react-router/src/searchMiddleware.ts
```diff
@@ -9,13 +9,14 @@ export function retainSearchParams<TSearchSchema extends object>(
   return ({ search, next }) => {
     const result = next(search)
     if (keys === true) {
-      return { ...search, ...result }
+      return { ...result, ...search }
```
schiller-manuel, line 12: I don't get this. `retainSearchParams` should only carry over search params that are NOT part of e.g. a `<Link search={...}>`. but this change here would mean that if the search params already contained `foo: 'hello'`, then the value from `<Link search={{foo:'bar'}}>` would be ignored

packages/react-router/src/searchMiddleware.ts
```diff
@@ -9,13 +9,14 @@ export function retainSearchParams<TSearchSchema extends object>(
...
     }
-    // add missing keys from search to result
+    // just like above, but only overwrite explicitly mentioned keys
```
schiller-manuel, line 14: same here, `result` should have precedence

packages/react-router/src/searchMiddleware.ts
FoHoOV: Shouldn't it? The value is retained, whatever it's initialized to. Imagine you go to route x with param=p1, then goto x/y with search param=p2, this would means the param is p2 when ur on x/y and reset back to p1 when go on x. This means the value is not retained on x/* route right? If this is not the intention, I might have misunderstood the point.

schiller-manuel: no, the intention of this search middleware is to keep search params around so they continue to live on with their previous value unless a Link overrides them

FoHoOV: Ooh, ok then. Feel free to close this PR, cuz I'm not home for next 2 or 3 days.

* next(search) returns the validated result (including what's in the links search prop) so it's not gonna be as easy as this fix.

schiller-manuel: no rush, just convert it to draft for the time being

FoHoOV:
@schiller-manuel Currenly, search middlewares call next which calls searchValidator, meaning, `searchValidator` result is the input to `retainSearchParams` middleware. Validator might return a default `x` and then in our `retainSearchParams`, we say ok `x` exists (which is the result of validator), so lets keep it and ignore what we have right now. But, I think it should be the other way around, validator should use the result of middlewares not middlewares using the result of validator. Currenly we have [this](https://github.com/TanStack/router/blob/412f6640dcc386bb2aba4b82af02fddbf8fda46d/packages/react-router/tests/link.test.tsx#L4197) test:
``` typescript
const postsRoute = createRoute({
  path: 'posts',
  validateSearch: z.object({
    foo: z.string().default('default'),
  }),
  search: {
    middlewares: [
      // @ts-expect-error we cannot use zodValidator here to due to circular dependency
      // this means we cannot get the correct input type for this schema
      stripSearchParams({ foo: 'default' }),
      retainSearchParams(true),
    ],
  },

  component: () => {
    const { foo } = postsRoute.useSearch()
    return (
      <>
        <h1>Posts</h1>
        <div data-testid="posts-search">{foo}</div>
        <Link data-testid="posts-link-new" to="/posts/new">
          new
        </Link>
      </>
    )
  },
})
```
here we are explicitly saying that this route should have `foo` with default of `default` but also stripping it in the middlewares. I think, the searchValidator should have higher proirity over what our middleware is saying, middleware strips it, but the validator should add it back, for example if it was stripping `foo` with value `x`, it should remove `x` and our validator should add `default` to it which makes sense (IMHO), this way middlewares are in the middle and not in the end. If it makes sense, and you are ok with it, I can make the change.

schiller-manuel:
> then why write it in the first place? just dont set the default value for foo
the reason for that is that you can then rely in your code that the value is always set and you don't need to check for `undefined`

FoHoOV:
Hi, thanks for your reply. 
This was my last possible solution that I could think of. I don't know any other solutions. I give up :D but the order of middlewares + searchValidator is really weird currently IMHO.
Thank you for your time >_<

PR #2862, opened 2024-11-26
feat(react-router): add support for ViewTransition types

asnaeb:


SeanCassiere (commented):
Is there anyway for there to have some sort of fallback detection to check if this ViewTransition Types API is supported in the browser?

Like a scenario where ViewTransitions is supported but the Types API is not supported, so just default to regular params of passing in a function.

SeanCassiere:
Also, please add a test in the `e2e/react/basic` where navigations are made using `viewTransiton=true` and `viewTransition={options}` just to make sure nothing explodes.

asnaeb:
> Is there anyway for there to have some sort of fallback detection to check if this ViewTransition Types API is supported in the browser?
> 
> Like a scenario where ViewTransitions is supported but the Types API is not supported, so just default to regular params of passing in a function.

One thing that comes to my mind is checking with `CSS.supports("selector(:active-view-transition-type(a)")`. Having this return `true` *should* mean that the browser supports the feature. I tested with supported and unsupported browsers and apparently the result is consistent. Do you think this may actually work?

SeanCassiere:
> One thing that comes to my mind is checking with `CSS.supports("selector(:active-view-transition-type(a)")`. Having this return `true` _should_ mean that the browser supports the feature. I tested with supported and unsupported browsers and apparently the result is consistent. Do you think this may actually work?

The selector approach sounds good.

If it is not supported, but they've opted into view transitions, then it should fallback and still make the transition.

asnaeb:
Okay, I've added the check for the feature support during construction, to avoid repeating it on every navigation. If no support is detected, it falls back to transition without types. Does it look good to you?

schiller-manuel:
is this a user facing API? if yes we need documentation

packages/react-router/src/router.ts
```diff
@@ -765,6 +770,9 @@ export class Router<
 
     if (typeof document !== 'undefined') {
       ;(window as any).__TSR__ROUTER__ = this
+
+      // Only check for viewTransition types support on construction
```
SeanCassiere, line 774: Please move this into the `.update` method on the `Router` class.

SeanCassiere (changes requested):
This looks pretty good!

I've left a note on a change to be made.

Also please check the `docs/` folder and update any documentation in there.

SeanCassiere (approved):

packages/react-router/src/router.ts
```diff
@@ -853,6 +858,12 @@ export class Router<
         },
       })
     }
+
+    if (!this.isServer) {
+      this.isViewTransitionTypesSupported = CSS.supports(
```
SeanCassiere, line 863: Looking at the failures with the unit tests, we probably should perform a check for `window` or something like `typeof window !== 'undefined' && 'CSS' in window` and then do the check as `window.CSS.supports(...)`.

asnaeb: Should I still include `!this.isServer` alongside these new checks?

SeanCassiere (approved):
@asnaeb thank you for working with us and pushing this through!

P.S. You may want to set up your code editor to pick up and format using the repo's prettier config 😉

asnaeb:
> @asnaeb thank you for working with us and pushing this through!
> 
> P.S. You may want to set up your code editor to pick up and format using the repo's prettier config 😉

It's been a pleasure and I'm looking forward to contribute more. I usually use eslint as a formatter but will definitely configure prettier for my next contributions :)

PR #2864, opened 2024-11-26
docs(router): usage of accessing the loaderData on the data-loading guide

arackaf:
Add info on accessing loader data from components in the data loading page

### DIFF
--- a/docs/framework/react/guide/data-loading.md
+++ b/docs/framework/react/guide/data-loading.md
@@ -92,7 +92,7 @@ To consume data from a `loader`, use the `useLoaderData` hook defined on your Ro
 const posts = Route.useLoaderData()
 ```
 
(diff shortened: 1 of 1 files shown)


docs/framework/react/guide/data-loading.md
```diff
@@ -84,6 +84,25 @@ The `loader` function receives a single object with the following properties:
...
 
+## Consuming data from `loader`s
+
+To consume data from a `loader`, use the `useLoaderData` hook defined on your Route object.
+
+```tsx
+const posts = Route.useLoaderData()
+```
+
+If you don't have ready access to your route object (ie you're deep in the component tree for the current route), you can use `getRouteApi` to access the same hook (as well as the other hooks on the Route object). This should be preferred over importing the Route object, which is likely to create circular dependencies.
```
SeanCassiere, line 95:
```suggestion
If you don't have ready access to your route object (i.e. you're deep in the component tree for the current route), you can use `getRouteApi` to access the same hook (as well as the other hooks on the Route object). This should be preferred over importing the Route object, which is likely to create circular dependencies.
```

PR #2866, opened 2024-11-26
test(router): adding test to verify route generation from `tsr.config.json`

davidturissini:
Simply adding a test to verify that the contents of `tsr.config.json` are being used to generate correct outputs. This PR adds one test to `packages/router-generator/tests/generator` that has the following configuration:

```
{
  "disableLogging": true,
  "routesDirectory": "./configured-routes-path",
  "generatedRouteTree": "./routeTree.gen.ts"
}
```

Like all other tests in this directory, this test verifies the snapshot output. `routesDirectory` is intentionally configured to a non-standard directory to verify that the contents of the config file are being read.

`generatedRouteTree` is configured as `routeTree.gen.ts` like in every other test, but I left this unchanged because I did not want to touch `.gitignore`.

Adding this PR because I noticed that there were no tests in the repo that actually loaded `tsr.config.json` from the file system(I could be wrong!). It would seem that a test verifying this would be a pre-requisite for supporting non-json config files (js, cjs, mjs, ts, etc)


packages/router-generator/tests/generator.test.ts
```diff
@@ -40,19 +40,25 @@ async function traverseDirectory(
   }
 }
 
+function shouldUseInlineConfig(folder: string) {
+  return folder !== 'config-json';
+}
```
davidturissini, line 45: Borrowing a pattern that I see in this file: [shouldThrow](https://github.com/TanStack/router/blob/656c56844983ede516b1c27fe312ad8e4261283a/packages/router-generator/tests/generator.test.ts#L202). This certainly works but I'm not confident it will scale beyond a half dozen test cases or so

packages/router-generator/tests/generator.test.ts
```diff
@@ -40,19 +40,25 @@ async function traverseDirectory(
...
   const dir = makeFolderDir(folder)
 
-  const config = await getConfig({
+  const defaultConfig = shouldUseInlineConfig(folder) === true ? {
     disableLogging: true,
     routesDirectory: dir + '/routes',
     generatedRouteTree: dir + generatedRouteTree,
     ...rest,
-  })
+  } : undefined;
```
davidturissini, line 59: If this test is specifically verifying config from the filesystem, bail on passing an inline config. Use the resolved config instead.

packages/router-generator/tests/generator/config-json/configured-routes-path/__root.tsx
```diff
@@ -0,0 +1,11 @@
...
+import { Outlet, createRootRoute } from '@tanstack/react-router'
+
+export const Route = createRootRoute({
+  component: () => (
+    <React.Fragment>
+      <div>Hello "__root"!</div>
+      <Outlet />
+    </React.Fragment>
+  ),
+})
```
davidturissini, line 11: Just using a very simple route configuration because not really concerned about the routes themselves.

schiller-manuel:
I am not sure this is the right approach. shouldn't we rather just test that the config is loaded correctly from generator? why would we need a full blown test like this?

davidturissini:
> I am not sure this is the right approach. shouldn't we rather just test that the config is loaded correctly from generator? why would we need a full blown test like this? 

Yes- if we can verify that the correct config is loaded without generating the routes that would be ideal. Do you have any suggestions for to tackle that? 

My  rationale for re-using the route generator tests is that it actually loads the config from the real filesystem. This is what a real end user would do so it seems like a closer approximation than using memfs, mocking fs APIs or something similar. Maybe a better strategy would be to just test getConfig directly

schiller-manuel:
> Maybe a better strategy would be to just test getConfig directly

yes that's what I meant. not sure what needs to be done to test this.

davidturissini:
> > Maybe a better strategy would be to just test getConfig directly
> 
> 
> 
> yes that's what I meant. not sure what needs to be done to test this.

Perhaps nothing- but what I really want to add is support for js files in config. In order for that to happen, there probably should be some test that verifies that json config files continue to work. Now maybe supporting js for config files isn't desired, and in that case yea, prob don't need to test this.

schiller-manuel:
what do you mean by "js files in config"?

schiller-manuel:
by using a vite like `define...` approach as we did for virtual file routes?

davidturissini:
> by using a vite like `define...` approach as we did for virtual file routes?

That api had crossed my mind yea. Something like:

```
module.exports = defineTanStackConfig({
  virtualRoutes: defineVirtualRoutes(...)
})
```

davidturissini:
@schiller-manuel do you think this is worth pursuing or no?

PR #2893, opened 2024-11-30
refactor(start): expose vinxi router middleware

jadedevin13:
Make use of vinxi strength by being able to use custom routers and middlewares on top of tanstack start and routers. To be used by advanced users only.


SeanCassiere (changes requested):

SeanCassiere (approved):
Removed the Vinxi router stuff in favor of `.addRouter`.

PR #2900, opened 2024-12-01
Update external-data-loading.md

stabildev:
Replace duplicate, confusing sentence


SeanCassiere (changes requested):
This text being taken out doesn't read well whilst going through the guide. I'd rather keep the original text in.

Unless you've got better verbiage that still correctly shows off the fact that the guide is leading into an example using TanStack Query this'd have to be closed.

SeanCassiere:
Closing

PR #2909, opened 2024-12-02
Update getting-started.md for type errors regarding routing functionality

leforgedroid:
Before the server can be started via `npm run dev`, the code for routeTree.gen.ts file needs to be created. This is done by, first, running `npm run build`.


schiller-manuel:
`npm run dev` should also create the file, does this not work for you?

SeanCassiere (changes requested):
This step should not be needed since the `npm run dev` command should generate the `routeTree.gen.ts` file.

Unless a bug has been discovered here, I'd rather not add any unneeded text on the "getting started" guide.

PR #2920, opened 2024-12-04
feat(start): create-start cli

timoconnellaus:
This is a new CLI package called `@tanstack/create-start` that scaffolds new TanStack Start applications. Here's how it works:

Key Features:
- Modular design - templates can be composed of multiple modules
- Modules can be imported and used as a package - the purpose is to allow programmatic scaffolding of apps (e.g. to create a model context protocol server / client for building tanstack start apps with generative AI)
- Modules can be run in isolation - the use case is adding a module to a project after the project is scaffolded. This isn't operational yet but the skeleton is there - e.g. `npx @tanstack/create-start add tanstack-query` or `@tanstack/create-start add convex`

Key CLI Features:
1. Interactive CLI prompts for project configuration
2. Multiple template options (barebones and showcase)
3. Support for different package managers (npm, yarn, pnpm, bun)
4. IDE setup (currently VSCode)
5. Git initialization
6. Dependency installation

Basic Usage:
```bash
# Using npm
npm create @tanstack/start@latest

# Using yarn
yarn create @tanstack/start

# Using pnpm
pnpm create @tanstack/start
```

The CLI flow:

1. Shows TanStack Start ASCII logo
2. Prompts for configuration:
   - Template selection (barebones or showcase)
   - Project directory 
   - Package name
   - Package manager preference
   - Whether to install dependencies
   - Whether to initialize git
   - IDE preference

3. The barebones template provides:
   - Basic routing setup
   - Minimal route structure
   - TypeScript configuration
   - Essential dependencies

4. The showcase template adds:
   - Example of server functions
   - More complex routing examples
   - File persistence demo

The CLI can also be run non-interactively with command line arguments:

```bash
npm create @tanstack/start@latest -- \
  --template barebones \
  --directory my-app \
  --package-name my-app \
  --package-manager npm \
  --install-deps \
  --init-git \
  --ide vscode
```

Key modules:
- `core` - Base project structure and dependencies
- `templates` - Different starter templates
- `ide` - IDE specific configurations
- `packageManager` - Package manager detection and commands

The package uses a modular architecture where each feature is a separate module that can:
1. Initialize configuration
2. Prompt for options
3. Validate settings
4. Apply changes

There are also unit tests + a minimal set of e2e tests


timoconnellaus:
This is what the CLI looks like when it runs
image

timoconnellaus:
There's still a bit more work to go on this before it's ready - I'm away until Monday but will get back to it then

schiller-manuel:
documentation needs to be updated to describe the usage

packages/create-start/build.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineBuildConfig } from 'unbuild'
+
+export default defineBuildConfig({
+  entries: ['src/index', 'src/cli-entry'],
```
schiller-manuel, line 4: why do we need two entry points?

packages/create-start/src/cli.ts
```diff
@@ -0,0 +1,104 @@
...
+    .addOption(options.directory)
+    .addOption(options.installDeps)
+    .addOption(options.noInstallDeps)
+    .addOption(options.initGit)
+    .addOption(options.noInitGit)
+    .addOption(options.hideLogo)
+    .addOption(options.ide)
+}
+
+// const addQueryCommand = addBaseOptions(
```
schiller-manuel, line 47: remove this?

packages/create-start/src/directory.ts
```diff
@@ -0,0 +1,81 @@
...
+  while (pathExists) {
+    folderName = `${DEFAULT_NAME}-${counter}`
+    absolutePath = getAbsolutePath(folderName)
+    pathExists = await doesPathExist(absolutePath)
+    counter++
+  }
+  return `./${folderName}`
+}
+
+export const newProjectDirectoryCliOption = createOption(
```
schiller-manuel, line 51: can we move the logic of `newProjectDirectoryCliOption` and `newProjectDirectoryPrompt` into one function to avoid repetition?

packages/create-start/src/modules/core/template/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "jsx": "react-jsx",
+    "moduleResolution": "Bundler",
+    "module": "ESNext",
+    "target": "ES2022",
+    "skipLibCheck": true,
+    "strictNullChecks": true
```
schiller-manuel, line 8: I would just use `strict: true` instead of `strictNullChecks`

packages/create-start/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "zod": "^3.23.8",
+    "zod-opts": "^0.1.8"
+  },
+  "peerDependencies": {
+    "@tanstack/react-router": "workspace:^",
+    "@tanstack/router-devtools": "workspace:^",
+    "@tanstack/start": "workspace:^",
+    "@types/react": "^18.3.12",
+    "@types/react-dom": "^18.3.1",
+    "react": "^18.3.1",
```
schiller-manuel, line 72: how would we handle an option that lets the user choose between react18 and react19?

packages/create-start/src/modules/ide.ts
```diff
@@ -0,0 +1,108 @@
...
+        ? {
+            error: `Failed to set up ${cfg.ide}`,
+            inProgress: `Setting up ${cfg.ide}`,
+            success: `${cfg.ide} set up`,
+          }
+        : undefined
+    },
+  })
+
+// export const ideModuleaa = createModule({
```
schiller-manuel, line 71: remove this?

packages/create-start/src/modules/vscode/index.ts
```diff
@@ -0,0 +1,37 @@
...
+        templateFolder: './template',
+        targetFolder: targetPath,
+        overwrite: false,
+      })
+    },
+    apply: async ({ targetPath }) => {
+      // Copy the vscode template folders into the project
+      const _ = initHelpers(__dirname, targetPath)
+
+      // TODO: Handle when the settings file already exists and merge settings
```
schiller-manuel, line 28: when would that file exist already?

packages/create-start/src/utils/runCmd.ts
```diff
@@ -0,0 +1,43 @@
+import spawn from 'cross-spawn'
+
+export async function runCmd(
```
schiller-manuel, line 3: when do we need `runCmd` instead of `runPackageManagerCommand`?

schiller-manuel: just for git?

packages/create-start/build.config.ts
timoconnellaus: This has been chagned to build the files using mkdist to retain the tree structure. This is required for the modules to understand relatively whre their templates are

packages/create-start/src/cli.ts
timoconnellaus: It will be uncommented soon. I started the add modules to test the UX but then commented them as there are no modules to add yet

packages/create-start/package.json
timoconnellaus: That's only used to get the semver. I think we can cross that bridge when we come to it - a work around or a new process for this case

packages/create-start/src/modules/vscode/index.ts
timoconnellaus: In the case the module is called when adding something to an existing project where that module needs to add vscode settings. This might not be called just for router vscode settings for example

packages/create-start/src/utils/runCmd.ts
timoconnellaus: Will simplify this

packages/create-start/src/directory.ts
timoconnellaus: I'm not sure what you mean

packages/create-start/src/modules/core/template/tsconfig.json
timoconnellaus: This is copied from the official docs https://tanstack.com/router/latest/docs/framework/react/start/getting-started

packages/create-start/src/modules/ide.ts
timoconnellaus: This is done

packages/create-start/src/utils/runCmd.ts
timoconnellaus: Pushed https://github.com/TanStack/router/pull/2920/commits/91fde4a4beed927d8bc040356ca81279368dcb60

thomasballinger:
This is great! When asked to enter the project name it was odd to me it didn't default to the project location, typically I make this match so that might be a better guess that `tanstack-start`.

PR #2921, opened 2024-12-04
fix(start): bump `vinxi` to `0.5.0`

SeanCassiere:


SeanCassiere (commented):
TanStack Config's deps still needs to support Vite 6.
Hopefull that'd fix these CI errors.

PR #2926, opened 2024-12-04
feat: static build caching for server functions

tannerlinsley:
This adds a new concept of "server function type". By default, all server functions are dynamic, meaning they will reach out to the server to execute. This addition brings a new type of `static`, which can be configured in a few places:

### As the default
```tsx
createServerFn({ type: 'static' })
```

### As an dynamic function of the data/context
```tsx
createServerFn().type(ctx => ctx.data > 5 ? 'static' : 'dynamic')
```

### At the call site of the serverFn
```tsx
const myFn = createServerFn()

myFn({ data: '...', type: 'static' })
```

It falls into this order with the rest of the builder API:

```tsx
createServerFn()
  .validate(...)
  .middleware(...)
  .type(...)
  .handler(...)
```

### How it works

- Development
  - Everything functions as a normal server function with zero caching
- Prerendering
  - Client
    - **Each invocation of the server function is determined if it is static or dynamic**
    - **The `type` is sent to the server with the request**
  - Server
    - The server function operates normally
    - **If static, the result of the server function is stored in the public output as a static JSON file**
  - Client
    - The result is received as normal
- Production
  - Client
    - Each invocation of the server function is determined if it is static or dynamic
    - If the server function is dynamic, the request is handled normally
    - **If the server function is static, the request is rerouted to fetch the static JSON file**

### Customizing the `serverFnStaticCache`

You can create and set your own `serverFnStaticCache` by using the `createServerFnStaticCache` and `setServerFnStaticCache` functions.

Imagine we might want to cache the output of the server function to giant massive JSON file (don't do this). We could do something like this:

```tsx
setServerFnStaticCache(() => {
  const cacheFile = path.join(process.cwd(), 'public/static-cache.json')
  const getKey = (ctx: MiddlewareCtx) => {
    return `${ctx.filename}-${ctx.functionId}-${JSON.stringify(ctx.data)}`
  }

  return createServerFnStaticCache({
    getItem: async (ctx) => {
      const key = getKey(ctx)
      const data = await readFile(cacheFile, 'utf8')
      const json = JSON.parse(data)
      return json[key]
    },
    setItem: async (ctx, response) => {
      const key = getKey(ctx)
      const data = await readFile(cacheFile, 'utf8')
      const json = JSON.parse(data)
      json[key] = response
      await writeFile(cacheFile, JSON.stringify(json, null, 2))
    },
    fetchItem: async (ctx) => {
      const key = getKey(ctx)
      const data = await fetch('/static-cache.json').then((res) => res.json())
      const json = JSON.parse(data)
      return json[key]
    },
  })
})

```

Imagine the other places you could technically create this cache:
- Database
- CDN
- File System
- In-memory Datastore (like Redis)

### More cool things
- You get middleware for free since it's based on server functions
- Any server context that you send back to the client is also saved with the JSON
- If errors are thrown, they are saved to the JSON as well under an `error` key
- The Start serializer is used, which means dates and undefined are properly serialized.


packages/start/src/client/createServerFn.ts
```diff
@@ -467,24 +543,77 @@ function serverFnBaseToMiddleware(
...
+            acc[curr] = value[curr]
+            return acc
+          }, {})
+      : value
+
+  // Convert JSON to string with sorted keys
+  const jsonString = JSON.stringify(json ?? '', sortedKeysReplacer)
+
+  // Replace characters invalid in filenames
+  return jsonString
```
SeanCassiere, line 616: Perhaps this could be simplified with a `encodeURIComponent`?

packages/start/src/config/index.ts
```diff
@@ -18,6 +18,7 @@ import { config } from 'vinxi/plugins/config'
 import { serverFunctions } from '@vinxi/server-functions/plugin'
 // @ts-expect-error
 import { serverTransform } from '@vinxi/server-functions/server'
+import { createNitro } from 'nitropack'
```
SeanCassiere, line 21: In the `package.json` we should move this from `devDependencies` to `dependencies`.

packages/start/src/config/index.ts
```diff
@@ -88,9 +90,9 @@ function mergeSsrOptions(options: Array<vite.SSROptions | undefined>) {
   return ssrOptions
 }
 
-export function defineConfig(
+export async function defineConfig(
```
SeanCassiere, line 93: We should probably confirm with @nksaraf that having the Vinxi export be a Promise that needs to be resolved won't break in the future.

examples/react/start-basic-static/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "tanstack-start-example-basic-static",
```
SeanCassiere, line 2: Keeping this here as a reminder that we should duplicate this example into the `e2e/start` directory and run the end-to-end test on the built output.

packages/start/src/client/createServerFn.ts
tannerlinsley: Worth a test. But not all url safe stuff is file-safe stuff

packages/start/src/config/index.ts
tannerlinsley: Yeah, we should

packages/start/src/config/index.ts
tannerlinsley: Yeah

examples/react/start-basic-static/package.json
tannerlinsley: Good idea

packages/start/src/server-runtime/index.tsx
```diff
@@ -152,7 +152,9 @@ export function createServerReference(_fn: any, id: string, name: string) {
   }
 
   return Object.assign(proxyFn, {
-    url: functionUrl.replace(fakeHost, ''),
+    url: functionUrl
+      .replace(fakeHost, '')
+      .replace(fakeHost.replace('https:', 'https:'), ''),
```
brenelz, line 157: Is this inner replace doing anything?

tannerlinsley: It is now.

packages/start/src/client/createServerFn.ts
tannerlinsley: Yeah, this change produced some invalid filenames that especially windows won't like. Keeping it for now

packages/start/src/config/index.ts
tannerlinsley: I think we're good.

PR #2941, opened 2024-12-07
fix(create-router): fix the create router scripts

SeanCassiere:


SeanCassiere:
Nope, this is not the avenue.

PR #2942, opened 2024-12-07
feat(start): createIsomorphicFn

EskiMojo14:
see https://github.com/TanStack/router/discussions/2937 for a full brief of this idea.

Allows creating a function with different implementations for server and client.

```ts
import { getRequestURL } from "vinxi/http"

const getIsomorphicUrl = createIsomorphicFn()
  .server(getRequestURL)
  .client(() => new URL(window.location.href));

// server
import { getRequestURL } from "vinxi/http"

const getIsomorphicUrl = getRequestURL

// client
const getIsomorphicUrl = () => new URL(window.location.href)
```

Any implementations not provided will default to no-ops.
```ts
const serverOnlyFn = createIsomorphicFn().server(() => "foo");

// server
const serverOnlyFn = () => "foo"

// client
const serverOnlyFn = () => {}
```

Not providing any implementations is allowed, but a warning will be logged as this is likely a mistake.
```ts
const noImpl = createIsomorphicFn()

// server
const noImpl = () => {}

// client
const noImpl = () => {}

// error logged during build (open to rewording)
// 
// createIsomorphicFn called without a client or server implementation! 
// This will result in a no-op function. Variable name: noImpl
```

Implementations should accept the same arguments, but can return different results.

### DIFF
--- a/e2e/start/basic/app/routes/isomorphic-fns.tsx
+++ b/e2e/start/basic/app/routes/isomorphic-fns.tsx
@@ -1,0 +1,61 @@
+import { createFileRoute } from '@tanstack/react-router'
+import { createIsomorphicFn, createServerFn } from '@tanstack/start'
+import { useState } from 'react'
+
+const getEnv = createIsomorphicFn()
+  .server(() => 'server')
+  .client(() => 'client')
+const getServerEnv = createServerFn().handler(() => getEnv())
+
+const getEcho = createIsomorphicFn()
+  .server((input: string) => 'server received ' + input)
+  .client((input) => 'client received ' + input)
+const getServerEcho = createServerFn()
+  .validator((input: string) => input)
+  .handler(({ data }) => getEcho(data))
+
+export const Route = createFileRoute('/isomorphic-fns')({
+  component: RouteComponent,
+})
+
+function RouteComponent() {
+  const [results, setResults] = useState<Partial<Record<string, string>>>()
+  async function handleClick() {
+    const env = getEnv()
+    const echo = getEcho('hello')
+    const [serverEnv, serverEcho] = await Promise.all([
+      getServerEnv(),
+      getServerEcho({ data: 'hello' }),
+    ])
+    setResults({ env, echo, serverEnv, serverEcho })
+  }
+  const { env, echo, serverEnv, serverEcho } = results || {}
+  return (
+    <div>
+      <button onClick={handleClick} data-testid="test-isomorphic-results-btn">
+        Run
+      </button>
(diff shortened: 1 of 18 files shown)


e2e/start/basic/app/routes/isomorphic-fns.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const echo = getEcho('hello')
+    const [serverEnv, serverEcho] = await Promise.all([
+      getServerEnv(),
+      getServerEcho({ data: 'hello' }),
+    ])
+    setResults({ env, echo, serverEnv, serverEcho })
+  }
+  const { env, echo, serverEnv, serverEcho } = results || {}
+  return (
+    <div>
```
datner, line 34: A cool application for isomorphic functions can be working with hooks.
```ts
const useAccount = createIsomorphicFn()
  .server(() => [getStorage('currentUser'), noop])
  .client(() => useState(null))


function RouteComponent() {
  const [account, setAccount] = useAccount()
  // ...
}
```
This should work as expected

packages/start-vite-plugin/src/index.ts
```diff
@@ -93,7 +94,9 @@ export function TanStackStartViteDeadCodeElimination(
 
       if (
         code.includes('createServerFn') ||
-        code.includes('createMiddleware')
+        code.includes('createMiddleware') ||
+        code.includes('serverOnly') ||
```
datner, line 98: slipped in? was missing?

datner (commented):
Such a robust and full-hearted PR, no corner cutting here 💪🏻
This is better work than I would do. Thank you for picking this up

e2e/start/basic/app/routes/isomorphic-fns.tsx
```diff
@@ -0,0 +1,61 @@
...
+  .client(() => 'client')
+const getServerEnv = createServerFn().handler(() => getEnv())
+
+const getEcho = createIsomorphicFn()
+  .server((input: string) => 'server received ' + input)
+  .client((input) => 'client received ' + input)
+const getServerEcho = createServerFn()
+  .validator((input: string) => input)
+  .handler(({ data }) => getEcho(data))
+
```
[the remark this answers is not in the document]
datner: ^might produce a false-positive because of the `createServerFn`, test without it

packages/start-vite-plugin/src/index.ts
EskiMojo14: seemed missing to me

PR #2963, opened 2024-12-09
fix(start): fix usage of `use()` with React 19

### LINKED ISSUE #2953: use of use() hook causes component to hang

Start

### Describe the bug

Using use() causes the component to hang, and never resolve the suspense boundary 

https://github.com/arackaf/tanstack-start-test-1/tree/feature/broken-use-repro

Clone the repo above - note the branch! -- feature/broken-use-repro
npm i 
Run TWO terminals. Do `npm run server` in one. Run `npm run dev` in the othder
hit http://localhost:3000/app/tasks

Navigating there as a client navigation runs fine. But refresh the page on that url. The outer page should render, but the suspense boundary should show indefinitely 

Image

### Expected behavior

See above 

### Screenshots or Videos

_No response_

### Platform

n/a

### Additional context

_No response_

schiller-manuel:
both router's `defer()` and React 19's `use()` hook were modifying the Promise object in a similar manner which caused a conflict.

See https://github.com/facebook/react/blob/v19.0.0/packages/react-reconciler/src/ReactFiberThenable.js#L162-L228 for how React 19 modifies the Promise object.

Both used an additional `status` property to track the Promise status, but router was using the value "success" and React was using the value "fulfilled" to mark a fulfilled Promise.

This change stores the router specific Promise tracking values using a symbol to avoid any potential collision.

fixes #2953


packages/react-router/src/awaited.tsx
```diff
@@ -14,38 +14,36 @@ export function useAwaited<T>({
...
           router.options.errorSerializer?.serialize ?? defaultSerializeError
-        )(promise.error),
+        )(promise[TSR_DEFERRED_PROMISE].error),
         __isServerError: true,
       }
     }
   }
-
-  return [promise.data as any, promise]
+  console.log('useAwaited', promise[TSR_DEFERRED_PROMISE])
```
AhmedBaset, line 45: this should be removed

packages/react-router/src/awaited.tsx
```diff
@@ -68,5 +66,7 @@ function AwaitInner<T>(
   },
 ): React.JSX.Element {
   const [data] = useAwaited(props)
+  console.log('AwaitInner', data)
```
AhmedBaset, line 69: same here

PR #2993, opened 2024-12-12
feat: Agnostic Server Fn Plugin

tannerlinsley:


packages/server-functions-vite-plugin/README.md
```diff
@@ -0,0 +1,5 @@
+<img src="https://static.scarf.sh/a.png?x-pxid=d988eb79-b0fc-4a2b-8514-6a1ab932d188" />
+
+# TanStack Start Vite Plugin
```
schiller-manuel, line 3: looks like this was copy / pasted and needs to be updated

packages/server-functions-vite-plugin/package.json
```diff
@@ -0,0 +1,84 @@
+{
+  "name": "@tanstack/server-functions-plugin",
+  "version": "1.87.3",
+  "description": "Modern and scalable routing for React applications",
+  "author": "Tanner Linsley",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/TanStack/router.git",
+    "directory": "packages/server-functions-plugin"
```
schiller-manuel, line 10: actual folder name is "server-functions-vite-plugin"

packages/server-functions-vite-plugin/src/ast.ts
```diff
@@ -0,0 +1,20 @@
+import { parse } from '@babel/parser'
+import type { ParseResult } from '@babel/parser'
+
+export type ParseAstOptions = {
```
schiller-manuel, line 4: we should really move this thing into a common package. we have this everywhere...

packages/server-functions-vite-plugin/tests/compiler.test.ts
```diff
@@ -0,0 +1,478 @@
...
+        return 'hello'
+      }
+    `
+
+    const client = compileServerFnClient({ ...clientConfig, code })
+    expect(client.compiledCode.code).toMatchInlineSnapshot(`
+      "import { createClientRpc } from "my-rpc-lib";
+      const fn = createClientRpc({
+        filename: "test.ts",
+        functionId: "test--fn_1"
```
schiller-manuel, line 62: why does this have an _1 suffix, but the `useServer` above does not?

packages/server-functions-vite-plugin/src/compilers.ts
```diff
@@ -0,0 +1,336 @@
...
+  > = {}
+  const counts: Record<string, number> = {}
+
+  babel.traverse(ast, {
+    Program: {
+      enter(programPath) {
+        const recordServerFunction = (nodePath: babel.NodePath) => {
+          const fnName = findNearestVariableName(nodePath)
+
+          const baseLabel = makeFileLocationUrlSafe(
```
schiller-manuel, line 151: what if you have `foo.tsx` and `foo.ts`? since you strip away the extension, the baseLabel could collide

packages/start/src/server-handler/index.tsx
```diff
@@ -23,28 +26,44 @@ async function handleServerAction(event: H3Event) {
 export async function handleServerRequest(request: Request, _event?: H3Event) {
   const method = request.method
   const url = new URL(request.url, 'http://localhost:3000')
+  // extract the serverFnId from the url as host/_server/:serverFnId
+  // Define a regex to match the path and extract the :thing part
+  const regex = /\/_server\/([^/?#]+)/
```
SeanCassiere, line 31: This base of `"_server"` needs to take in the inject environment variable of `"TSS_SERVER_BASE"` like its being done in `packages/start/src/client-runtime/getBaseUrl.tsx`.

packages/start/src/server-handler/index.tsx
```diff
@@ -23,28 +26,44 @@ async function handleServerAction(event: H3Event) {
...
 
   if (process.env.NODE_ENV === 'development')
-    console.info(`ServerFn Request: ${serverFnId} - ${serverFnName}`)
-  if (process.env.NODE_ENV === 'development') console.info()
+    console.info(`\nServerFn Request: ${serverFnId}`)
+
+  let action: Function
+  if (process.env.NODE_ENV === 'development') {
+    action = (await (globalThis as any).app
+      .getRouter('server')
```
SeanCassiere, line 52: Is this `.getRouter` call a Vinxi thing? or is it from Nitro?

Best to leave a comment here with any reference links that are available.

packages/server-functions-vite-plugin/src/index.ts
```diff
@@ -0,0 +1,246 @@
+import { fileURLToPath, pathToFileURL } from 'node:url'
+
+import { mkdirSync, readFileSync, writeFileSync } from 'node:fs'
```
SeanCassiere, line 3: Can we take inspiration from Astro here and propagate the system-reliant node modules like "fs" and "path" from the root config? https://github.com/withastro/astro/blob/98f9e833a5610cbd6f966a827451318435d31483/packages/astro/src/core/config/config.ts#L95

We've had issues opened in the past where the use of a certain system-specific node module for items like "path" and "fs" have broken things like the reading of the client-manifest.
\+ for the future (if we want to), it would let us customize this or use a dependency injection strategy based on the user's system should it be required.

---

Also, just a preference thing on my part, perhaps, we can just rely on synchronous calls? instead of mixing between sync and async. For me, the blocking nature of sync calls makes more sense for these plugins, since they often dependant on a previous plugin/step and can be the decider for the plugin after the current one.

packages/server-functions-vite-plugin/src/index.ts
```diff
@@ -0,0 +1,246 @@
...
+        onDirectiveFnsById: (d) => {
+          // When directives are compiled, save them so we
+          // can create a manifest
+          Object.assign(directiveFnsById, d)
+        },
+      }),
+      (() => {
+        let serverFunctionsManifest: Record<string, DirectiveFn>
+        return {
+          name: 'tanstack-start-server-fn-vite-plugin-build-',
```
SeanCassiere, line 155: Remove trailing hyphen (`"-"`).

packages/server-functions-vite-plugin/src/ast.ts
tannerlinsley: Let's do it later

packages/server-functions-vite-plugin/tests/compiler.test.ts
```diff
@@ -0,0 +1,478 @@
...
+    functionId: ${JSON.stringify(opts.functionId)},
+  })`,
+}
+
+const serverConfig = {
+  root: './test-files',
+  filename: 'test.ts',
+}
+
+describe('server function compilation', () => {
```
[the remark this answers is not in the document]
tannerlinsley: Not yet.

packages/server-functions-vite-plugin/src/index.ts
tannerlinsley: I would love to do what Astro is doing, but what I'm doing here is at least a feature parity, good enough for now.

packages/start/src/config/index.ts
```diff
@@ -204,31 +218,42 @@ export function defineConfig(
...
+                ...(viteConfig.userConfig.optimizeDeps || {}),
+                ...(ssrViteConfig.userConfig.optimizeDeps || {}),
+                // include: ['@tanstack/start/server-runtime'],
+              },
+            }),
+            TanStackStartVitePlugin({
+              env: 'server',
+            }),
+            TanStackServerFnsPlugin.ssr,
+            TanStackRouterVite({
```
SeanCassiere, line 240: When using the output from `@tanstack/router-plugin`, there's a toggle named [`enableRouteGeneration`](https://github.com/TanStack/router/blob/40f9f49e4b1648884da3aa1addbe267194d938ee/packages/router-plugin/src/core/router-generator-plugin.ts#L70) (that defaults to `true`) which controls whether the router-generator is run thereby running the file-system reads and writes. It has no bearing on the automatic code splitting.

Would it make sense to only have this flag active on a single vinxi router? and therefore disabled on the others to save having duplicated reads and writes on the user's machine.

tannerlinsley:
Absolutely

Tanner Linsley
TanStack LLC

packages/start-plugin/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      if (
+        !includesToken
+        // includesEitherFunc
+        // /node_modules/.test(id)
+      ) {
+        return null
+      }
+
+      if (code.includes('@react-refresh')) {
```
SeanCassiere, line 55: Do we still need this?

packages/start/src/server-handler/index.tsx
```diff
@@ -12,40 +12,108 @@ import {
...
+  if (!action) {
+    console.log('serverFnManifest', serverFnManifest)
+    console.log('fnModule', fnModule)
+    throw new Error(
+      `Server function module export not resolved for serverFn ID: ${serverFnId}`,
+    )
+  }
 
   const response = await (async () => {
     try {
```
SeanCassiere, line 119: This may be outside the scope of this PR, but since I noticed it, I figured I'd bring it up.

Currently, here we directly importing `defaultTransformer` from `@tanstack/react-router` and consuming. What happens when the user has declared the custom transformer like SuperJSON ([as mentioned here in the docs](https://tanstack.com/router/latest/docs/framework/react/guide/ssr#data-transformers))?

Perhaps this can be solved since we require the user to pass in a `createRouter` function in the `app/routes/ssr.tsx` file which can be used here in the server 🤔... Question is, are there any changes that need to happen on the client fetcher...

packages/start-plugin/src/index.ts
tannerlinsley: Yeah. It ensures that the plugins are in the right order.

PR #2994, opened 2024-12-12
fix(react-router): allow union when assigning

chorobin:


GiovaniGuizzoYUB:
This PR has broken my application's search params. I have a param A that's optional from the root route and a param B that's also optional from a page. It essentially makes the page's search param types to be something like this:

`{ A: {}; } | { B: string; } | { A: {}; B: string; } | undefined`

This type is basically unusable and TypeScript throws an error when I try to use search param B:

`Property 'B' does not exist on type '{ A: {}; }'`

PR #2996, opened 2024-12-12
feat(react-router): implement hash change "Scroll Into View" option

joshuaslate:
I am working on an API documentation page that has a number of anchor links. The sidebar contains `Link` elements that expand/collapse sections based on if its hash is active. It works like a table of contents. I also update the hash based on the user's scroll position. If they've scrolled past an anchor, it updates the hash. The issue I was hitting is that updating the hash with `navigate` or a number of other methods actually runs `scrollIntoView`, which takes over the user's scrolling in these cases.

The change I'm proposing is to add an option to `navigate` and `Link` that will allow the consumer to choose to opt out of the default `scrollIntoView` behavior, or even allow them to customize the scroll behavior if they would like to (e.g., for smooth scrolling instead of instant).

Discussed in Discord here: https://discord.com/channels/719702312431386674/1316559407919796334

Another relevant discussion: https://github.com/TanStack/router/discussions/1155

### DIFF
--- a/docs/framework/react/api/router/RouterType.md
+++ b/docs/framework/react/api/router/RouterType.md
@@ -93,6 +93,7 @@ Commits a new location object to the browser history.
     location: ParsedLocation & {
       replace?: boolean
       resetScroll?: boolean
+      hashChangeScrollIntoView?: boolean | ScrollIntoViewOptions
       ignoreBlocker?: boolean
     },
   ) => Promise<void>
@@ -112,6 +113,12 @@ Commits a new location object to the browser history.
     - Optional
     - Defaults to `true` so that the scroll position will be reset to 0,0 after the location is committed to the browser history.
     - If `false`, the scroll position will not be reset to 0,0 after the location is committed to history.
+  - `hashChangeScrollIntoView`
+    - Type: `boolean | ScrollIntoViewOptions`
+    - Optional
+    - Defaults to `true` so the element with an id matching the hash will be scrolled into view after the location is committed to history.
+    - If `false`, the element with an id matching the hash will not be scrolled into view after the location is committed to history.
+    - If an object is provided, it will be passed to the `scrollIntoView` method as options.
   - `ignoreBlocker`
     - Type: `boolean`
     - Optional
--- a/docs/framework/react/api/router/RouterType.md
+++ b/docs/framework/react/api/router/RouterType.md
@@ -93,7 +93,7 @@ Commits a new location object to the browser history.
     location: ParsedLocation & {
       replace?: boolean
       resetScroll?: boolean
-      hashChangeScrollIntoView?: boolean | ScrollIntoViewOptions
+      hashScrollIntoView?: boolean | ScrollIntoViewOptions
       ignoreBlocker?: boolean
     },
   ) => Promise<void>
@@ -113,12 +113,13 @@ Commits a new location object to the browser history.
     - Optional
     - Defaults to `true` so that the scroll position will be reset to 0,0 after the location is committed to the browser history.
     - If `false`, the scroll position will not be reset to 0,0 after the location is committed to history.
-  - `hashChangeScrollIntoView`
+  - `hashScrollIntoView`
     - Type: `boolean | ScrollIntoViewOptions`
     - Optional
     - Defaults to `true` so the element with an id matching the hash will be scrolled into view after the location is committed to history.
     - If `false`, the element with an id matching the hash will not be scrolled into view after the location is committed to history.
     - If an object is provided, it will be passed to the `scrollIntoView` method as options.
+    - See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) for more information on `ScrollIntoViewOptions`.
   - `ignoreBlocker`
     - Type: `boolean`
     - Optional
--- a/e2e/react-router/basic-file-based/src/routes/anchor.tsx
+++ b/e2e/react-router/basic-file-based/src/routes/anchor.tsx
@@ -61,7 +61,11 @@ function AnchorSection({ id, title }: { id: string; title: string }) {
 
   return (
     <div id={id} className="p-2 min-h-dvh">
-      <h1 className="font-bold text-xl pt-10" ref={elementRef}>
+      <h1
+        data-testId={`heading-${id}`}
+        className="font-bold text-xl pt-10"
+        ref={elementRef}
+      >
         {title}
         {hasShown ? ' (shown)' : ''}
       </h1>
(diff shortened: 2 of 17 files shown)


docs/framework/react/api/router/RouterType.md
```diff
@@ -112,6 +113,12 @@ Commits a new location object to the browser history.
     - Optional
     - Defaults to `true` so that the scroll position will be reset to 0,0 after the location is committed to the browser history.
     - If `false`, the scroll position will not be reset to 0,0 after the location is committed to history.
+  - `hashChangeScrollIntoView`
```
schiller-manuel, line 116: we should also add a `defaultHashChangeScrollIntoView` (or whatever this thing will be named finally) into router (defaults to `true`)

examples/react/basic/src/main.tsx
```diff
@@ -82,6 +86,179 @@ function IndexComponent() {
   )
 }
 
+export const anchorRoute = createRoute({
```
schiller-manuel, line 89: move this into `e2e/react-router/basic-file-based` and add a new `scrollIntoView.spec.ts` test (test name just a proposal)

packages/react-router/src/router.ts
```diff
@@ -717,6 +717,7 @@ export class Router<
     Math.random() * 10000000,
   )}`
   resetNextScroll = true
+  scrollOnNextHashChange = true
```
schiller-manuel, line 720: remove this

docs/framework/react/api/router/RouterType.md
joshuaslate: I added it to the router options and implemented it as I saw some of the other default options. Thanks!

examples/react/basic/src/main.tsx
joshuaslate: Done, thanks!

packages/react-router/src/router.ts
joshuaslate: Done, thanks!

e2e/react-router/basic-file-based/tests/scrollIntoView.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+test.beforeEach(async ({ page }) => {
+  await page.goto('/anchor')
+})
+
+// Testing the `Link` component with the `hashChangeScrollIntoView` prop
+test('Navigating via anchor `Link` with default hash scrolling behavior', async ({
+  page,
+}) => {
+  await expect(
+    page.getByRole('heading', { name: 'Default Anchor' }),
```
schiller-manuel, line 12: please use getByTestId instead of getByRole (this needs a `data-testId` prop on the element you want to find

joshuaslate: Updated, thanks.

schiller-manuel:
this looks great. let's wait for @SeanCassiere to review

schiller-manuel (approved):
LGTM

packages/react-router/src/router.ts
```diff
@@ -313,6 +313,14 @@ export interface RouterOptions<
    * @link [API Docs](https://tanstack.com/router/latest/docs/framework/react/api/router/RouterOptionsType#defaultviewtransition-property)
    */
   defaultViewTransition?: boolean | ViewTransitionOptions
+  /**
+   * The default `hashChangeScrollIntoView` a route should use if no hashChangeScrollIntoView is provided while navigating
+   *
+   * See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) for more information on `ScrollIntoViewOptions`.
```
SeanCassiere, line 319: The MDN link should also be placed in the markdown files.

- RouterOptionsType.md
- RouterType.md

e2e/react-router/basic-file-based/tests/scrollIntoView.spec.ts
```diff
@@ -0,0 +1,213 @@
+import { expect, test } from '@playwright/test'
```
SeanCassiere, line 1: Could we rename the file to `scroll-into-view.spec.ts`

e2e/react-router/basic-file-based/src/routes/anchor.tsx
```diff
@@ -0,0 +1,225 @@
...
+            if (!toHash) {
+              return
+            }
+
+            const hashChangeScrollIntoView = withScroll
+              ? ({
+                  behavior: formData.get('scrollBehavior') ?? 'instant',
+                  block: formData.get('scrollBlock') ?? 'start',
+                  inline: formData.get('scrollInline') ?? 'nearest',
+                } as ScrollIntoViewOptions)
```
SeanCassiere, line 121: Please use `satisfies` here.

SeanCassiere (approved):
I'm not at home at the moment, so I'll have to check the sandbox locally at home tomorrow.

I'd also like us to think a bit more on the property name. The name `hashChangeScrollIntoView` just feels a bit off. Maybe even `hashScrollIntoView`?

Approving for now, but I'd like to get a chance to play with the sandbox before merging in.

joshuaslate:
I believe I've addressed all of the feedback that's come in so far, please let me know if anything else comes up. I added the `/anchor` route to the `basic-file-based` example for easy manual testing in addition to the e2e tests.

packages/react-router/src/router.ts
```diff
@@ -1838,6 +1846,9 @@ export class Router<
         }
       }
 
+      nextHistory.state.__hashScrollIntoViewOptions =
+        hashScrollIntoView ?? this.options.defaultHashScrollIntoView ?? true
```
SeanCassiere, line 1850: Would this (the `?? true`) break any existing apps?

joshuaslate: These options are new, so the value will default to true if they aren't provided.

I read on MDN for scrollIntoView that true is the default argument value, so passing true by default should maintain the behavior that exists today. I don't believe this is a breaking change.

I'm not at my computer right now, but I believe I added a test case and Link on the example that use the default behavior, and it seems to function the same as the released version does for me.

SeanCassiere (approved):
Tried out the sandbox and it looks good.

@schiller-manuel OK to merge from my side.

joshuaslate:
I was investigating the test failure from scroll restoration and found that if browser (not router)-controlled navigation happens, the `router.state.location.state.__hashScrollIntoViewOptions` option isn't being set, because it doesn't run through `commitLocation`.

I could add another nullish coalescing operator to set the value if it's undefined, like so:
```
if (typeof document !== 'undefined' && (document as any).querySelector) {
        const hashScrollIntoViewOptions =
          router.state.location.state.__hashScrollIntoViewOptions ?? true

        if (hashScrollIntoViewOptions && router.state.location.hash !== '') {
          const el = document.getElementById(router.state.location.hash)
          if (el) {
            el.scrollIntoView(hashScrollIntoViewOptions)
          }
        }
      }
```

This fixes the test and is working for me in the scroll restoration sandbox.

@SeanCassiere, does that sound ok to you?

SeanCassiere:
> @SeanCassiere, does that sound ok to you?

Worth giving it a try.
I'll also be investigating what's going on here.

SeanCassiere (approved):

SeanCassiere:
cc @schiller-manuel

PR #2998, opened 2024-12-12
fix(react-router): use correct type for TRouter generic

squirly:
I think this is a typo.

There are many instances with `TRouter extends AnyRouter`: https://github.com/search?q=repo%3ATanStack%2Frouter+%22TRouter+extends+AnyRouter%22&type=code

There are a few other instances with `TRouter extends RegisteredRouter` which should probably be audited: https://github.com/search?q=repo%3ATanStack%2Frouter+%22TRouter+extends+RegisteredRouter%22&type=code.


schiller-manuel:
I think this is a type performance thing, but let's ask @chorobin

PR #2999, opened 2024-12-13
fix(history): browser transitions being skipped for `router.history` navigation calls

### LINKED ISSUE #2983: View Transition API not working for `router.history` navigation calls

Router

### Describe the bug

When navigating through browser history manually using the Tanstack Router `BrowserHistory` API, native browser transitions (from the new View Transitions API) are not happening. They do occur when using the native `window.history` API. 

Example use case would be a "Back" button that navigates the user back in history. 

https://stackblitz.com/edit/github-axehctcy

1. Load the example app on the "Home" page
2. Navigate to the "About" page (note the browser default fade transition)
3. Click the "Back to Previous Page with router.history.back()" button text
4. Navigation will have no transition

### Expected behavior

Navigation should use the same transition that navigating back with the window.history.back() or native browser back button uses. Notice the other button in the example does transition.

### Screenshots or Videos

_No response_

### Platform

- OS: Mac OS
- Browser: Chrome
- Version: 131.0.6778.109

### Additional context

The `notify()` call in the browser history source appears to override or cancel out the default transition animation. Is it possible that we do not need to call browser history subscribers in some history events, since native browser events like the user clicking the back button do not do so either?

scarabcoder:
Fixes #2983 

When using `back()`, `forward()` and `go()` on `router.history` for the browser history implementation, native transitions are not occurring.

This is caused by `notify()` being called twice, once for the `back()` from the default history implementation and again for when the window event listener is called by the browser for a navigation event.

This PR fixes it by moving calls for `notify()` to the implementation for the history API, so that the browser history implementation does not call `notify()` (instead relying on the event subscriber), while the memory implementation calls `notify()` on every history navigation call.

(fix suggested by @schiller-manuel )

### DIFF
--- a/packages/history/src/index.ts
+++ b/packages/history/src/index.ts
@@ -119,32 +119,27 @@ export function createHistory(opts: {
       state = assignKey(state)
       tryNavigation(() => {
         opts.pushState(path, state)
-        notify()
       }, navigateOpts)
     },
     replace: (path, state, navigateOpts) => {
       state = assignKey(state)
       tryNavigation(() => {
         opts.replaceState(path, state)
-        notify()
       }, navigateOpts)
     },
     go: (index, navigateOpts) => {
       tryNavigation(() => {
         opts.go(index)
-        notify()
       }, navigateOpts)
     },
     back: (navigateOpts) => {
       tryNavigation(() => {
         opts.back()
-        notify()
       }, navigateOpts)
     },
     forward: (navigateOpts) => {
       tryNavigation(() => {
         opts.forward()
-        notify()
       }, navigateOpts)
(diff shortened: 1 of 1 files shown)


schiller-manuel:
we have another PR open that touches history

https://github.com/TanStack/router/pull/1790

this has an `ignoreNextPop` flag we can use to solve this issue here. 
so let's first merge #1790 and then fix the issue here

ferretwithaberet:
#3017 might fix the same issue, to be checked

schiller-manuel:
yes, https://github.com/TanStack/router/pull/3017 will fix this

PR #3016, opened 2024-12-15
feat(react-router): allow to preload route chunk

akhmadshin:
Adds ability to preload js chunk of the page


packages/react-router/src/router.ts
```diff
@@ -2666,11 +2636,12 @@ export class Router<
                             status: 'error',
                             isFetching: false,
                           }))
+                          return Promise.resolve();
                         }
 
                         // Last but not least, wait for the the components
                         // to be preloaded before we resolve the match
-                        await route._componentsPromise
+                        return route._componentsPromise
```
schiller-manuel, line 2644: why not `await` ?

packages/react-router/src/router.ts
```diff
@@ -2666,11 +2636,12 @@ export class Router<
                             status: 'error',
                             isFetching: false,
                           }))
+                          return Promise.resolve();
```
schiller-manuel, line 2639: why do we need this?

packages/react-router/src/router.ts
```diff
@@ -2821,6 +2791,40 @@ export class Router<
     this.clearCache({ filter })
   }
 
+  preloadRouteChunk = (route: AnyRoute) => {
```
schiller-manuel, line 2794: this should return `route._componentsPromise`

PR #3017, opened 2024-12-15
feat(react-router): add useCanGoBack

### LINKED ISSUE #2983: View Transition API not working for `router.history` navigation calls

tomrehnstrom:
fixes #2983

### DIFF
--- a/packages/history/src/index.ts
+++ b/packages/history/src/index.ts
@@ -8,7 +8,7 @@ export interface NavigateOptions {
 
 type SubscriberHistoryAction =
   | {
-      type: HistoryAction | 'ROLLBACK'
+      type: HistoryAction
     }
   | {
       type: 'GO'
@@ -51,17 +51,12 @@ export interface ParsedPath {
 
 export interface HistoryState {
   key?: string
+  index: number
 }
 
 type ShouldAllowNavigation = any
 
-export type HistoryAction =
-  | 'PUSH'
-  | 'POP'
-  | 'REPLACE'
-  | 'FORWARD'
-  | 'BACK'
-  | 'GO'
+export type HistoryAction = 'PUSH' | 'REPLACE' | 'FORWARD' | 'BACK' | 'GO'
 
 export type BlockerFnArgs = {
   currentLocation: HistoryLocation
@@ -107,7 +102,7 @@ export function createHistory(opts: {
   createHref: (path: string) => string
   flush?: () => void
   destroy?: () => void
-  onBlocked?: (onUpdate: () => void) => void
+  onBlocked?: () => void
   getBlockers?: () => Array<NavigationBlocker>
   setBlockers?: (blockers: Array<NavigationBlocker>) => void
 }): RouterHistory {
@@ -119,11 +114,8 @@ export function createHistory(opts: {
     subscribers.forEach((subscriber) => subscriber({ location, action }))
   }
 
-  const _notifyRollback = () => {
+  const _setLocation = () => {
     location = opts.getLocation()
-    subscribers.forEach((subscriber) =>
-      subscriber({ location, action: { type: 'ROLLBACK' } }),
-    )
   }
 
   const tryNavigation = async ({
@@ -149,7 +141,7 @@ export function createHistory(opts: {
           action: actionInfo.type,
         })
         if (isBlocked) {
-          opts.onBlocked?.(_notifyRollback)
+          opts.onBlocked?.()
           return
         }
       }
@@ -174,7 +166,10 @@ export function createHistory(opts: {
       }
     },
     push: (path, state, navigateOpts) => {
-      state = assignKey(state)
+      // This is a fallback for when updating the router and there are history entries without index
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      const currentIndex = location.state.index ?? 0
+      state = assignKey({ ...state, index: currentIndex + 1 })
       tryNavigation({
         task: () => {
           opts.pushState(path, state)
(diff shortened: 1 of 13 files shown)


packages/history/src/index.ts
```diff
@@ -174,7 +166,10 @@ export function createHistory(opts: {
       }
     },
     push: (path, state, navigateOpts) => {
-      state = assignKey(state)
+      // This is a fallback for when updating the router and there are history entries without index
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      const currentIndex = location.state.index ?? 0
+      state = assignKey({ ...state, index: currentIndex + 1 })
```
schiller-manuel, line 172: can we put this logic into `assigKey()`?

packages/history/src/index.ts
```diff
@@ -187,7 +182,10 @@ export function createHistory(opts: {
       })
     },
     replace: (path, state, navigateOpts) => {
-      state = assignKey(state)
+      // This is a fallback for when updating the router and there are history entries without index
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      const currentIndex = location.state.index ?? 0
+      state = assignKey({ ...state, index: currentIndex })
```
schiller-manuel, line 188: same as above, can we put this logic into `assigKey()`?

packages/history/src/index.ts
```diff
@@ -551,6 +568,7 @@ export function createMemoryHistory(
       entries[index] = path
     },
     back: () => {
+      console.log('back', index)
```
schiller-manuel, line 571: remove

packages/history/src/index.ts
```diff
@@ -51,17 +52,12 @@ export interface ParsedPath {
 
 export interface HistoryState {
   key?: string
+  index: number
```
schiller-manuel, line 55: shouldn't this be `index?: number` as the index is not set for the initial entry?

also, i think we should use a more ... obscure name to not collide with user data

e.g. `__TSR_index` ?

ferretwithaberet: It's always available after we run through `parseHref`, this is because, if the route has no state (should be valid only for initial route), we default it to `{ index: 0 }`. About the naming, sure, but by that logic, shouldn't the `key` also be renamed?

schiller-manuel: ideally, yes. but the `key` has been there since the beginning, but adding an `index` now could introduce a breaking change

packages/react-router/tests/router.test.tsx
```diff
@@ -1091,3 +1093,161 @@ describe('route id uniqueness', () => {
...
+  })
+
+  const router = createRouter({
+    routeTree: rootRoute.addChildren([indexRoute, postsRoute]),
+  })
+
+  return { router }
+}
+
+describe('history: History gives correct notifcations and state', () => {
```
schiller-manuel, line 1143:
```suggestion
describe('history: History gives correct notifications and state', () => {
```

packages/react-router/tests/router.test.tsx
```diff
@@ -1091,3 +1093,161 @@ describe('route id uniqueness', () => {
...
+        <React.Fragment>
+          <h1>Posts</h1>
+          <button onClick={() => navigate({ to: '/' })}>Index</button>
+        </React.Fragment>
+      )
+    },
+  })
+
+  const router = createRouter({
+    routeTree: rootRoute.addChildren([indexRoute, postsRoute]),
```
schiller-manuel, line 1137: we should run the same test for all history types

packages/history/src/index.ts
```diff
@@ -119,11 +118,9 @@ export function createHistory(opts: {
     subscribers.forEach((subscriber) => subscriber({ location, action }))
   }
 
-  const _notifyRollback = () => {
-    location = opts.getLocation()
-    subscribers.forEach((subscriber) =>
-      subscriber({ location, action: { type: 'ROLLBACK' } }),
-    )
+  const handleIndexChange = (action: SubscriberHistoryAction) => {
+    if (opts.notifyOnIndexChange ?? true) notify(action)
```
schiller-manuel, line 122: can you please always use `{ }` even for one liners?

packages/history/src/index.ts
```diff
@@ -375,9 +377,10 @@ export function createBrowserHistory(opts?: {
     }
   }
 
-  const onPushPop = () => {
+  // NOTE: this function can probably be removed
```
schiller-manuel, line 380: so can we remove it?

MTG2000:
Small question.

Is this new `__TSR_index` property in `HistoryState` expected to be manually set every time we set a value for the `state` property on `navigate`, `<Link>`, ...etc?

Cause after updating my router version I started getting this error:
```
navigate({ state: { foo: "bar" } })
^
"Type '{ foo: "bar" }' is not assignable to type 'true | NonNullableUpdater<HistoryState> | undefined'.
  Property '__TSR_index' is missing in type '{ key: string; }' but required in type 'HistoryState'.ts(2322)"
```

Should I manually assign it with some value (like `0`), or is there a better way?

schiller-manuel:
no this is meant to be automatically set

@ferretwithaberet we need to make it optional

ccapndave:
After updating I'm getting this error all over my code too.

```
 Property '__TSR_index' is missing in type '{ stackId: string; }' but required in type 'HistoryState'.
```

igor9silva:
Hi, sorry to bother.

Just wanted to say I love you guys. Your working is amazing.

PR #2778, opened 2024-11-16
test(start): e2e test to check for allowing a null return

SeanCassiere:
E2E test for #2776

PR #3048, opened 2024-12-20
fix(start): returning `null` from server functions

### LINKED ISSUE #2776: Start: createServerFn cannot return `null`

Router

### Describe the bug

All details are found in this discussion thread on Discord - https://discord.com/channels/719702312431386674/1306957125246914590

The crust of the matter is that you aren't able to return `null` from a `createServerFn` call.

```ts
const fn = createServerFn().handler(async () => {
    return null
})

await fn(); // undefined
```

https://discord.com/channels/719702312431386674/1306957125246914590

Create a server function returning `null` and see that at runtime it does not return the value, instead, it returns `undefined`.

### Expected behavior

The returned `null` value from the server function should correctly get piped down to the caller.

### Screenshots or Videos

_No response_

### Platform

N/A

### Additional context

_No response_

alakhpc:
Fixes #2776, pass through null returns from serverfns


schiller-manuel:
can you please cherry-pick this on top? https://github.com/TanStack/router/pull/2778

alakhpc:
I had to update the test from

```ts
getServerResult ?? '-'
```
to 
```ts
getServerResult !== undefined ? getServerResult : '-'
```

should be good now

e2e/start/basic/app/routes/-server-fns/allow-fn-return-null.tsx
```diff
@@ -0,0 +1,67 @@
...
+            getServerResult !== undefined ? getServerResult : '-',
+          )}
+        </span>
+      </p>
+      <p>
+        {`POST: $allow_return_null_postFn returns`}
+        <br />
+        <span data-testid="allow_return_null_postFn-response">
+          {JSON.stringify(
+            postServerResult !== undefined ? postServerResult : '-',
```
SeanCassiere, line 50: Maybe the `useState` values should be initialized with something else that's not falsy. Like `-` or something else. I don't like the fact that we are doing conditional checking here in the test.
This test component really should just be stringifying whatever the stored result is.

alakhpc: updated

packages/start/src/client/createServerFn.ts
```diff
@@ -365,7 +365,10 @@ const applyMiddleware = (
         context,
         sendContext,
         headers,
-        result: userResult?.result ?? (mCtx as any).result,
+        result:
```
SeanCassiere, line 368: I'm not going to lie. I still don't understand why this change fixes the issue. Since at runtime, both conditions should resolve the same thing. 
¯\_(ツ)_/¯

SeanCassiere (approved):

packages/start/src/client/createServerFn.ts
TkDodo: previously, when userResult.result was `null`, we retuned mCtx.result because ?? returns the fallback for both null and undefined. 

Now, since we check for undefined explicitly, null will be "passed through".

SeanCassiere: Omg... I totally thought that `??` only ever passed through when it was `undefined` 🤯.

SeanCassiere:
@alakhpc thank you for contributing this fix 🎉

PR #3053, opened 2024-12-21
refactor(react-router): expose `scrollBehavior` on the `ScrollRestoration` component

joshuaKnauber:
Exposes the scroll behavior option for the ScrollRestoration component.

Currently the scrollTo only receives the x and y which makes it animate smoothly. If we always want to force an instant jump to the last point we should be able to set the behavior.


SeanCassiere (approved):

SeanCassiere (changes requested):
@joshuaKnauber looks like your changing of the import ordering in the `packages/react-router/src/scroll-restoration.tsx` file has broken eslint. Please use the eslint suggestions and fix the ordering.

schiller-manuel:
please update docs

docs/framework/react/guide/scroll-restoration.md

docs/framework/react/guide/scroll-restoration.md
```diff
@@ -137,3 +137,20 @@ function Component() {
   )
 }
 ```
+
+## Scroll Behavior
+
+To control the scroll behavior when navigating between pages, you can use the `scrollBehavior` option. This allows you to make the transition between pages instant instead of a smooth scroll.
```
SeanCassiere, line 143:
```suggestion
To control the scroll behavior when navigating between pages, you can use the `scrollBehavior` option. This allows you to make the transition between pages instant instead of a smooth scroll. The global configuration of scroll restoration behavior has the same options as those supported by the browser, which are `smooth`, `instant`, and `auto` (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#behavior) for more information).
```

SeanCassiere (changes requested):

SeanCassiere (approved):

SeanCassiere:
@joshuaKnauber thank you for contributing this in 🎉

joshuaKnauber:
> @joshuaKnauber thank you for contributing this in 🎉

Thank you for reviewing and for all your guys work!

PR #3055, opened 2024-12-23
Feat server fn public paths

tannerlinsley:
Builds on #2993


thehouseisonfire:
Would there be a timeline for this?

PR #3082, opened 2024-12-27
fix(start): fixed route scripts filtering

3fuyang:
The filtering before is like:

```ts
// ...
select: (state) => {
  // Assume here:
  // `[[undefined], undefined]`
  const scriptsOfEachMatch = state.matches.map((match) => match.scripts)
  // Filtered:
  // `[[undefined]]`
  const truthyScriptsOfEachMatch = scriptsOfEachMatch.filter(Boolean)
  // Turns out to be:
  // `[undefined]`
  // which will cause `TypeError` when `undefined` gets destructured below
  const truthyScripts = truthyScriptsOfEachMatch.flat(1)

  return {
    scripts: truthyScripts.map(({ children, ...script }) => ({ // <- TypeError: Cannot destructure property 'children' of 'undefined' as it is undefined.
      // ...
    })),
  }
},
// ...
```

Which will cause error when specifying `scripts` like:

```ts
  head: () => ({
    // ...
    scripts: [undefined], // Passes type check but will panic at runtime
  }),
  // ...
})
```

This PR moves the `filter(Boolean)` after flattening. But I'm not sure if the following patterns are equivalent:

```js
arr.filter(Boolean).flat(1).filter(Boolean) // Semantically correct

arr.flat(1).filter(Boolean) // Less code
```

### DIFF
--- a/e2e/start/basic/app/routeTree.gen.ts
+++ b/e2e/start/basic/app/routeTree.gen.ts
@@ -15,6 +15,7 @@ import { Route as UsersImport } from './routes/users'
 import { Route as StatusImport } from './routes/status'
 import { Route as ServerFnsImport } from './routes/server-fns'
 import { Route as SearchParamsImport } from './routes/search-params'
+import { Route as ScriptsImport } from './routes/scripts'
 import { Route as PostsImport } from './routes/posts'
 import { Route as LinksImport } from './routes/links'
 import { Route as IsomorphicFnsImport } from './routes/isomorphic-fns'
@@ -63,6 +64,12 @@ const ServerFnsRoute = ServerFnsImport.update({
 const SearchParamsRoute = SearchParamsImport.update({
   id: '/search-params',
   path: '/search-params',
+  getParentRoute: () => rootRoute,
+} as any)
+
+const ScriptsRoute = ScriptsImport.update({
+  id: '/scripts',
+  path: '/scripts',
   getParentRoute: () => rootRoute,
 } as any)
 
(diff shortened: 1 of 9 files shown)


schiller-manuel:
we need an e2e test for this

3fuyang:
Hi @schiller-manuel, it seems the existing e2e are all related to examples, so I'm not sure where to put the e2e case for this. But I found the `start` package already includes unit tests for `scripts`, so I added it there instead. PTAL

schiller-manuel:
I would put an e2e test in this project: https://github.com/TanStack/router/tree/main/e2e/start/basic

add new routes for this setup and then access them via the playwright test and add some assertions

3fuyang:
Hi @schiller-manuel , thanks for the guidance!

I've added a new `/scripts` route and the corresponding case to the `start/basic` example. Here are a few tips for review:

- Two empty script files `script.js` and `script2.js` are added to `public` for testing.
- I have to include the added script files to `TSConfig.include` so that it does not conflict with the global ESLint configuration for typed linting.
- I found that [a case](https://github.com/TanStack/router/pull/3082/commits/523739ff5ee2a71badd1e3dd464caea55c44b414#diff-1be0197cd6a94732428eea7195281e42557356c49390484bc8a29126cdbcb0edL196) was marked with `only`, thus only this case gets evaluated when testing. I removed the `only` declaration and please let me know if this was intentional.
- The `invoking a server function with custom response status code` case fails when running e2e tests, so I modified the [assertion](https://github.com/TanStack/router/pull/3082/commits/a22446bf512adcdf70faf8c808b08acf3d8cabd9#diff-1be0197cd6a94732428eea7195281e42557356c49390484bc8a29126cdbcb0edL93) to make it pass.

BTW please let me know whether I should remove the UT added to the `start` package.

3fuyang:
Oops, the test seems to behave differently between headless and UI mode, ill check it

schiller-manuel:
very nice, thanks a lot!

PR #3083, opened 2024-12-27
Move `@tanstack/router-devtools` to `@tanstack/react-router-devtools`

birkskyum:
The current router-devtools package is specific to react.

This PR moves the react devtools to react-router-devtools, allowing for addition of e.g. solid-router-devtools, and for collection of shared functions in an agnostic core devtools package to the extend it makes sense in this context.

Since the `@tanstack/react-router-devtools` isn't published, this PR doesn't change the docs (.md files) or the examples/* folder which rely on specific published versions (rather than workspace packages)


schiller-manuel:
we would need to still have `router-devtools` package as an almost empty shell so people can continue to use it (and then deprecate it once we have `react-router-devtools` in place). we did the same for the router vite plugin

SeanCassiere (commented):
This may not be the best approach here. At the moment, this would be a 'no-go' since it'd break people currently using the `@tanstack/router-devtools` package with no fallback.

---

As a whole, however, it'd be worth asking @ardeora on his thoughts here, since there's potential for following Query's approach for the devtools as a whole.

If done so, we could have the following package structure, which I believe would be the better approach here.

- `router-devtools-core`
- `react-router-devtools`
- `router-devtools` (proxies everything from `react-router-devtools` with a console warning asking them to switch packages and commitment that'd we'd drop this package in v2)

schiller-manuel:
while we are at restructuring, we should also keep in mind start's devtools (hopefully coming soon @ardeora  :D)

birkskyum:
I have added a `router-devtools` package which is a re-export of `react-router-devtools`. The `router-devtools` is used in all /e2e and /examples, so make sure nothing is broken by this change.

birkskyum:
@SeanCassiere , right. I hope one of the other people can figure this, because I'm at my wits end on it. I've looked at how query does it, but it's using tsup /esbuild rather than vite / rollup. Also it's not likely that the setup from query is better at this, it might just be less strict, similar to if the tsconfig.json didn't cover the ./dist folder that complains here by adding an `include: ["src"]`. I went down a rabit hole with  dts /  https://api-extractor.com/ and the [bundledPackages](https://github.com/qmhc/vite-plugin-dts/issues/302#issuecomment-2581815879), but I couldn't make it give the right result. And there's the `@tanstack/config` with the [preserveModules](https://github.com/TanStack/config/blob/20bb895342d50f255e5b9e373f67f9ed8f965301/packages/config/src/vite/index.js#L102) too, which I could worry is also working against this attempt to inline the shared package.

Basically, the only solution I've been able to make work here, that doesn't carbon-copy the router-core into e.g. react-router putting us where we started, is to have a symlink doing the same thing. I.e. a link from react-router/src/core -> router-core/src, so that instead of the dependency on @tanstack/router-core, the react-router just imports `from './core/index'`. 🤷🏻‍♂️

birkskyum:
Superseded by
- #3636

PR #3086, opened 2024-12-28
Begin `solid-router` and `router-core`

Brendonovich:
I've copied `react-router` into `solid-router` and replaced most of the React APIs with their Solid equivalent, and also extracted a lot of code that doesn't rely on Solid to `router-core`, which we can later use in `react-router`.
`solid-router` still has a lot of issues (and seems to be running into https://github.com/solidjs/solid/issues/2352), but a high number of tests are already passing.


MengLinMaker:
The `render` function from `@solidjs/testing-library` does not update the router state, nor render the route related html elements.
This causes many unit tests to fail.

packages/solid-router/src/index.tsx
```diff
@@ -0,0 +1,377 @@
...
+  createLazyRoute,
+  createLazyFileRoute,
+} from './fileRoute'
+export type {
+  FileRoutesByPath,
+  FileRouteTypes,
+  LazyRouteOptions,
+} from './fileRoute'
+
+export * from '@tanstack/router-core'
```
birkskyum, line 41: I'm not sure this is great - it's a result of moving the ./history.ts file over

packages/router-core/src/routeInfo.ts
```diff
@@ -0,0 +1,24 @@
...
+    ? TAcc
+    : TChildren extends ReadonlyArray<any>
+      ? ParseRoute<TChildren[number], TAcc | TChildren[number]>
+      : ParseRoute<
+          TChildren[keyof TChildren],
+          TAcc | TChildren[keyof TChildren]
+        >
+  : TAcc
+
+export type ParentPath<TOption> = 'always' extends TOption
```
birkskyum, line 14: These functions can't be extracted this easily - they depend on the AnyRouter now:

https://github.com/TanStack/router/blob/255f31603d0aed10bfdfd25bf403524c6238c5ac/packages/react-router/src/routeInfo.ts#L62-L75

SeanCassiere:
@Brendonovich FYI on the `router-core` package present on `"main"` which has centralized some of the sharable parts of `packages/react-router`.

Brendonovich:
This is being continued on #3276, I'm happy to have got the ball rolling 😄

PR #3099, opened 2025-01-02
fix: make `to` required if `from` is not set

### LINKED ISSUE #2797: Typescript issue when navigating without passing the to

Router

### Describe the bug

Considering a route like:
/users/$userId/details
/users/$userId/posts
/dummy/$dummyId

If we have a component that is used in the users routes that knows how to navigate to a different user, keeping the current location (details or posts),
then we can use the navigate returned from useNavigate to do so. It was suggested not to provide a `to` to have this working properly, and it works but Typescript complaints:
* with `navigate({ params: { userId: value }});` it complains that `Object literal may only specify known properties`, and userId does not exist in type 
* with `navigate({ params: (prev) => ({...prev, userId: value })});` it complains that the` { userId: string } is not assignable to type never`

Using the `from` or the `to`, parameters solves the typescript issue, but we need to use a specific route otherwise we lose the current location,

One possibility is to use, but it might be overcomplicating the solution.
```
const matches = useMatches();
const to = matches[matches.length - 1].fullPath;

navigate({to, params: { userId: value });
```

This issue is a followup of the discussion on [Discord](https://discord.com/channels/719702312431386674/1263245142991831132)

https://stackblitz.com/edit/tanstack-router-xlb7ip?file=app%2Fcomponents%2FNextUserNavigator.tsx

Use navigate to change only one parameter keeping any sub-route.

### Expected behavior

A cleaner way to change only the parameter without TS complaints.

### Screenshots or Videos

_No response_

### Platform

- OS: Linux
- Browser: Chrome
- Version: 130

### Additional context

_No response_

chorobin:
Fixes #2797


aboqasem:
does this mean `to` must be a valid path? if not, would be possible to allow any string? because after upgrading we have to suppress typescript:

```ts
import { createFileRoute, redirect } from "@tanstack/react-router";

export const Route = createFileRoute("/ar/$")({
	beforeLoad: (opts) => {
        throw redirect({
            // @ts-expect-error -- if the path doesn't exist, 404 will be shown
			to: `/${opts.params._splat ?? ""}`,
			...opts,
		});
	},
});

```

PR #3106, opened 2025-01-03
Mocked example of Link without "from" behavior change

Cornally:
👋 

After stubbing an example, it makes sense as to why things behave the way they do — It just caught me off guard ([this](https://github.com/TanStack/router/compare/v1.81.9...v1.81.10#diff-5fd7a0f05758a78cf8b73c09e4d71d21af63fe3d9b312b0ee75b04041ef078e7R654-R663) changes existing behavior).

I have a layout-level component (a user navigation dropdown). My application spawn modals based on a search param, `modalId`. For example, I can be on `/_layout/route-b` and click a link in the user dropdown (mounts inside `_layout.tsx`) intended to stay on the current visible route with the only change being a modal spawned on top of the current content.

I was triggering modals by way of ```<Link to="." search={{(prev) => ({...prev, modalId: 'sign-out-confirmation' }) }} />```
This worked up until `v1.81.9`. As of `v1.81.10`, I'm passing an extra `from={undefined}` to retain the same behavior. Without this change, modals fall back to using the root.

### DIFF
--- a/examples/react/basic-file-based-codesplitting/src/linkWrapper.tsx
+++ b/examples/react/basic-file-based-codesplitting/src/linkWrapper.tsx
@@ -1,0 +1,9 @@
+import { Link, type LinkProps } from '@tanstack/react-router';  
+
+export function LinkWrapper({ className, children, search, to }: { className?: string; children: React.ReactNode; to: string, search: (prev: Record<string, string>) => Record<string, string> }) {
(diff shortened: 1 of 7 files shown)


schiller-manuel:
can you please check if this issue still persists in the latest version?

SeanCassiere:

PR #3107, opened 2025-01-03
fix(react-router): navigate if redirect is thrown in component

schiller-manuel:


packages/react-router/src/CatchBoundary.tsx
```diff
@@ -9,13 +12,23 @@ export function CatchBoundary(props: {
...
       onCatch={props.onCatch}
       children={({ error, reset }) => {
         if (error) {
+          if (isRedirect(error)) {
+            const redirect = router.resolveRedirect({
+              ...error,
+              _fromLocation: router.state.location,
+            })
+            return (
+              <Navigate {...redirect} replace={true} ignoreBlocker={true} />
```
SeanCassiere, line 29: What's the rationale for setting `replace=true` here? Is this something that's required to make it work?
Or is it purely for allowing the user to go from `1` to `-1` in a single back button click when considering the user being at position `1` in `[ -1 , 0 , 1 ]`?

Asking, because if this is not something required on our side to make it functional at runtime, then perhaps, we shouldn't be enforcing `replace=true` on our side forcing a history replacement. Maybe this should be something the user should opt-in to?

packages/react-router/tests/redirect.test.tsx
```diff
@@ -252,6 +252,84 @@ describe('redirect', () => {
       expect(await screen.findByText('Final')).toBeInTheDocument()
       expect(window.location.pathname).toBe('/final')
     })
+
+    test('when `redirect` is thrown in route component', async () => {
```
SeanCassiere, line 256: I think we should test the back button behaviour here since it'd likely be one of the few buttons that'd be front of mind for the end-user to consider immediately hitting next.

LadyBluenotes:
Closing this since the router's redirect shape and `CatchBoundary` flow have changed, so this branch no longer applies.

PR #3109, opened 2025-01-04
fix(router-generator) use explicit routes export by default

esjs:
There is an issue with Virtual File Router when [config provided as as path](https://tanstack.com/router/latest/docs/framework/react/guide/virtual-file-routes#configuration-via-the-tanstackrouter-plugin).

For some reason there is an explicit default export added
Screenshot 2025-01-04 at 12 11 09

an there is clearly no default export provided in original config:
```ts
import { index, rootRoute, route } from '@tanstack/virtual-file-routes';

export const routes = rootRoute('root.tsx', [
  index('index.tsx'),
  route('about-us', 'about-us/index.lazy.tsx'),
]);

```

---

I've found a similar [issue](https://github.com/privatenumber/tsx/issues/427) in `tsx` package repo, and looks like it has something to do with internal module resolution of NodeJS, see [minimal reproduction](https://stackblitz.com/edit/stackblitz-starters-sykrfb?file=index.js) provided in discussion there.

Looks like we can easily solve this problem by using explicit `routes` export from `routes.ts` file.

### DIFF
--- a/packages/router-generator/src/filesystem/virtual/getRouteNodes.ts
+++ b/packages/router-generator/src/filesystem/virtual/getRouteNodes.ts
@@ -108,7 +108,7 @@ async function getVirtualRouteConfigFromFileExport(
   }
 
   const virtualRouteConfig =
-    'default' in exports ? exports.default : exports.routes
+    'routes' in exports ? exports.routes : exports.default
 
   return virtualRootRouteSchema.parse(virtualRouteConfig)
 }


SeanCassiere (changes requested):
I'm not sure I understand the problem here. Is this an issue with our implementation? or a bug in `tsx`?

Currently, from my point of view, this looks to be a workaround for a bug in `tsx`.

I'd prefer prioritizing the `default` export over the named one to have some consistency in how configuration files are writen.

esjs:
@SeanCassiere yes, it's clearly not an issue on TanStack/router side.

I just thought that "fixing" this inconvenience would be a good thing, especially for new users like myself.

Feel free to close this issue if you don't agree with proposed changes.

SeanCassiere (approved):

PR #3112, opened 2025-01-05
feat(react-router): add NotFoundErrorData interface for improved type safety

Sheraff:
Related discussion thread: https://github.com/TanStack/router/discussions/3113. This PR implements *one of the possible solutions* mentioned in the discussion.

---

Here's a demo of this PR, throwing type-safe `notFound` errors: https://stackblitz.com/edit/tanstack-router-fyrpy35d?file=src%2Froutes%2Findex.tsx

--- 

Currently the `notFound()` function accepts a params object with the shape of `NotFoundError` that is then passed as props to the `notFoundComponent`.

Arbitrary data can be send through this mechanism with the `data?: any` key of `NotFoundError`, but this `any` prevents us from creating type-safe "not found" errors.

This PR proposes that users should be able to define their own `data` shape through the global augmentation of a new interface `NotFoundErrorData`. For example:
```ts
  interface NotFoundErrorData {
    data: {
      a: string
      b: string
    }
  }
}
```
```ts
  interface NotFoundErrorData {
    data: 'foo' | 'bar'
  }
}
```
```ts
  interface NotFoundErrorData {
    data:
      | { type: 'foo', a: number }
      | { type: 'bar', b: string }
  }
}
```

This PR uses this new interface in a way that will preserve `data?: any` as the default if it is not globally augmented:
```ts
data?: NotFoundErrorData extends { data: infer TData } ? TData : any
```

---

Closes #3113


EskiMojo14:
might be worth extending the existing Register interface already used for registering the router type, rather than adding a new interface for the same purpose

SeanCassiere (approved):
Please add this into the reference docs for `notFound`.

Sheraff:
> might be worth extending the existing Register interface already used for registering the router type, rather than adding a new interface for the same purpose

I would love a 2nd opinion on this. I'm definitely not against this and it does kinda make sense. But it's also not uncommon for libraries to expose multiple `interface`s  that can be extended, and by using a "dedicated" one, we're kinda scoping the concerns.

Again, it also does make complete sense to me to reuse the `Register`. So anybody wants to weigh in?

schiller-manuel:
cc @chorobin please have a look
especially whether we want this configurable per route

LadyBluenotes:
Closing this since `NotFoundError` has moved into `router-core`, so this branch conflicts and the React-only type surface no longer lines up with the shared API.

Added this as an issue #7774 if we want to revisit it

PR #3140, opened 2025-01-10
fix(start): preserve multiple values for same key in `FormData` serialization

EskiMojo14:
Also adds Error and FormData to the allowed list of "serializable" values in Typescript, and make createTransformer more helpful with its types.

### DIFF
--- a/e2e/start/basic/app/routes/-server-fns/serialize-formdata-fn-call.tsx
+++ b/e2e/start/basic/app/routes/-server-fns/serialize-formdata-fn-call.tsx
@@ -28,16 +28,9 @@ export const greetUser = createServerFn()
       pets: pets.map((pet) => pet.toString()),
     }
   })
-  .handler(
-    ({
-      data: {
-        name,
-        age,
-        pets: [pet1, pet2],
-      },
-    }) =>
-      `Hello, ${name}! You are ${age + testValues.__adder} years old, and your favorite pets are ${pet1} and ${pet2}.`,
-  )
+  .handler(({ data: { name, age, pets } }) => {
+    return `Hello, ${name}! You are ${age + testValues.__adder} years old, and your favorite pets are ${pets.join(',')}.`
+  })
 
 // Usage
 export function SerializeFormDataFnCall() {
@@ -52,7 +45,7 @@ export function SerializeFormDataFnCall() {
           <pre data-testid="expected-serialize-formdata-server-fn-result">
             Hello, {testValues.name}! You are{' '}
             {testValues.age + testValues.__adder} years old, and your favorite{' '}
-            pets are {testValues.pet1} and {testValues.pet2}.
+            pets are {testValues.pet1},{testValues.pet2}.
           </pre>
         </code>
       </div>
--- a/e2e/start/basic/app/routes/-server-fns/serialize-formdata-fn-call.tsx
+++ b/e2e/start/basic/app/routes/-server-fns/serialize-formdata-fn-call.tsx
@@ -4,6 +4,8 @@ import { createServerFn } from '@tanstack/start'
 const testValues = {
   name: 'Sean',
   age: 25,
+  pet1: 'dog',
+  pet2: 'cat',
   __adder: 1,
 }
 
@@ -14,19 +16,27 @@ export const greetUser = createServerFn()
     }
     const name = data.get('name')
     const age = data.get('age')
+    const pets = data.getAll('pet')
 
-    if (!name || !age) {
-      throw new Error('Name and age are required')
+    if (!name || !age || pets.length === 0) {
+      throw new Error('Name, age and pets are required')
     }
 
     return {
       name: name.toString(),
       age: parseInt(age.toString(), 10),
+      pets: pets.map((pet) => pet.toString()),
     }
   })
   .handler(
-    ({ data: { name, age } }) =>
-      `Hello, ${name}! You are ${age + testValues.__adder} years old.`,
+    ({
+      data: {
+        name,
+        age,
+        pets: [pet1, pet2],
+      },
+    }) =>
+      `Hello, ${name}! You are ${age + testValues.__adder} years old, and your favorite pets are ${pet1} and ${pet2}.`,
   )
 
 // Usage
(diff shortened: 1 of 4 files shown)


packages/react-router/src/transformer.ts
```diff
@@ -162,9 +160,14 @@ export type TransformerParse<T, TSerializable> = T extends TSerializable
     ? ReadableStream
     : { [K in keyof T]: TransformerParse<T[K], TSerializable> }
 
+export type DefaultSerializeable = Date | undefined | Error | FormData
```
EskiMojo14, line 163: to avoid this being missed in future, it *could* be extracted directly from the transformers array:
```ts
type DefaultTransformers = (typeof transformers)[number]

export type DefaultSerializeable = {
  [K in DefaultTransformers['key']]: Extract<
    DefaultTransformers,
    { key: K }
  > extends { stringifyCondition: (value: any) => value is infer T }
    ? T
    : never
}[DefaultTransformers['key']]
```
but it makes the compiled output all icky and i suspect the performance will be impacted slightly, so it's a tradeoff

EskiMojo14: you could also flip it, and check the transformers array *against* the desired types:
```ts
type DefaultSerializeableByFrequency = [undefined, Date, Error, FormData]

type TransformersFor<T extends Array<any>> = {
  [K in keyof T]: { stringifyCondition: (value: any) => value is T[K] }
}

const transformers = [...] as const satisfies TransformersFor<DefaultSerializeableByFrequency>

export type DefaultSerializeable = DefaultSerializeableByFrequency[number]
```
this isn't quite as nice for DX since it's manually specified instead of inferred, but the build output will be nicer (and it'll still complain if there's a mismatch)

packages/react-router/src/transformer.ts
```diff
@@ -80,11 +80,11 @@ export const defaultTransformer: RouterTransformer = {
...
 
-const createTransformer = <T extends string>(
-  key: T,
-  check: (value: any) => boolean,
-  toValue: (value: any) => any = (v) => v,
-  fromValue: (value: any) => any = (v) => v,
+const createTransformer = <TKey extends string, TFrom, TTo>(
+  key: TKey,
+  check: (value: any) => value is TFrom,
+  toValue: (value: TFrom) => TTo = (v) => v as never,
```
EskiMojo14, line 86: these being optional seems odd - they're used in all of the current transformers, and I can't think of a situation where i'd want one and not the other, or neither

packages/react-router/src/transformer.ts
SeanCassiere: Could you instead just add a comment at the top of the array of transformers asking any future adventurers' to add the type into this union?

e2e/start/basic/app/routes/-server-fns/serialize-formdata-fn-call.tsx
```diff
@@ -4,29 +4,40 @@ import { createServerFn } from '@tanstack/start'
...
     return {
       name: name.toString(),
       age: parseInt(age.toString(), 10),
+      pets: pets.map((pet) => pet.toString()),
     }
   })
-  .handler(async ({ data: { name, age } }) => {
-    return `Hello, ${name}! You are ${age + testValues.__adder} years old.`
-  })
+  .handler(
```
SeanCassiere, line 31: Please switch this to using explicit returns with their brackets.
Also, don't restructure the pets array, instead just join with a comma.

```js
.handler(async ({ data: { name, age, pets } }) => {
	return `Hello, ${name}! You are ${age + testValues.__adder} years old, and your favourite pets are ${pets.join(',')}.`
})
```

packages/react-router/src/transformer.ts
SeanCassiere: cc @chorobin @schiller-manuel

packages/react-router/src/transformer.ts
```diff
@@ -130,21 +130,19 @@ const transformers = [
...
+    (v): v is FormData => v instanceof FormData,
     // To
-    (v: FormData) => {
-      const entries: Record<string, any> = {}
-      v.forEach((value, key) => {
-        entries[key] = value
-      })
+    (v) => {
+      const entries: Record<string, Array<any>> = {}
+      v.forEach((value, key) => (entries[key] ??= []).push(value))
```
SeanCassiere, line 137: Is there anything that can be done here where all keys aren't by default an array? I'd rather switch it to an array **only** if multiple instances of the same `name` are encountered.

If you've got better reasoning for sticking with this approach, I'm open to pushback on this one.

SeanCassiere (changes requested):

packages/react-router/src/transformer.ts
EskiMojo14: just because it simplifies the logic, and it's only two extra characters when serialised. I can add the logic to switch it so you can see the difference

e2e/start/basic/app/routes/-server-fns/serialize-formdata-fn-call.tsx
EskiMojo14: can do, but why, out of curiosity?

EskiMojo14: the linting was complaining about `async` with no `await`, fwiw

packages/react-router/src/transformer.ts
EskiMojo14: updated

e2e/start/basic/app/routes/-server-fns/serialize-formdata-fn-call.tsx
EskiMojo14: updated, sans async keyword

SeanCassiere (approved):

e2e/start/basic/app/routes/-server-fns/serialize-formdata-fn-call.tsx
SeanCassiere: > can do, but why, out of curiosity?

Legibility mostly. For a lib, the more legible the better.

SeanCassiere:
Please see the CI type error.

EskiMojo14:
should be fixed - it's another case of the workspace version of Typescript being 5.7, but the tests requiring all source code work in 5.2 (which makes sense for compiled code, but means you miss out on useful advancements in TS when writing source code)

SeanCassiere (approved):
I'm fine with the changes being made here.

@schiller-manuel could you confirm the changes regarding the generics are OK?

packages/react-router/src/transformer.ts
schiller-manuel: Can you please choose different names than `TFrom` and `TTo` ? those are usually used in the navigation API for the `from` / `to` props

EskiMojo14: renamed

PR #3149, opened 2025-01-12
Experiment with using findReferencedIdentifiers to avoid extra aggressive DCE

EskiMojo14:
this is pretty messy currently, but the idea based on https://github.com/pcattori/babel-dead-code-elimination/issues/34#issuecomment-2585541436 is to find some way of getting the referenced identifiers *before* Start does any transformations, so it can use them when doing dead code elimination. this would mean that the only dead code that gets eliminated is any that wasn't there before Start.

I also have no idea how to test that a server side effect happened in Playwright, so that'll be a fun hurdle to cross.

from research i have basically found two "accepted" ways of communicating between plugins:

- attaching things to `options` in one plugin and getting them out in the other
- exposing methods on one plugin and finding the plugin from the other using buildStart

This PR currently uses the first, because buildStart seems to be receiving an empty object in my tests.

### DIFF
--- a/e2e/start/basic/app/routes/-server-fns/dead-code-fn-call.tsx
+++ b/e2e/start/basic/app/routes/-server-fns/dead-code-fn-call.tsx
@@ -1,0 +1,37 @@
+import { createServerFn } from '@tanstack/start'
+import { useState } from 'react'
+// by using this we make sure DCE still works - this errors when imported on the client
+import { getRequestHeader } from 'vinxi/http'
+
+function sideEffect() {
+  console.log('side effect')
+  return true
+}
+
+const serverFn = createServerFn().handler(() => {
+  // eslint-disable-next-line unused-imports/no-unused-vars
+  const test = sideEffect()
+  const header = getRequestHeader('X-Test')
+  return header
+})
+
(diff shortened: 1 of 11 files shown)


SeanCassiere:
@EskiMojo14 do you have any test files to be run through? With it, we can move this PR along.

EskiMojo14:
ignoring the very large amount of conflicts caused by the recent reshuffle of the packages - I honestly don't know the right way to test this. It relies on the interplay between the two vite plugins, meaning unit testing it the way the others do isn't really possible, and it's only demonstrable by adding some code that would normally get tree shaken out. The E2E test I've added proves it works, but I have no idea how to assert that a console log showed up on the server side.

schiller-manuel:
> The E2E test I've added proves it works, but I have no idea how to assert that a console log showed up on the server side.

could we instead of logging write to a file on the server in a side effect? and then in another server function return that previously written content?

schiller-manuel:
now that https://github.com/TanStack/router/pull/3387 is merged, recreating this PR would be great!

EskiMojo14:
remade and merged in #3422

PR #3155, opened 2025-01-13
fix(react-router): Pass SearchParamError#cause in validateSearch

KubaJastrz:
# What

Ensure that `SearchParamError#cause` contains the original parsing error thrown/returned by the `validateSearch` implementation configured by the user. Notable example: passing valibot schema directly without an adapter. 

```ts
validateSearch: object({ ... }) // when this failed, it did not contain the proper `cause` field on the error
```

# Why

It's useful to know what happened, because the error can be accessed with `Route.errorComponent` or `Route.onError`. Also for consistency with other implementations of `validateSearch`.

# How

1. Add missing `cause` to the `SearchParamError` thrown by `validateSearch`
2. Do not wrap instances of `SearchParamError` in another `SearchParamError`
3. Add tests for `~standard`, validator fn, validator obj

### DIFF
--- a/packages/react-router/tests/router.test.tsx
+++ b/packages/react-router/tests/router.test.tsx
@@ -19,6 +19,7 @@ import {
   createRouter,
   useNavigate,
 } from '../src'
+import type { StandardSchemaValidator } from '../src/validators'
 import type {
   AnyRoute,
   AnyRouter,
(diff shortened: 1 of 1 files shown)


schiller-manuel:
can you add a test please?

KubaJastrz:
Thanks for the suggestion. Adding the test for `~standard` parsing revealed another problem that would prevent my initial change from working. Apparently the `SearchParamError` thrown by `validateSearch` were wrapped in another `SearchParamError`, messing with the `cause` field.

PR #3171, opened 2025-01-14
refactor(router-core): move router-core into a separate package

birkskyum:
This is a preparation step for:
- #3086 

It moves pieces of shared logic from react-router to router-core, and imports it from there.

The exports in react-router/src/index.tsx are kept as is, by re-exporting functions from router-core, to avoid breaking changes.


SeanCassiere (commented):
I'm changing this PR's status to "draft" for now as there are still parts that haven't been moved over.

Also, leaving this note here to add the new package in to the `scripts/publish.js` script to distribute the `router-core` package.

SeanCassiere:
@schiller-manuel @chorobin any idea how this affects the global module declaration merging that we use for type-safety.

birkskyum:
> I'm changing this PR's status to "draft" for now as there are still parts that haven't been moved over.

@SeanCassiere let me know if there are some specific areas that I should look into - I do find it a bit challenging to evaluate what can and can't be shared, with only the react-router currently present as consumer of router-core.

birkskyum:
@SeanCassiere , it seems a React.JSX.Element made it's way to the router-core 

https://github.com/TanStack/router/pull/3171/commits/a2cdc677b279a65a44d477328d3b192b13df4b1b#diff-75e49878ce10e98f01a0e36eb7367c1580cd137288a0c53422e829751ade2c30R176-R180

SeanCassiere:
@schiller-manuel @chorobin any thoughts as to where this declaration merge type error is coming from?
https://cloud.nx.app/runs/W720HeqFWm/task/%40router-mono-simple%2Frouter%3Abuild

birkskyum:
@SeanCassiere , I find it helps to add `"@tanstack/router-core": "^1.97.1",` inside

examples/react/router-monorepo-react-query/packages/router/package.json

SeanCassiere:
> @SeanCassiere , I find it helps to add `"@tanstack/router-core": "^1.97.1",` to the deps of

The problem here is that it shouldn't require this. 

+ this'd be considered a breaking change.

birkskyum:
@SeanCassiere , right. I've looked at how query does it, but it's using tsup /esbuild rather than vite / rollup. I don't think the setup from query is better at this, it might just be less strict, similar to if the examples/../router/tsconfig.json had an `include: ["src"]` and thus didn't conflict with the output from the /dist folder.

I went down a rabit hole with  dts /  https://api-extractor.com/ and the [bundledPackages](https://github.com/qmhc/vite-plugin-dts/issues/302#issuecomment-2581815879), and I couldn't make it give the right result. Then there's the `@tanstack/config` with the [preserveModules](https://github.com/TanStack/config/blob/20bb895342d50f255e5b9e373f67f9ed8f965301/packages/config/src/vite/index.js#L102) too, which I could worry is also working against this attempt to inline the shared package.

Basically, the only solution I've been able to make work here, that doesn't carbon-copy the router-core into e.g. react-router putting us where we started, is to have a symlink doing the same thing. I.e. a link from react-router/src/core -> router-core/src, so that instead of the dependency on @tanstack/router-core, the react-router just imports `from './core/index'`.

SeanCassiere:
Since what's currently broken in this one, is just the module declaration (so a typescript issue really), once that fix goes through this one should be fine.

birkskyum:
I think basically we're facing a hoisting issue, similar to:

- https://github.com/nuxt/nuxt/issues/14146

birkskyum:
@SeanCassiere , this change to react-router vite.config will make the type declarations work as expected by resolving the files:

- https://github.com/TanStack/router/pull/3171/files#diff-6e75c2bdd7e92c4bb5ecf84ce748a2729c52a0400d6fb72f9ad46e3bde174b0a

For the vitest test.alias, it'll work for both:

`replacement: path.resolve(__dirname, '../router-core/src/index.ts')`

And
`replacement: path.resolve(__dirname, './node_modules/@tanstack/router-core/src/')`

birkskyum:
Since we don't have a test suite, the examples we can run manually that does use `"use server";` that we for a start can verify still works are:
- [ ] hackernews
- [ ] todomvc
- [ ] notes
- [ ] with-auth
- [ ] with-drizzle
- [ ] with-prisma

birkskyum:
As stated [here](https://github.com/nuxt/nuxt/issues/14146), users can either add `public-hoist-pattern[]=@tanstack/router-core` to .npmrc, or:
`Generated types are not fully functional without hoisted node_modules`

SeanCassiere (approved):

SeanCassiere:
@birkskyum thanks for your work on this one. I've pushed this to the team for review and we should be able to merge tomorrow.

---

In response to your comment [above](https://github.com/TanStack/router/pull/3171#issuecomment-2607171122), we'll wait to see the common practice going forward since this pattern is currently in use by TanStack Table as well, so we'll be to apply whatever changes to the TanStack ecosystem as whole when the "correct" way has been figured out.

birkskyum:
Wait, how did the CI just pass? Did the tsconfig change do the trick, making react-router an isolated module?

SeanCassiere:
> ... And the CI just passed?

One of your commits prior to https://github.com/TanStack/router/pull/3171/commits/538b8b9d3b6cc884c6ce52bdb8bb528565b83552 did the trick. All that remains were eslint errors.

Issue #3162, opened 2025-01-14
Cursor jumps to end of search params bound input


Nick-Lucas:
### Which project does this relate to?

Router

### Describe the bug

It's common to bind a Filter input to search params so that URLs can be shared, and a user should be able to type anywhere in the input and it behave predictably.

Current behaviour of this isn't predictable though, if you type into the end of an input it behaves fine, but if you type into the middle, the cursor jumps to the end.

I suspect this is because startTransition is called by default under the hood now, but there doesn't appear to be a way to disable this behaviour or any past github issue which describes a workaround or fix

### Your Example Website or App

Example project: search-validator-adapters

### Steps to Reproduce the Bug or Issue

1. Boot up the search-validator-adapters example project
2. Navigate to a tab (I like the zod one 😉)
3. Type some text into the input
4. Change cursor position to middle of the text
5. type 1 character
6. the cursor jumps to the end 😱

### Expected behavior

Typing anywhere in the input types where your cursor is and retains cursor position

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/22396c2b-7362-4d8d-9f6a-8baa4da6643e)

### Platform

- OS: Mac and Windows
- Browser: Chrome (tested)

Nick-Lucas:
I've added a failing test here. I'm probably not the right person to fix it simply because I understand why startTransition is now the default, but I can see this being a tricky one to balance the need for startTransition with the need for binding inputs successfully

schiller-manuel:
as the router state is outside of react, React seems to not be able to track the cursor position as it would with a local `useState`. 

This works around that:

```tsx
const Zod = () => {
  const search = Route.useSearch({
    select: (search) => search.search ?? '',
  })
  const navigate = useNavigate({ from: Route.fullPath })

  const [state, setState] = React.useState(search)
  return (
    <>
      <Header title="Zod" />
      <Content>
        <Search
          search={state}
          onChange={(v) => {setState(v); navigate({search: { search: v } })}}
        />
        <React.Suspense>
          <Users search={search} />
        </React.Suspense>
      </Content>
    </>
  )
}
```

Nick-Lucas:
So maybe the bugfix is the react-router package needs to have its own in-react cache which synchronises out to the router-core? Possibly with useOptimistic instead of useState?

Would you accept a PR for that?

Nick-Lucas:
Done some investigation today. `useSearch` and likely `useMatch` have had this bug at least as far back as v1.26.8

[...]

PR #3174, opened 2025-01-15
Add test case for cursor resetting to the end of inputs

Nick-Lucas:
Adds a failing test case for #3162 

I'm not sure about the solution, it's likely the default startTransition at fault but there doesn't seem to be a way to disable that via the public API, and it does make sense why that's a sensible default


Nick-Lucas:
Closing as the bug seems to have been resolved at some stage

PR #3181, opened 2025-01-16
fix(react-router): ensure fetch abort works correctly even with pendingComponent

### LINKED ISSUE #3014: AbortController signal is not called after `pendingMs`

Router

### Describe the bug

AbortController signal is not called after `pendingMs` and user leaves the route.

The following sandbox simulates a long data fetching of 10 seconds when entering `profile` route
for the `profile` route and a 5 seconds pendingMs, the signal doesn't get called after 5 seconds.

https://codesandbox.io/p/devbox/wm3ffd

1. Click on profile link (profile loader will simulate a 10 seconds request)
2. Wait for pendingComponent to render (`pendingMs`: 5 seconds for testing purposes)
3. Click on home link right after pendingComponent renders.
4. AbortController signal didn't got called and `profile` loader successfully fulfilled (should had been aborted as user left /profile route)

(abortController signal get's called just fine within `pendingMs` time)

### Expected behavior

When user leaves a route, the `abortController` property of the `loader`  should be called whenever there's a pending promise on it's loader.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 1.89.1

### Additional context

_No response_

leesb971204:
fixes #3014


schiller-manuel:

leesb971204:
> can you please add a test for this?

I’m sorry, but this is my first time writing test codes. Could you please explain which files I should write the test codes in and how I can execute them? Thank you so much!

leesb971204:
> can you please add a test for this?

I have added and written the test cases as requested. Since this is my first attempt at writing test cases, I am unsure if they have been written correctly. I would greatly appreciate it if you could kindly review them and provide your feedback. For your reference, it seems that the code below needs to be placed at the very top of the file in my local environment for the test cases to work properly. Thank you so much in advance for your help!
```typescript
/**
 * @vitest-environment jsdom
 */
import '@testing-library/jest-dom/vitest'
```

leesb971204:
> can you please add a test for this?

I apologize for the misunderstanding earlier. To clarify the issue, when pendingMs is 0, the abort function operates correctly if there is no pendingComponent. However, if a pendingComponent is present, the abort function does not work as expected. I have made modifications to address this behavior. Considering that this might be the intended functionality, I believe it may be necessary to revise the title of the PR accordingly.

schiller-manuel:
this does not work with preloads. hover over a link, let the preload start. then click, wait for the pending component, navigate away. preload is not cancelled.

leesb971204:
> this does not work with preloads. hover over a link, let the preload start. then click, wait for the pending component, navigate away. preload is not cancelled.


https://github.com/user-attachments/assets/7b13bede-1f8b-46d7-9fc8-dc4f9164f0ce

If the issue you mentioned about preload not working aligns with the behavior in the video, it seems unrelated to my work. I have reproduced this behavior in the latest main branch, and it does not appear to be a side effect of my changes. While I agree that preload(fetching **posts** in my video) should be aborted, the environment in which I worked had defaultPreload set to false, and I believe addressing this behavior should be treated as a separate issue. Therefore, I think that the work regarding preload not functioning properly should be addressed in a new PR. What are your thoughts on this?

PR #3191, opened 2025-01-18
fix: Fix scroll restoration

schiller-manuel:


packages/react-router/src/scroll-restoration.tsx
```diff
@@ -87,110 +190,84 @@ export function useScrollRestoration(options?: ScrollRestorationOptions) {
...
+  const getKey = props.getKey || defaultGetKey
+  const userKey = getKey(router.latestLocation)
+  const resolvedKey =
+    userKey !== defaultGetKey(router.latestLocation) ? userKey : null
+
+  return (
+    <script
+      suppressHydrationWarning
+      dangerouslySetInnerHTML={{
+        __html: `(${restoreScroll.toString()})(${JSON.stringify(storageKey)},${JSON.stringify(resolvedKey)});__TSR__.cleanScripts()`,
```
SeanCassiere, line 267: Perhaps this bit should be optionally checked since Start isn't guaranteed to be there?
```suggestion
        __html: `(${restoreScroll.toString()})(${JSON.stringify(storageKey)},${JSON.stringify(resolvedKey)});__TSR__?.cleanScripts?.()`,
```

schiller-manuel: it's also called `__TSR_SSR__` now

packages/react-router/src/Transitioner.tsx
```diff
@@ -87,74 +88,50 @@ export function Transitioner() {
...
-          }
-        }
-      }
     }
   }, [isAnyPending, previousIsAnyPending, router])
 
   return null
 }
+
+export function getLocationChangeInfo(routerState: {
```
SeanCassiere, line 127: Could this be exported out of the `router-core`?

Purely, so that we have less work when we eventually move more into the `router-core` package from `react-router` *(like the `Router` and `Route` classes, +others).

packages/react-router/src/ScrollRestoration.tsx
```diff
@@ -0,0 +1,54 @@
...
+  getCssSelector,
+  scrollRestorationCache,
+} from './scroll-restoration'
+import type { ScrollRestorationOptions } from './scroll-restoration'
+import type { ParsedLocation } from '@tanstack/router-core'
+
+/**
+ * @deprecated use createRouter's `scrollRestoration` option instead
+ */
+export function ScrollRestoration(_props: ScrollRestorationOptions) {
```
SeanCassiere, line 13: Is this considered a breaking change? Where do we draw the line on this one?

Like builds won't fail since the component is still being exported... But at the same-time, any existing scroll restoration functionality will break on the user/app facing experience.

packages/react-router/src/scroll-restoration.tsx
```diff
@@ -1,241 +1,315 @@
...
-
-type Cache = {
-  state: CacheState
-  set: (updater: NonNullableUpdater<CacheState>) => void
+export type ScrollRestorationOptions = {
+  getKey?: (location: ParsedLocation) => string
+  scrollBehavior?: ScrollToOptions['behavior']
 }
 
+export const storageKey = 'tsr-scroll-restoration-v3'
```
SeanCassiere, line 22:
```suggestion
// TODO: Bump `v1` to `v2` when performing the library major bump.
export const storageKey = 'tsr-scroll-restoration-v1-3'
```

And then in v2, change it to `tsr-scroll-restoration-v2-1`.

packages/react-router/src/Transitioner.tsx
tannerlinsley: Yep

rolftimmermans:
I found this change broke native browser scroll restoration in our app. This is because `setupScrollRestoration()` is called even when `scrollRestoration` is set to `false`. Is this intentional, or could there be a way to opt out entirely? I was happy with browser-native scroll restoration.

tannerlinsley:
It’s called because it also handles hash scrolling. The actual restoration logic internally only runs if it’s turned on.

rolftimmermans:
> It’s called because it also handles hash scrolling. The actual restoration logic internally only runs if it’s turned on.

Thanks for your response! Much appreciated.

I guess the issue is: if no restoration happens, the page is scrolled to the top (see https://github.com/TanStack/router/blob/4d74d55da848c69c58635b4f9b489a3b953bf8b0/packages/react-router/src/scroll-restoration.tsx#L155). This counteracts native scroll restoration. And it seems to be different behaviour than before this PR. Ideally I'd like a way to opt out of that altogether. Also OK if it means opting out of hash scrolling.

I'd be happy to create a PR for it. But of course only if you agree this is a valid option to provide. And what exactly the option should do – opt out of hash scrolling as well? or only opt out of the default scrolling to the top in case there is nothing to do?

mihkelmark:
This release also broke the scroll restoration in my app. 

After 30 minutes of reading the docs and trying to figure out how to have my app (with simple use cases) support the new version, I haven't succeeded. So I'm either thick or the documentation is not clear about this. Which is unlike the tanstack suite, so I'm guessing it's me. 

I'm reverting back to the old version for now. Reluctantly, but will have to do it to move forward.

tannerlinsley:
Would you have time to jump on a call soon to debug with me?

tannerlinsley:
Dm me in discord.

rolftimmermans:
For @mihkelmark or others in this thread; the reason things worked with native scroll restoration in the past was more or less accidental.
Architectural changes in @tanstack/react-router after 1.98 require scroll restoration to be configured. In my case this was a matter of adding:

```
const router = createRouter({
    scrollRestoration: true,
    // ...
})
```

And in the virtualized tables in our application:

```
const restoration = useElementScrollRestoration({getElement: () => window})
const virtualizer = useWindowVirtualizer({
  initialOffset: restoration?.scrollY ?? 0,
  // ...
})
```

That is all!

Returning `window` from `getElement()` was not possible before yesterday, but @tannerlinsley graciously fixed this when investigating this use case with `useWindowVirtualizer()`.

schiller-manuel:
@rolftimmermans we should add this to the docs !
want to provide a PR to the docs?

rolftimmermans:
> @rolftimmermans we should add the window part to the docs ! want to provide a PR to the docs? with a small example?

Sure! #3752

Issue #3193, opened 2025-01-19
ParsedHistoryState doesn't work with useLocation


guy-ecton:
### Which project does this relate to?

Router

### Describe the bug

After the change in https://github.com/TanStack/router/pull/3119
The following code throws error on `1.97.3` (didn't throw error on `1.95.1`):

```typescript
declare module '@tanstack/history' {
  interface HistoryState {
    test?: string
  }
}


//


const { test } = useLocation({
    select: (location) => location.state,
  })
```

```
Property 'test' does not exist on type 'ParsedLocation<{ [x: string]: unknown; }>'.
```

---
Also the documentation still says `HistoryState`:
https://github.com/TanStack/router/blob/18610d8ebf945a5fd4f44ee97435bec50fdcf054/docs/framework/react/api/router/ParsedLocationType.md?plain=1#L14

https://github.com/TanStack/router/blob/18610d8ebf945a5fd4f44ee97435bec50fdcf054/packages/react-router/src/location.ts#L9

### Your Example Website or App

-

### Steps to Reproduce the Bug or Issue

```typescript
declare module '@tanstack/history' {
  interface HistoryState {
    test?: string
  }
}


//


const { test } = useLocation({
    select: (location) => location.state,
  })
```

### Expected behavior

As with version `1.95.1` this shouldn't throw an error

### Platform

- Version: 1.97.3

ferretwithaberet:
You have to extend the `HistoryState` interface in the `@tanstack/react-router` module instead of in the `@tanstack/history` one. Good catch on the docs though, I will work on them now. Also, by chance, are you using `pnpm`, maybe the way it structures the `node_modules` folder is the reason why you cannot extend it in the `@tanstack/history` package.

guy-ecton:
When I try 
```typescript
declare module '@tanstack/react-router' {
  interface HistoryState {
    test?: string
  }
}
```

it breaks everything.
Like
```
src/App.tsx:11:10 - error TS2305: Module '"@tanstack/react-router"' has no exported member 'Navigate'.

11 import { Navigate, RouterProvider, createRouter } from '@tanstack/react-router'
```

Anyway, I found this (https://github.com/TanStack/router/discussions/284#discussioncomment-9341612) and will try to check my local setup (I use npm and Vite BTW).

ferretwithaberet:
Can you provide a full minimal reproduction? I created a completely new project and added the code above with the fix I mentioned and it works perfectly, I cannot reproduce your issue.

[...]

PR #3194, opened 2025-01-19
docs(router): fix docs related to parsed state

ferretwithaberet:
Fixes docs for `ParsedLocation` type, adds docs for `ParsedHistoryState` type and clarifies how `HistoryState` interface should be extended

`HistoryState` should be extended in the `@tanstack/react-router` package, as trying to extend it in `@tanstack/history` might error due to the way package managers like pnpm structure the `node_modules/` folder.

image


ferretwithaberet:
Fixes #3193

PR #3196, opened 2025-01-19
feat(router): add useCanGoForward

ferretwithaberet:


matthiasschwarz:
The `length` property of `RouterHistory` should return the total entries in the history. See [History.length](https://developer.mozilla.org/en-US/docs/Web/API/History/length).
Since `useCanGoForward` needs the index of the current entry in the history there could be a new `index` property in `RouterHistory` used instead.

ferretwithaberet:
> The `length` property of `RouterHistory` should return the total entries in the history. See [History.length](https://developer.mozilla.org/en-US/docs/Web/API/History/length). Since `useCanGoForward` needs the index of the current entry in the history there could be a new `index` property in `RouterHistory` used instead.

Rather, it needs the last known index that was still in the app. My implementation is overriding the `length` property to only track the indexes from tanstack router instead of the whole browser tab history. I made it this way because I doubt we need to know the browser history length, we only need to know the router's length, and even then, when we need the browser history length we can use `window.history.length`.
I could try reimplementing it to store the last window history index that was in the app to session storage and use that instead, but I will wait for other opinions first.

nukeop:
Any chance this will get merged?

SeanCassiere:

PR #3216, opened 2025-01-22
fix: router stream injection

tannerlinsley:


packages/react-router/src/router.ts
```diff
@@ -456,19 +401,6 @@ export interface RouterOptions<
    * @link [Guide](https://tanstack.com/router/latest/docs/framework/react/guide/not-found-errors#default-router-wide-not-found-handling)
    */
   defaultNotFoundComponent?: NotFoundRouteComponent
-  /**
-   * The transformer that will be used when sending data between the server and the client during SSR.
-   *
-   * @link [API Docs](https://tanstack.com/router/latest/docs/framework/react/api/router/RouterOptionsType#transformer-property)
-   * @link [Guide](https://tanstack.com/router/latest/docs/framework/react/guide/ssr#data-transformers)
-   */
-  transformer?: RouterTransformer
```
SeanCassiere, line 465: At this current stage, the removal of the user-configurable `transformer` property would be a breaking change. 

If the argument is that since this has the most impact on Start and that it's fine to make these breaking changes, could we perhaps move these Start-specific options under their own key?

Something like this perhaps?

```ts
createRouter({
	// ...
	server: { // or framework / start
		serializer: {}
	}
})
```

---

docs/framework/react/guide/ssr.md
```diff
@@ -201,28 +201,32 @@ This pattern can be useful for pages that have slow or high-latency data fetchin
...
 
-The following example shows usage with [SuperJSON](https://github.com/blitz-js/superjson), however, anything that implements [`Router Transformer`](../api/router/RouterOptionsType.md#transformer-property) can be used.
+<!-- The following example shows usage with [SuperJSON](https://github.com/blitz-js/superjson), however, anything that implements [`Start Serializer`](../api/router/RouterOptionsType.md#serializer-property) can be used. -->
 
 ```tsx
 import { SuperJSON } from 'superjson'
 
 const router = createRouter({
-  transformer: SuperJSON,
+  serializer: SuperJSON,
```
SeanCassiere, line 229: As far as I can see, at this current time, this action of setting `serializer: SuperJSON` is not possible since the `serializer` property (previously named `transformer`) no longer exists on `RouterOptions`.

SeanCassiere (commented):
The removal of a user-configurable serializer/transformer is quite a breaking change for Start.

Should this go forward "as-is", we'll need to update it on the breaking changes tracking list (https://github.com/TanStack/router/discussions/2863).

docs/framework/react/guide/ssr.md
SeanCassiere: 👍🏼

Are we committing to removing `transformer/serializer` though?

If not, I think it'd make sense to **introduce** a bug where we leave the `RouterOptions` interface "as-is" but disable functionality in code. Then wire it back up in another PR with any renaming that needs to be done.

schiller-manuel: for now it cant be configured. but this functionality should be re-added at some point.

packages/react-router/src/router.ts
schiller-manuel: I think we need a different API approach to configure server stuff. Ideally, it does not touch the router at all.

arolson101:
the `hydrate` and `dehydrate` methods on router were removed - this change breaks `@ssrx/plugin-tanstack-router`

PR #3244, opened 2025-01-26
docs(examples): add lazy loaded monorepo example

beaussan:
This PR adds a monorepo with lazy loading example

Key files to look at:
- [Re exported types and functions from the router to have type augmentation](examples/react/router-monorepo-simple-lazy/packages/router/src/index.ts)
	- Build the router definition, with loader but no components
	- Expose functions and utils from tanstack router, but with type augmentation
- [Example exposed route from a feature library](examples/react/router-monorepo-simple-lazy/packages/post-feature/src/PostIdPage.tsx)
	- Because it uses the `createLazyRoute` function from the internal router, type augmentation applies and the path is typesafe
- [Where we plug the lazy routes with the routes](examples/react/router-monorepo-simple-lazy/packages/app/src/main.tsx)
	- Ensure the route.lazy() is linked to the right one (aka path === path) to avoid runtime error


SeanCassiere (changes requested):
Could you update the monorepo examples to use Tailwindcss v4? Currently, it's just using the distribution from their CDN.

If having to add Tailwindcss into the build for the monorepo packages is too much, it's fine, we can push this "as-is".

beaussan:
> @SeanCassiere  Could you update the monorepo examples to use Tailwindcss v4? Currently, it's just using the distribution from their CDN.
> 
> If having to add Tailwindcss into the build for the monorepo packages is too much, it's fine, we can push this "as-is".

Sure thing

Do you prefer this as a follow up pr or within this one ? I will need to provide custom content paths given tailwind cannot follow the within monorepo content, it shouldn't be a huge work but I want to make sure the core logic of the lazy loaded example can be consider good practice

beaussan:
Done in [`d2d397c`](https://github.com/TanStack/router/pull/3244/commits/d2d397c35f95e37474a28d602478999b90321244) + https://github.com/TanStack/router/pull/3247 for other monorepo examples (to avoid the noise in this already big pr)

SeanCassiere (approved):

PR #3251, opened 2025-01-27
ci(root): add automatic labeler workflow

SeanCassiere:
Adds the "automatic labeler" workflow that's used by Query.


SeanCassiere:
I thought about adding the `labeler-generate` script  into the "autofix" CI workflow, however, the workflow doesn't seem to have the ability to write into the `.github` folder.

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  - pull_request_target
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/labeler@v4.3.0
+        with:
+          repo-token: ${{ secrets.GH_TOKEN }}
```
SeanCassiere, line 12: Not sure if this is correct. Query uses `secrets.GITHUB_TOKEN`.

SeanCassiere, line 12:
```suggestion
          repo-token: ${{ secrets.GITHUB_TOKEN }}
```

This should be fine:
- https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication
- https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

PR #3253, opened 2025-01-27
fix(start): server functions to handle direct submission of `FormData`

SeanCassiere:
Sometime in the big rewrite, the handler lost the ability to accept incoming `FormData` when submitted via a `<form>` using the functions URL (accessed via `serverFn.url`).


e2e/start/basic/app/routes/-server-fns/submit-formdata-fn.tsx
```diff
@@ -0,0 +1,55 @@
...
+    if (!name) {
+      throw new Error('Name is required')
+    }
+
+    return {
+      name: name.toString(),
+    }
+  })
+  .handler(({ data: { name } }) => {
+    return new Response(`Hello, ${name}!`)
```
SeanCassiere, line 23: @schiller-manuel @chorobin could use help fixing the type-level error here not allowing `Response` to be returned. It works at runtime.

PR #3256, opened 2025-01-27
fix: headers

tannerlinsley:


nick22985:
Hey @tannerlinsley,
It seems to still be an issue but now seems to work on the server buth then gets refreshed on the client and gets rid of the data. Video attached just updated the example in this issue to replicate

https://gyazo.com/eb5086a22b05d4b28cf11f4c0eae9d74

PR #3258, opened 2025-01-28
fix(start-server): remove debugging timeout

schiller-manuel:


packages/start-server/src/transformStreamWithRouter.ts
```diff
@@ -174,6 +168,7 @@ export function transformStreamWithRouter(
   readStream(appStream, {
     onData: (chunk) => {
       const text = decodeChunk(chunk.value)
+      console.log('#### appStream onData', text)
```
SeanCassiere, line 171: Remove?

PR #3271, opened 2025-01-29
chore(examples): fix nx graph detection

beaussan:
On Nx 20.3.2 and 20.3.3, this issue got fixed:  https://github.com/nrwl/nx/issues/29486
However, on Nx 20.4 [Nx reverted the fix](https://github.com/nrwl/nx/pull/29762), causing the graph to be missing the link between the feature library and the application

The temporary workaround is to define a `implicitDependencies` on the project to force the link between the two projects


SeanCassiere:
cc-ing @schiller-manuel since the exclude in the root package.json was added in https://github.com/TanStack/router/pull/3270.

PR #3285, opened 2025-01-30
fix(start): ensure generic h3 functions remain generic, including new overloads

EskiMojo14:
previously, type parameters weren't being passed through to wrapper functions, meaning things like `useSession<Data>(config)` didn't work properly.

This PR ensures the original signature is kept intact, along with providing specifically typed overloads for wrappers that need it.

breaking changes:
- getEvent returns H3Event, or throws an error if it couldn't find one. Previously it returned `any`, and didn't check whether event was truthy.


schiller-manuel:
we should use those functions in the e2e test to ensure this still builds

PR #3289, opened 2025-01-31
fix(router-generator): reset route-group Regex when calling `handleNode`

### LINKED ISSUE #3280: "/"  goes to an empty page, instead of index.tsx

Router

### Describe the bug

using file/dir based routing

when using group routes, and having no index.ts directly inside src, "/" goes to an empty page instead of a 404 or the nearest index.tsx

https://github.com/FatahChan/tanstack-router-bugs

1. pnpm install
2. pnpm dev
3. go to localhost:defaultport

renders empty page, check the router dev tools

### Expected behavior

As a user, I expected that the nearest  index.tsx page to the routes to be render.

### Screenshots or Videos

Image

Image

### Platform

- OS: mac
- Browser: chrome
- Version: 132.0.6834.111

### Additional context

_No response_

SeanCassiere:
This was a weird one 🤔. 

Since the Regex we use for checking if a "route-node" is a possible "route-group" uses the global flag, it carries some state across runs. As such, it falsly returned `false` (instead of `true`) on some nodes that were in fact representing a "route-group".

Closes #3280


SeanCassiere:
> @SeanCassiere thank you so much, that was relatively was quick, I was hesitant to open an issue, I thought, there was something I didn't grasp yet

It was certainly a weird one 😅

PR #3294, opened 2025-01-31
feat(start): serialize `BigInt` for server functions

ryanagillie:
Adds the ability to serialize / parse bigints with serverFns.

I was using `BigInt`s as table IDs with postgres and was getting errors when sending over rows with `createServerFn`, this should fix that.


ryanagillie:
Any holdups on this? Looks ready to go

PR #3296, opened 2025-01-31
tests: create shared, non-published package for e2e utils

schiller-manuel:


e2e/e2e-utils/package.json
```diff
@@ -0,0 +1,29 @@
...
+        "types": "./dist/esm/index.d.ts",
+        "default": "./dist/esm/index.js"
+      },
+      "require": {
+        "types": "./dist/cjs/index.d.cts",
+        "default": "./dist/cjs/index.cjs"
+      }
+    },
+    "./package.json": "./package.json"
+  },
```
TkDodo, line 26: I think we should structure this package a bit differently: There isn’t an actual need to build this, as we are never publishing this. So we could have `exports` point to `src` directly:

- https://github.com/TanStack/router/pull/3301

PR #3318, opened 2025-02-03
fix: peer dependencies

CarsonF:
With these peer dependencies declared, Yarn PnP correctly stores one version of each tanstack package.
Without each one of these additions, yarn will duplicate one or more tanstack packages.
Which leads to context problems like #2540


packages/router-plugin/package.json
```diff
@@ -122,6 +122,7 @@
   },
   "peerDependencies": {
     "@rsbuild/core": ">=1.0.2",
+    "@tanstack/react-router": "workspace:^",
```
schiller-manuel, line 125: why is this required? there is nothing used from that package in router-plugin?

CarsonF: I believe it's because the router generator is declared as a dependency and it has that peer dependency declared, so this needs to forward that along.

schiller-manuel: interesting that this is not handled automatically...

can you please add 

```json 
"peerDependenciesMeta": {
    "@tanstack/react-router": {
      "optional": true
    }
  }
  ```

CarsonF: Sure!

CarsonF:
@schiller-manuel That PR test failure isn't related to my change right?

schiller-manuel:
no this is not related, CI was somehow broken. fixed that now

Issue #3326, opened 2025-02-04
Duplicate route segment at runtime breaks route types and from parameter


Sheraff:
### Which project does this relate to?

Router

### Describe the bug

With a specific configuration of pathless route, layout route, and index, the runtime of tanstack/router can end up with an incorrect `routeId` for a route.

For example:
```
routes/
  foo/
    _layout/
      route.tsx <- empty createFileRoute(), so basically just an <Outlet/>
      index.tsx
```
when we log `useMatches().at(-1).routeId`  from `index.tsx` we get 
```
/foo/foo/_layout/
```
instead of the expected
```
/foo/_layout/
```
Notice the `/foo` appearing twice

This example mentions `useMatches().at(-1).routeId`, but the same is true for `Route.id` in the same file.

If it was all "under the hood" it wouldn't necessarily be a huge deal, but route ids are used in the `from` parameters of many hooks (`useSearch({ from: '<some-route-id>' })`), and because of this bug, 
- if you use the "expected" route id as the `from` param, you get the classic "invariant violation" error
- if you use the "real" route id (the one with the duplicate segment), then you don't get any type inferences

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-tjgs9sry

### Steps to Reproduce the Bug or Issue

1. go to the stackblitz
2. click on the "reproduction" link

### Expected behavior

The `/foo` segment should not be duplicated

### Platform

- OS: macOS
- Browser: Chrome
- Version: 1.99.0

### Additional context

Interestingly, the `routeTree.gen.ts` file seems to be correct, no "segment duplication" (but maybe I just don't really know how this all gets assembled at runtime, and it's possible the duplication is already present, just not obvious).

Another interesting thing to note is that adding a `foo/route.tsx` file (even if it's just an empty `createFileRoute`) removes the issue
```
routes/
  foo/
    route.tsx   <- empty createFileRoute()
    _layout/
      route.tsx <- empty createFileRoute()
      index.tsx
```

For anyone stumbling onto this issue, this is probably your best bet at a temporary fix until this gets resolved.

schiller-manuel:
would be great if you can add one or more (failing) unit test(s) as a PR

Sheraff:
> would be great if you can add one or more (failing) unit test(s) as a PR

I think this is the correct failing test: https://github.com/TanStack/router/pull/3333. I added one in `packages/router-generator` because I think that's where the issue is actually from, but I'm not 100% sure.

[...]

PR #3333, opened 2025-02-04
test(router-generator): add failing test for #3326

Sheraff:
Maybe this is the correct failing test?

Screenshot 2025-02-04 at 23 52 12


packages/router-generator/tests/generator/no-duplicate-route-segment/routeTree.snapshot.ts
```diff
@@ -0,0 +1,159 @@
...
+const FooRoute = FooImport.update({
+  id: '/foo',
+  path: '/foo',
+  getParentRoute: () => rootRoute,
+} as any)
+
+const FooLayoutRouteRoute = FooLayoutRouteImport.update({
+  id: '/_layout',
+  getParentRoute: () => FooRoute,
+} as any)
```
Sheraff, line 34: Here the generated code (with 1.99, still with the bug) would generate
```ts
const FooLayoutRouteRoute = FooLayoutRouteImport.update({
  id: '/foo/_layout',
  getParentRoute: () => FooRoute,
} as any)
```

PR #3336, opened 2025-02-05
fix(react-router): build correct `href` when reloading the document with `Router.navigate`

nklhtv:
Reload the document with `navigate` function now builds the correct location href, in cases where hash history is used.

Similar to: https://github.com/TanStack/router/pull/1414


schiller-manuel:
we should test this

SeanCassiere (approved):

PR #3344, opened 2025-02-06
refactor(react-router): remove unsed `router` prop from `Route` class

nklhtv:
Route's `router` prop is always undefined as it is not initialized anywhere.
Should it be assigned in `init()`?

### DIFF
--- a/packages/react-router/src/route.ts
+++ b/packages/react-router/src/route.ts
@@ -721,7 +721,6 @@ export class Route<
   >
 
   // The following properties are set up in this.init()
-  router!: AnyRouter
   parentRoute!: TParentRoute
   private _id!: TId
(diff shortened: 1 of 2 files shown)


SeanCassiere (commented):
This is probably just some dead-code from when the configuration of routes required passing in the router instance. This can probably be safely removed since there's no logic that references this.

nklhtv:
If its just me that needs this we can close it and i can import the router in a different way.

SeanCassiere:
Actually, do you mind removing the `router?: AnyRouter` field and undoing the other stuff? Would help lighten the bundle a little.

SeanCassiere:
Thank you @nklhtv!

PR #3364, opened 2025-02-07
fix(react-router): simpler useForwardedRef using useImperativeHandle

Sheraff:
If I'm not missing some weird edge-case, I believe the `useForwardedRef` hook can be simplified using the native `useImperativeHandle`.

This hook is only used inside `useLinkProps` of `react-router`


packages/react-router/src/utils.ts
```diff
@@ -131,15 +131,6 @@ export function useIntersectionObserver<T extends Element>(
...
-  React.useEffect(() => {
-    if (!ref) return
-    if (typeof ref === 'function') {
-      ref(innerRef.current)
-    } else {
-      ref.current = innerRef.current
-    }
-  })
-
+  React.useImperativeHandle(ref, () => innerRef.current!, [])
```
Sheraff, line 134:
```suggestion
  React.useImperativeHandle(ref, () => innerRef.current!)
```

Maybe we should leave the dep array out, if we want to be sure this reproduces the same behavior as the `useEffect` it replaces

SeanCassiere (commented):
REMINDER to @SeanCassiere,

Check if `preload: 'viewport'` still works before merging.

schiller-manuel:
@SeanCassiere don't we have an e2e test for ` preload: 'viewport'`?

SeanCassiere:
@schiller-manuel probably not an e2e test. What we do have is a mocked unit test.

Sheraff:
> @schiller-manuel probably not an e2e test. What we do have is a mocked unit test.
> 
> Edit: I remembered why we don't have an e2e test. We could figure out a way of testing that the viewport functionality was working in playwright. TBH, I still don't know how we'd test that.

This seems to do the trick, assuming the targeted link is out of the viewport:
```ts
test('scrolling into viewport a link with preload=viewport to a route should preload route', async ({
  page,
}) => {
  const fetched: Array<string> = []
  page.on('request', (request) => {
    fetched.push(request.url())
  })

  await page.goto('/')
  await page.waitForLoadState('networkidle')

  // it hasn't been loaded yet
  const alreadyLoaded = fetched.some((url) => url.includes('without-loader.tsx?tsr-split'))
  expect(alreadyLoaded).toBe(false)
  fetched.length = 0

  const promise = page.waitForRequest(() => true)
  await page.getByRole('link', { name: 'without-loader' }).scrollIntoViewIfNeeded()
  await promise

  // it loaded on scroll
  expect(fetched[0]).toEqual(expect.stringContaining('without-loader.tsx?tsr-split'))
})
```

And to make sure the link was out of the viewport I just did this
```tsx
<div style={{ height: '200vh' }} />
<Link
  preload="viewport"
  to="/without-loader"
>
  without-loader
</Link>
```

SeanCassiere:
> Here's a PR for this #3372

Merged and this branch has been updated.

packages/react-router/src/utils.ts
Sheraff: I tested with an empty deps array and it seemed fine: https://stackblitz.com/edit/vitejs-vite-usxaksas?file=src%2FApp.tsx&terminal=dev

The cost of removing the deps array and re-executing this on every render is probably low, but in this case (a regular DOM ref, and not an object with a closure scope that matters), using `[]` is fine because the value is static. At least that's my understanding.

SeanCassiere: Are you satisfied with the PR? If so, I'm happy to merge.

Sheraff: Yeah let's go, i hope I'm not breaking anything, but I'm pretty confident this is correct.

PR #3365, opened 2025-02-07
fix(react-router): memo of useLinkProps

Sheraff:
> [!IMPORTANT]
> I'm not saying we should merge this, I just want to demonstrate *how it would look* to enable memoization in this hook

## The issue

Currently, the `useLinkProps` hook is either called from the `<Link>` component, or explicitly in userland.
- from the `<Link>` component (which is equivalent to `createLink`):
  ```tsx
  (props, ref) => {
    const { _asChild, ...rest } = props
    useLinkProps(rest, ref)
  ```
- in userland (just a guess as to how it's probably used)
  ```tsx
  useLinkProps({ foo, bar, baz })
  ```

In both cases, the `options` object will *always* be a new object on every render.

In addition to this, the `options` object is also internally overwritten which forces it to always be a new object:
```ts
  options = {
    from: parentRouteId,
    ...options,
  }
```

This means that some of the memoization inside `useLinkProps` is ineffective. For example:
```ts
  const next = React.useMemo(
    () => router.buildLocation(options as any),
    [router, options, currentSearch],
  )
```
or
```ts
  const doPreload = React.useCallback(() => {
    router.preloadRoute(options as any).catch((err) => {
      console.warn(err)
      console.warn(preloadWarning)
    })
  }, [options, router])
```

And this instability cascades down to more things, like `useLayoutEffect` and `useIntersectionObserver`

## Possible solutions

Because of this "instability" of the `options` object, using `useMemo` / `useCallback` adds cost, and doesn't bring any benefits.

There are 2 possible avenues for improving this situation:

1. stabilize the `options` object to make memoization possible. This is what this PR does, but it also demonstrates that it could potentially be verbose, and maybe pointless (some objects might have a shape that makes them somewhat inappropriate for memoization like `search`, `params`, ... unless we start resorting to ugly patterns of dynamic dependency arrays). It's not ideal, but for some situations it could be better than nothing.

2. remove most of the memoization. Since it is currently pure cost / no benefits, we could reduce the cost. This means that `useLayoutEffect` and `useIntersectionObserver` will keep re-executing on every render for now. This solution is most likely a bet on the `react-compiler` to solve this in the near future.

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -454,6 +454,18 @@ type LinkCurrentTargetElement = {
   preloadTimeout?: null | ReturnType<typeof setTimeout>
 }
 
+const observerOptions = { rootMargin: '100px' }
+
+const composeHandlers =
+  (handlers: Array<undefined | ((e: any) => void)>) =>
+  (e: { persist?: () => void; defaultPrevented: boolean }) => {
+    e.persist?.()
+    handlers.filter(Boolean).forEach((handler) => {
+      if (e.defaultPrevented) return
+      handler!(e)
+    })
+  }
+
 export function useLinkProps<
   TRouter extends AnyRouter = RegisteredRouter,
   const TFrom extends string = string,
@@ -482,6 +494,14 @@ export function useLinkProps<
     startTransition,
     resetScroll,
     viewTransition,
+    params,
+    search,
+    hash,
+    state,
+    mask,
+    reloadDocument,
+    _fromLocation,
+    from,
     // element props
     children,
     target,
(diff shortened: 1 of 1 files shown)


schiller-manuel:
> some objects might have a shape that makes them somewhat inappropriate for memoization like search, params

what problem does their shape cause?
we try to keep referential stability for those using replaceEqualDeep as much as possible, so them being object should not be an issue.

Sheraff:
> > some objects might have a shape that makes them somewhat inappropriate for memoization like search, params
> 
> what problem does their shape cause? we try to keep referential stability for those using replaceEqualDeep as much as possible, so them being object should not be an issue.

These are userland objects, they're not coming from hooks with structural sharing. For example:
```tsx
<Link
  to="/foo/$a/$b"
  params={{ a: 'a', b: 'b' }}
  search={{ hello: 'world' }}
>
```

schiller-manuel:
oh right... then they need to memoized by the caller :(

Sheraff:
> oh right... then they need to memoized by the caller :(

Yeah, but I doubt most codebases will be *this* stringent on memoization (until everyone has the `react-compiler` that is!)

Sheraff:
Fixed in https://github.com/TanStack/router/pull/4668

We did not handle "non stable userland objects" (like `search` and `params`). This won't be an issue on codebases that use the react compiler anyway.

PR #3366, opened 2025-02-07
fix(router-generator): prevent route file being auto-formatted during updates

benjamesfleming:
This PR adds a new configuration option (`formatter`) that allows the user to opt-out of having their route files auto formatted by Prettier.

This is useful when using a formatter other than Prettier (e.g. ESLint Stylistic, Biome, etc.)

### DIFF
--- a/docs/router/framework/react/guide/file-based-routing.md
+++ b/docs/router/framework/react/guide/file-based-routing.md
@@ -424,9 +424,9 @@ The following options are available for configuration via the `tsr.config.json` 
   - See the [using automatic code-splitting](./code-splitting.md#using-automatic-code-splitting) guide.
 
 - **`quoteStyle`**
-  - (Optional, **Defaults to `single`**) whether to use `single` or `double` quotes when formatting the generated route tree file.
+  - (Optional, **Defaults to `single`**) whether to use `single` or `double` quotes when formatting the generated files.
 - **`semicolons`**
-  - (Optional, **Defaults to `false`**) whether to use semicolons in the generated route tree file.
+  - (Optional, **Defaults to `false`**) whether to use semicolons in the generated files.
 - **`apiBase`**
   - (Optional) The base path for API routes. Defaults to `/api`.
   - This option is reserved for future use by the TanStack Start for API routes.
--- a/packages/router-generator/src/generator.ts
+++ b/packages/router-generator/src/generator.ts
@@ -1,9 +1,9 @@
 import path from 'node:path'
 import * as fs from 'node:fs'
 import * as fsp from 'node:fs/promises'
-import * as prettier from 'prettier'
 import {
   determineInitialRoutePath,
+  format,
   logging,
   multiSortBy,
   removeExt,
@@ -76,12 +76,6 @@ export async function generator(config: Config, root: string) {
 
   const TYPES_DISABLED = config.disableTypes
 
-  const prettierOptions: prettier.Options = {
-    semi: config.semicolons,
-    singleQuote: config.quoteStyle === 'single',
-    parser: 'typescript',
-  }
-
   let getRouteNodesResult: GetRouteNodesResult
 
   if (config.virtualRouteConfig) {
@@ -138,17 +132,22 @@ export async function generator(config: Config, root: string) {
 
     if (!routeCode) {
       const _rootTemplate = ROUTE_TEMPLATE.rootRoute
-      const replaced = fillTemplate(_rootTemplate.template(), {
+      const replaced = await fillTemplate(config, _rootTemplate.template(), {
         tsrImports: _rootTemplate.imports.tsrImports(),
         tsrPath: rootPathId,
         tsrExportStart: _rootTemplate.imports.tsrExportStart(),
         tsrExportEnd: _rootTemplate.imports.tsrExportEnd(),
       })
 
-      logger.log(`🟡 Creating ${node.fullPath}`)
-      fs.writeFileSync(
+      await writeIfDifferent(
         node.fullPath,
-        await prettier.format(replaced, prettierOptions),
+        '', // Empty string because the file doesn't exist yet
+        replaced,
+        {
+          beforeWrite: () => {
+            logger.log(`🟡 Creating ${node.fullPath}`)
+          },
+        },
       )
     }
   }
@@ -208,7 +207,8 @@ export async function generator(config: Config, root: string) {
         if (node.isLazy) {
           // Check by default check if the user has a specific lazy route template
           // If not, check if the user has a route template and use that instead
-          replaced = fillTemplate(
+          replaced = await fillTemplate(
+            config,
             (config.customScaffolding?.lazyRouteTemplate ||
               config.customScaffolding?.routeTemplate) ??
               tLazyRouteTemplate.template(),
@@ -227,7 +227,8 @@ export async function generator(config: Config, root: string) {
             !node.isPendingComponent &&
             !node.isLoader)
         ) {
-          replaced = fillTemplate(
+          replaced = await fillTemplate(
+            config,
             config.customScaffolding?.routeTemplate ??
               tRouteTemplate.template(),
             {
@@ -260,17 +261,11 @@ export async function generator(config: Config, root: string) {
           )
       }
 
-      await writeIfDifferent(
-        node.fullPath,
-        prettierOptions,
-        routeCode,
-        replaced,
-        {
-          beforeWrite: () => {
-            logger.log(`🟡 Updating ${node.fullPath}`)
-          },
+      await writeIfDifferent(node.fullPath, routeCode, replaced, {
+        beforeWrite: () => {
+          logger.log(`🟡 Updating ${node.fullPath}`)
         },
-      )
+      })
     }
 
     if (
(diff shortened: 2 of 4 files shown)


packages/router-generator/src/format.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export async function format(source: string, config: Config): Promise<string> {
+    switch (config.formatter) {
+        case 'prettier':
+            const preitterOptions: prettier.Config = {
+                semi: config.semicolons,
+                singleQuote: config.quoteStyle === 'single',
+                parser: 'typescript',
+            };
+            return prettier.format(source, preitterOptions);
```
schiller-manuel, line 12: preitterOptions => typo

packages/router-generator/src/format.ts
```diff
@@ -0,0 +1,17 @@
+import { Config } from './config'
+import * as prettier from 'prettier'
+
+export async function format(source: string, config: Config): Promise<string> {
```
schiller-manuel, line 4: why async?

docs/router/framework/react/guide/file-based-routing.md
```diff
@@ -416,6 +416,9 @@ The following options are available for configuration via the `tsr.config.json`
   - (Optional, **Defaults to `single`**) whether to use `single` or `double` quotes when formatting the generated route tree file.
 - **`semicolons`**
   - (Optional, **Defaults to `false`**) whether to use semicolons in the generated route tree file.
+- **`formatter`**
```
schiller-manuel, line 419: we need a more descriptive name for this config option since this only affects route files

packages/router-generator/src/config.ts
```diff
@@ -27,6 +27,7 @@ export const configSchema = z.object({
   generatedRouteTree: z.string().optional().default('./src/routeTree.gen.ts'),
   quoteStyle: z.enum(['single', 'double']).optional().default('single'),
   semicolons: z.boolean().optional().default(false),
+  formatter: z.enum(['prettier', 'none']).optional().default('prettier'),
```
schiller-manuel, line 30: maybe this should accept a function and defaults to a function that uses prettier?
that way you can format using your own setup instead of just turning it off
should still offer a way to turn it off completely though

packages/router-generator/src/format.ts
benjamesfleming: Fixed

packages/router-generator/src/format.ts
benjamesfleming: In the default case this is just a wrapper around the existing prettier.format(...) call, which is an async function.

packages/router-generator/src/config.ts
benjamesfleming: I did think about going this way, but then it only be configurable via the build plugins (e.g. TanStackRouterVite), and not the tsr.config.json file used by the `tsr generate` command.

docs/router/framework/react/guide/file-based-routing.md
benjamesfleming: Yeah... I was wondering about this because both existing prettier options (quoteStyle and semicolons) are documented as settings for the generated route tree file, but from what I can tell they are only used to format the route files.

Maybe there has been a regression in the docs?

Did you have a name suggestion? How about `routeFileFormatter`?

schiller-manuel: `routeFileFormatter` is better!

@SeanCassiere often has good naming ideas, so let's wait for him

SeanCassiere: > Yeah... I was wondering about this because both existing prettier options (quoteStyle and semicolons) are documented as settings for the generated route tree file, but from what I can tell they are only used to format the route files.

That's probably a mistake in the docs. Those settings (semicolons and quotes) are used for both the `routeTree` and generated route files.

It's generally not an issue for the route files, since as soon as user content comes in, the router-generator should be pretty hands-off at that point. @schiller-manuel, I hate to say it, but a lot of this could be avoided, if these transformations in the route file, were done using AST transformations.

packages/router-generator/src/generator.ts
```diff
@@ -246,17 +237,11 @@ export async function generator(config: Config, root: string) {
...
-      await writeIfDifferent(
-        node.fullPath,
-        prettierOptions,
-        routeCode,
-        replaced,
-        {
-          beforeWrite: () => {
-            logger.log(`🟡 Updating ${node.fullPath}`)
-          },
+      await writeIfDifferent(node.fullPath, config, routeCode, replaced, {
```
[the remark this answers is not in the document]
benjamesfleming: There were two places in generator.ts where writes to the file system were happening outside of the `writeIfDifferent` function, in both cases new files were being written. I've moved both of these to use the `writeIfDifferent` function.

packages/router-generator/src/config.ts
benjamesfleming: Done. A custom `(source: string) => string` function is now also accepted

docs/router/framework/react/guide/file-based-routing.md
benjamesfleming: > That's probably a mistake in the docs.

Yeah. The existing options, and the new `formatter` option, apply to both the route tree file and the generated route files. I've updated the docs to reflect this.

> I hate to say it, but a lot of this could be avoided, if these transformations in the route file, were done using AST transformations.

Just using AST transformations would still run into formatting issues. If the AST transformation updates the route id to one that exceeds the line's character limit, then the user would still have to run their formatter over the transformed code to ensure compliance.

SeanCassiere: > Just using AST transformations would still run into formatting issues. If the AST transformation updates the route id to one that exceeds the line's character limit, then the user would still have to run their formatter over the transformed code to ensure compliance.

Either way, it falls on the user, even if a `formatter='none'` option was introduced. Since that too, would require the user to run their format on the output.

Since the goal of `formatter='none'`, would be to save the file wihtout making any **stylistic** changes (i.e what we are doing now with prettier), it'd make more sense (in my point of view) to make those changes via AST transformations and let the user's formatted do the necessary work on that side. This was, we really are guaranteeing that we aren't saving the route files with any opinions on code-style.

benjamesfleming: The goal of `formatter='none'` is to allow the user to opt-out of the stylistic changes currently made to route files by TanStack Router, without introducing any breaking changes.

> it'd make more sense (in my point of view) to make those changes via AST transformations 

Maybe I'm missing something, but wouldn't this just add extra dependencies/complexity, and require extra parsing options (e.g. babel config) to be configurable in the tsr config? How is this better than a simple string replacement?

---

My question is - What's the simplest way to prevent TanStack Router formatting route files? I think a `formatter='none'` option is the answer to that question.

Is there a pathway to getting this PR merged?

SeanCassiere: > The goal of formatter='none' is to allow the user to opt out of the stylistic changes currently made to route files by TanStack Router, without introducing any breaking changes.

In agreement with you here 👍🏼

> Maybe I'm missing something, but wouldn't this just add extra dependencies/complexity, and require extra parsing options (e.g. babel config) to be configurable in the tsr config? 

Definitely not simpler but it would be the safer method of performing these changes. We've just be taking the existing babel utils be have in `router-utils`.

> How is this better than a simple string replacement?

Currently, the string replacement is using `Regex`. This is a bottleneck of sorts since it inherently expects a certain level of formatting. Like the `createFileRoute('/posts/')` being on one line and not broken up like this due to a formatting change.

```js
// We could have parsing issues with this.
	'/posts' // 🧠 imagine a longer route ID here
)
```

---

@benjamesfleming Overall, I'm in agreement and support the goal of this PR 💪🏼.

I just want to make sure we correctly approaching the fix for it. 
cc @schiller-manuel

SeanCassiere: @benjamesfleming speaking with @schiller-manuel, we've decided to go ahead with this PR, with one minor exception. The removal of the `formatter` option entirety.

The behavior would be as follows:

-When the generator is saving the route file during an "update" event, then the formatter shouldn't be used.
- However, on first write of the route file and any operations on the routeTree file, the formatter should still be in control.

This would make this the new behavior for all and would be classified as fix.

benjamesfleming: This makes sense 👍 

I've updated the PR to match the described behaviour.

SeanCassiere (approved):

packages/router-generator/src/utils.ts
```diff
@@ -123,34 +124,42 @@ export function removeExt(d: string, keepExtension: boolean = false) {
...
+ * @param source The content to format
+ * @param config The configuration object
+ * @returns The formatted content
+ */
+export async function format(source: string, config: Config): Promise<string> {
+  const prettierOptions: prettier.Config = {
+    semi: config.semicolons,
+    singleQuote: config.quoteStyle === 'single',
+    parser: 'typescript',
+  }
```
honzahruby, line 159: ❓ Just curious, why don't we use `prettier.resolveConfig` to get the user's actual config? We could use `config.generatedRouteTree` as the resolution path. Seems more accurate than having two random settings in the generator config. Or is it meant to be generic for other pluggable formatters in the future?

cc @schiller-manuel @SeanCassiere

Issue #3110, opened 2025-01-04
`useLocation` re-renders before navigating


Sheraff:
### Which project does this relate to?

Router

### Describe the bug

Assuming a simple route like this:
```ts
export const Route = createFileRoute('/about')({
  component: AboutComponent,
});

function AboutComponent() {
  const {pathname} = useLocation()
  console.log(pathname)
  return (
    <div className="p-2">
      <h3>About</h3>
    </div>
  );
}
```

My naive assumption would be that the only value `pathname` can ever have here is `"/about"`, but in reality, this will also re-render just before navigating away, with `pathname` having the value of the **next page**.

Alternatively, we can also use the following:
```ts
const pathname = useRouterState({ select: (s) => s.resolvedLocation.pathname });
```

But in that case, `pathname` will *initially* have the value of the **previous page** before immediately re-rendering with the current page.

One way to access the `pathname` that seems stable is this:
```ts
const pathname = useMatch({ strict: false, select: (s) => s.pathname });
```

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-fpv611d5?file=src%2Froutes%2F_yo%2Fabout.tsx

### Steps to Reproduce the Bug or Issue

1. Go to '/about' in the demo above
2. open the console to see what gets logged
3. Go to any other page

This will show that the `console.log(pathname)` (similar to the one in the code sample above) logs the *next route* even though it's only in the route `component` of /about.

The `about.tsx` file contains other commented-out methods of "logging the pathname" that demonstrate the various render behaviors.

### Expected behavior

I'm not entirely sure what is "expected behavior" here. But my naive intuition would be that for all of these methods, since they're used inside the "/about" route, they should only ever render with the "/about" pathname.

### Platform

- OS: macOS
- Browser: Chrome
- Version: 1.95.0

nullberri:
https://tanstack.com/router/latest/docs/framework/react/guide/search-params#zod

At some recent update they changed what you should be doing for Zod validation adapting. If you just swap the adapter for the import in the example it fixes it.

hobi9:
I’m on version 1.97.3 and noticed an issue with useMatch. When using the following code:

`const pathname = useMatch({ strict: false, select: (s) => s.pathname }); `

It initially returns '/' but does not update when the route changes. As a temporary workaround, I switched to this approach:

`const { pathname } = useMatches().at(-1)!;`

This resolves the issue for now. Is this the intended behavior, or is this a bug in the useMatch implementation?

[...]

PR #3371, opened 2025-02-09
docs(router): add pitfall when navigating for the `useLocation` hook

xsjcTony:
Relates to: https://github.com/TanStack/router/issues/3110#issuecomment-2645886152

This is the only part I'm confident about, and I'm not sure about the exact router lifecycle, so I'll just leave it to someone else rather than the potentially misleading details.

### DIFF
--- a/docs/router/framework/react/api/router/useLocationHook.md
+++ b/docs/router/framework/react/api/router/useLocationHook.md
@@ -5,6 +5,11 @@ title: useLocation hook
 
 The `useLocation` method is a hook that returns the current [`location`](./ParsedLocationType.md) object. This hook is useful for when you want to perform some side effect whenever the current location changes.
 
+> [!WARNING]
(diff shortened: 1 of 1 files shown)


brenelz:
What is the status of this PR. Is this docs change something we still want to do?

xsjcTony:
> What is the status of this PR. Is this docs change something we still want to do?

No idea. Seems the original issue is not getting any attention either. Still a significant behaviour difference from my perspective, hopefully it's getting fixed soon.

aerialist7:
@tannerlinsley, hello! Would you merge this PR according to your previous https://github.com/TanStack/router/issues/3110#issuecomment-2645886152 ?

LadyBluenotes:
Sorry, this ended up getting lost in all the PR work and such.

I'm going to close this because the docs page moved, so this PR no longer applies cleanly, and the `pendingComponent` explanation needs to be revisited against current navigation behavior.  

#3110 remains open so if you can, please open a fresh PR with if you want to continue this documentation work!

PR #3375, opened 2025-02-09
fix(start): align middleware context types with runtime

chorobin:
The types for middleware context can be improved to match the runtime better.

The return type of `next` in `.client` should have the following properties:
 - context - **This is the merged client context from the middleware chain (including any context sent to the client from the server) plus the inferred context passed to next**
 - sendContext - **This is the merged context which is sent to the server from the middleware chain plus the inferred sendContext passed to next**

The options in `.client` should have the following properties:
- context - **This is the merged client context from the middleware chain without any sent client context from the server**

The return type of `next` in `.server` should have the following properties:
 - context - **This is the merged server context from the middleware chain (including any context sent to the server from the client) plus the inferred context passed to next**
 - sendContext - **This is the merged context which is sent to the client from the middleware chain plus the inferred sendContext passed to next**

The options in `.server` should have the following properties:
- context - **This is the merged server context from the middleware chain with sent sever context from the client**


chorobin:
- [ ] Type tests!!
- [ ] Optimize types
- [ ] Fix remaining type issues with `context` options for `.server` and `.client`

tannerlinsley:
Looks great! Let’s get it updated

Issue #1686, opened 2024-05-31
URIError: URI malformed


hckhanh:
### Describe the bug

I access my route with this url: `http://localhost:3000/login?redirect=http:/%localhost:3000/standalone/policies-new/15MRZKHBCC/CAR_COMPREHENSIVE`

### Your Example Website or App

N/A

### Steps to Reproduce the Bug or Issue

1. Define a normal route

### Expected behavior

It should not throw any internal bugs

### Screenshots or Videos

![image](https://github.com/TanStack/router/assets/6380436/72e19998-7d4d-436b-b3fc-794dd50651f8)

![image](https://github.com/TanStack/router/assets/6380436/05671536-b9ec-44ba-8100-d61ea4dbc843)

### Platform

- OS: Windows
- Browser: Chrome
- Version: 125.0.6422.113

schiller-manuel:
decoding `http:/%localhost:3000/standalone/policies-new/15MRZKHBCC/CAR_COMPREHENSIVE` with `decodeURIComponent` yields this error, so not a router problem:

![Screenshot 2024-06-03 at 19 02 52](https://github.com/TanStack/router/assets/6340397/16b7df9c-3eac-46cc-89bc-018fe23edfaf)

Looks like you got a `%` instead of a `/` in your search param and thus its not correctly encoded.

hckhanh:
I mean if someone intend to make this error. How can I prevent this. My Sentry service just make an alarm about this all the time.

hckhanh:
I hate the way you close the issue like this before you can clearly understand the context.

schiller-manuel:
please provide all the necessary context in the issue description in future

schiller-manuel:
what would you expect to happen when you pass a wrongly encoded query string?

PR #3381, opened 2025-02-10
fix(router-core): patch qss to attempt encoding value to prevent malformed URI errors

leesb971204:
maybe related #1686 

In my case, an error occurred when using the navigate.to function with a URL containing special characters like %.
	1. Searching on a path like myApp.com/depth updates the URL to myApp.com/depth?search=100%.
	2. To retain the search query, searchStr is appended to useLocation.pathname and passed to navigate.to.
	3. Navigating to myApp.com/depth/1?search=100% causes a URI malformed error.
Actually, this issue didn’t occur when using useRouter in Next.js, but it happens after migrating to React.js with TanStack Router.

---

The earlier search/query param decode pattern used `decodeURIComponent(value)` which could throw an error when a value like '100%' was inputted. This change makes it safe for such actions to be attempted and gracefully be handled by the router.

### DIFF
--- a/packages/router-core/src/path.ts
+++ b/packages/router-core/src/path.ts
@@ -183,7 +183,7 @@ export function parsePathname(pathname?: string): Array<Segment> {
 
       return {
         type: 'pathname',
-        value: decodeURI(part),
+        value: decodeURI(encodeURI(part)),
       }
     }),
   )
--- a/packages/router-core/src/path.ts
+++ b/packages/router-core/src/path.ts
@@ -1,4 +1,4 @@
-import { hasUriEncodedChars, last } from './utils'
+import { last } from './utils'
 import type { MatchLocation } from './RouterProvider'
 import type { AnyPathParams } from './route'
 
@@ -183,9 +183,7 @@ export function parsePathname(pathname?: string): Array<Segment> {
 
       return {
         type: 'pathname',
-        value: hasUriEncodedChars(part)
-          ? decodeURI(part)
-          : decodeURI(encodeURI(part)),
+        value: decodeURI(part),
       }
     }),
   )
--- a/packages/router-core/src/path.ts
+++ b/packages/router-core/src/path.ts
@@ -1,4 +1,4 @@
-import { last } from './utils'
+import { hasUriEncodedChars, last } from './utils'
 import type { MatchLocation } from './RouterProvider'
 import type { AnyPathParams } from './route'
 
@@ -183,7 +183,9 @@ export function parsePathname(pathname?: string): Array<Segment> {
 
       return {
         type: 'pathname',
-        value: decodeURI(encodeURI(part)),
+        value: hasUriEncodedChars(part)
+          ? decodeURI(part)
+          : decodeURI(encodeURI(part)),
       }
     }),
   )
--- a/packages/router-core/src/qss.ts
+++ b/packages/router-core/src/qss.ts
@@ -48,7 +48,7 @@ export function encode(obj: any, pfx?: string) {
  */
 function toValue(mix: any) {
   if (!mix) return ''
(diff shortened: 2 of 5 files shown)


patrick-jessen:
@leesb971204 Why was this closed?
We are experiencing the same issue - any sign of `%`, `%25`, `%2525` causes the router (and `navigate()`) to crash.
I have tried patching in your changes, and it seems to solve the issue.

leesb971204:
> @leesb971204 Why was this closed? We are experiencing the same issue - any sign of `%`, `%25`, `%2525` causes the router (and `navigate()`) to crash. I have tried patching in your changes, and it seems to solve the issue.

I think this job doesn't seem to work very well and at the same time, it didn't pass the test, so I closed it. I'll reopen it, so I'd appreciate it if you could help me solve the problem.

SeanCassiere (approved):
@leesb971204, for path params, you cannot have a URL in this format `/perc/100%`. It is an invalid URI and the expected behaviour on the web for `decodeURIComponent('100%')` is for it to throw an error.

As such, for now we won't be accepting changes for path params regarding the `%` character.

leesb971204:
> @leesb971204, for path params, you cannot have a URL in this format `/perc/100%`. It is an invalid URI and the expected behaviour on the web for `decodeURIComponent('100%')` is for it to throw an error.
> 
> As such, for now we won't be accepting changes for path params regarding the `%` character.

@SeanCassiere 
You’re right. I was mistaken. So in the end, the work will only be done on searchParam, right?

SeanCassiere:
> ... So in the end, the work will only be done on searchParam, right?

Yes, these changes will only affect search/query parameters.

SeanCassiere:
@leesb971204 thank you for opening this PR and contributing 🎉

patrick-jessen:
> @leesb971204, for path params, you cannot have a URL in this format `/perc/100%`. It is an invalid URI and the expected behaviour on the web for `decodeURIComponent('100%')` is for it to throw an error.
> 
> As such, for now we won't be accepting changes for path params regarding the `%` character.

After the latest modifications, this PR no longer solves our issue, as we are using path param.
Check out the following cases - I would expect all of these to be valid, don't you agree?

- `website.com/100%24` -> okay, param is `100$`
- `website.com/100%2524` -> okay, param is `100$`
- `website.com/100%25` -> error `URI malformed`
- `website.com/100%2525` -> error `URI malformed`

leesb971204:
> > @leesb971204, for path params, you cannot have a URL in this format `/perc/100%`. It is an invalid URI and the expected behaviour on the web for `decodeURIComponent('100%')` is for it to throw an error.
> > As such, for now we won't be accepting changes for path params regarding the `%` character.
> 
> After the latest modifications, this PR no longer solves our issue, as we are using path param. Check out the following cases - I would expect all of these to be valid, don't you agree?
> 
> * `website.com/100%24` -> okay, param is `100$`
> * `website.com/100%2524` -> okay, param is `100$`
> * `website.com/100%25` -> error `URI malformed`
> * `website.com/100%2525` -> error `URI malformed`

I missed these cases. You mean `100%25` is valid value becase of '%' is '%25' by percent encoding, right?

patrick-jessen:
Yes - `encodeURIComponent("100%") === '100%25'`.
I would really appreciate if this could be fixed as well, as it breaking for our product 🙏.

SeanCassiere:
> Yes - `encodeURIComponent("100%") === '100%25'`. I would really appreciate if this could be fixed as well, as it breaking for our product 🙏.

I'm honestly not sure how to go about getting this to work 🤔.

@patrick-jessen Please create an issue for this with a reproduction, perhaps Manuel or Tanner would have a better idea about implementing a proper for this one.

PR #3390, opened 2025-02-11
Update build-from-scratch.md - Fix typo in charset meta

lb-:
I believe this should be `charset` not `charSet`.

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#browser_compatibility
- https://www.w3.org/TR/2010/WD-html-markup-20101019/meta.charset.html

Build output when following the documentation.
Screenshot 2025-02-11 at 4 24 37 PM

### DIFF
--- a/docs/start/framework/react/build-from-scratch.md
+++ b/docs/start/framework/react/build-from-scratch.md
@@ -212,7 +212,7 @@ export const Route = createRootRoute({
   head: () => ({
     meta: [
       {
-        charSet: 'utf-8',
+        charset: 'utf-8',
       },
       {
         name: 'viewport',


SeanCassiere (changes requested):
Blocking this using "request changes" due to the following:

This change would cause a typescript error. Since it wouldn't match the types, as the `RouteMatch` interface expects the following:

```tsx
interface RouteMatch {
	// ...
	meta?: Array<React.JSX.IntrinsicElements['meta'] | undefined>
	// ...
}
```

lb-:
OK looks like this value should be `charSet` when declared with React, however, React SSR will not convert it to lower case.

See
* https://github.com/facebook/react/issues/21571
* https://github.com/facebook/react/pull/11110
* https://github.com/facebook/react/issues/10863

As HTML attribute names are case insensitive, the React team have decided to not convert this (See PR above).

https://html.spec.whatwg.org/multipage/semantics.html#character-encoding-declaration

Although it feels inconsistent it's nothing to do with TanStack, just a quirk of React DOM generation when pre-rendering HTML.

PR #3418, opened 2025-02-13
Server fn mocks

EugenEistrach:
add basic mock functionality for client. we can execute this somewhere in our code to change the function to be mocked, e.g.:  
```someServerFn.mock.setImpl(async () => {
  console.log('Default mock!')
  return { message: 'Hello from default mock!' }
})
```
not sure if this is suitable or we need some other approach. 
also might need to think if this should be controlled by some flag or whatever.. 

also have create this package that does allow mock in userland without having to add the mock logic to the server functions themself https://www.npmjs.com/package/tanstack-mock-server-fn


SeanCassiere:

PR #3419, opened 2025-02-13
fix(router-core): allow `%25` in path params

### LINKED ISSUE #3402: Path params do not support `%`

Router

### Describe the bug

This is a two-fold issue.
- `navigate()` does not support path params with `%`
- The router cannot resolve paths with `%25`

https://stackblitz.com/edit/github-dnzkk3ai?file=src%2Froutes%2Findex.tsx

### Regarding `navigate()`:
`navigate()` will throw `URI malformed` if a param contains `%`.
Example: `navigate({..., params: {q: "100%"}})`
Example: `<Link to="100%">` and `<Link to="/$q" params={{q:"100%"}}>`

**Working around the issue**
This can be circumvented by manually encoding (though I would expect the library to handle this).
Example: `navigate({..., params: {q: encodeURIComponent("100%")}})` - (`q` then becomes `"100%25"`)

The latter will successfully pushState, although the `%` ends up double-encoded.
Example: `http://website.com/search/100%2525`

`useParams` does successfully double-decode, so `const params = useParams()` yields `{q: "100%"}`

However..

### Regarding route resolution
The router cannot resolve paths (unless pushState) containing `%25` nor `%2525`.
Examples:
`http://website.com/search/100%25` - throws `URI malformed` in SSR
`http://website.com/search/100%2525` - works in SSR but throws `URI malformed` client side 

### Expected behavior

- I would expect `navigate` to handle URI encoding of `%` for me.
- I would expect the URL to not need double-encoded `%2525`.
- I would expect the router (SSR and CSR) to be able to handle `%25` in path.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Version: 1.102.5

### Additional context

Related PR for handling `%` in query-params: https://github.com/TanStack/router/pull/3381

leesb971204:
fixes #3402


schiller-manuel:
not saying this is the correct fix, rather wait for @SeanCassiere  to look at this.
however, we need tests, tests and some more tests.

SeanCassiere (commented):
This doesn't seem like the "correct fix", since it's mostly just accounting for a particular use case with a single encoded character in mind (the `%` character).

leesb971204:
> This doesn't seem like the "correct fix", since it's mostly just accounting for a particular use case with a single encoded character in mind (the `%` character).

Can you provide some use cases that explains why this fix is not correct?

SeanCassiere:
> Can you provide some use cases that explains why this fix is not correct?

This PR is surgically only handling the case of `%`. This is a single use case patch.

This will probably make more sense once you start writing test cases. To start with you can replicate some tests in `packages/react-router/tests/router.test.tsx` to handle the navigation and parsing of `100%25` (converts to `100%`) and `100%26` (converts to `100&`).

leesb971204:
> > Can you provide some use cases that explains why this fix is not correct?
> 
> This PR is surgically only handling the case of `%`. This is a single use case patch.
> 
> This will probably make more sense once you start writing test cases. To start with you can replicate some tests in `packages/react-router/tests/router.test.tsx` to handle the navigation and parsing of `100%25` (converts to `100%`) and `100%26` (converts to `100&`).

Thinking about it again, converting `100%25` to `100%` doesn’t seem to be the right approach. The issue was not about the encoding format, but rather that a standalone `%` character in the path parameter causes a malformed error. If we proceed with the conversion, `100%2F200` would be transformed into `100/200`. Wouldn’t it be more appropriate to simply allow the `%` character to be included in the path parameter?

leesb971204:
> > Can you provide some use cases that explains why this fix is not correct?
> 
> This PR is surgically only handling the case of `%`. This is a single use case patch.
> 
> This will probably make more sense once you start writing test cases. To start with you can replicate some tests in `packages/react-router/tests/router.test.tsx` to handle the navigation and parsing of `100%25` (converts to `100%`) and `100%26` (converts to `100&`).

So, I modified the code to ensure that `%25` in the path parameter does not cause an error. The original issue was that strings like `%24` did not cause errors, but `%25` required passing through encodeURI to be included in the path parameter, which is why I made the modification. Also, I believe that having a standalone `%` in a URI is illegal like `100%`.

patrick-jessen:
Thank you 🙏, this does work for our scenarios - I would love to see this merged 🙂.

I do want to point out though, that there are still some *theoretical* pitfalls:
Example: `/%F0%9F%9A%80to%2Fthe%2Fmoon%25`
Throws: `InvalidCharacterError: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.`
Here is Google handling that particular scenario: https://www.google.com/search?q=%F0%9F%9A%80to%2Fthe%2Fmoon%25

leesb971204:
> Thank you 🙏, this does work for our scenarios - I would love to see this merged 🙂.
> 
> I do want to point out though, that there are still some _theoretical_ pitfalls: Example: `/%F0%9F%9A%80to%2Fthe%2Fmoon%25` Throws: `InvalidCharacterError: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.` Here is Google handling that particular scenario: https://www.google.com/search?q=%F0%9F%9A%80to%2Fthe%2Fmoon%25

@patrick-jessen 
First of all, the scenario you provided does not fit with this work. The scenario you provided is about the search parameter, and I worked on the path parameter. I think it will work the same if you put it in the search parameter.

But the error message was something I didn't encounter, even passed the test code. Is it an error that occurs in the ssr environment?

patrick-jessen:
@leesb971204 
Ah yes you are right - just disregard that then, is a super edge case either way.
(The error did in fact originate from an `URI malformed`).

Could `decodeURI(encodeURI('%25'))` not just be `'%25'`?
Anyway, I am sorry, I don't want to hold you up any more - thank you for fixing this issue 🙏

leesb971204:
> @leesb971204 Ah yes you are right - just disregard that then, is a super edge case either way. (The error did in fact originate from an `URI malformed`).
> 
> Could `decodeURI(encodeURI('%25'))` not just be `'%25'`? Anyway, I am sorry, I don't want to hold you up any more - thank you for fixing this issue 🙏

Oh, that’s right! I added unnecessary computations. Thanks for pointing that out!
My work hasn’t been merged yet, but I’m glad it was helpful.
Since this seems to be an issue with many factors to consider, I’ll wait for feedback.

patrick-jessen:
Any update on this?

schiller-manuel:
@patrick-jessen the comments about this probably not being the right fix (see above) still stand.

leesb971204:
> @patrick-jessen the comments about this probably not being the right fix (see above) still stand.

@schiller-manuel 
The issue occurs only when ‘%25’ is included, whereas a standalone ‘%’ cannot be used as a path parameter. This is why I handled it this way. Could you explain why this is not the right fix?

SeanCassiere:
For me, my original comment still stands. I still don't think that this is the "correct" fix to this problem.

This patch only accounts for a single use case where `%25` is present as a path param. So, for example, if `%26` was present, it'd likely cause the same existing behaviour of throwing an exception.

---

In my opinion, I think that the fix for this is probably going to be a variation of what was applied in https://github.com/TanStack/router/pull/3381, adapted to make it work for path params.

patrick-jessen:
@SeanCassiere 
%25 is the only encoding that causes errors.
%24, %26 etc. work just fine

leesb971204:
> For me, my original comment still stands. I still don't think that this is the "correct" fix to this problem.
> 
> This patch only accounts for a single use case where `%25` is present as a path param. So, for example, if `%26` was present, it'd likely cause the same existing behaviour of throwing an exception.
> 
> In my opinion, I think that the fix for this is probably going to be a variation of what was applied in #3381, adapted to make it work for path params.

I agree that this patch considers only one case,, but in my execution environment, an error does not occur when `%26` is present. I would appreciate it if you could explain the situation in detail.

SeanCassiere:
> I agree that this patch considers only one case,, but in my execution environment, an error does not occur when `%26` is present. I would appreciate it if you could explain the situation in detail.

I'll do some local testing and add some more unit tests to confirm.

SeanCassiere (approved):
I still don't really understand why params with `%25` throw an exception at runtime whilst others like `%26` do not 🤷🏼‍♂️

That being said, in meantime we'll have the testing infrastructure in place to make sure we don't see a regression with the pesky `%25` character.

cc @schiller-manuel merge at your discretion for this one.

PR #3428, opened 2025-02-14
fix: add failing test for anon default export for server fns

brenelz:
When using an anonymous default export the directives plugin produces invalid javascript like the following

```
function() {
}
```

When it has an identifier its fine:

```
function Home() {
}
```


brenelz:
Addressed in this PR https://github.com/TanStack/router/pull/3428

PR #3444, opened 2025-02-15
feat(start): data guard when sending binary data to GET server functions

nikolovlazar:
Implementing a data guard that checks if the data we're sending to a GET server function contains binary data (unserializable in query params).

[Video Walkthrough](https://www.youtube.com/watch?v=UBMAvqzCr8w)
👆 Video walkthrough

This PR covers two scenarios:
- When we're sending binary data to a GET server function (lines 40-44)
- When we're sending binary data to a POST server function, but within a payload instead of FormData (lines 61-65)

This is my first time contributing to Start, so let me know if I'm missing anything (I didn't write tests for this behavior 😬), or if I should push the example that I showed in the video.


schiller-manuel:
we should also prevent this at the type level I guess?

nikolovlazar:
@schiller-manuel didn't realize we could. Could you give me some pointers?

schiller-manuel:
this would need to cause a typescript error if the data is not serialzable as query and GET is the method. 

let's ask @chorobin

nikolovlazar:
I can't seem to figure this out 😅

webdevcody:
this would be great! just spent time debugging why my FormData File was becoming a string on my server action just to forget to just a POST instead of a GET

nikolovlazar:
I decided to close this since there hasn't been any movement. I have to close it since it prevents me from keeping my own fork in sync.

PR #3447, opened 2025-02-15
fix(directive-functions-plugin): handle anonymous default exports

schiller-manuel:


brenelz:
Confirmed this would cause this test to pass in SolidStart
https://github.com/solidjs/solid-start/pull/1808

PR #3450, opened 2025-02-16
fix(react-router): lazyRouteComponent reload guard key fallsback to importer's import.meta.url on Safari

### LINKED ISSUE #3327: Auto-reload on failed lazy route loading can only work once per session on Safari

Router

### Describe the bug

This issue is a follow-up of https://github.com/TanStack/router/pull/3262. The 4th issue mentioned in this PR hasn't been solved yet.

Basically: the system to auto-reload the app on navigation if a lazy route file failed to load is based on the *error message*. This error message is used as a `sessionStorage` key to ensure that we don't enter a reload loop if the same file fails to load twice during the same user session.

However, Safari does not include the file name in the error message, only Chrome and Firefox do. So on Safari
- user opens the app
- on the server: we deploy a new version of the app, all files get new hashes
- the user tries to navigate, fails to load the route file
- we store the error message in sessionStorage and reload the page (so far, everything is OK)
- on the server: we deploy another version of the app
- the user tries to navigate, fails to load the route file
- the error message is the same as before, we already have it as a key in sessionStorage, we do not reload the page

All this might make more sense looking at the source code: https://github.com/TanStack/router/blob/main/packages/react-router/src/lazyRouteComponent.tsx

see below

Reproducing this on tanstack is a little complicated. But I have a repro that shows the network error: https://stackblitz.com/edit/vitejs-vite-ztyqvabt?file=src%2FApp.tsx

Just open that link on any browser and you'll see the error message directly on the page.

### Expected behavior

A user on Safari should be able to have their session "survive" more than 1 deployment, like in other browsers.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Safari
- Version: 1.99.0

### Additional context

Thread on discord where Manuel Shiller and I were talking about this issue: https://discord.com/channels/719702312431386674/1333557517254398065

Sheraff:
Fixes https://github.com/TanStack/router/issues/3327

This PR might be easier to review as commit-by-commit (because there are many changes in generated files `packages/router-plugin/tests/code-splitter/snapshots/...`).

- `e2e/react-router/basic-file-based-code-splitting`
    Add an E2E playwright spec for validating the reload behavior of pages where a lazy import fails to load
- `packages/react-router/src/lazyRouteComponent.tsx`
    Accept a 4th parameter to provide a unique key to the sessionStorage entry used as a reload guard on a failed lazy load
- `packages/router-generator/src/generator.ts` and `packages/router-plugin/src/core/code-splitter/compilers.ts`
    Provide 4th parameter to `lazyRouteComponent` calls as `import.meta.url`. This is native in browsers, so should be "bundler independant" (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta).

> [!NOTE]
> The playwright spec added by this PR tests a behavior that might be browser-specific. However, the E2E tests run on chromium only. Maybe it might be worth it to enable multiple "projects" in playwright?
> 
> See implementation:
> https://github.com/TanStack/router/blob/adcdde4d576a9242303400f4112a883328af4554/packages/react-router/src/lazyRouteComponent.tsx#L10-L20

---

In a future PR, it might be a good idea to apply the same (or similar) behavior on import fail to the `lazyFn` since it's used in a way that feels similar to `lazyRouteComponent`, but has no error handling.

https://github.com/TanStack/router/blob/adcdde4d576a9242303400f4112a883328af4554/packages/react-router/src/router.ts#L3060-L3073

### DIFF
--- a/e2e/react-router/basic-file-based-code-splitting/src/main.tsx
+++ b/e2e/react-router/basic-file-based-code-splitting/src/main.tsx
@@ -10,6 +10,7 @@ const router = createRouter({
   defaultPreload: 'intent',
   defaultStaleTime: 5000,
   scrollRestoration: true,
+  defaultErrorComponent: () => <div>This is the error component</div>,
 })
 
 // Register things for typesafety
--- a/e2e/react-router/basic-file-based-code-splitting/tests/app.spec.ts
+++ b/e2e/react-router/basic-file-based-code-splitting/tests/app.spec.ts
@@ -31,3 +31,41 @@ test('Navigating to a not-found route', async ({ page }) => {
   await page.getByRole('link', { name: 'Start Over' }).click()
   await expect(page.getByRole('heading')).toContainText('Welcome Home!')
 })
+
+test('Navigating to a route where the lazy component fails to load', async ({ page }) => {
+  // block (and count) all requests to the posts.index route component
+  let requested = 0
+  await page.route('**/assets/posts.index-*', (route) => {
+    requested++
+    return route.fulfill({
+      status: 404,
+      contentType: 'text/plain',
+      body: 'Not Found!'
+    })
+  })
+
+  // count how many times the page reloads
+  let reloaded = 0
+  page.on('load', () => {
+    reloaded++
+  })
+
+  // navigate to the posts page
+  await page.getByRole('link', { name: 'Posts' }).click()
+
+  // will reload only once, despite failing twice, because the name of the module is the same both times
+  await expect(() => {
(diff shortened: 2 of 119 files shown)


brenelz:
This has a ton of conflicts now

LadyBluenotes:
Closing this because of all the conflicts and the current patch probably needs to be reworked. 

The Safari reload guard still looks worth addressing, though, but the fix needs framework-specific handling and WebKit coverage.

PR #3276, opened 2025-01-29
feat(solid-router): Add a SolidJS driver for TanStack Router

birkskyum:
Supersedes:
- #3086 

Continuation of:
- #3171 

This adds a solid-router package, which is a copied and modified version of the react-router.

The branch isn't complete just yet - the PR is made to increase collaboration.

Blockers
- https://github.com/TanStack/router/issues/3360

PR #3481, opened 2025-02-19
docs(solid-router): add getting-started pages

birkskyum:
Blocked by:
- #3276 

Related to:
- https://github.com/TanStack/tanstack.com/pull/346


docs/router/framework/solid/overview.md
```diff
@@ -0,0 +1,147 @@
...
+- [TRPC](https://trpc.io/)
+- [Remix](https://remix.run)
+- [Chicane](https://swan-io.github.io/chicane/)
+- [Next.js](https://nextjs.org)
+
+We acknowledge the investment, risk and research that went into their development, but are excited to push the bar they have set even higher.
+
+## Let's go!
+
+Enough overview, there's so much more to do with TanStack Router. Hit that next button and let's get started!
```
tannerlinsley, line 147: I swear there is a way to have pages like this work across frameworks so we only have to maintain one. @schiller-manuel @SeanCassiere ?

SeanCassiere: Yeah, the replacement syntax should still be there in Query's repo. We'll need to move some of the docs as "core" and then use the replacement syntax we have.

- For settings items as "core" in `docs/config.json` (https://github.com/TanStack/virtual/blob/main/docs/config.json)
- Replacement syntax to not need to duplicate the docs (https://github.com/TanStack/query/blob/main/docs/framework/vue/overview.md?plain=1)

birkskyum: The reason i made a second overview page was to take out all the comparisons to other react routers.

birkskyum:
@SeanCassiere , how to i make the solid appear in the framework drop down?

docs/router/framework/solid/quick-start.md
```diff
@@ -0,0 +1,217 @@
...
+# or
+deno add npm:@tanstack/solid-router npm:@tanstack/router-plugin
+```
+
+#### Configure the Vite Plugin
+
+```tsx
+// vite.config.ts
+import { defineConfig } from 'vite'
+import solid from 'vite-plugin-solid'
```
brenelz, line 38: do we want to make a note about installing vite-plugin-solid?

docs/router/framework/solid/quick-start.md
```diff
@@ -0,0 +1,217 @@
...
+```
+
+#### `src/main.tsx`
+
+Regardless of whether you are using the `@tanstack/router-plugin` package and running the `npm run dev`/`npm run build` scripts, or manually running the `tsr watch`/`tsr generate` commands from your package scripts, the route tree file will be generated at `src/routeTree.gen.ts`.
+
+Import the generated route tree and create a new router instance:
+
+```tsx
+import { render } from 'solid-js'
```
brenelz, line 123: should this be solid-js/web?

docs/router/framework/solid/quick-start.md
birkskyum: there's not note on the react vite plugin, so i think let's just keep the pages as similar as possible.

brenelz: good point i realized the other one didn't have it

docs/router/framework/solid/quick-start.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### `src/routes/__root.tsx`
+
+```tsx
+import { createRootRoute, Link, Outlet } from '@tanstack/solid-router'
+
+export const Route = createRootRoute({
+  component: () => (
+    <>
+      <div className="p-2 flex gap-2">
```
brenelz, line 69: will want to replace className with class

PR #3485, opened 2025-02-19
feat(react-router): final type utilities

chorobin:
This PR addresses performance issues as well as type assertion issues for the new type utilities


SeanCassiere (commented):
Could you check if the content in `navigation.md` and `type-safety.md` as correct and synced with the stuff here?

chorobin:
> Could you check if the content in `navigation.md` and `type-safety.md` is correct and synced with the stuff here?
> 
> Edit: In case, it wasn't clear, my comment is purely on the documentation side.

I don't think either of these docs show how to write your own custom things with types so I think they are ok to keep as they are. What do you think?

docs/router/framework/react/guide/type-utilities.md
```diff
@@ -0,0 +1,197 @@
...
+  return (
+    <>
+      <h1>{props.title}</h1>
+      <Link {...props.linkOptions} />
+    </>
+  )
+}
+```
+
+A more permissive overload of `HeadingLink` is used to avoid type assertions you would otherwise have to do with the generic signature.
```
schiller-manuel, line 34: "more permissive" than what?

schiller-manuel: "type assertions you would otherwise have to do with the generic signature." 
either that needs an example or just remove that. it confuses me :D

docs/router/framework/react/guide/type-utilities.md
```diff
@@ -0,0 +1,197 @@
...
+
+All type parameters for utilities are optional but for the best TypeScript performance `TRouter` should always be specified for the public facing signature. And `TOptions` should always be used at inference sites like `HeadingLink` to infer the `linkOptions` to correctly narrow `params` and `search`
```
schiller-manuel, line 36: > And `TOptions` should always be used at inference sites like `HeadingLink` to infer the `linkOptions` to correctly narrow `params` and `search`

can you bring a negative example? so people see what to avoid

docs/router/framework/react/guide/type-utilities.md
```diff
@@ -0,0 +1,197 @@
...
+        <li>
+          <Link {...item} from={props.from} />
+        </li>
+      ))}
+    </ul>
+  )
+}
+```
+
+`ValidateLinkOptionsArray` allows you to fix `from` by providing an extra type parameter. The result is a type safe array of `Link` options providing relative navigation with a fixed `from`
```
schiller-manuel, line 115: seems like a repetition of the previous sentence somehow?

docs/router/framework/react/guide/type-utilities.md
chorobin: The last overload is the most loose function signature so type assertions are not necessary when spreading options to `Link`

packages/react-router/src/link.tsx
```diff
@@ -937,8 +937,8 @@ export type LinkOptionsFnOptions<
   TRouter extends AnyRouter = RegisteredRouter,
 > =
   TOptions extends ReadonlyArray<any>
```
birkskyum, line 939: should this also be applied in the solid-router?

birkskyum: It's here:
- https://github.com/TanStack/router/pull/3505

PR #3493, opened 2025-02-20
fix(start-client): preload route chunks prior to hydration

schiller-manuel:


sebasptsch:
Thanks for this fix, it fixed an SSR to Client hydration flash for my application!

PR #3508, opened 2025-02-21
media: update header image for solid

incognitojam:


incognitojam:
I don't think the colour profile is the same after importing/exporting with figma :confused:

PR #3509, opened 2025-02-21
docs(router): backfill guides for `solid-router`

SeanCassiere:
Using the replacement strategies we've got for TanStack.com to make the existing docs, work for `solid-router`.


docs/router/framework/solid/guide/custom-link.md
```diff
@@ -0,0 +1,25 @@
+---
+ref: docs/router/guide/custom-link.md
+replace: { 'react-router': 'solid-router' }
+---
+
+[//]: # 'BasicExampleImplementation'
+
+```tsx
+// TODO: Do the Solid equivalent of this.
```
SeanCassiere, line 9: @birkskyum @brenelz could you please add in the Solid specific typings for this?

birkskyum: It would look like this for solid:

```tsx
import { createLink, createRootRoute, Link, LinkComponent, Outlet } from "@tanstack/solid-router";
import { Component, JSX, Suspense } from "solid-js";

  component: RootComponent
});


type BasicLinkProps = JSX.IntrinsicElements["a"] & {
  // Add any additional props you want to pass to the anchor element
};

const BasicLinkComponent: Component<BasicLinkProps> = (props) => (
  <a {...props} class="block px-3 py-2 text-red-700">
    {props.children}
  </a>
);


const CreatedLinkComponent = createLink(BasicLinkComponent)


export const CustomLink: LinkComponent<typeof BasicLinkComponent> = (props) => {
  return <CreatedLinkComponent preload={'intent'} {...props} />
}
```

docs/router/framework/solid/guide/type-utilities.md
```diff
@@ -0,0 +1,5 @@
+---
+ref: docs/router/guide/type-utilities.md
+replace:
+  { 'react-router': 'solid-router', 'React.ReactNode': 'Solid.SolidNode' }
```
SeanCassiere, line 4: @birkskyum @brenelz the equivalent of `React.ReactNode` as an output type.

```tsx
function ReactRuntimeValue(): React.ReactNode {
// 	returns  a valid value that can be used at runtime in react
}
```

docs/router/guide/type-utilities.md
```diff
@@ -173,7 +173,7 @@ export function useConditionalNavigate<
 export function useConditionalNavigate(
   navigateOptions: ValidateNavigateOptions,
 ): UseConditionalNavigateResult {
-  const [enabled, setEnabled] = React.useState(false)
+  const [enabled, setEnabled] = useState(false)
```
birkskyum, line 176: For solid, this `useState` call would be `createSignal`

docs/router/guide/path-params.md
```diff
@@ -0,0 +1,128 @@
...
+If we add a component to our `postRoute`, we can access the `postId` variable from the URL by using the route's `useParams` hook:
+
+```tsx
+export const Route = createFileRoute('/posts/$postId')({
+  component: PostComponent,
+})
+
+function PostComponent() {
+  const { postId } = Route.useParams()
+  return <div>Post {postId}</div>
```
birkskyum, line 67: i believe in solid, it would be a syntax like

```ts
  const params = Route.useParams()
  return <div>Post {params.postId()}</div>
  ```

Comment (SeanCassiere) [reply docs/router/guide/path-params.md:67]: Could this also work?

```ts
<div>{post()}</div>
```

docs/router/framework/solid/guide/scroll-restoration.md
```diff
@@ -0,0 +1,46 @@
+---
+ref: docs/router/guide/scroll-restoration.md
+replace: { 'react-router': 'solid-router' }
+---
+
+[//]: # 'ManualRestorationExample'
+
+```tsx
+// TODO: Port over to Solid
```
SeanCassiere, line 9: cc @birkskyum

SeanCassiere (commented):
With exception to the following guides, everything else had been brought in.

- SSR - https://tanstack.com/router/latest/docs/framework/react/guide/ssr
- Render Optimizations - https://tanstack.com/router/latest/docs/framework/react/guide/render-optimizations

Let me know if those also should be brought in.

---

Other than that, I've backfilled all the other guides with a single comment still being left open.

docs/router/framework/solid/guide/scroll-restoration.md
birkskyum: I'd do it like this:

```tsx
function Component() {
  // We need a unique ID for manual scroll restoration on a specific element
  // It should be as unique as possible for this element across your app
  const scrollRestorationId = 'myVirtualizedContent'

  // We use that ID to get the scroll entry for this element
  const scrollEntry = useElementScrollRestoration({
    id: scrollRestorationId,
  })

  // Let's use TanStack Virtual to virtualize some content!
  let virtualizerParentRef: any;
  const virtualizer = createVirtualizer({
    count: 10000,
    getScrollElement: () => virtualizerParentRef,
    estimateSize: () => 100,
    // We pass the scrollY from the scroll restoration entry to the virtualizer
    // as the initial offset
    initialOffset: scrollEntry?.scrollY,
  })


  return (
    <div
      ref={virtualizerParentRef}
      // We pass the scroll restoration ID to the element
      // as a custom attribute that will get picked up by the
      // scroll restoration watcher
      data-scroll-restoration-id={scrollRestorationId}
      class="flex-1 border rounded-lg overflow-auto relative"
    >
      ...
    </div>
  )
}
```

tannerlinsley:
Lgtm. Merge when you’re ready.

tannerlinsley (approved):

PR #3515, opened 2025-02-22
feat(solid-cross-context): add package

birkskyum:
Port of:
- [react-cross-context](https://github.com/TanStack/router/tree/main/packages/react-cross-context)

Necessary for the solid versions of:
- start-client
- start-server


SeanCassiere (commented):
Please run the `labeler-generate` script at the root and commit its output.

PR #3524, opened 2025-02-22
feat(history): access history entries

matthiasschwarz:
Added a new `entries` function for the memory history to be able to access all entries.
An entry has an `index` property to get the current index of the entry in the entries or `-1` when it was removed. With `getState` the history state of the entry can be received.
Also added a new `entries` test case. 

The api is inspired by the [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation).

### DIFF
--- a/packages/history/src/index.ts
+++ b/packages/history/src/index.ts
@@ -39,6 +39,15 @@ export interface RouterHistory {
   _ignoreSubscribers?: boolean
 }
 
+export interface RouterMemoryHistory extends RouterHistory {
+  entries: () => Array<HistoryEntry>
+}
+
+export interface HistoryEntry {
+  index: number;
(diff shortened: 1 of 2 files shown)


schiller-manuel:
interesting! we should add this to browser history as well by recording the history in session storage

matthiasschwarz:
> interesting! we should add this to browser history as well by recording the history in session storage

I have updated this pull request to allow access to the entries for the browser history as well.

schiller-manuel:
quite a big PR!
can you please explain what this does in detail?

we also need some e2e tests for this

LadyBluenotes:
Closing this as stale. The underlying need is still worth preserving, but `RouterHistory.entries` needs to be brought up to date with main.

Successor issue is #7777.

PR #3526, opened 2025-02-22
preload type fix (naive type change)

ducin:
This PR is meant to bring the `preload` options to reflect the possible values listed in both `defaultPreload` (below) and the docs (https://tanstack.com/router/v1/docs/framework/react/guide/preloading#supported-preloading-strategies).

Currently the `preload` in `createFileRoute` allows only to pass `boolean` which seems wrong (see above):

`UpdatableRouteOptions`:
```
preload?: boolean
```

vs

`RouterOptions`:
```
  defaultPreload?: false | 'intent' | 'viewport' | 'render'
```

# state of this PR

due to flaky e2e tests I'm unable to get them running. I've got one e2e server functions test (`nx run tanstack-start-e2e-server-functions:test:e2e`, unrelated to the change) failing on 30s timeout: https://nx.app/runs/0YaAbzz0il

I need help here.

### DIFF
--- a/packages/react-router/src/route.ts
+++ b/packages/react-router/src/route.ts
@@ -395,7 +395,7 @@ export interface UpdatableRouteOptions<
   pendingMinMs?: number
   staleTime?: number
   gcTime?: number
-  preload?: boolean
+  preload?: false | 'intent' | 'viewport' | 'render'
   preloadStaleTime?: number
   preloadGcTime?: number
   search?: {


ducin:
Hopefully this gets unblocked after https://github.com/TanStack/router/pull/3528 gets merged by @schiller-manuel

schiller-manuel:
we are currently moving stuff into router-core since we now not only have the react impl but also the solid one (more to come...)

so i am a bit hesitant to merge this. if we merge this, we would also need to apply the same change to the solid one.

ducin:
@schiller-manuel how about doing it in the router-core: is it doable now or when (what is the status of the refactor?

schiller-manuel:
the change makes sense. we should however probably have only a single type that we reference for both the default and the route level prop.
the refactor is ongoing, @chorobin  is on it. cannot give you an ETA but we are hoping to have it soon since the duplication right now will not work out.

schiller-manuel:
the refactor that once blocked this is long done, do you want to pick this up again?

schiller-manuel:
the route level `preload` is not the same as the `defaultPreload`. it just allows disabling preloads of routes

PR #3527, opened 2025-02-22
CONTRIBUTING.md: added missing `pnpm exec playwright install`

ducin:
a small addition to contributing.

I faced lack of `pnpm exec playwright install` myself when trying to improve types.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,8 @@
   - `pnpm install`
   - This installs dependencies for all of the packages in the monorepo, even examples!
   - Dependencies inside of the packages and examples are automatically linked together as local/dynamic dependencies.
+- Install test dependencies
+  - `pnpm exec playwright install` (required for e2e tests)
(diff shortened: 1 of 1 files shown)


schiller-manuel:
is playwright not doing this automatically the first time you use it?

ducin:
@schiller-manuel nope

airtonix:
not only this but you'll also (at least in ubuntu linux) need to run this : 

```
sudo -E env "PATH=$PATH" ./node_modules/.bin/playwright install-deps
```

Around the internet you'll see this form given instead: 

```
sudo npx playwright install-deps
```

But this is an issue for a few those of us that don't install nodejs to the root of the OS and instead choose to use tools like mise or asdf (anything that's specific to the current users "$PATH")

I use the above in some helper tools in a few monorepos i run to get people up and running quickly.

schiller-manuel:
@airtonix want to add these instructions as a PR?

PR #3530, opened 2025-02-22
chore: correct deps

schiller-manuel:


packages/directive-functions-plugin/package.json
```diff
@@ -73,12 +73,15 @@
     "@babel/traverse": "^7.26.8",
     "@babel/types": "^7.26.8",
     "@tanstack/router-utils": "workspace:^",
+    "babel-dead-code-elimination": "^1.0.9",
```
SeanCassiere, line 76: Not sure if the renovate bot is correctly doing this or not, but we can probably make use of [pnpm's catalog feature](https://pnpm.io/catalogs) to make sure that the correct versions of babel and other common packages are at the same version across the monorepo.

I'll look into this sometime.

PR #3536, opened 2025-02-23
test: add failing test on passing a server function to useMutation

Balastrong:
This test passes on start `v1.102.1` and fails since `v1.102.2`

### DIFF
--- a/packages/start-client/src/tests/createServerFn.test-d.tsx
+++ b/packages/start-client/src/tests/createServerFn.test-d.tsx
@@ -388,3 +388,19 @@ test('createServerFn can validate FormData', () => {
       >
     >()
 })
+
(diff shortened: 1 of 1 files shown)


Balastrong:
A bit more context as this was initially mentioned on discord, the following code throws a TS error:
```ts
  const updateUserMutation = useMutation({
    mutationFn: updateUser,
  })
```
image


A temporary workaround for `1.111.7`

```ts
const updateUser = createServerFn()
  .validator(UserMetaSchema)
  .handler(async ({ data }) => { ... })

const updateUserMutation = useMutation({
  mutationFn: (data: Parameters<typeof updateUser>[0]) => updateUser(data),
})
```

schiller-manuel:
fixed with https://github.com/TanStack/router/pull/3532

Issue #3521, opened 2025-02-22
SSR with `@solidjs/start`


mrlubos:
### Which project does this relate to?

Router

### Describe the bug

Hey! I am super excited about Solid Router as seen by my timely [comment](https://github.com/TanStack/router/discussions/975#discussioncomment-12269226). I assume this is not currently supported so perhaps a feature request, but how to use the router in SolidStart? Is it the SSR functionality I am requesting? Am I the problem?

### Your Example Website or App

https://stackblitz.com/edit/github-grada5-vrkkfmhj?file=app.config.ts,src%2Fapp.tsx

### Steps to Reproduce the Bug or Issue

Here's a list of changes I've made to an existing project.

1. Modify `app.config.ts`

```ts
import { defineConfig } from "@solidjs/start/config";
import { TanStackRouterVite } from "@tanstack/router-plugin/vite";

export default defineConfig({
  vite: {
    plugins: [
      TanStackRouterVite({
        autoCodeSplitting: true,
        target: "solid",
      }),
    ],
  },
});
```

2. Create `routes/__root.tsx`

```ts
import { Outlet, createRootRoute } from '@tanstack/solid-router'

export const Route = createRootRoute({
  component: RootComponent,
})

function RootComponent() {
  return (
    <>
      <div>Hello "__root"!</div>
      <Outlet />
    </>
  )
}
```

3. Modify `app.tsx`

```ts
import { Router } from "@solidjs/router";
import { FileRoutes } from "@solidjs/start/router";
import { createRouter } from "@tanstack/solid-router";
import { ErrorBoundary, Suspense } from "solid-js";

import { routeTree } from '~/routeTree.gen'

const router = createRouter({ routeTree });

export default function App() {
  return (
    <Router
      root={props => (
        <ErrorBoundary>
          <Suspense>
            {props.children}
          </Suspense>
        </ErrorBoundary>
      )}
    >
      <FileRoutes />
    </Router>
  );
}
```

This is enough to show the following error trace when I load the root page

```
Error: Client-only API called on the server side. Run client-only code in onMount, or conditionally run client-only component with .
    at notSup (file:///path-to-project/node_modules/.pnpm/solid-js@1.9.4/node_modules/solid-js/web/dist/server.js:1138:9)
    at file:///path-to-project/node_modules/.pnpm/@tanstack+solid-router@1.110.0_solid-js@1.9.4/node_modules/@tanstack/solid-router/dist/esm/CatchBoundary.js:3:30
[...]

PR #3541, opened 2025-02-24
fix: solid-router for ssr

brenelz:
Looked at what Tanstack form does and this seems to fix part of the SSR issue. Doesn't solve the whole thing but gets us partially there.

https://github.com/TanStack/form/blob/main/packages/solid-form/tsconfig.build.json

Fixes part of the problem discussed here
https://github.com/TanStack/router/issues/3521


brenelz:
@birkskyum can you see if this maybe fixes the issue you were seeing?

PR #3557, opened 2025-02-25
Devinxi

tannerlinsley:


packages/start-plugin/src/nitro/plugins/dev-server-plugin.ts
```diff
@@ -0,0 +1,124 @@
...
+        viteServer.middlewares.use(async (req, res) => {
+          const template = await viteServer.transformIndexHtml(
+            req.originalUrl as string,
+            `<html>
+              <head></head>
+              <body></body>
+            </html>`,
+          )
+
+          console.log('template, maybe use?', template)
```
SeanCassiere, line 45: If we can, we probably should use this, since it'd have the necessary `@vitejs/plugin-react` bindings baked into it. So we can do away with this workaround and just let the React plugin do its thing.
https://github.com/TanStack/router/blob/4a9d7717dd32f507bc146adafeb7939bd479f95b/packages/start-router-manifest/src/index.ts#L34-L39

packages/start-plugin/src/nitro/nitro-plugin.ts
```diff
@@ -0,0 +1,257 @@
...
+              baseName: 'public',
+              dir: clientOutputPath,
+            },
+          ]
+          nitroConfig.renderer = rendererEntry
+
+          if (ssrBuild) {
+            if (isWindows) {
+              // Write out the renderer manually because
+              // Windows doesn't resolve the aliases
```
SeanCassiere, line 112: `pathe` may be a soluton to this
https://www.npmjs.com/package/pathe

packages/start-plugin/src/nitro/plugins/dev-server-plugin.ts
birkskyum: For solid, I had to remove this section. It's the only five lines making `start-router-manifest` non-agnostic. Alternatively, if the react-refresh could be added conditioned on the vite plugin target (react|solid).

PR #3574, opened 2025-02-26
docs(react-router): update import from `viteReact` to `react`

KoolieKangarooMjolner:
the plugin imported from `@vitejs/plugin-react` is called `react`, not `viteReact`.


schiller-manuel:
its a default import, so you can call it whatever you want

KoolieKangarooMjolner:
aah missed that

KoolieKangarooMjolner:
Anyway: the npm page uses `react` - https://www.npmjs.com/package/@vitejs/plugin-react

SeanCassiere (approved):
Although this change may not necessarily be needed, it'd bring it inline with what's being done for the solid-router setup.

PR #3590, opened 2025-02-27
docs(solid-router) fix examples that import workspace packages

birkskyum:
Follow up to
- https://github.com/TanStack/router/pull/3585

@SeanCassiere the new examples import from workspace which breaks when opening them in stackblitz in the docs. 

This import the published versions, so it works

I also cleaned up references to react router devtools

### DIFF
--- a/examples/solid/basic-file-based/package.json
+++ b/examples/solid/basic-file-based/package.json
@@ -9,9 +9,7 @@
     "start": "vite"
   },
   "dependencies": {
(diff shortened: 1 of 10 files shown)


SeanCassiere:
Cool beans!

Btw, just FYI. Once a code-release goes through, Config would perform a rewrite of the `package.json` files for the examples and automatically write in the new published versions.

Good call on the router-devtools. I mostly left it out of laziness by not removing it from the `package.json` files and root routes 😅

PR #3592, opened 2025-02-27
fix(solid-router): remove delete linkProps.disabled

brenelz:
We were getting a TypeError around this in the Tanstack Solid Start when working with it.

I am not sure if this is a proper fix as maybe something won't work as expected then?

The error we get is:

> TypeError: Cannot delete property 'disabled' of #<Object>

### DIFF
--- a/packages/solid-router/src/link.tsx
+++ b/packages/solid-router/src/link.tsx
@@ -540,12 +540,6 @@ export const Link: LinkComponent<'a'> = (props: any) => {
         })
       : rest.children
 
-  if (typeof local._asChild === 'undefined') {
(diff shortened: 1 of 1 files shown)


birkskyum:
Tests introduced here:
 https://github.com/TanStack/router/pull/1544

Tests refactored here:
https://github.com/TanStack/router/pull/1590

SeanCassiere (approved):
> from discord
> ... I think there was an issue with Anchor Elements in React's runtime receiving the disabled prop

If this works fine for Solid, then 👍🏼

birkskyum:
Excellent - it does.

PR #3607, opened 2025-02-28
feat(react-start-api-routes): Write the matched route path as unenumerable to request object

lforst:
This comes a bit out of nowhere but I am currently building the official Sentry SDK for TSStart and one thing that is really important for tracing/spans/opentelemetry is being able to group stuff by the parameterized route.

I haven't found a nice way in the public API yet to allow reading the parameterized route _in a central location_ that doesn't require users to wrap every single route they create.

This is probably a bit of a hack but it seemed unintrusive enough for me (since it is unenumerable) and is getting the job done wonderfully.


birkskyum:
@lforst , a small heads up, the ts start will likely be having packages for different framworks, so a react-start, [solid-start](https://github.com/TanStack/router/pull/3276) etc., just when choosing name for sentry packages it might be worth putting a "react-" in the sentry sdk for `@tanstack/react-start`.

birkskyum:
@lforst , just an FYI there will likely be tanstack start for multiple frameworks, in case the sentry sdk is react-specific you might want to add `react-` somehwere in the package name. See i.e. early work here on a [solid version](https://github.com/TanStack/router/pull/3522).

lforst:
> @lforst , just an FYI there will likely be tanstack start for multiple frameworks, in case the sentry sdk is react-specific you might want to add `react-` somehwere in the package name. See i.e. early work here on a [solid version](https://github.com/TanStack/router/pull/3522).

That's a great point. I somehow missed that tss will be generic. I'll publish a package with an updated name. Thanks!

lforst:
> for API route I hope we have a single package in the future.

That's dope 👌 

> w.r.t. to the sentry integration: I think it would make sense to offer a generic customization point where sentry could hook in. then this request tag would solely live in e.g. sentry

Just gauging interest in this change in general, would it make sense to add this interim until there is a proper public API? I would be like 0% mad if you all decide to break this change when you switch over to a generic customization.

I am also happy to add a hook somewhere if you give me some guidance on how you would want it to look like 😄 This is kinda super duper mission critical for Sentry so I am happy to sacrifice some time on this.

birkskyum:
> > for API route I hope we have a single package in the future.

> That's dope 👌

@lforst the future is now. 

https://npmjs.com/@tanstack/start-api-routes

lforst:
Noice. How shall we proceed here?

birkskyum:
This PR would have to be rebased, so the change is applied instead to the shared start-api-routes package

lforst:
> This PR would have to be rebased, so the change is applied instead to the shared start-api-routes package

rebased 👍

schiller-manuel:
we'll be adding middleware support for API routes which should cover this use case

lforst:
> we'll be adding middleware support for API routes which should cover this use case

Is there a timeline for that?

lforst:
Discussed on Discord w/ @schiller-manuel that this change will likely become obsolete with global api middlewares.

PR #3622, opened 2025-03-01
feat: provide location in onEnter/onStay/onLeave hooks

jprosevear:
https://discord.com/channels/719702312431386674/1345441363831033960

### DIFF
--- a/packages/router-core/src/route.ts
+++ b/packages/router-core/src/route.ts
@@ -8,7 +8,12 @@ import type {
   RouteMatch,
 } from './Matches'
 import type { RootRouteId } from './root'
-import type { FullSearchSchema, ParseRoute, RouteById, RoutePaths } from './routeInfo'
+import type {
+  FullSearchSchema,
+  ParseRoute,
+  RouteById,
+  RoutePaths,
+} from './routeInfo'
 import type { AnyRouter, RegisteredRouter } from './router'
 import type { BuildLocationFn, NavigateFn } from './RouterProvider'
 import type {
@@ -1035,8 +1040,9 @@ export interface UpdatableRouteOptions<
     {
       location,
     }: {
-      location: 
-        ParsedLocation<ResolveFullSearchSchema<TParentRoute, TSearchValidator>>
+      location: ParsedLocation<
+        ResolveFullSearchSchema<TParentRoute, TSearchValidator>
+      >
     },
   ) => void
   onStay?: (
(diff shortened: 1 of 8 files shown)


examples/react/basic/src/main.tsx
```diff
@@ -14,6 +14,15 @@ import type { ErrorComponentProps } from '@tanstack/react-router'
 import './styles.css'
 
 const rootRoute = createRootRoute({
+  onEnter: (match, { location }) => {
```
jprosevear, line 17: All changes to this file can be removed, just makes testing easier

packages/solid-router/tests/route.test.tsx
```diff
@@ -191,6 +191,27 @@ describe('onEnter event', () => {
 
     expect(fn).toHaveBeenCalledWith({ foo: 'bar' })
   })
+
+  it('should have location defined in router.load()', async () => {
```
schiller-manuel, line 195: rather `should have location defined in onEnter`?

please also add a test for onStay and onLeave

packages/solid-router/src/router.ts
```diff
@@ -791,7 +791,7 @@ export class Router<
     if (
       typeof window !== 'undefined' &&
       'CSS' in window &&
-      typeof window.CSS?.supports === 'function'
+      typeof window.CSS.supports === 'function'
```
schiller-manuel, line 794: don't remove the `?`

packages/router-core/src/route.ts
```diff
@@ -865,6 +865,7 @@ export interface UpdatableRouteOptions<
       >,
       TLoaderDeps
     >,
+    { location }: { location: ParsedLocation<{}> },
```
schiller-manuel, line 868: we should be able to supply the search type here (same for the others as well)

packages/solid-router/tests/route.test.tsx
schiller-manuel: also applies to react-router of course

jprosevear: Was just copying the style from above

packages/router-core/src/route.ts
jprosevear: i think this additional typing is a little out of my immediate grasp - can you give some more guidance

packages/solid-router/tests/route.test.tsx
jprosevear: its in an onEnter describe block

packages/router-core/src/route.ts
schiller-manuel: before we do that, what is the semantic of location in each of these hooks?

I think in case of onLeave location is anything BUT the match, so we cannot statically type the search here. 
whereas for the other hooks, the route's search is present in location

LadyBluenotes:
Thanks for the PR. I'm closing it because it has fallen behind main (sorry it got lost in the shuffle), and the lifecycle implementation has since moved into router-core.

The feature is still useful, so we’ve opened #7778 to track a fresh implementation. A new PR based on that issue would be welcome if you are still interested!

PR #3638, opened 2025-03-02
fix(react-router,solid-router): prevent crashes if sessionStorage isn't available

### LINKED ISSUE #3634: Version 1.98.1 stops working with uncaught SecurityError exception

Router

### Describe the bug

Updating a router package to version 1.98.1 completely breaks router initialization in native context. In my example it breaks it in Figma Plugin context with following error
```
Uncaught SecurityError: Failed to read the 'sessionStorage' property from 'Window': Storage is disabled inside 'data:' URLs.
```

https://github.com/asevich/tanstack-router-bug

1. Install Figma 
2. Build plugin
```
pnpm -r install && pnpm -r build
```
3. Open any Figma document or create a new one
4. Go to Figma menu -> Plugins -> Development -> Import new plugin from manifestImage
5. Select `manifest.json` from the example plugin project. That should show menu. Just click on the `router-kaput` plugin and it should show empty UI. The error is in the console. To open console press Shift + P and select/search for Show/Hide console. 

If router is downgraded to 1.98.0 and plugin is rebuilt then UI will show `working` without any plugin errors in console. 

### Expected behavior

As a user I expect router to work in native context after minor version update

### Screenshots or Videos

_No response_

### Platform

- OS: Windows with WSL2 
- Browser: Figma Sandbox - The sandbox is a minimal JavaScript environment and does not expose browser APIs - https://www.figma.com/plugin-docs/how-plugins-run/
- Version: from Feb 27, 2025

### Additional context

MIght be related to https://github.com/TanStack/router/pull/3191

avendiart:
Fixes #3634 

Simply checking the type of `window.sessionStorage` is apparently enough to trigger `SecurityError`. To address it we use try/catch and set the var to true only if it's indeed available.

### DIFF
--- a/packages/react-router/src/scroll-restoration.tsx
+++ b/packages/react-router/src/scroll-restoration.tsx
@@ -20,7 +20,10 @@ export type ScrollRestorationOptions = {
 }
 
 export const storageKey = 'tsr-scroll-restoration-v1_3'
-const sessionsStorage = typeof window !== 'undefined' && window.sessionStorage
+let sessionsStorage = false
+try {
+  sessionsStorage = typeof window !== 'undefined' && typeof window.sessionStorage === 'object'
+} catch { }
 const throttle = (fn: (...args: Array<any>) => void, wait: number) => {
   let timeout: any
   return (...args: Array<any>) => {
--- a/packages/react-router/src/scroll-restoration.tsx
+++ b/packages/react-router/src/scroll-restoration.tsx
@@ -22,8 +22,9 @@ export type ScrollRestorationOptions = {
 export const storageKey = 'tsr-scroll-restoration-v1_3'
 let sessionsStorage = false
 try {
-  sessionsStorage = typeof window !== 'undefined' && typeof window.sessionStorage === 'object'
-} catch { }
+  sessionsStorage =
+    typeof window !== 'undefined' && typeof window.sessionStorage === 'object'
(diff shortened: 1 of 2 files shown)


birkskyum:
@avendiart are you able to add a test for it?

schiller-manuel:
a better fix would be to only try to access session storage if scroll restoration is enabled

avendiart:
@birkskyum I'm not sure the effort is justified in that particular case nor do I really have that much time for it at the moment, unfortunately. We are talking about an e2e test here with a very specific condition: https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage#securityerror.

I suppose the "easiest" one to reproduce the error would be an origin with `file://` scheme? The test would then need to assert that a) sessionStorage is indeed unavailable and b) no exception is being thrown and the rest of the code works as expected.

avendiart:
@schiller-manuel In my view those are two separate issues. There might be users who would want to use scroll restoration in a context where sessionStorage is not available, in that case it still should not crash. I haven't spent much time looking at the code here, but in general I'd agree that the checks/scroll restoration logic should only run on-demand.

schiller-manuel:
@avendiart a test could be using router in an iframe inside of playwright with ` sandbox="allow-scripts"`

SeanCassiere (approved):

PR #3640, opened 2025-03-02
refactor(router-core): move AnyRoute and AnyRouter to router-core

chorobin:


lo1tuma:
@chorobin will there be something like `AnyReactRouter` and `AnyReactRoute`? Because an instance of the react-router has some methods and properties that are not included by `AnyRouter` from `router-core`. For example `viewTransitionPromise` or `hasNotFoundMatch()`.

PR #3642, opened 2025-03-03
feat: add support for `shouldThrow` in RouteApi

### LINKED ISSUE #3482: shouldThrow option for useParams not working

Router

### Describe the bug

My current route - `'/'`
Im trying to get page params like this:
```
const route = getRouteApi('/_auth/users/$userId')

export function UsersList() {
  const { userId } = route.useParams({ shouldThrow: false })
  ...
}
```
but im getting error - Error: Invariant failed: Could not find an active match from "/_auth/users/$userId"
Image

Also types for `shouldThrow` prop are wrong, it accepts only `true` but shoud accept any `boolean` value
Image

Expecting behavior, that im just get undefined in my `userId` value

-

1. go to '/' url in your project
2. Call `useParams` with `{ shouldThrow: false }` prop for another route for example `'/users/$userId'`
3. get the error

### Expected behavior

expected not getting arror just get undefined value

### Screenshots or Videos

_No response_

### Platform

- OS:  macOS
- Browser: Chrome
- Version: latest

### Additional context

_No response_

schiller-manuel:
resolves #3482

WIP
needs runtime tests

### DIFF
--- a/packages/react-router/src/fileRoute.ts
+++ b/packages/react-router/src/fileRoute.ts
@@ -204,21 +204,21 @@ export class LazyRoute<TRoute extends AnyCoreRoute> {
   }
 
   useSearch: UseSearchRoute<TRoute['id']> = (opts) => {
-    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
     return useSearch({
       select: opts?.select,
       structuralSharing: opts?.structuralSharing,
+      shouldThrow: opts?.shouldThrow,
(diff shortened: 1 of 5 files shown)


LadyBluenotes:
@schiller-manuel did you still want this in? I think it’s close. It would need a rebase, the runtime coverage noted in the draft, and a decision on whether we should fix the same route-bound API in Solid and Vue at the same time.

If you need me to take it over, just lmk.

Issue #3639, opened 2025-03-02
bug(solid-router): useRouterState break client nav


birkskyum:
### Which project does this relate to?

Router

### Steps to Reproduce the Bug or Issue

Open an example:

react-router is working
https://stackblitz.com/github/tanstack/router/tree/main/examples/react/basic-file-based

solid-router is not working
https://stackblitz.com/github/tanstack/router/tree/main/examples/solid/basic-file-based

Add the call to `useRouterState` in the top of the RootComponent

```
function RootComponent() {
  const routerState = useRouterState();

```

And try to change pages using the top navigation

### Expected behavior

That the pages changes as normal

### Platform

- OS:  macOS
- Browser: Chrome

### Additional context

This is a blocker for solid-router devtools

birkskyum:
Narrowed this down to solid-start breaking on:

useRouterState()
useRouterState({ select: (s) => s })

while it works with 

useRouterState({ select: (s) => JSON.stringify(s) })
useRouterState({ select: (s) => s.status })

PR #3643, opened 2025-03-03
feat(solid-router) add structural sharing

birkskyum:
attempt to close:
- #3639


birkskyum:
@chorobin it seems like we don't need structuralSharing for solid, so the only value this PR potentially holds is if these changes, which make solid-router more like react-router, can somehow ease the transition to router-core.

schiller-manuel:
let's close this then?

PR #3644, opened 2025-03-03
fix(history): `parseLocation` in createHashHistory to handle search params correctly

### LINKED ISSUE #3595: The hash history's parseLocation fails to locate search params

Router

### Describe the bug

The hash history's _parseLocation_ method assumes that the search parameters come after the hash, and this can result in losing search parameters if the order is different. See [the code here](https://github.com/TanStack/router/blob/main/packages/history/src/index.ts#L548-L551).

I have an app that's rendered in an iframe, and I was working on implementing the Tanstack Router to it. To construct the app's URL, I used the native `URL` API. But then, when I used its method `toString()`, it put the search parameters before the hash, for example `?q=p#/routePath`. Because of that Tanstack Router kept saying that the required search params were missing although they were in the URL and could be found in the `location.search`.

https://stackblitz.com/edit/github-ky3m9sh1?file=index.html

1. Create an app using Tanstack Router and its hash history
2. Create a route, for example `/hello`.
3. Add a required search param to the route in (2), for example `param: number`
4. Navigate to `${baseUrl}?param=1#/hello`
5. The app fails to render and throws the type error.

### Expected behavior

The order between the hash and the search parameters shouldn't matter when using the hash history.

### Screenshots or Videos

_No response_

### Platform

- OS: MacOS
- Browser: Chrome
- Version: 133.0.6943.127 

### Additional context

_No response_

ohpyupi:
`parseLocation` in createHashHistory failed to parse search params when they appeared before the hash. This PR updates the parsing logic so that search params are correctly extracted and processed, regardless of their position.
fixes #3595

### DIFF
--- a/packages/history/tests/createHashHistory.test.ts
+++ b/packages/history/tests/createHashHistory.test.ts
@@ -4,35 +4,29 @@ import { createHashHistory } from '../src'
 
 describe('createHashHistory', () => {
     describe('parseLocation', () => {
-        test('neither search params nor hash', () => {
-            const history = createHashHistory()
-            window.history.pushState({}, "", "/")
-            expect(history.location.pathname).toBe('/')
-            expect(history.location.search).toBe('')
-        })
-        test('hash present, no search params', () => {
-            const history = createHashHistory()
-            window.history.pushState({}, "", "/#hello")
-            expect(history.location.pathname).toBe('/hello')
-            expect(history.location.search).toBe('')
-        })
-        test('search params present, no hash', () => {
-            const history = createHashHistory()
-            window.history.pushState({}, "", "/?search=params")
-            expect(history.location.pathname).toBe('/')
-            expect(history.location.search).toBe('?search=params')
-        })
-        test('both hash and search params present, in that order', () => {
-            const history = createHashHistory()
-            window.history.pushState({}, "", "/#hello?search=params")
-            expect(history.location.pathname).toBe('/hello')
-            expect(history.location.search).toBe('?search=params')
-        })
-        test('both search params and hash present, in that order', () => {
-            const history = createHashHistory()
-            window.history.pushState({}, "", "/?search=params#hello")
-            expect(history.location.pathname).toBe('/hello')
-            expect(history.location.search).toBe('?search=params')
+        describe.each([
+            ['/', {pathname: '/', search: ''}, 'neither search params nor hash'],
+            ['/#hello', {pathname: '/hello', search: ''}, 'hash present, no search params'],
+            ['/?search=params', {pathname: '/', search: '?search=params'}, 'search params present, no hash'],
+            ['/#hello?search=params', {pathname: '/hello', search: '?search=params'}, 'both hash and search params present, in that order'],
+            ['/?search=params#hello', {pathname: '/hello', search: '?search=params'}, 'both search params and hash present, in that order'],
+        ])('check for %s', (...[path, exp, desc]) => {
+            test(`onLoad with ${path} (${desc})`, () => {
+                const mockWindow = {
+                    addEventListener: window.addEventListener,
+                    history: window.history,
+                    location: new URL(`https://www.example.com${path}`)
+                }
+                const history = createHashHistory({window: mockWindow})
+                expect(history.location.pathname).toBe(exp.pathname)
+                expect(history.location.search).toBe(exp.search)
+            })
+            test(`onNavigate with ${path} (${desc})`, () => {
+                const history = createHashHistory()
+                window.history.pushState({}, "", path)
+                expect(history.location.pathname).toBe(exp.pathname)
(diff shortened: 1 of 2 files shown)


packages/history/tests/createHashHistory.test.ts
```diff
@@ -0,0 +1,24 @@
+import { describe, expect, test } from 'vitest'
+
+import { createHashHistory } from '../src'
+
+describe('createHashHistory', () => {
+  test('parseLocation', () => {
```
SeanCassiere, line 6: Please break this one large test function into individual ones.

```ts
describe('createHashHistory', () => {
	describe('parseLocation',  () => {
		test('...', () => {})

		test('...', () => {})

		// ...
	})
})
```

Maybe add some into a matrix where it checks both "on load's `parseLocation` call" and "after pushState's `parseLocation`" values.

SeanCassiere (changes requested):
Thank you for making this contribution. 

Just need a few adjustments to the tests.

packages/history/tests/createHashHistory.test.ts
ohpyupi: Sure. I'll break one test into smaller ones. Can you elaborate more on the other suggestion "add some into a matrix ...". I don't think I understand that.

SeanCassiere: Maybe the word matrix wasn't the right word here. Essentially, what I'm talking about is checking the behaviours in both "on load" and "on navigate".

Something like this for example.

```ts
describe.each([
	['/', { pathname: '', search: '' }],
	// ...
])('check for %i', ([path, exp]) => {
	// check if parseLocation works in initial load
	test(`onLoad with ${path}`, () => {
		window.location.href=path
		const history = createHashHistory()
		expect(...).toBe(exp['...'])
	})

	// check if parseLocation works after navigate
	test(`onNavigate with ${path}`, () => {
		const history = createHashHistory()
		window.history.pushState({}, null, path)
		expect(...).toBe(exp['...'])
	})
})
```

ohpyupi: Oh okay, thanks for the explanation. Sure will work on that.

packages/history/tests/createHashHistory.test.ts
```diff
@@ -0,0 +1,32 @@
...
+            ['/#hello', {pathname: '/hello', search: ''}, 'hash present, no search params'],
+            ['/?search=params', {pathname: '/', search: '?search=params'}, 'search params present, no hash'],
+            ['/#hello?search=params', {pathname: '/hello', search: '?search=params'}, 'both hash and search params present, in that order'],
+            ['/?search=params#hello', {pathname: '/hello', search: '?search=params'}, 'both search params and hash present, in that order'],
+        ])('check for %s', (...[path, exp, desc]) => {
+            test(`onLoad with ${path} (${desc})`, () => {
+                const mockWindow = {
+                    addEventListener: window.addEventListener,
+                    history: window.history,
+                    location: new URL(`https://www.example.com${path}`)
```
ohpyupi, line 18: @SeanCassiere The search params update triggers the page refresh, but such functionality isn't implemented in JSDOM, and it just throws an error. Thus, it requires mocking over the location.

I found [this stackover flow post](https://stackoverflow.com/questions/54021037/how-to-mock-window-location-href-with-jest-vuejs) and took its idea to use the URL API to mock the location API. Please let me know if you have some concerns or other better ideas for it. Thanks.

SeanCassiere: It might just be missing the correct setup for the testing env.

Could you refactor to using `window.location` and then ping me? I'll fix up the deps and get the testing environment working.

ohpyupi: So the JSDOM didn't implement the navigation functionality by design (see [here](https://github.com/jsdom/jsdom/issues/2112)).

Basically, the `window` object is an object mocked by JSDOM because it exists only in the browser environment. Are you saying that the such mocking exists under the `packages/react-router`? For all test cases under that package, it uses `history` object instead of window.location. Could you point me to the code block if I missed it?

SeanCassiere: You are right, my mistake. I was thinking about this https://github.com/TanStack/router/blob/35c7c495f38f89675a5570a9b4fa35e63f0a3525/packages/react-router/tests/link.test.tsx#L51

In that case, don't mock `window`, just do a `window.(push/replace)State` before calling `createHashHistory` for the "on-load" tests.

ohpyupi: Alright, I've updated the PR. Please take a look when you get some time. Thanks.

SeanCassiere (approved):

SeanCassiere:
@ohpyupi thank you for your contribution!!!

ohpyupi:
@SeanCassiere Thanks!

PR #3651, opened 2025-03-03
fix: solid router crashing with useRouterState

### LINKED ISSUE #3639: bug(solid-router): useRouterState break client nav

brenelz:
Seems to fix the issue of `useRouterState` causing the client side navigation to break. Not sure why exactly this causes it to break but all the unit tests seem to pass without it.

Also helps unblock the devtools that @birkskyum is working on

Fixes #3639

### DIFF
--- a/e2e/react-router/basic-file-based/src/routes/__root.tsx
+++ b/e2e/react-router/basic-file-based/src/routes/__root.tsx
@@ -5,6 +5,7 @@ import {
   createRootRoute,
   useCanGoBack,
   useRouter,
+  useRouterState,
 } from '@tanstack/react-router'
 import { TanStackRouterDevtools } from '@tanstack/router-devtools'
 
@@ -23,6 +24,8 @@ export const Route = createRootRoute({
 function RootComponent() {
   const router = useRouter()
   const canGoBack = useCanGoBack()
+  // test useRouterState doesn't crash client side navigation
+  const _state = useRouterState()
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


schiller-manuel:
we probably don't have tests for this particular scenario. but I think this is needed if you trigger a preload of a route by e.g. hovering a link and then quickly click the link (or have a sloooow beforeLoad)
then you have an overlap of loading the same route twice and you need to finish the first load with this line.

brenelz:
Interesting. Something is going wrong here  as it was getting a maximum call stack exceeded error. 

My assumption was someone it was getting in a loop of creating promises and resolving them

birkskyum (approved):

birkskyum:
Additional work on fixing this applied here:
- https://github.com/TanStack/router/commit/dff7e7a8e6823ca2c317e22d9c9a1135834fdf09

PR #3669, opened 2025-03-05
chore(examples): add authenticated-routes example using firebase

downright-development:
I struggled greatly getting firebase client-side authentication to play nicely inside TSR. This is a quick example on setting up your project. In this example, I'm using firebase oauth with Github. :) For this example you work you *MUST* create a firebase project and setup github oauth.  

I *personally* prefer using the directory format below for organizational reasons, but I left it flat to align with other demos.

_auth/
  protected-route-1 (dashboard)
(public)/
  login.tsx
  other public routes like signup

*Due to conflicts with AuthProvider being a type in firebase, I've updated the context to "AuthContextProvider" in this example*

The index.tsx in this example is a great place to host your sales/landing page... if a user is logged in and attempts to hit the root index, they'll get routed back to /dashboard, otherwise they can directly route to index OR the /login page.


downright-development:
Update - I realized Zustand wasn't actually necessary. Resolved to remove zustand dependency!

examples/react/authenticated-routes-firebase/.gitignore
```diff
@@ -0,0 +1,5 @@
+node_modules
+.DS_Store
+dist
+dist-ssr
+*.local
```
SeanCassiere, line 5: Please add `.env` but exclude `.env.example`. Not sure if this is already available further up the tree for the monorepo, but since this example needs to be clonable, it'd be best to add it in.

examples/react/authenticated-routes-firebase/.env.example
```diff
@@ -0,0 +1,6 @@
+VITE_FIREBASE_API_KEY=
```
SeanCassiere, line 1: Please add a comment that says pretty much this:

```
# copy this file over as `.env.local` and fill in these values from the Firebase console.
```

examples/react/authenticated-routes-firebase/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "serve": "vite preview",
+    "start": "vite"
+  },
+  "dependencies": {
+    "@tanstack/react-router": "^1.112.13",
+    "@tanstack/router-devtools": "^1.112.13",
+    "@tanstack/router-plugin": "^1.112.13",
+    "autoprefixer": "^10.4.20",
+    "firebase": "^11.4.0",
+    "lucide-react": "^0.477.0",
```
SeanCassiere, line 17: Can we drop the icons library?

+ `lucide-react` has frequent updates

examples/react/authenticated-routes-firebase/src/routes/__root.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import { Outlet, createRootRouteWithContext } from '@tanstack/react-router'
+import { TanStackRouterDevtools } from '@tanstack/router-devtools'
+import type { AuthContextType } from '../auth'
+
+
```
SeanCassiere, line 6: If possible, please keep the `MyRouterContext` with auth as a key.

Essentially, similar to what we already have going. Currently, this would make it difficult to inject Query into the context.

i.e.
```ts
export type MyRouterContext = {
	auth: AuthContextType
}
```

examples/react/authenticated-routes-firebase/src/main.tsx
```diff
@@ -0,0 +1,71 @@
...
+		login: () => Promise.resolve(), // This will be set after we wrap the app in AuthContextProvider
+		logout: () => Promise.resolve(), // This will be set after we wrap the app in AuthContextProvider
+  },
+})
+
+// Register things for typesafety
+declare module '@tanstack/react-router' {
+  interface Register {
+    router: typeof router;
+    context: {
```
SeanCassiere, line 29: This should not be needed. Since the generic in `createRootRouteWithContext` should be enough.

examples/react/authenticated-routes-firebase/src/routes/login.tsx
```diff
@@ -0,0 +1,106 @@
...
+function LoginComponent() {
+  const router = useRouter();
+	const { login } = useAuth();
+
+	const handleSignIn = async (provider: "github" | "apple" | "google") => {
+		console.log(`Clicked ${provider} sign in!`);
+		try {
+			const providers = {
+				google: new GoogleAuthProvider(),
+				github: new GithubAuthProvider(),
```
SeanCassiere, line 40: Probably best to just leave in GitHub. The pattern would already be established and the user can add whatever else they need.

Atmost, GitHub and Google.

SeanCassiere:
Looking at the test failure, it looks like you might want to add a `src/vite.d.ts` file with all the environment variables.

```ts
/// <reference types="vite/client" />

interface ImportMetaEnv {
  readonly VITE_CUSTOM_ENV_VARIABLE: string
}

interface ImportMeta {
  readonly env: ImportMetaEnv
}
```

examples/react/authenticated-routes-firebase/package.json
downright-development: Yeah we can drop this, I think I'm only using this for a loading spinner TBH

examples/react/authenticated-routes-firebase/src/main.tsx
downright-development: I think I noticed in the IDE that "return <RouterProvider router={router} context={{ ...auth }} />" on line 51 was red squiggled without this. I'll double check though and can report back

SeanCassiere: See my comment in `__root.tsx`, this should solve this.

examples/react/authenticated-routes-firebase/package.json
SeanCassiere: Even just `"Loading..."` is enough 👍🏼

examples/react/authenticated-routes-firebase/.gitignore
downright-development: .env.* is uptree in the monorepo, but this is resolved in next commit !

examples/react/authenticated-routes-firebase/.env.example
downright-development: Resolved in next commit!

examples/react/authenticated-routes-firebase/src/routes/__root.tsx
downright-development: I can update to put stuff under auth - I only went the other direction because it became annoying TBH, BUT... good shout about my approach being more difficult to inject query into the context... shoving under "auth" is less annoying now :D Let me work on resolving this one.

examples/react/authenticated-routes-firebase/src/main.tsx
downright-development: Okay, ignore me. I must've been decaffeinated - this will be resolved in next commit. 👍

examples/react/authenticated-routes-firebase/src/routes/login.tsx
downright-development: Resolved, removed google and apple left a comment you can add other providers there though.

examples/react/authenticated-routes-firebase/src/routes/__root.tsx
downright-development: Resolved!

examples/react/authenticated-routes-firebase/package.json
downright-development: Removed and replaced!

PR #3672, opened 2025-03-05
chore: replace `<React.Fragment>` by `<>`

schiller-manuel:


packages/router-generator/src/template.ts
```diff
@@ -59,7 +59,7 @@ export function getTargetTemplate(target: Config['target']): TargetTemplate {
               '%%tsrImports%%',
               '\n\n',
               '%%tsrExportStart%%{\n component: RootComponent\n }%%tsrExportEnd%%\n\n',
-              'function RootComponent() { return (<React.Fragment><div>Hello "%%tsrPath%%"!</div><Outlet /></React.Fragment>) };\n',
+              'function RootComponent() { return (<><div>Hello "%%tsrPath%%"!</div><Outlet /></>) };\n',
```
SeanCassiere, line 62: Just FYI, the removal of `React.Fragment` here leaves you with `React is defined by not used.`. The Eslint default is to error it out I believe (not too sure on this one).

Of course, this is eslint setup-dependent, but I'd leave it in. No need to create a new Route file and instantly have red squigglys.

PR #3678, opened 2025-03-06
feat(router-generator): disable formatting of routeTree

honzahruby:
Based on discussion in https://github.com/TanStack/router/discussions/3664, I'm adding an option to disable formatting of the generated routeTree. The file is not meant for human consuption and disabling the formatter can significantly speed up the route generation process.


schiller-manuel:
wouldnt we want to disable formatting by default?

also the config name is pretty long, how about just `formatRouteTree` ?

docs/router/api/file-based-routing.md
```diff
@@ -202,3 +203,9 @@ By default, this value is set to:
 This option allows you to disable the generation of the manifest.
 
 By default, this value is set to `false`.
+
+### `disableGeneratedRouteTreeFormatting`
+
+This options turns off formatting of the generated route tree which can be time-consuming for large project.
```
SeanCassiere, line 209:
```suggestion
This option turns off the formatting function on the generated route tree file, which can be time-consuming for large projects.
```

SeanCassiere (commented):
As much as the name is quite verbose, it's very explicit in what it says it's going to do. 

I'd probably keep it the same, or if anything shorten it down to `disableRouteTreeFormatting`.

---

Naming is hard 🫣

SeanCassiere:
> wouldnt we want to disable formatting by default?

This is actually interesting 🤔 

Maybe just `enableRouteTreeFormatting`  and have it be defaulted to `true`.

That'd actually be a good pattern going forward and to refactor for in v2. Don't disable, just enable. I'll probably use it for disabling route manifest generation by default.

honzahruby:
Sounds good, but I don't really want to introduce a breaking change, so for now I will rename it to `disableRouteTreeFormatting` and update the docs. Hope that's okay with you 🙏

SeanCassiere:
@honzahruby please rename to `enableRouteTreeFormatting` and default it to `true`. I'll be renaming the other `disable*` flags to `enable*`, with backwards compact whilst we are in the v1 cycle.

Your introduction here, won't be a breaking change since it's a new feature that'd be defaulted to  `true` which is the current behaviour.

For your use case, you can set it to `false` in your project.

docs/router/api/file-based-routing.md
```diff
@@ -25,6 +25,7 @@ The following options are available for configuring the file-based routing:
 - [`routeTreeFileHeader`](#routetreefileheader)
 - [`routeTreeFileFooter`](#routetreefilefooter)
 - [`disableManifestGeneration`](#disablemanifestgeneration)
+- [`disableRouteTreeFormatting`](#disablegeneratedroutetreeformatting)
```
SeanCassiere, line 28: 👀

SeanCassiere (changes requested):

SeanCassiere (approved):

SeanCassiere:
@honzahruby thank you for this 🚀

honzahruby:
Thanks for your help and support, I’m happy to see this land and speed up our project further 💪

PR #3695, opened 2025-03-08
fix(router-devtools-core): lazy load panel

birkskyum:


birkskyum:
superseded by 
- #3696

Issue #3479, opened 2025-02-19
Hash history router doesn't scroll to top when visiting a new page


kottkrig:
### Which project does this relate to?

Router

### Describe the bug

When using hash history via `createHashHistory`, navigation to new pages doesn't scroll to the top of the page. Instead it seems to keep the scroll position from the original page.

### Your Example Website or App

https://stackblitz.com/edit/github-4rybtq-5mekkq2j?file=src%2Froutes%2Findex.tsx,src%2Froutes%2Fabout.tsx,src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

1. Scroll down
2. Click on "Go to about"
3. Confirm that you are in a scrolled state

### Expected behavior

As a user, I expect to go to the top of /about when I haven't been there before. Instead I end up scrolled to the middle of the page.

### Platform

- OS: macOS
- Browser: Arc
- Version: Chromium Engine Version 133.0.6943.35

### Additional context

The incorrect behaviour is only visible when using "hash history". When using default browser history the navigation behaves as expected.

SeanCassiere:
Had to revert #3700 in #3702 due to an unintended bug that it introduced in Start.

BrendanC23:
@SeanCassiere Do you have any updates about when this might get merged again?

BrendanC23:
Thanks for the fix!

PR #3702, opened 2025-03-09
revert(react-router,solid-router): revert #3700

SeanCassiere:
#3700 introduced an unintended side-effect with the scroll restoration process in TanStack Start.

This revets the code changes in `@tanstack/react-router` and `@tanstack/solid-router` and short-circuits the tests relevant test scripts introduced in the offending PR.

### DIFF
--- a/e2e/react-router/scroll-restoration-sandbox-vite/package.json
+++ b/e2e/react-router/scroll-restoration-sandbox-vite/package.json
@@ -10,7 +10,7 @@
     "serve": "vite preview",
     "start": "vite",
     "test:e2e:browser": "VITE_APP_HISTORY=browser playwright test --config=playwright.browser.config.ts --project=chromium",
-    "test:e2e:hash": "VITE_APP_HISTORY=hash playwright test --config=playwright.hash.config.ts --project=chromium",
+    "test:e2e:hash": "exit 0; VITE_APP_HISTORY=hash playwright test --config=playwright.hash.config.ts --project=chromium",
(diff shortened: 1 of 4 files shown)


flipvh:
Hi @SeanCassiere, great to see you working even in the weekends 🫡 . I was also investigating a hash issue but not for hash history but for URLs with hash in general. Since a few days the first when clicking Link or doing `navigate` in my app, it doesnt seem to directly move to anchor. It seems to only happen on fresh page init and on second click it does work. Could this be related? I'd love to help and provide a minimal reproduction, what can I use best to achieve this - perhaps there is already an example with hash links? Thnks!

SeanCassiere:
> Hi @SeanCassiere, great to see you working even in the weekends 🫡 . I was also investigating a hash issue but not for hash history but for URLs with hash in general. Since a few days the first when clicking Link or doing `navigate` in my app, it doesnt seem to directly move to anchor. It seems to only happen on fresh page init and on second click it does work. Could this be related? I'd love to help and provide a minimal reproduction, what can I use best to achieve this - perhaps there is already an example with hash links? Thnks!

Possibly could be. 

We have a couple issues open regarding scroll restoration. Could you check those out? 

If yours doesn't fit those, then yes please, a reproduction would be great.

BrendanC23:
@SeanCassiere Can you describe the problems caused by https://github.com/TanStack/router/pull/3700 and what the "unintended side-effect" of it was? I'd like to work on a fix for https://github.com/TanStack/router/issues/3479, but I don't know what went wrong with the original PR.

Issue #60, opened 2021-11-13
Example - Basic with React Query is throwing an error


burhanuday:
**Describe the bug**
Example is crashing with an error
```
_reactLocation.useIsNextLocation is not a function
```

**To Reproduce**
1. Go to [Basic with react query example](https://react-location.tanstack.com/examples/basic-with-react-query)
2. See the sandbox that loads
3. See error

**Expected behavior**
Example should not crash

**Screenshots**
<img width="1091" alt="image" src="https://user-images.githubusercontent.com/25934701/141604167-280f0436-5546-4b87-8f94-74ec2eb4ff07.png">

**Desktop (please complete the following information):**
 - OS: MacOS Big Sur
 - Browser: Firefox, Chrome
 - Version: 94

**Smartphone (please complete the following information):**
 - Browser: [e.g. stock browser, safari]
 - Version: [e.g. 22]

**Additional context**

tannerlinsley:
Should be fixed now.

PR #3708, opened 2025-03-09
fix: extract props from validator issues

jkazimierczak:
## Changes Summary

Replaces naive JSON serialization of validator issues by extracting [known properties](https://github.com/standard-schema/standard-schema?tab=readme-ov-file#the-interface) first, and then serializing extracted values instead of the error itself.

This makes is possible to avoid errors such as validation library having non-serializable errors.

Changes introduced based on the [Discord discussion](https://discord.com/channels/719702312431386674/1346288913521705002).

## How

Introduces [utilities](https://github.com/standard-schema/standard-schema/tree/main/packages/utils) provided by Standard Schema to the core package, adds `path` attribute which schema defines, exports the normalization function and uses it within the `execValidator.`

## Result

The validation result is provided under message key in the error response:
```json
{
  "$error": {
    "message": "{\n  \"root\": [\n    \"must be an object (was a string)\"\n  ],\n  \"issues\": {}\n}",
    "stack": "... error stacktrace ...",
    "cause": { "$undefined": 0 }
  }
}
```

The validation result is split into two parts - root and issues. `root` aggregates all validation messages that happened at the root (as explained in [When is Issue.path allowed to be undefined? #60](https://github.com/standard-schema/standard-schema/issues/60)), while `issues` contains all errors for which the `path` was present.

### Docs

Maybe it would be good to reword a [docs](https://github.com/standard-schema/standard-schema/issues/60), to mention that `validator` integrate seamlessly with external validators, that implement Standard Schema?

```diff
-Validators also integrate seamlessly with external validators, if you want to use something like Zod.
+Validators also integrate seamlessly with external validators that implement Standard Schema, which includes Zod, Valibot or Arktype among others. Full list of spec-compliant libraries can be found [here](https://github.com/standard-schema/standard-schema?tab=readme-ov-file#what-schema-libraries-implement-the-spec).
```

### DIFF
--- a/packages/react-start-client/src/createServerFn.ts
+++ b/packages/react-start-client/src/createServerFn.ts
@@ -1,6 +1,7 @@
 import { default as invariant } from 'tiny-invariant'
 import { default as warning } from 'tiny-warning'
 import { isNotFound, isRedirect } from '@tanstack/react-router'
+import { normalizeValidatorIssues } from '@tanstack/router-core'
 import { mergeHeaders } from './headers'
 import { globalMiddleware } from './registerGlobalMiddleware'
 import { startSerializer } from './serializer'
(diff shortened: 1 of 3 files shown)


schiller-manuel:
yes the docs update as mentioned makes sense!

LadyBluenotes:
Thanks for putting this together! I’m closing this PR because it's quite stale, and validation now runs through start-client-core rather than the file changed here.

The serialization problem still matters, so I’ve opened #7779 to track a current implementation that covers both server-function and search validation, safe issue normalization, tests, documentation, and bundle-size validation. A fresh PR based on that issue would be welcome :)

PR #3718, opened 2025-03-10
chore(start-router-manifest): agnostic package

birkskyum:
It's not necessary to have react/solid packages for the router manifest - it rely only on the shared router-core


birkskyum:
package is deleted in devinxi, so no reason to work on it

PR #3726, opened 2025-03-10
refactor: use router-core in server-functions packages

birkskyum:


packages/react-start-client/src/createServerFn.ts
```diff
@@ -1,4 +1,4 @@
-import { isNotFound, isRedirect } from '@tanstack/react-router'
+import { isNotFound, isRedirect } from '@tanstack/router-core'
```
graup, line 1: This throws an error for me: `The requested module '@tanstack/router-core' does not provide an export named 'isNotFound'`
(on @tanstack/router-core 1.114.7, @tanstack/react-router 1.114.7, @tanstack/react-start 1.114.10)

birkskyum: There'll be 1.114.12 of everything in a moment - that should resolve it.

Issue #3719, opened 2025-03-10
Fresh create-rs-router app or


jimjeffers:
### Which project does this relate to?

Router

### Describe the bug

When I attempt to launch the local server using `yarn start` (via fresh create-ts-router app) or `yarn dev` (via ts router quick start) I get the following error when I add the `TanStackRouterVite` plugin:

```
Error: Build failed with 2 errors:
node_modules/@tanstack/react-router-devtools/dist/esm/TanStackRouterDevtools.js:2:43: ERROR: Could not resolve "@tanstack/router-devtools-core"
node_modules/@tanstack/react-router-devtools/dist/esm/TanStackRouterDevtoolsPanel.js:3:48: ERROR: Could not resolve "@tanstack/router-devtools-core"
    at failureErrorWithLog (/Users/jimjeffers/Sites/my-app/node_modules/esbuild/lib/main.js:1477:15)
    at /Users/jimjeffers/Sites/my-app/node_modules/esbuild/lib/main.js:946:25
    at /Users/jimjeffers/Sites/my-app/node_modules/esbuild/lib/main.js:1355:9
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  errors: [Getter/Setter],
  warnings: [Getter/Setter]
}
```

In my config I have the following:

```
import { defineConfig } from "vite";
import viteReact from "@vitejs/plugin-react";
import { TanStackRouterVite } from '@tanstack/router-plugin/vite'

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [TanStackRouterVite({
    target: 'react',
    autoCodeSplitting: true,
  }),
  viteReact()],
  test: {
    globals: true,
    environment: "jsdom",
  },
  build: {
    rollupOptions: {
      external: ['@tanstack/router-devtools-core']
    }
  }
});
```

Note I tried to move dev tools into external deps and still had no luck. When I clicked the link below for the file-based routes starting point on stackblitz the build fails with the same error without making any changes!

```
> dev
> vite --port 3000


♻️  Generating routes...
✅ Processed routes in 233ms

  VITE v6.2.1  ready in 1606 ms

  ➜  Local:   http://localhost:3000/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
✘ [ERROR] Could not resolve "@tanstack/router-devtools-core"

    node_modules/@tanstack/react-router-devtools/dist/esm/TanStackRouterDevtools.js:2:43:
[...]

PR #3732, opened 2025-03-10
fix(solid-router-devtools,react-router-devtools): add core package and solid-js as deps

birkskyum:
Attempt to close
- https://github.com/TanStack/router/issues/3719

### DIFF
--- a/packages/react-router-devtools/package.json
+++ b/packages/react-router-devtools/package.json
@@ -61,6 +61,10 @@
   "engines": {
     "node": ">=12"
   },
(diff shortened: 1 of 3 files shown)


graup:
Wait, why is solid-js a dependency of the react-router-devtools package? It's a copy-paste error right?

birkskyum:
The devtools is a solidjs app that is mounted from react/solid/other-framework

graup:
Oh, I see, thanks. :+1: I was confused by the new dependency in my react app.

PR #3737, opened 2025-03-11
fix: view transitions in solid-router

brenelz:


brenelz:
@birkskyum does this make sense?

brenelz:
Doesn't seem to be the right fix

PR #3753, opened 2025-03-12
Add "response" option details to server-functions.md

Aleksion:
Added information about "response" option and returning non-standard return types


docs/start/framework/react/server-functions.md
```diff
@@ -472,6 +511,65 @@ export const getServerTime = createServerFn({ method: 'GET' }).handler(
 )
 ```
 
+For more advanced control over responses, especially with streaming data, you can use the `response: 'raw'` option:
```
schiller-manuel, line 514: please change the above example as well. you MUST specify `response: 'raw'` when you want to return a raw response

Aleksion: Done

> please change the above example as well. you MUST specify `response: 'raw'` when you want to return a raw response

Issue #3746, opened 2025-03-12
TanStackRouterDevtools: `toggleButtonProps` is no longer applied


melv-n:
### Which project does this relate to?

Router

### Describe the bug

This is a regression.

Passing `toggleButtonProps` to `TanStackRouterDevtools` no longer works, at least the `className` doesn't.

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-fvj7mimg?file=src%2Froutes%2F__root.tsx

### Steps to Reproduce the Bug or Issue

1. Inspect the button html for the TanStackRouterDevTools
2. No className applied

### Expected behavior

I expect `toggleButtonProps` to pass on the `className` like it used to.

### Platform

macOS / Arc

schiller-manuel:
cc @birkskyum

PR #3768, opened 2025-03-13
fix(router-devtools-core): propagate toggleButtonProps properly

birkskyum:
Closes https://github.com/TanStack/router/issues/3746

With this change, setting on the DevTools the properties:
- panelProps
- closeButtonProps
- toggleButtonProps

Will propagate those values, such as `style` or `class`  correctly to the respective DOM elements.


melv-n:
Thanks @birkskyum!

PR #3779, opened 2025-03-15
fix(history): double `/` being added in hash routing

### LINKED ISSUE #3699: Context provided in a route's `beforeLoad` is not provided to `loader` functions when using hash history

Router

### Describe the bug

When providing a context via a route's `beforeLoad`, its own or any child route's `loader` function does not receive the updated context values, but only when using the hash history.

https://stackblitz.com/edit/github-cpnrbu8i?file=src%2Fmain.tsx

1. Open the preview page in its own tab (you need to do this to manipulate the hash route)
2. When viewing the root page (no # value), you should see that the string "success" is available via route context hook and also in the `loader` function.
3. Navigate to the "About" page, then *refresh* so that the router initializes with a # value
4. You should see "Loader Value is: undefined" and the message "beforeLoadValue from context is: undefined" in console because the value wasn't set from the `beforeLoad`. Note that it is available in the useRouterContext hook, just not the route's `loader` function.

### Expected behavior

The values set from the `beforeLoad` function should be available in the context passed to `loader`.

### Screenshots or Videos

_No response_

### Platform

- OS: Linux
- Browser: Firefox
- Version: 136.0

### Additional context

Tanstack router version 1.114.1

Note again that this only happens when using the hash history, the browser history works fine.

### LINKED ISSUE #3739: When using hash history, a slash is added to the href when changing window.location.hash and search params before the hash get added to href

Router

### Describe the bug

PR #3644 for issue #3595 introduced two breaking changes in [v1.112.8](https://github.com/TanStack/router/releases/tag/v1.112.8) :
- an additional slash is added in some cases
- search params are mixed between "server search params" and "hash search params"

I have to agree with the opinion of Tanner in #1449, maybe this breaking change should not have been introduced.
I think the additional slash can be safely removed, and maybe the new search behaviour can be an option ?

I'll try and create a PR for this if that's OK

https://stackblitz.com/edit/github-ujeikeh5?file=src%2Froutes%2Findex.tsx

1. Go to About
2. Navigate to Other using the yellow or orange links
3. See that the href displayed is //other instead of /other (also the header link is not "active")
4. If you add a search param in the preview url (for instance https://wmaibpqbzggithub-fcsj--3000--495c5120.local-corp.webcontainer.io/**?toto=tutu**#/about), navigating using window.location.hash or <a> tags results in an href value of //other?toto=tutu

### Expected behavior

The previous behaviour was that when using window.location.hash = '#/about' or a regular <a>, the href would have only one /.

Also the search params sent to the server (before the # character) are now received by the router and are not separate from the "search params" that come after the # character. I do have a use case in my application where I use regular search params so that the server can dynamically change the source of my SPA script, and "hash search params" so that my @tanstack/react-router routes can use them for navigation. I do agree it is a pretty uncommon use case

### Screenshots or Videos

_No response_

### Platform

- OS: [Windows]
- Browser: [Edge, Firefox]
- Version: [Edge 134.0.3124.51, Firefox 136.0]

I don't think the issue is platform specific

### Additional context

_No response_

AlexProgrammerDE:
This more cleanly and safely constructs a href with search parameters and prevents incorrectly adding a double slash to the start of the URL that was introduced in #3644.

Fixes #3739
Fixes #3699

### DIFF
--- a/packages/history/src/index.ts
+++ b/packages/history/src/index.ts
@@ -546,11 +546,12 @@ export function createHashHistory(opts?: { window?: any }): RouterHistory {
   return createBrowserHistory({
     window: win,
     parseLocation: () => {
-      const hashSplit = win.location.hash.split("#").slice(1)
-      const pathPart = hashSplit[0] ?? "/"
+      const hashSplit = win.location.hash.split('#').slice(1)
+      const pathPart = hashSplit[0] ?? '/'
       const searchPart = win.location.search
       const hashEntries = hashSplit.slice(1)
-      const hashPart = hashEntries.length === 0 ? "" : `#${hashEntries.join("#")}`
+      const hashPart =
+        hashEntries.length === 0 ? '' : `#${hashEntries.join('#')}`
       const hashHref = `${pathPart}${searchPart}${hashPart}`
       return parseHref(hashHref, win.history.state)
     },
--- a/packages/history/src/index.ts
+++ b/packages/history/src/index.ts
@@ -546,9 +546,12 @@ export function createHashHistory(opts?: { window?: any }): RouterHistory {
   return createBrowserHistory({
     window: win,
     parseLocation: () => {
-      const search = win.location.search
-      const pathname = win.location.hash.substring(1)
-      const hashHref = search ? `/${pathname}${search}` : `/${pathname}`
+      const hashSplit = win.location.hash.split("#").slice(1)
+      const pathPart = hashSplit[0] ?? "/"
+      const searchPart = win.location.search
+      const hashEntries = hashSplit.slice(1)
+      const hashPart = hashEntries.length === 0 ? "" : `#${hashEntries.join("#")}`
+      const hashHref = `${pathPart}${searchPart}${hashPart}`
       return parseHref(hashHref, win.history.state)
     },
     createHref: (href) =>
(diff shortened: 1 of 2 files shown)


schiller-manuel:
we need some unit tests to prevent future regressions

AlexProgrammerDE:
It seems the original unit tests the author made were faulty at testing. I'll have to investigate where it went wrong.

AlexProgrammerDE:
The original tests the author made were incorrect. They didn't use actual hash based routing. They tested that for example `/#hello` resolves to -> `/hello`, however that's not how hash based routing works. Instead `/#/hello` -> `/hello` should be tested. That's what I've now modified the tests to do.

AlexProgrammerDE:
Question to the maintainers: How *should* `/#hello` be treated in this case? Should `/` be added by the router is it's missing after the hash or should it remain undefined like before?

SeanCassiere:
> Question to the maintainers: How _should_ `/#hello` be treated in this case? Should `/` be added by the router if `/` missing after the hash or should it remain undefined behaviour like before?

The intended behaviour is definitely for there to be a `/` after the root hash when using hash history.

i.e: `/hello-world` would become `/#/hello-world`.

AlexProgrammerDE:
Alright. Then the PR is good like this.

PR #3800, opened 2025-03-18
refactor(router-core): moving internals of Router to router core

birkskyum:
The router-core has a new RouterCore class, that the frameworks' Router class just extend

If there were framework specific methods on the Router, it would be trivial to overwrite those this way

This reduces the size of the adapter packages by a third (e.g. packages/react-router/src from 7338 -> 4844 LOC)


birkskyum:
There is a challenge here.

The Link component lose it's type inference, for instance on the onClick event of e2e/react-start/basic/src/components/DefaultCatchBoundary.tsx, because it can't tell any longer it's in a react context.

We have to explicit type it

chorobin:
> @chorobin , it's _only_ types that are failing here. If I added the types in the user code e2e/example, then it works fine. would you be able to help debug how we can make the react-router Link have react types again?

Unfortunately `createRouter` needs to return `RouteCore` in core. This is because `RegisteredRouter` checks for this, not the specific `Router` type. I know that `Router` in extends `RouterCore` and there is a way to get this work but at the cost of large structural comparisons causing some large regressions in TS performance. Specific framework methods need to be augmented into `RouterCore` currently to avoid this

PR #3813, opened 2025-03-19
fix(solid-router): nav flickering and view transitions

brenelz:
I think this should fix the nav flicker some people have brought up. This also seems to fix the view transitions issue

Basically a tweak of this pr https://github.com/TanStack/router/pull/3589/files

### DIFF
--- a/packages/solid-router/src/lazyRouteComponent.tsx
+++ b/packages/solid-router/src/lazyRouteComponent.tsx
@@ -96,19 +96,22 @@ export function lazyRouteComponent<
       throw error
     }
 
-    const [compResource] = createResource(load, {
-      initialValue: comp,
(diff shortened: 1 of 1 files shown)


brenelz:
@birkskyum can you review this?

brenelz:
This happens because the resource causes it to suspend. I think we only need to do this if there is not a comp

birkskyum:
Repro:

Flicker: https://tanstack.com/router/latest/docs/framework/solid/examples/basic-solid-query-file-based
Doesn't Flicker: https://tanstack.com/router/latest/docs/framework/solid/examples/kitchen-sink-file-based

birkskyum (approved):
Does resolve the flickering

PR #3830, opened 2025-03-22
refactor: Make (zod|valibot|arktype)-adapter agnostic

AhmedBaset:
`@tanstack/react-router` remains a dev dependency because it's still used in tests

### DIFF
--- a/packages/arktype-adapter/package.json
+++ b/packages/arktype-adapter/package.json
@@ -65,6 +65,7 @@
   },
   "devDependencies": {
     "@tanstack/react-router": "workspace:*",
+    "@tanstack/router-core": "workspace:^",
     "arktype": "^2.1.7",
     "@testing-library/jest-dom": "^6.6.3",
     "@testing-library/react": "^16.2.0",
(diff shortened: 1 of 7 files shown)


schiller-manuel:
hmm we should deprecate valibot and arktype adapter since they are not required anymore. zod we still need, but just for the fallback. 

@SeanCassiere 
@chorobin

SeanCassiere:
> hmm we should deprecate valibot and arktype adapter since they are not required anymore. zod we still need, but just for the fallback.

Yup, I agree.

We should be able to merge this "as-is" and ask Tanner to mark `@tanstack/arktype-adapter` and `@tanstack/valibot-adapter` as deprecated on NPM.

@schiller-manuel maybe after devinxi we delete the zombie packages from the repo as a whole? or just throw an error asking the user to uninstall the package since it's no longer needed?

schiller-manuel:
wouldn't throw an error, just delete them and not publish them anymore

SeanCassiere:
> wouldn't throw an error, just delete them and not publish them anymore

👍🏼

birkskyum:
> @schiller-manuel maybe after devinxi we delete the zombie packages from the repo as a whole?

I reckon we can do that now, if it hasn't already been cleaned up?

LadyBluenotes:
I’m closing this because the adapter direction has changed since it was opened: ArkType and Valibot adapters should be retired rather than made more broadly consumable.

They are still referenced by examples and documentation, so I’ve opened #7780 to track whether we still want to go through with the migration and deprecation work against the current package layout.

PR #3833, opened 2025-03-22
fix(docs): Update link for the The Router Configuration: Preloading & Caching

agussetyar:
There are broken links in the Tanstack Start "Build from Scratch: The Router Configuration" section.
I believe this should goes into Tanstack Router documentation.


birkskyum:
this is a consequence of Start docs being broken out from Router docs.

PR #3856, opened 2025-03-25
feat: add biome lint ignore comment to routegen

danteissaias:
Biome just dropped their [2.0 beta](https://biomejs.dev/blog/biome-v2-0-beta/) and you can now add a comment to ignore all lint rules in a file. It would be cool if Tanstack router should included that comment in the generated routes file, like it does for ESLint.

### DIFF
--- a/packages/router-generator/src/config.ts
+++ b/packages/router-generator/src/config.ts
@@ -27,6 +27,7 @@ export const configSchema = z.object({
       '/* eslint-disable */',
       '// @ts-nocheck',
       '// noinspection JSUnusedGlobalSymbols',
+      '// biome-ignore-all'
     ]),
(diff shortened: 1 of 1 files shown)


SeanCassiere (commented):
Normally, I wouldn't be opposed to adding this in, but we've been asking people to add any custom rules by utilizing the `routeTreeFileHeader` option, since we plan to drop most of these default directives in V2.

The plan for v2 is to drop these rules/directives and explicitly point the users at an FAQ page (or something similar to what we have [here](https://tanstack.com/router/latest/docs/framework/react/routing/installation-with-vite#ignoring-the-generated-route-tree-file)) essentially pointing users at a guide to ignore the `routeTree.gen.ts` file.

danteissaias:
Ah cool, that makes sense. I'll close this PR then, thanks! 👍

PR #3868, opened 2025-03-27
chore(examples): add i18n example with Lingui

depsimon:
This PR adds an example of tanstack start with i18n. This is using [Lingui](https://lingui.dev/).


schiller-manuel:
please strip this example down to the bare minimum that is necessary to show i18n using lingui.
i.e. remove API routes, most of the other routes etc.

LadyBluenotes:
Thanks for putting this together, and sorry it took us a while to get back to you!

We’re still open to a Lingui example. Could you update this PR against current main and current dependencies? Once it is up to date, we can take another look.

PR #3872, opened 2025-03-27
fix(react-router-with-query): skip client-side redirect handling during ssr

actuallyjamez:
When a query function throws a redirect during ssr, it causes mismatch errors on the client. This seems to be because `react-router-with-query` applies its client-side redirect handling in server contexts.

This change skips the client-side redirection handling during ssr, preventing these mismatches.


schiller-manuel:
we need e2e tests for this

actuallyjamez:
closing this. I haven't had time to validate this is still a problem since the new with query package.

PR #3893, opened 2025-03-30
solid-router-with-query

birkskyum:
There's an issue with solid-query missing:

CreateQueryOptions.queryKey
CreateQueryOptions.queryKeyHashFn


birkskyum:
I'm working on getting solid-query up to speed with react-query so that this can proceed.

Issue #3655, opened 2025-03-04
Search Params plus and whitespace issue


marvinroger:
### Which project does this relate to?

Router

### Describe the bug

If you have a `http://localhost:3001/search?q=red%2Byellow+orange` URL, according to search params serialization, `q` should be `red+yellow orange`. However, inside `validateSearch`, we currently end up with `red+yellow+orange`, which, in my understanding, is invalid.

So I have to replace `+` with ` `, but the actual `+` would also be replaced with a whitespace, making `+` actually unusable in search params.

Am I missing something?

### Your Example Website or App

https://stackblitz.com/edit/tanstack-react-router-whitespace-issue?file=src%2Froutes%2Findex.tsx

### Steps to Reproduce the Bug or Issue

Create a route and log the param in `validateSearch`. Try with `?q=red%2Byellow+orange` and you'll see printed `{ q: 'red+yellow+orange' }`.

### Expected behavior

I would have expected to see printed `{ q: 'red+yellow orange' }` or the raw `{ q: 'red%2Byellow+orange' }`.

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/0896e431-1689-42fe-8b7f-507012e0d6ce)

### Platform

- OS: macOS 15.3.1
- Browser: Chromium 133.0.6943.142
- Version: 1.112.8

### Additional context

Thank you for this great library, the type-safety is a game-changer!

IanVS:
For now as a workaround I'm using `search.mystring.replaceAll('+', ' ')`, but that won't work if I actually want to have a `+` in the final string.  I don't think it's possible to completely work around this since encoding information is being lost inside of tanstack router itself.

nbibler:
I'm running into this issue as well.

```javascript
export const Route = createRootRoute({
  component: RootComponent,
  errorComponent: ErrorComponent,
  notFoundComponent: NotFoundComponent,
  validateSearch: zodValidator(
    z.object({ name: z.string().optional() }),
  ),
});
```

Given a request to `/?name=Test+Value`:

```javascript
const { name } = Route.useSearch(); // => "Test+Value"
```

Expected it to be "Test Value", received "Test+Value". Ultimately, `+` signs are not being unescaped when being read.

schiller-manuel:
ideally can you provide a failing unit test as a PR?

nbibler:
```diff
diff --git a/packages/zod-adapter/tests/index.test.tsx b/packages/zod-adapter/tests/index.test.tsx
[...]

PR #3894, opened 2025-03-31
fix(router-core): search parsing of `+` using `URLSearchParams`

marvinroger:
Closes https://github.com/TanStack/router/issues/3655

It turns out the `qss` package has a bug when handling the plus sign.

```typescript
const queryString = 'q=red%2Byellow+orange'

// { q: 'red+yellow orange' }
const nativeParsing = new URLSearchParams(queryString)

// { q: 'red+yellow+orange' }
const qssParsing = qss.decode(queryString)
```

Notice the `+` between `yellow` and `orange`.

This PR fixes the issue by using `URLSearchParams` directly, while doing a bit more work to allow the special handling of arrays.

I think [URLSearchParams is widely available enough](https://caniuse.com/urlsearchparams), but I'm not sure what are the rules about browser compatibility here.


SeanCassiere (commented):
I'm not opposed to this.

A minor level of concern that this would would on the old browsers (as mentioned by @marvinroger), but there's a higher chance of React 19 causing issues first I'd think.

I'd also still keep the license and mention that we're taking the API/interfaces for `qss` and reimplementing them with modern browser standards whilst still keeping the same interface.

---

I'm wondering whether this bit in `toValue` is still needed 🤔

```ts
  const str = hasUriEncodedChars(mix)
    ? decodeURIComponent(mix)
    : decodeURIComponent(encodeURIComponent(mix))
```

@marvinroger could you test this out?

SeanCassiere (approved):

marvinroger:
Sorry, I did not find the time to make these adjustments. Thanks for the last mile!

PR #3912, opened 2025-04-01
fix(router-core): return meta if match.status notFound

### LINKED ISSUE #3657: Route head function not called if rendering notFoundComponent

Router

### Describe the bug

Note: 
we are attempting to bump from `1.86.1` to `1.112.11`

Context:
We are running our own SSR server and have not yet migrated to start BUT the repro provided is using start.

The issue only occurs on SSR (test case is simulating google bot/crawlers without JS)

I have a case where we have a head fn on a route that can provide meta data for both matched or notFound depending on the data return from loaderData. The notFound is thrown from a child route and bubbles up as I would expect it to BUT the head fn on the route is no longer called as I would expect. This is only an issue when we are rendering the notFoundComponent, the standard use case still works as expected.

This had been working previously without issues. Any advice?

https://github.com/GregoryCollett/tanstack-ssr-head-bug

* Run app
* Set network conditions to simulate google bot
* Disable Javascript
* Go to posts
* Navigate through posts and see the title is changing as expected
* Navigate through to non-existing-post and see the meta data is not updated as expected and the head function is not called.

### Expected behavior

As a developer, I expect the head function to be called so that I can render relevant meta data but I am seeing that the head function is not called at all meaning I cannot output relevant meta data e.g. Post not found

### Screenshots or Videos

_No response_

### Platform

- OS: ALL
- Browser: ALL
- Version: N/A

### Additional context

_No response_

roduyemi:
Fixes #3657

Updates the match `meta`, `links`, `script` etc. if  in case of `match.status` being `notFound` or an error.

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -1443,7 +1443,7 @@ export class RouterCore<
       // If it's already a success, update headers and head content
       // These may get updated again if the match is refreshed
       // due to being stale
-      if (match.status === 'success') {
+      if (match.status === 'success' || match.status === 'notFound') {
         match.headers = route.options.headers?.({
           loaderData: match.loaderData,
         })
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -1440,18 +1440,19 @@ export class RouterCore<
         }
       }
 
-      // If it's already a success or notFound, update headers and head content
+      // If the match status is already error, notFound or success, update headers and head content
       // These may get updated again if the match is refreshed
       // due to being stale
-      if (match.status === 'success' || match.status === 'notFound') {
+      if (['error', 'notFound', 'success'].includes(match.status)) {
+        const { loaderData } = match
         match.headers = route.options.headers?.({
-          ...(match?.loaderData && { loaderData: match.loaderData }),
+          ...(loaderData && { loaderData }),
         })
         const assetContext = {
           matches,
           match,
           params: match.params,
-          loaderData: match.loaderData,
+          ...(loaderData && { loaderData }),
         }
         const headFnContent = route.options.head?.(assetContext)
         match.links = headFnContent?.links
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -1444,16 +1444,13 @@ export class RouterCore<
       // These may get updated again if the match is refreshed
       // due to being stale
       if (['error', 'notFound', 'success'].includes(match.status)) {
-        const { loaderData } = match
-        match.headers = route.options.headers?.({
-          ...(loaderData && { loaderData }),
-        })
         const assetContext = {
           matches,
           match,
           params: match.params,
-          ...(loaderData && { loaderData }),
+          loaderData: match.loaderData,
         }
+        match.headers = route.options.headers?.(assetContext)
         const headFnContent = route.options.head?.(assetContext)
         match.links = headFnContent?.links
         match.headScripts = headFnContent?.scripts
@@ -2644,9 +2641,7 @@ export class RouterCore<
                           const headScripts = headFnContent?.scripts
 
                           const scripts = route.options.scripts?.(assetContext)
-                          const headers = route.options.headers?.({
-                            loaderData,
-                          })
+                          const headers = route.options.headers?.(assetContext)
 
                           updateMatch(matchId, (prev) => ({
                             ...prev,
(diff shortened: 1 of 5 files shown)


packages/router-core/src/router.ts
```diff
@@ -1443,7 +1443,7 @@ export class RouterCore<
       // If it's already a success, update headers and head content
       // These may get updated again if the match is refreshed
       // due to being stale
-      if (match.status === 'success') {
+      if (match.status === 'success' || match.status === 'notFound') {
```
schiller-manuel, line 1446: while this will invoke the meta fns, the types of those functions are now incorrect I think, as loaderdata is not necessarily defined

roduyemi: Thanks @schiller-manuel, may you please help with the correct fix for this? We're keen to update to the latest version but this is a blocker for us 🙏🏿

schiller-manuel: I am not sure. just making the loader data optional would be the easiest ...
@SeanCassiere  why don't we pass in the full `AssetFnContextOptions` into `headers`?

roduyemi: If it always updated the meta tags (success or error) that'd be great! So it should update for `success`, `error` and `notFound` but not `redirected` or `pending`?

schiller-manuel: probably? unless there is a reason not to update them?

roduyemi: Nope, will make those changes and thanks for the speedy response!

packages/router-core/src/router.ts
```diff
@@ -1440,18 +1440,19 @@ export class RouterCore<
...
-      // If it's already a success, update headers and head content
+      // If the match status is already error, notFound or success, update headers and head content
       // These may get updated again if the match is refreshed
       // due to being stale
-      if (match.status === 'success') {
+      if (['error', 'notFound', 'success'].includes(match.status)) {
+        const { loaderData } = match
         match.headers = route.options.headers?.({
-          loaderData: match.loaderData,
+          ...(loaderData && { loaderData }),
```
SeanCassiere, line 1449: Why is this bit changing? Same question for the bit down below?

schiller-manuel:
please pass `AssetFnContextOptions` into `headers` as well
then make `loaderData` in `AssetFnContextOptions`optional

packages/router-core/src/router.ts
schiller-manuel: yeah seems not necessary.

roduyemi:
> please pass `AssetFnContextOptions` into `headers` as well then make `loaderData` in `AssetFnContextOptions`optional

Thank you for clarifying the change you’re expecting.

packages/router-core/src/router.ts
schiller-manuel: no sorry i was referring to the user facing types

packages/router-core/src/router.ts
```diff
@@ -2643,9 +2641,7 @@ export class RouterCore<
                           const headScripts = headFnContent?.scripts
 
                           const scripts = route.options.scripts?.(assetContext)
-                          const headers = route.options.headers?.({
-                            loaderData,
-                          })
+                          const headers = route.options.headers?.(assetContext)
```
schiller-manuel, line 2644: are the meta functions down here really executed when an error is thrown? dont we have to run them in the catch block as well?

if my hunch is right, why are your tests not failing?

roduyemi: The match.meta is updated here as expected but is there another change you're expecting? I can extend the test to add scripts and links etc too if you want?

schiller-manuel: I was just wondering what happens if the loader throws. it would not reach those lines, so headers/links/ etc would not get executed, right?

roduyemi: These lines are executed now because the condition has been updated to include a match.status of `error` or `notFound` 🙂

schiller-manuel: are you on discord? let's discuss there

https://discord.com/invite/WrRKjPJ

roduyemi: Please let me know if there's anything you'd like me to do to get this moved forward 🙏🏿

schiller-manuel:
superseded by https://github.com/TanStack/router/pull/4189

PR #3914, opened 2025-04-01
fix(react-router): restore scroll position onResolved

### LINKED ISSUE #3741: <Link> with hash prop not working as expected since 1.106.0

Router

### Describe the bug

I noticed for a while that a Link with hash prop isnt working anymore to scroll to that position in the scroll window. In my own code it works when I click it twice. But in the link below it doesnt work at all since 1.106.0: https://github.com/TanStack/router/compare/v1.105.5...v1.106.0

https://stackblitz.com/edit/tanstack-router-wzvsy46k?file=package.json

1. Go to page 'Layout B'
2. Click 'middle' on the top
3. Notice it doesnt scroll to the anchor.
4. Now downgrade to 1.105.5
5. repeat step 1 and 2 and notice it works.

### Expected behavior

Using <Link  hash=""> should scroll to the anchor.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser:  Chrome
- Version: 135

### Additional context

_No response_

t1nky:
Fixes #3741 

Added hash scroll handling directly in the React `Transitioner` component's `onResolved`, similar to how it's implemented in the Solid version

### DIFF
--- a/packages/router-core/src/scroll-restoration.ts
+++ b/packages/router-core/src/scroll-restoration.ts
@@ -317,6 +317,14 @@ export function setupScrollRestoration(router: AnyRouter, force?: boolean) {
   })
 }
 
+/**
+ * @internal
+ * Handles hash-based scrolling after navigation completes.
+ * To be used in framework-specific <Transitioner> components during the onResolved event.
+ *
+ * Provides hash scrolling for programmatic navigation when default browser handling is prevented.
+ * @param router The router instance containing current location and state
+ */
 export function handleHashScroll(router: AnyRouter) {
   if (typeof document !== 'undefined' && (document as any).querySelector) {
     const hashScrollIntoViewOptions =
--- a/packages/router-core/src/scroll-restoration.ts
+++ b/packages/router-core/src/scroll-restoration.ts
@@ -316,3 +316,17 @@ export function setupScrollRestoration(router: AnyRouter, force?: boolean) {
     }
   })
 }
+
+export function handleHashScroll(router: AnyRouter) {
+  if (typeof document !== 'undefined' && (document as any).querySelector) {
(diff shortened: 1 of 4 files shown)


SeanCassiere (commented):
If I'm not mistaken, this exact same logic is present on `solid-router`.

Perhaps this is logic that can be exported out of `router-core` and then imported (plus setup) in both `react-router` and `solid-router`?

t1nky:
> If I'm not mistaken, this exact same logic is present on `solid-router`.
> 
> Perhaps this is logic that can be exported out of `router-core` and then imported (plus setup) in both `react-router` and `solid-router`?

Good point. I've extracted it into scroll-restoration

birkskyum (approved):

packages/router-core/src/scroll-restoration.ts
```diff
@@ -316,3 +316,17 @@ export function setupScrollRestoration(router: AnyRouter, force?: boolean) {
     }
   })
 }
+
+export function handleHashScroll(router: AnyRouter) {
```
SeanCassiere, line 320: Couple of small changes here please:

- Rename to `handleTransitionerHashScroll`
- Annotate the function with some JSDoc comments marking it as `@internal` with a short description saying that this is to be setup in the `<Transitioner>` component and what it actually does.

SeanCassiere (changes requested):

packages/router-core/src/scroll-restoration.ts
SeanCassiere: Optionally, the rename is not necessary if just the JSDoc comment being there.

SeanCassiere (approved):

SeanCassiere:
@t1nky thank you for contributing this fix!

Issue #3680, opened 2025-03-06
scrollRestoration not working the first time


anthony2261:
### Which project does this relate to?

Router

### Describe the bug

With `scrollRestoration: true`, the scroll won't get reset on first page load, but after opening another page via `<Link>` or similar, it will work. I think the bug might not always happen.

### Your Example Website or App

https://stackblitz.com/edit/github-yeosauxb?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

- Open stackblitz app, scroll to the bottom of the home page (which load by default)
- Click on `Click Me`
- It should open the bottom of the About page, even though scrollRestoration is enabled
- If the top of the About page opens, try again by going to the home page, refreshing the app, and repeating the steps above

### Expected behavior

I expect the page I just navigated to to start from the top, but instead my scroll is at the bottom.

### Screenshots or Videos

https://github.com/user-attachments/assets/35b41d4f-f67f-462e-9933-ec8983ee5a99

### Platform

- OS: WSL2
- Browser: Edge, Firefox
- Version: 1.112.13

schiller-manuel:
please provide a complete minimal example by forking one of the existing examples on stackblitz

anthony2261:
@schiller-manuel added stackblitz example, reproduction steps and a video 🤝

Krasnopir:
i have another  case, when scroll restoration does not scroll to top when i trigger navigate to the  same url second  and  more times.
I have a button with click handler, where i make some post request to get id, and  then navigate to /checkout/${id}
after history.back() and trying to press that  button again i see  that my checkout page is not on the  top.

I find that it  start from 1.112.13.
In 1.112.12  it  works  fine

So maybe  this commit  broke some logic ? 

https://github.com/TanStack/router/pull/3659/files

JoshuaHD:
> i have another case, when scroll restoration does not scroll to top when i trigger navigate to the same url second and more times. I have a button with click handler, where i make some post request to get id, and then navigate to /checkout/${id} after history.back() and trying to press that button again i see that my checkout page is not on the top.
> 
> I find that it start from 1.112.13. In 1.112.12 it works fine
> 
> So maybe this commit broke some logic ?
> 
> https://github.com/TanStack/router/pull/3659/files

I can confirm the issue. @Krasnopir Thank you for pointing out the working version. I rolled back to 1.112.12 after trying the entire day way the scroll behavior was so weird and random.

denys-isaichenko:
[...]

Issue #3874, opened 2025-03-28
Can't render in non-dom environments


nick-potts:
### Which project does this relate to?

Router

### Describe the bug

I'd like to use this as the router for my Shopify Extensions, however they restrict the rendering environment and don't allow dom elements. 

This package nearly works in memory mode, however the <Match/> component renders as <script/> tag. Ideally there would be a way to run this library in a way where it never uses dom elements. In this particular case, hydration never occurs so this script tag is not required.

https://github.com/TanStack/router/blob/main/packages/react-router/src/Match.tsx#L141

### Your Example Website or App

n/a

### Steps to Reproduce the Bug or Issue

Setting up a Shopify Extension project isn't super easy. But reproducing is easy, because it always renders a <script/> tag in the root <Match/>

### Expected behavior

I'd imagine this is already the expected behaviour, but an option to disable would be great.

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

### Additional context

https://www.npmjs.com/package/@remote-ui/react
Here's what shopify uses for remote rendering. Basically they provide elements you can use to write to the dom, but you cannot use anything outside of that (like regular dom elements).

schiller-manuel:
did you enable scroll restoration? because only then that script tag will be rendered.
=> so not enabling scroll restoration should work for you

nick-potts:
Can confirm it works fine with that enabled. I don't know how I missed that.

nick-potts:
Weirdly still doesn't work after a single navigation despite using a memory router - gives "Invariant failed: Could not find routeId for matchId".

PR #3918, opened 2025-04-02
fix(react-router): fix issue with scroll restoration

denys-isaichenko:
fixes https://github.com/TanStack/router/issues/3680

It seems like `resolvedLocation` does not always point to the latest location state in the ref callback; sometimes it points to the previous location and `onRendered` is not triggered.

### DIFF
--- a/packages/react-router/src/Match.tsx
+++ b/packages/react-router/src/Match.tsx
@@ -113,10 +113,10 @@ export const Match = React.memo(function MatchImpl({
           </ResolvedCatchBoundary>
         </ResolvedSuspenseBoundary>
       </matchContext.Provider>
-      {parentRouteId === rootRouteId && router.options.scrollRestoration ? (
+      {parentRouteId === rootRouteId ? (
         <>
           <OnRendered />
-          <ScrollRestoration />
+          {router.options.scrollRestoration && <ScrollRestoration />}
         </>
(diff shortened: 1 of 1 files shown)


packages/react-router/src/Match.tsx
```diff
@@ -113,10 +113,10 @@ export const Match = React.memo(function MatchImpl({
           </ResolvedCatchBoundary>
         </ResolvedSuspenseBoundary>
       </matchContext.Provider>
-      {parentRouteId === rootRouteId && router.options.scrollRestoration ? (
```
schiller-manuel, line 116: this change is a bit problematic

see 

https://github.com/TanStack/router/discussions/3469

https://github.com/TanStack/router/issues/3874

denys-isaichenko: @schiller-manuel It's just a bit strange that if I don't enable scrollRestoration, I won't be able to subscribe to onRendered router events. From my perspective, those shouldn't be connected. I'm not sure what the best solution here is. Maybe it can be just useEffect since we don't need to have access to DOM. What do you think?

schiller-manuel: i think it's just meant as an internal event really. so for now please dont apply this change until we clarify this.

PR #3933, opened 2025-04-03
fix(router-core): Fixed decoding of "+" in search params

DreierF:
Hi,

While trying to migrate an existing React Router application to TanStack Router I ran into a bug in qss, which I first reported in qss (https://github.com/lukeed/qss/issues/15), but then noticed that react-router already uses a fork of it, so I fixed the issue in the fork here.


schiller-manuel:
we already have this PR lined up: https://github.com/TanStack/router/pull/3894
so this wont be necessary here I guess

DreierF:
Ah, perfect 🙂

PR #3936, opened 2025-04-04
Fix: Return query data from loader in Router Context docs

johurul-haque:
```diff
export const Route = createFileRoute('/todos')({
  component: Todos,
  loader: async ({ context }) => {
-   await context.queryClient.ensureQueryData({
+   return await context.queryClient.ensureQueryData({
      queryKey: ['todos', { userId: user.id }],
      queryFn: fetchTodos,
    })
```

### DIFF
--- a/docs/router/framework/react/guide/router-context.md
+++ b/docs/router/framework/react/guide/router-context.md
@@ -161,7 +161,7 @@ Then, in your route:
 export const Route = createFileRoute('/todos')({
   component: Todos,
   loader: async ({ context }) => {
-    await context.queryClient.ensureQueryData({
+    return await context.queryClient.ensureQueryData({
       queryKey: ['todos', { userId: user.id }],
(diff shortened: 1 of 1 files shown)


SeanCassiere:
@johurul-haque thank you for the contribution, but in this instance the docs are correct. 

If you are using TanStack Query as your asynchronous-state-manager, then you should not be consuming the returned value from `Route.useLoaderData()`. Instead you should be accessing the data/values directly from TanStack Query using the `useQuery` or `useSuspenseQuery` hooks.

The reason for this is because, if you were to consume the value of `Route.useLoaderData()`, then there is a high potential for there to be a mismatched between the value returned from the route's `loader` versus what's actually in the `QueryCache`. Plus, you lose out on the features/behaviours inherent to Query like their background refetching, cache invalidation and refetching, etc.

PR #3938, opened 2025-04-04
docs: validation path params

raphaelcunha:
Add Validation Path params instructions.

### DIFF
--- a/docs/router/framework/react/guide/path-params.md
+++ b/docs/router/framework/react/guide/path-params.md
@@ -111,6 +111,39 @@ function Component() {
 
 Notice that the function style is useful when you need to persist params that are already in the URL for other routes. This is because the function style will receive the current params as an argument, allowing you to modify them as needed and return the final params object.
 
+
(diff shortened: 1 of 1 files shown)


SeanCassiere (commented):
This doesn't seem to comply with what's actually happening in the types
https://github.com/TanStack/router/blob/d1e78bb5f6c2883a72568f9fa8c8275ffecdf8c4/packages/router-core/src/route.ts#L180-L183
Or at runtime
https://github.com/TanStack/router/blob/d1e78bb5f6c2883a72568f9fa8c8275ffecdf8c4/packages/router-core/src/router.ts#L1202-L1203

schiller-manuel:
@SeanCassiere should work because zod schemas have a `.parse` method

SeanCassiere:
> @SeanCassiere should work because zod schemas have a `.parse` method

My issue is that this makes it looks like a StandardSchema Input.

Versus actually specifying the `.stringify` and `.parse` methods "correctly".

PR #3967, opened 2025-04-09
feat(router): Add useHistoryState hook for type-safe state management

naoya7076:
# Proposal
- https://github.com/TanStack/router/discussions/4113
# Overview
This PR introduces the new `useHistoryState` hook and related functionality to enable type-safe history state management in TanStack Router.

- https://github.com/TanStack/router/discussions/975#discussioncomment-10414075
- https://github.com/TanStack/router/discussions/284#discussioncomment-9344482

## Background

In the current implementation of TanStack Router, to pass type-safe state (History State) between routes that isn't part of the path or params, we need to define types globally. This leads to two main issues:

1. Type pollution - types appear in places where they aren't needed
2. Interface bloat - as more states are added, the global interface grows excessively

For example:

```tsx
  interface HistoryState {
    name?: string;
    project?: MyProjectType;
  }
}
```

https://github.com/TanStack/router/discussions/284
When migrating from react-router to TanStack Router, developers want to be able to handle state in a type-safe manner, similar to how they could with react-router:

```jsx
const history = useHistory();
history.replace({ pathname: '/', state: { name: 'Lily' }});
...
// do something with location.state...
```

This PR addresses these issues by adding route-specific, type-safe history state management.

## Changes

- Added `useHistoryState` hook export in the main package index
- Extended the `FileRoute` class to support state validation with `TStateValidator` generic parameter
- Created a full example in `examples/react/basic-history-state` demonstrating usage patterns
- The example shows how to:
  - Define state schemas using Zod
  - Pass state during navigation with Link and navigate

## Key features of useHistoryState

- Type-safe access to history state passed during navigation
- Support for default values and optional properties
- Ability to select specific state properties
- Non-strict mode for accessing raw state object

## Example usage

Users can define state validation on routes:
```tsx
validateState: (input) =>
  z.object({
    color: z.enum(['white', 'red', 'green', 'blue']).default('white'),
    visited: z.boolean().default(false),
    message: z.string().optional(),
  }).parse(input),
```

Then use it in components:
```tsx
// From route directly
const state = postRoute.useHistoryState()

// With select to extract specific values
const color = useHistoryState({
  from: '/posts/post',
  select: (state) => state.color
})
```

## Summary by CodeRabbit

* **New Features**
  * Read and access history state passed during navigation, including route‑scoped access, selectable transforms, and structural‑sharing support.
  * Route‑level state validation with validators and clear validation error reporting.
  * Devtools: inspect validated state in a new "State Params" panel.

* **Documentation**
  * Added comprehensive docs for the history‑state hook with examples and validation guidance.

* **Examples**
  * New React example demonstrating history state usage, validation, and UI integration.

* **Tests**
  * Extensive type and runtime tests covering history‑state behavior and typings.


docs/router/framework/react/api/router/useHistoryStateHook.md
```diff
@@ -0,0 +1,148 @@
+---
+id: useHistoryStateHook
+title: useHistoryState hook
+---
+
+The `useHistoryState` method returns the state object that was passed during navigation to the closest match or a specific route match.
```
AhmedBaset, line 6:
```suggestion
The `useHistoryState` hook returns the state object that was passed during navigation to the closest match or a specific route match.
```

docs/router/framework/react/api/router/useHistoryStateHook.md
```diff
@@ -0,0 +1,148 @@
...
+
+- Type: `string`
+- Optional
+- The route ID to get state from. If not provided, the state from the closest match will be used.
+
+### `opts.strict` option
+
+- Type: `boolean`
+- Optional - `default: true`
+- If `true`, the state object type will be strictly typed based on the route's `stateSchema`.
```
AhmedBaset, line 22:
```suggestion
- If `true`, the state object type will be strictly typed based on the route's `validateState`.
```

docs/router/framework/react/api/router/useHistoryStateHook.md
naoya7076: Tanks a lot!

schiller-manuel:
thanks for your PR. we are currently focusing on finalizing our devinxi efforts. as soon as this branch is merged we'll revisit this PR.

hleekeeper:
any update for this PR? I'm waiting for this feature to be merged soon

packages/router-core/src/router.ts
```diff
@@ -1201,6 +1201,46 @@ export class RouterCore<
           return [parentSearch, {}, searchParamError]
         }
       })()
+      const [preMatchState, strictMatchState, stateError]: [
```
schiller-manuel, line 1537: this is almost the same as search validation. can we move the identical parts into a reusable function?

packages/router-core/src/router.ts
```diff
@@ -2962,6 +3033,33 @@ export function getInitialRouterState(
     statusCode: 200,
   }
 }
+function validateState(validateState: AnyValidator, input: unknown): unknown {
```
schiller-manuel: same here, can we move the validation of both state and search into a reusable function?

packages/router-core/src/validators.ts
```diff
@@ -89,6 +89,21 @@ export type ResolveSearchValidatorInput<TValidator> =
         ? ResolveSearchValidatorInputFn<TValidator['parse']>
         : ResolveSearchValidatorInputFn<TValidator>
 
+export type ResolveStateValidatorInputFn<TValidator> = TValidator extends (
```
schiller-manuel: this is the same type as for search, right? if yes, we should reuse
maybe just rename the search type to something generic?

packages/solid-router/src/useHistoryState.tsx
```diff
@@ -0,0 +1,96 @@
...
+  ThrowOrOptional<UseHistoryStateResult<TRouter, TFrom, TStrict, TSelected>, TThrow>
+> {
+  return useMatch({
+    from: opts.from!,
+    strict: opts.strict,
+    shouldThrow: opts.shouldThrow,
+    select: (match: any) => {
+      const matchState = match.state;
+      const filteredState = Object.fromEntries(
+        Object.entries(matchState).filter(([key]) => !(key.startsWith('__') || key === 'key'))
```
schiller-manuel: the logic to filter internal state should be put into one single central location and be reused

schiller-manuel:
can you please add type tests, similar to what we do for search?

also, what happens if I open up /posts directly and it has a state validator, but no state is present?
does this mean all state needs to be optional?

packages/router-core/src/router.ts
naoya7076: I unified the shared logic between validateState and validateSearch, including the filter step. Splitting it further would require a helper with too many parameters and little added value, so I’d prefer to keep it as is.

packages/router-core/src/router.ts
naoya7076: Resolved in [this commit](https://github.com/TanStack/router/pull/3967/commits/d4d105c308f8901287254b1f3f4a12a8c0442801): the validation logic for state/search has been extracted to a common function.

packages/router-core/src/validators.ts
naoya7076: Fixed in [`67f84d1` (#3967)](https://github.com/TanStack/router/pull/3967/commits/67f84d18643b3751df97364e1681a71d9ba046ff) &[`88c7126` (#3967)](https://github.com/TanStack/router/pull/3967/commits/88c7126d498c736397a855fa0a9751048da9dbaf) — both search and state now share one generic validator type.

packages/solid-router/src/useHistoryState.tsx
naoya7076: Fixed in [`7847424` (#3967)](https://github.com/TanStack/router/pull/3967/commits/7847424581902156f89854bc4b75df8c115fb74c) and [`799541f` (#3967)](https://github.com/TanStack/router/pull/3967/commits/799541f5a9d3100412d2b1d74a2e362a233b3eb4) 

I pulled the internal-state filter into a single helper omitInternalKeys inside @tanstack/history and switched every call-site to use it.

naoya7076:
@schiller-manuel 

### > can you please add type tests, similar to what we do for search?

added tests [`481df10` (#3967)](https://github.com/TanStack/router/pull/3967/commits/481df10b649a587c352c910007a828c3ab16beb7) and [`5315139` (#3967)](https://github.com/TanStack/router/pull/3967/commits/531513918e5ec7096f52cc4742693a35fc3ebabb)


### > what happens if I open up /posts directly and it has a state validator, but no state is present?

This behavior is same as  how `useSearch` works.

### Current Behavior
When you navigate directly to a /posts with a state validator but no state present:
- The validator receives an empty object `{}`
- Validation outcome depends on your validator implementation

### Example

```tsx
validateState: (input) =>
  z.object({
    example: z.string().default('default-example'),
    count: z.number().default(0),
    options: z.array(z.string()).default([]),
  }).parse(input),
```

### Type System Handles This by `strict`

- `strict: true` → required properties (`{ page: number }`)
- `strict: false` → optional properties (`{ page?: number }`)

When `strict: false`, properties become optional.

### Answer to Your Question

> Does this mean all state needs to be optional?

No.You have the same options as with search params.

hleekeeper:
Hi @naoya7076, thanks for putting this up together.

I have a question. How can one access to the route-specific history state type? For example, currently to access the type of history state, the `HistoryState` type is used. But this is just an aggregation of all possible history states. But let say I want to find the history state type of a specific path `/example`, would it be possible to do something like `HistroyState['/example']`?

naoya7076:
@hleekeeper 

Currently, This PR(useHistoryState) doesn't support accessing route-specific history state types like
  `HistoryState['/example']`. 

  The only available approach currently is to manually define the type:
  ```typescript
  type ExampleRouteState = {
    color: string
    // Same structure as defined in validateState
  }
```

  You're correct that the current HistoryState is an aggregation of all route states, and there's no
  type-level API to extract specific route state types

**Comment by hleekeeper** [comment]
> @hleekeeper
> 
> Currently, This PR(useHistoryState) doesn't support accessing route-specific history state types like `HistoryState['/example']`.
> 
> The only available approach currently is to manually define the type:
> 
> ```ts
> type ExampleRouteState = {
>   color: string
>   // Same structure as defined in validateState
> }
> ```
> 
> You're correct that the current HistoryState is an aggregation of all route states, and there's no type-level API to extract specific route state types

I guess we can access the route-specific history state type by `RouterType['routeTree']['/example']['types']['fullStateSchema']`? But I may be wrong. I'll check it out when the PR is merged and get released. Anyway, thanks for the great work!

**Comment by naoya7076** [comment]
@hleekeeper 
Yes, your understanding is spot on.
In `examples/react/basic-history-state/src/main.tsx`, I tried to access the route-specific history-state type like this:

```ts
type DestinationRouteStateSchema = RouteById<
  (typeof router)['routeTree'],
  '/state-examples/destination'
>['types']['fullStateSchema'];
```

This resolves to:
```:ts
type DestinationRouteStateSchema = {
  example: string;
  count: number;
  options: string[];
};
```

naoya7076:
Hi @schiller-manuel,

About 3 weeks ago, I have addressed all the feedback and [replied your comments](https://github.com/TanStack/router/pull/3967#issuecomment-2994272891). All the requested changes have been implemented:

-  Refactored shared validation logic between `validateState` and `validateSearch` 
-  Unified validator types to use generic approach
-  Added comprehensive type tests as requested

The PR is ready for re-review already. Could you please take another look when you have a chance? I appreciate your time and feedback.

Thanks!

schiller-manuel:
sorry for being slow to review. will do so soon

hleekeeper:
any update on the PR's review?

packages/react-router/tests/useHistoryState.test.tsx
```diff
@@ -0,0 +1,352 @@
...
+
+  test('throws when match not found and shouldThrow=true', async () => {
+    function RootComponent() {
+      try {
+        useHistoryState({ from: '/non-existent', shouldThrow: true })
+        return <div>No error</div>
+      } catch (e) {
+        return <div>Error occurred: {(e as Error).message}</div>
+      }
+    }
```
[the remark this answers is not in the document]
naoya7076: resolved by [this](https://github.com/TanStack/router/pull/3967/commits/dd93cb174c051f7ef4a71d2c00a1cbd8cbd0363d)

packages/router-core/src/route.ts
```diff
@@ -155,6 +167,35 @@ export type ResolveSearchSchema<TSearchValidator> =
           ? ResolveSearchSchemaFn<TSearchValidator['parse']>
           : ResolveSearchSchemaFn<TSearchValidator>
 
+export type ParseSplatParams<TPath extends string> = TPath &
+  `${string}$` extends never
+  ? TPath & `${string}$/${string}` extends never
+    ? never
+    : '_splat'
+  : '_splat'
```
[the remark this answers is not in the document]
naoya7076: removed by [this commit](https://github.com/TanStack/router/pull/3967/commits/fe9d1dc4c8511beefb85c9a5ed9eaa3682e4bce1)

packages/router-core/src/router.ts
```diff
@@ -1197,6 +1198,39 @@ export class RouterCore<
...
+        } catch (err: any) {
+          const stateValidationError = err
+
+          if (opts?.throwOnError) {
+            throw stateValidationError
+          }
+
+          return [filteredState, {}, stateValidationError]
+        }
+      })()
```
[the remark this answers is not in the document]
naoya7076: fixed by [this commit](https://github.com/TanStack/router/pull/3967/commits/16825f3e5d7b9f73bba9e7e1713acb14bcb4389b)

packages/router-core/src/router.ts
```diff
@@ -1575,6 +1620,26 @@ export class RouterCore<
...
+                }) ?? {}),
+              }
+            }
+          } catch {
+            // ignore errors here because they are already handled in matchRoutes
+          }
+        })
+        nextState = validatedState
+      }
+
```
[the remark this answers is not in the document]
naoya7076: fixed by [this commit](https://github.com/TanStack/router/pull/3967/commits/3cb74f4ba57b5011637780155448094a79f3935e)

naoya7076:
@schiller-manuel 

Thank you for your daily development work on TanStack Router! I'm also looking forward to the development of TanStack Start.

It's been over a month since the previous comment. I've also addressed the CodeRabbit comments, so please review when you have a chance.

hleekeeper:
> any update on the PR's review?

Is this PR still under review or blocked? Seems like a major improvement to the type-safety, and I'm looking forward to having this change.

naoya7076:
Hi @schiller-manuel  @SeanCassiere  , just checking in on this PR. All the review feedback from June has been addressed and CI is passing. I've also been keeping the branch up to date with main (resolved conflicts, fixed tests for the TStateValidator type param addition, etc.).

Let me know if anything else needs to be changed!

SalahAdDin:
@naoya7076 conflicts.

LadyBluenotes:
Thanks for all the work on this! Sorry it's taken a while to get back to this, the feature is still useful, and we’d like to keep moving it forward.

Before we can merge, `validateState` needs to guarantee that `useHistoryState` never exposes values that do not match its validated type. Please also wire state validation into committed navigation, decide and cover the direct-load/SSR behavior when required state is absent, and bring Vue support in line with the other frameworks or scope it out explicitly, please!

Once those pieces are addressed, we can continue the review :)

naoya7076:
@LadyBluenotes 

Thanks for the detailed feedback! I've merged the latest main and addressed all four points.

1. validateState now guarantees what useHistoryState returns
The hook was reading match.state, which is the raw history state merged with the validator output, and casting it to the validated type. So unvalidated keys leaked through at runtime even though the type said otherwise.
It now reads _strictState, which only contains what the validateState validators on the route chain returned, so the return value always matches the validated type.

2. State validation is wired into committed navigation
It's now enabled in the same places as the search equivalent: buildAndCommitLocation, the server-side beforeLoad path, and the three Transitioners. Validator defaults and transforms end up in the state that actually gets pushed into history.
Internal router keys like __TSR_index, and any keys no validator claimed, are still preserved.

3. Direct load / SSR when required state is absent
I went with erroring. match.stateError was written but never read, so validation failures were silently dropped. It now goes through handleSerialError next to searchError, so the route's errorComponent renders and the server returns a 500.

History state doesn't survive a fresh document load, so the validator always gets {} on a direct load, a refresh, or during SSR. A schema with required fields will therefore always fail there. Skipping validation instead would hand the component {} typed as having required fields, which is the same unsoundness as point 1, so erroring seemed like the better trade. The workarounds are documented.

4. Vue
  I'd like to keep the scope and the diff of this PR down, so the Vue hook isn't part of it. The only Vue changes here are the minimum needed to keep the package compiling against the new router-core generics and in step with the shared core behaviour. It's noted in the hook docs. Once this is approved and merged I'll add Vue support in a follow-up PR.

Other
<Link state> typing wasn't working at all. true | ResolveRelativePath<TFrom, TTo> extends infer TPath parses as (true | ResolveRelativePath<...>) extends infer TPath, so TPath ends up including true, the TPath extends string branch never matches, and it always fell back to the loose HistoryState. Fixed that, and switched from stateSchema to fullStateSchema so state inherited from parent routes gets typed too.

Tests
Added table-driven tests across the three validator shapes, matching the existing search validation tests.

One existing test changed: it expected an invalid color: 'yellow' to reach the component. That's the leak from point 1, so it now checks that the error boundary renders instead.

PR #3969, opened 2025-04-09
feat(react-router): Export `ClientOnly`

zrosenbauer:
Exposes the `ClientOnly` component for preventing rendering & hydration errors.

## Background

When you attempt to prevent content from rendering using something like `typeof document !== 'undefined'` as that type of conditional will cause hydration errors.

## API

```tsx
import { ClientOnly, createRoute } from '@tanstack/react-router'
import { Charts, FallbackCharts } from './charts-that-break-server-side-rendering'

const Route = createRoute({
  // ... other route options
  path: '/dashboard',
  component: Dashboard,
})

function Dashboard() {
  return (
    <div>
      <p>Dashboard</p>
      <ClientOnly fallback={<FallbackCharts />}>
        <Charts />
      </ClientOnly>
    </div>
  )
}
```

> [!NOTE]
> This component as is (other than I removed the `children` as a `function`) in production on [joggr.io](https://joggr.io)


docs/router/framework/react/api/router/clientOnlyComponent.md
```diff
@@ -0,0 +1,50 @@
...
+### `props.children` prop
+
+The component to render if the JS is loaded in the client.
+
+## Returns
+
+- Returns the component's children if the JS is loaded in the client.
+- Returns the `fallback` component if the JS is not yet loaded in the client.
+
+## Examples
```
SeanCassiere, line 25: Switch the example to use file-based routing.

```tsx
// src/routes/dashboard.tsx
import { ClientOnly, createFileRoute } from '@tanstack/react-router'
import {
  Charts,
  FallbackCharts,
} from './charts-that-break-server-side-rendering'

export const Route = createFileRoute('/dashboard')({
  component: Dashboard,
  // ... other route options
})

function Dashboard() {
  return (
    <div>
      <p>Dashboard</p>
      <ClientOnly fallback={<FallbackCharts />}>
        <Charts />
      </ClientOnly>
    </div>
  )
}
```

SeanCassiere (changes requested):

SeanCassiere:
cc @birkskyum about possibly adding an equivalent in the Solid adapter.

SeanCassiere (commented):
I've applied the changes I had suggested.

@birkskyum Do you want to add this solid equivalent onto this PR? or follow it up in a separate PR?

birkskyum:
I can just add it since - i have most of it working, there is just one test (the renderToString one), where there is bit more work before i put up a PR for solid, but it doesn't have to hold back react-router.

PR #3979, opened 2025-04-11
fix(router-core): Accept AnyRouter in `redirect`

navelencia:
PR #2243 introduced a tighter type for the `TRouter` type parameter in `redirect`, making it not possible to pass a different router than the currently registered one.

This PR brings back the previous behaviour, and allows any `Router` to be passed, with a default of `RegisteredRouter`. `TTo` also needed to be changed as the first parameter became optional. 

First time contributing here, do I need to create an issue for this?

### DIFF
--- a/packages/router-core/src/redirect.ts
+++ b/packages/router-core/src/redirect.ts
@@ -51,8 +51,8 @@ export type ResolvedRedirect<
 }
 
 export function redirect<
-  TRouter extends RegisteredRouter,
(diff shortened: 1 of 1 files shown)


SeanCassiere (approved):

SeanCassiere:
@navelencia Welcome! and thank you for finding and making a PR for this fix. Much appreciated!

PR #3980, opened 2025-04-11
fix(router-plugin): avoid blocking webpack build command

vkunz:
hi,

I pulled in the router into an existing webpack based project. When trying to build the project, webpack was hanging indefinitely as it was a development mode run.

I tracked the problem down to a chokidar watcher that is registered independent of the webpack command used.

This pr fixes that problem by restricting the watcher to only be registered when webpack [watchMode](https://webpack.js.org/api/compiler-hooks/#watching) is enabled.

Thanks,
Valentin

### DIFF
--- a/packages/router-plugin/src/core/router-generator-plugin.ts
+++ b/packages/router-plugin/src/core/router-generator-plugin.ts
@@ -2,6 +2,7 @@ import { isAbsolute, join, normalize, resolve } from 'node:path'
 import { generator, resolveConfigPath } from '@tanstack/router-generator'
 
 import { getConfig } from './config'
+import type { FSWatcher } from 'chokidar'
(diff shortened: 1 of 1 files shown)


SeanCassiere (commented):
@vkunz could you test this with rspack (you can find an example in the monorepo under `examples/react/quickstart-rspack`)? Since `rspack` and `webpack` have quite a bit of overlap, its possible that this fix could be applicable for it as well.

vkunz:
@SeanCassiere I tries rspack, and it does contain the same problem, also the same solution

I put the fix for rspack into the same commit

SeanCassiere (approved):

SeanCassiere:
@vkunz thank you for finding and contributing a fix for this!

Issue #3962, opened 2025-04-08
Backslash in path parameters are incorrectly restored from URL


DreierF:
### Which project does this relate to?

Router

### Describe the bug

When navigating to a route with a path parameter (or splash) where the value contains a backslash, everything works at first. But when reloading the page or navigating back to the page the path restored from the URL differs from the original URL.

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-yoii3nte?file=src%2Froutes%2Fb.%24postId.tsx

### Steps to Reproduce the Bug or Issue

1. Open the preview in a new tab (so that you can see the full URL)
2. Click one of the "Splat route" or "Path param route" links
3. Reload the page
4. Observe that the URL changes and the page content changes as well.

### Expected behavior

As a user I expect that reloading a page shows the same route and same parameters than before.

### Screenshots or Videos

https://github.com/user-attachments/assets/a7cb6f5b-1493-487b-9f47-e179b2aad271

### Platform

- OS: macOS
- Browser: Chrome
- Version: 135

nlynzaad:
resolved in #5663

PR #3987, opened 2025-04-14
fix(router-core): Backslash in path parameters are incorrectly restored from URL

leesb971204:
fixes: #3962


schiller-manuel:
can you please remove the irrelevant commits from this PR?

leesb971204:
@schiller-manuel 

I think we need to write separate tests for paths containing special characters, depending on whether we use `$` with params or define the path directly.
For example, when accessing a path like `%C3%A9` using `$` and `params` props in `<Link>`, the path remains `%C3%A9`, but when it’s defined directly, it gets `decoded` and becomes `é`.
This behavior was the same before and after my changes, but I don’t understand why the tests are failing only after my modification...

SeanCassiere (commented):
Aside of the test regressions present in the unit testing, I'm not sure if this is the right approach.

I'm wondering whether we should be doing any decoding of the named path parameters (not the splat parameter) at all. Perhaps, we shouldn't be encoding the path parameters at all? 🤔

Whilst, URI encoding works for the URL Search Parameters, perhaps we shouldn't be doing this for the path parameters. Essentially we'd leave them "as-is".

leesb971204:
> Aside of the test regressions present in the unit testing, I'm not sure if this is the right approach.
> 
> I'm wondering whether we should be doing any decoding of the named path parameters (not the splat parameter) at all. Perhaps, we shouldn't be encoding the path parameters at all? 🤔
> 
> Whilst, URI encoding works for the URL Search Parameters, perhaps we shouldn't be doing this for the path parameters. Essentially we'd leave them "as-is".

```typescript
// path.ts
value: part.includes('%25')
          ? part
              .split('%25')
              .map((segment) => decodeURI(segment))
              .join('%25')
          : hasUriEncodedChars(part)
            ? part
            : decodeURI(part),

👇

value: part
```

SeanCassiere:
Yea, I'm wondering if that'd be the trick. Also, whether or not we should be doing the `encodeUri...` for the path parameters.

Nothing concrete. Just thinking of ideas to go about this without breaking everything for everyone 😅

leesb971204:
> Yea, I'm wondering if that'd be the trick. Also, whether or not we should be doing the `encodeUri...` for the path parameters.
> 
> Nothing concrete. Just thinking of ideas to go about this without breaking everything for everyone 😅

Exactly. The decoding logic does seem pretty complex. For encoding, I think it might be reasonable to delegate that responsibility to the user when it comes to path parameters. But if we go that route, it would likely introduce a lot of changes to the existing codebase.

DreierF:
What is the current state of this PR? 
Is the PR currently just blocked by the test failure or do you consider to rework the whole parameter encoding approach? Is there anything I could contribute to move this forward?

As we are currently waiting for a fix for the issue, I would favor if the current fix would be released and a potential bigger rework could be implemented as a follow-up PR.

SeanCassiere:
We are currently locked in a bit of code-freeze so we can minimize the number of conflicts while we complete the large rewrites happening in the `devinxi` branch.

'As-is', after the code freeze, even disregarding the lack of tests, I think I'd still have reservations as to the flow and logic in how this fix is being implemented in this PR and stand by comment from above - https://github.com/TanStack/router/pull/3987#pullrequestreview-2766320734.

---

> What is the current state of this PR? Is the PR currently just blocked by the test failure or do you consider to rework the whole parameter encoding approach? Is there anything I could contribute to move this forward?
> 
> As we are currently waiting for a fix for the issue, I would favor if the current fix would be released and a potential bigger rework could be implemented as a follow-up PR.

That said, @DreierF if you'd be willing to look at going about this using the "no encoding at all approach" and have some unit tests in place, then I can look at having your changes pushed through and I'd resolve the conflicts later on.

nlynzaad:
closing in favour of #5663

PR #4003, opened 2025-04-16
fix(router-core): Pathless Layout Route Renders Empty HTML

### LINKED ISSUE #3843: Pathless Layout Route Renders Empty HTML

Router

### Describe the bug

When defining a pathless layout route (`_appLayout.tsx`) inside a folder (app/), TanStack Router incorrectly treats it as a route. Instead of returning a Not found when accessing `/app`, it renders an empty HTML page.

https://stackblitz.com/edit/github-aeat3k26

1. go to the `/app` route

### Expected behavior

**Expected**: `/app` should return Not found page (since `_appLayout.tsx` is pathless).

**Actual**: `/app` renders an empty HTML page instead.

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 91.1]

### Additional context

_No response_

leesb971204:
fixes #3843

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -1040,6 +1040,13 @@ export class RouterCore<
         // Sort by original index
         return a.index - b.index
       })
+      .filter((d) => {
+        if (!d.child.children) return true
+        return (
+          Array.from(d.child.children).length > 0 &&
+          Array.from(d.child.children).every((c) => (c as AnyRoute).path)
+        )
+      })
       .map((d, i) => {
         d.child.rank = i
         return d.child


schiller-manuel:
can you add a test for this?

leesb971204:
> can you add a test for this?

Sure. But I’d like to ask where it would be appropriate to write the test code. Would it be better to create a new file named `notfound.test.tsx` under the `/react-router/tests` path? Or would it be better to add it to an existing file?

schiller-manuel:
also wondering if this maybe needs a fix in the router generator instead? what's the idea of your fix here?

schiller-manuel:
yes I think so. the generator currently generates a "virtual route" in this situation. this is then matched at runtime

leesb971204:
> yes I think so. the generator currently generates a "virtual route" in this situation. this is then matched at runtime

However, if we handle this in the generator, I think the auto-completion feature might not work when creating a pathless_layout if it’s the only file in the route folder.
I’m planning to solve this by not returning a node if the only child is a pathless_layout. That way, it will result in a not found state.
What do you think about this approach?

leesb971204:
> yes I think so. the generator currently generates a "virtual route" in this situation. this is then matched at runtime

Got it. I updated the condition for `isNonPath` in the generator.
But I think additional work might be needed.
Do we also need to update the `path` or any other values of the layout node?
What about the type definitions in `routeTree.gen.ts`?

leesb971204:
> yes I think so. the generator currently generates a "virtual route" in this situation. this is then matched at runtime

I believe the issue can be resolved by updating the `isNonPath` value when a node’s only child is a `pathless_layout`.
What do you think about this approach?

schiller-manuel:
maybe, I did not look into it deeply yet. can you add an e2e test into one of the existing e2e test setups for react router for this please? then we can check whether the fix solves it

leesb971204:
> maybe, I did not look into it deeply yet. can you add an e2e test into one of the existing e2e test setups for react router for this please? then we can check whether the fix solves it

Got it. But I think additional changes might be needed.
For example, if a node has both a `pathless_layout` and `static` children, should we still treat it as `isNonPath: true?`
What do you think?

```
L app  👈 isNonPath ??
  L _appLayout.tsx  
  L index.tsx  
  ```

**Comment by leesb971204** [comment]
> Been thinking about this more and we'd need some comprehensive sandbox (end-to-end) testing for this.
> 
> Plus, my point earlier about _pathless layout routes_ being assigned a _path_ isn't correct.



> Been thinking about this more and we'd need some comprehensive sandbox (end-to-end) testing for this.
> 
> Plus, my point earlier about _pathless layout routes_ being assigned a _path_ isn't correct.

Yes, that’s correct. When I run it locally, `_appLayout` also doesn’t have the `/app` path assigned.
What are your thoughts on my comment above?

**Comment by leesb971204** [comment]
@SeanCassiere 
Could you review the last commit?
I only modified the logic related to the path part, so the e2e tests may not pass, but based on my local testing, it seems to be working fine.
Additional changes might be needed, but I’d appreciate it if you could check whether the issue is resolved.

**Review by SeanCassiere — COMMENTED** [review]
@leesb971204 its partially resolved. 

With your changes, these scenarios now work.

```sh
src/routes/
	app/
		_appLayout.tsx
		_appLayout.dashboard.tsx
		_appLayout.something.tsx
```

```
src/routes/
	app/
		_appLayout.tsx
		_appLayout.dashboard.tsx
		_appLayout.something.tsx
		_foo.tsx
		_foo.index.tsx
```

However, these two scenarios break the route tree.

1. For this scenario, the `routeTree` file gets put into a state where it tries to create an import from `'./routes/app'` which of-course does not exist on the file-system.
```
src/routes/
	app/
		_appLayout.tsx
		_appLayout.dashboard.tsx
		_appLayout.something.tsx
		_foo.tsx
		_foo.something.tsx
```

2. For this scenario, the `routeTree` file adds an additional `/app` into the routes. So, instead of it being `/app/dashboard`, it instead becomes `/app/app/dashboard`.
```sh
src/routes/
	app/
		route.tsx
		_appLayout.tsx
		_appLayout.dashboard.tsx
		_appLayout.something.tsx
		_foo.tsx
		_foo.index.tsx
```

---

I wish I had better feedback to give, as to how you can approach this, but this is something @schiller-manuel and I have discussed earlier at length with the current system being 'best-effort' compromise that we came to.

The shortcut, to get your original sandbox working tree, would be to throw a `redirect` in `src/routes/app/index.tsx` to a different route or to set a `notFoundComponent` and immediately throw a `notFound`.
```diff
src/routes/
	app/
+		index.tsx
		_appLayout.tsx
		_appLayout.dashboard.tsx
		_appLayout.something.tsx
```

leesb971204:
> @leesb971204 its partially resolved.
> 
> With your changes, these scenarios now work.
> 
> ```shell
> src/routes/
> 	app/
> 		_appLayout.tsx
> 		_appLayout.dashboard.tsx
> 		_appLayout.something.tsx
> ```
> 
> ```
> src/routes/
> 	app/
> 		_appLayout.tsx
> 		_appLayout.dashboard.tsx
> 		_appLayout.something.tsx
> 		_foo.tsx
> 		_foo.index.tsx
> ```
> 
> However, these two scenarios break the route tree.
> 
> 1. For this scenario, the `routeTree` file gets put into a state where it tries to create an import from `'./routes/app'` which of-course does not exist on the file-system.
> 
> ```
> src/routes/
> 	app/
> 		_appLayout.tsx
> 		_appLayout.dashboard.tsx
> 		_appLayout.something.tsx
> 		_foo.tsx
> 		_foo.something.tsx
> ```
> 
> 2. For this scenario, the `routeTree` file adds an additional `/app` into the routes. So, instead of it being `/app/dashboard`, it instead becomes `/app/app/dashboard`.
> 
> ```shell
> src/routes/
> 	app/
> 		route.tsx
> 		_appLayout.tsx
> 		_appLayout.dashboard.tsx
> 		_appLayout.something.tsx
> 		_foo.tsx
> 		_foo.index.tsx
> ```
> 
> I wish I had better feedback to give, as to how you can approach this, but this is something @schiller-manuel and I have discussed earlier at length with the current system being 'best-effort' compromise that we came to.
> 
> The shortcut, to get your original sandbox working, would be to throw a `redirect` in `src/routes/app/index.tsx` to a different route or to set a `notFoundComponent` and immediately throw a `notFound`.
> 
> ```diff
> src/routes/
> 	app/
> +		index.tsx
> 		_appLayout.tsx
> 		_appLayout.dashboard.tsx
> 		_appLayout.something.tsx
> ```

Thanks for the review.
I’m really sorry to bother you again, but I need some clarification on the last part.
Do you see this scenario not as something that requires changes to the generator or other internal code, but rather as an issue where the user is expected to add an `index.tsx` themselves to handle the `redirect`?
If so, I’ll go ahead and close this PR.

SeanCassiere:
> ... Do you see this scenario not as something that requires changes to the generator or other internal code, but rather as an issue where the user is expected to add an `index.tsx` themselves to handle the `redirect`? If so, I’ll go ahead and close this PR.

It is kind of two-fold.

Router works in a tree-like structure, which is important since pathless layout routes (which do not have a 'path') requires it having to know what its parent is, for everything to _work_ at runtime.

As such, in this scenario, it creates a route in the tree which doesn't have a component and/or any logic attached to it. Its purely there, _just_ for the pathless layout route to use for it refer to it as its parent.

So, I say this problem is two-fold, since we could either tackle this problem by.

1. Adding code in the generator that detects this "empty" route scenario, setup a virtual `index` for the given route which throws a `redirect` to `/` if the parent layout route's children do not contain an existing index route.
2. Or move that stuff into a user-space and ask them to set it up themselves (using an `index.tsx`).

If you are open to tackling **option 1**, that'd be great! If not, no worries.

leesb971204:
> > ... Do you see this scenario not as something that requires changes to the generator or other internal code, but rather as an issue where the user is expected to add an `index.tsx` themselves to handle the `redirect`? If so, I’ll go ahead and close this PR.
> 
> It is kind of two-fold.
> 
> Router works in a tree-like structure, which is important since pathless layout routes (which do not have a 'path') requires it having to know what its parent is, for everything to _work_ at runtime.
> 
> As such, in this scenario, it creates a route in the tree which doesn't have a component and/or any logic attached to it. Its purely there, _just_ for the pathless layout route to use for it refer to it as its parent.
> 
> So, I say this problem is two-fold, since we could either tackle this problem by.
> 
> 1. Adding code in the generator that detects this "empty" route scenario, setup a virtual `index` for the given route which throws a `redirect` to `/` if the parent layout route's children do not contain an existing index route.
> 2. Or move that stuff into a user-space and ask them to set it up themselves (using an `index.tsx`).
> 
> If you are open to tackling **option 1**, that'd be great! If not, no worries.

I’ve implemented the first approach — it now redirects to `/` when there’s no `index` route in the children.
Would it make sense to update the test cases to reflect this change?

packages/router-generator/src/generator.ts
```diff
@@ -518,6 +518,26 @@ export class Generator {
...
+
+        if (node.children && node.children.length > 0) {
+          if (node.children.some((child) => child.cleanedPath === '/')) {
+            return // already has index child → skip
+          }
+        }
+
+        node.isVirtualRedirectIndex = true
+        node.redirectTo = '/'
+      })
```
leesb971204: We iterate through acc.routeNodes to find virtual layout routes that do not have an `index` route as a children, and redirect them to `'/'`.

leesb971204: With the following directory structure, accessing `/foo/bar` currently results in a redirect to `'/'`.
If this behavior is incorrect and `/foo/bar` should be accessible, additional changes are needed.
Please share your thoughts.

```
src/routes/
	foo/
		_fooLayout.tsx
		bar/	
			index.tsx
```

schiller-manuel: surely this is a valid case where /foo/bar should be accesible
don't see where the pathless layout comes into play here?

schiller-manuel:
I am wondering what the actual issue is that we need to solve. it seems the problem is that at runtime router is not aware that those routes should not be used for matching. if my understanding is correct, shouldn't we rather fix this when processing the route tree in the router?
we probably can't detect this at that point in time (or can we?). if we cannot, we could issue a marker to the route in the generator to set a prop using Route.update which would then later get read when processing the route tree upon router initialization

leesb971204:
> I am wondering what the actual issue is that we need to solve. it seems the problem is that at runtime router is not aware that those routes should not be used for matching. if my understanding is correct, shouldn't we rather fix this when processing the route tree in the router? we probably can't detect this at that point in time (or can we?). if we cannot, we could issue a marker to the route in the generator to set a prop using Route.update which would then later get read when processing the route tree upon router initialization

Are you suggesting that redirecting pathless layout routes to `'/'` may not be the appropriate solution?

schiller-manuel:
yes. I think ideally the route matching would just fail and the not found handling would kick in.

leesb971204:
> yes. I think ideally the route matching would just fail and the not found handling would kick in.

In that case, it sounds more appropriate to handle those specific paths as `notFound` within the router itself.
I’ll try that approach instead.

packages/router-core/src/router.ts
```diff
@@ -3225,6 +3225,17 @@ export function getMatchedRoutes<TRouteLike extends RouteLike>({
...
+      if (route.children && route.children.length > 0) {
+        if (
+          route.children.find(
+            (child) => trimPathLeft(child.fullPath) === route.fullPath,
+          )
+        ) {
+          return true
+        }
+      }
+
```
leesb971204, line 3238: If a route does not have a parent's index route among its children, it is marked as `notFound`.

packages/router-core/src/router.ts
```diff
@@ -3225,6 +3225,13 @@ export function getMatchedRoutes<TRouteLike extends RouteLike>({
     routeParams = getMatchedParams(foundRoute)!
   } else {
     foundRoute = flatRoutes.find((route) => {
+      // If there are child routes, check if any of them have the same path as the current route
+      if (route.children && route.children.length > 0) {
+        if (!route.children.find((child) => child.path === '/')) {
+          return false
+        }
+      }
+
```
leesb971204: If a route does not have a parent index route among its children, it is excluded from matching.

schiller-manuel: what's "parent index route" when it comes to children?

leesb971204: > what's "parent index route" when it comes to children?

This refers to a route with an `index` path among the children under a parent route.

```
src/routes/
	foo/
		_fooLayout.tsx
		index.tsx 👈
		bar/	
			index.tsx
```

schiller-manuel: what if you have foo.tsx or route.tsx inside /foo ?

schiller-manuel:
we need to add tests for this, ideally in router core.

packages/router-core/src/router.ts
leesb971204: > what if you have foo.tsx or route.tsx inside /foo ?

I’m not sure if this can be fully solved within the router core alone.
As you mentioned, we need to correctly match both cases:
– when a route file with the same name exists in another directory (foo.tsx)
– when a `route.tsx` file exists inside the directory

But it seems difficult to handle both cases purely in the router core.
I initially tried to determine this based on the presence of `options.component`,
but this approach fails in cases where the component is lazy.

Should we consider adding a specific flag in the generator to help the router core handle it properly or using `update`?

packages/router-generator/src/generator.ts
```diff
@@ -583,6 +600,7 @@ export class Generator {
               `id: '${node.path}'`,
               !node.isNonPath ? `path: '${node.cleanedPath}'` : undefined,
               `getParentRoute: () => ${findParent(node, exportName)}`,
+              node.isVirtualLayout ? `isVirtualLayout: true` : undefined,
```
leesb971204, line 612: A value is set via `update()` in the `router core` to determine whether a route should be matched.

leesb971204: @schiller-manuel 
I wanted to implement this in a way that keeps the logic hidden from the user’s code, but I couldn’t come up with a better solution. 😕
I’d appreciate any further suggestions you might have.
If you need a test case, I’d be happy to write one.

schiller-manuel:
sorry for being slow to respond here.

we should definitely add a test for this before progressing. can you please add an e2e test to the `e2e/react-router/basic` project?

leesb971204:
> sorry for being slow to respond here.
> 
> we should definitely add a test for this before progressing. can you please add an e2e test to the `e2e/react-router/basic` project?

I thought it made the most sense to add it under `e2e/react-router/basic-file-based`, so I went ahead and placed it there.
If you think it should also be added to the basic path, please let me know!

leesb971204:
Would it be necessary to add a test under the `e2e/react-router/basic` path as well?

schiller-manuel:
closing in favor of https://github.com/TanStack/router/pull/6120

PR #4031, opened 2025-04-22
feat(router-core): allow dynamic `ViewTransition` types

burakcan:
Context: https://github.com/TanStack/router/discussions/4027

Changes the ViewTransitionOptions["types"] to also accept a function. This enables automatically applying different animations (for example sliding left or right) based on history navigation. When navigating via `history.back()/forward()`, it’s very common to want a "slide-left" vs "slide-right" effect based on direction.

This technically also works for `Link`s but it's not as useful there since the prop can already be dynamic. This is mostly for navigations that are triggered via other sources than links.

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -2183,20 +2183,19 @@ export class RouterCore<
         const next = this.latestLocation
         const prevLocation = this.state.resolvedLocation
 
-        let types = shouldViewTransition.types
-
-        if (typeof types === 'function') {
-          types = types(
-            getLocationChangeInfo({
-              resolvedLocation: prevLocation,
-              location: next,
-            }),
-          )
-        }
+        const resolvedViewTransitionTypes =
+          typeof shouldViewTransition.types === 'function'
+            ? shouldViewTransition.types(
+                getLocationChangeInfo({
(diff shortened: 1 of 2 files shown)


packages/router-core/src/router.ts
```diff
@@ -2172,9 +2180,23 @@ export class RouterCore<
         typeof shouldViewTransition === 'object' &&
         this.isViewTransitionTypesSupported
       ) {
+        const next = this.latestLocation
+        const prevLocation = this.state.resolvedLocation
+
+        let types = shouldViewTransition.types
```
SeanCassiere, line 2186: Please inline this one with a ternary.

```ts
const resolvedViewTransitionTypes = typeof showViewTransition.types === 'function'
	? showViewTransition.types(getLocationChangeInfo({ resolvedLocation: prevLocation, location: next }))
	: showViewTransition.types 
```

SeanCassiere (changes requested):

SeanCassiere:
@burakcan thank you for adding this in!

SeanCassiere (approved):

burakcan:
Thank you for the quick review @SeanCassiere

PR #4059, opened 2025-04-27
feat(router-core): allow disabling scroll restoration for specific containers

schiller-manuel:


SeanCassiere (changes requested):
Could you please add some end-to-end tests (in `e2e/react-router/scroll-restoration-sandbox-vite` and `e2e/solid-router/scroll-restoration-sandbox-vite`)?

Something that tests this behaviour.

LadyBluenotes:
This was fixed in `bc811688fd`.

PR #4080, opened 2025-05-03
fix(directive-functions-plugin): do not outlaw generator functions

SeanCassiere:


packages/directive-functions-plugin/tests/compiler.test.ts
```diff
@@ -185,12 +210,18 @@ describe('server function compilation', () => {
       const namedFunction_createServerFn_namedFunction = createServerRpc("test_ts--namedFunction_createServerFn_namedFunction", function namedFunction() {
         return 'hello';
       });
+      const namedGeneratorFunction_createServerFn_namedGeneratorFunction = createServerRpc("test_ts--namedGeneratorFunction_createServerFn_namedGeneratorFunction", function* namedGeneratorFunction() {
+        return 'hello';
```
theturtle32, line 214: For test, it might be good to also `yield` a value before returning

theturtle32:
@SeanCassiere Wow, thanks for hopping on this so quickly! I'm curious, do you know why generator functions were outlawed previously? Is there something else downstream that may not be expecting to have to handle generator functions?

SeanCassiere:
> @SeanCassiere Wow, thanks for hopping on this so quickly! I'm curious, do you know why generator functions were outlawed previously? Is there something else downstream that may not be expecting to have to handle generator functions?

I'm not quite sure. TBF though, I still barely understand generators 😄

packages/directive-functions-plugin/tests/compiler.test.ts
SeanCassiere: Yup makes sense! I'll add this.

theturtle32:
> I'm not quite sure. TBF though, I still barely understand generators 😄

I feel you. Guess we'll find out together! 😄

PR #4082, opened 2025-05-03
Adds unhead types and head tag sorting

IHIutch:
From this conversation: https://discord.com/channels/719702312431386674/1367798166493335622

This adds stronger typing for head tags and sorts them based on [unhead](https://github.com/unjs/unhead/blob/main/packages/unhead/src/utils/sort.ts)/capo


packages/solid-router/src/HeadContent.tsx
```diff
@@ -5,6 +5,33 @@ import { useRouter } from './useRouter'
 import { useRouterState } from './useRouterState'
 import type { RouterManagedTag } from '@tanstack/router-core'
 
+const isTruthy = (val?: string | boolean) => val === '' || val === true
```
schiller-manuel, line 8: we should move the whole sorting logic into router-core
also add a comment please where that is originating from

schiller-manuel:
please also add some type tests that use the new types

packages/solid-router/src/HeadContent.tsx
schiller-manuel: thinking about this more, maybe we should just expose an API that a user could hook into to apply sorting using existing libraries instead of copying the stuff into router.

IHIutch: > also add a comment please where that is originating from

Can you clarify what you mean?

LadyBluenotes:
Thanks for putting this together. The head ordering work needs a fresh decision on whether we own sorting logic or expose an extension point, and this draft is now too far behind to carry that discussion forward.

The concrete `stylesheet` versus `modulepreload` ordering bug is tracked in #6749, so I’m closing this in favour of that focused follow-up.

PR #4095, opened 2025-05-05
feat: add bindings for `Route.Link` and `routeApi.Link`

chorobin:
This allows `Route.Link` and `routeApi.Link` bindings instead of manual `from`.

We never used to be able to do this due to circular issues but it seems this is no longer an issue since all the changes to the types


TkDodo (approved):
I added a quick test for `routeApi.Link` - looks great 🙌

PR #4122, opened 2025-05-09
preloadIntentProximity

birkskyum:
This is the react impl.

Solid impl. missing
Tests missing.


schiller-manuel:
docs are also missing :)

brenelz:
@birkskyum is this something we can maybe get into the alpha. not sure if there is still some work to do

birkskyum:
think it's okay to introduce this with the new alpha branch

PR #4172, opened 2025-05-15
fix: resolve custom `client` and `server` entry paths

### LINKED ISSUE #4138: Cannot find module 'srcserver.tsx' imported from '/~start/server-entry.tsx'

Start

### Describe the bug

Getting error 
```
An error occured while server rendering /index.html:

	Cannot find module 'srcserver.tsx' imported from '/~start/server-entry.tsx'
```

when creating a `server.tsx` file

https://github.com/ammarmbe/tanstack-start

clone and run dev

### Expected behavior

`server.tsx` files should work according to [the updated docs](https://github.com/TanStack/router/blob/alpha/docs/start/framework/react/learn-the-basics.md#the-server-entry-point-optional).

### Screenshots or Videos

_No response_

### Platform

@tanstack/start [v1.120.4-alpha.5](https://github.com/TanStack/router/commit/016e5f0089d697371ba4d01ead392c934094b263)

### Additional context

_No response_

SeanCassiere:
This PR fixes it, so that a user can define a custom framework entry for both the `client` and `server` (in either `src/client.tsx` or `src/server.tsx`).

This PR also removes the unneeded `src/server.ts` files present as part of the `devinxi` migration.

Additionally, the React client entry has been change to wrap `hydrateRoot` with `startTransition`. This change useful, since after the SSR-rendered markup is sent to the client, `startTransition` lets React know to defer the hydration in favor of user events.

fixes #4138


packages/start-plugin-core/src/plugin.ts
```diff
@@ -125,7 +138,7 @@ export function TanStackStartVitePluginCore(
           define: {
             ...injectDefineEnv('TSS_PUBLIC_BASE', startConfig.public.base),
             ...injectDefineEnv('TSS_CLIENT_BASE', startConfig.client.base),
-            ...injectDefineEnv('TSS_CLIENT_ENTRY', startConfig.clientEntryPath),
+            ...injectDefineEnv('TSS_CLIENT_ENTRY', `/@fs${getClientEntryPath(startConfig)}`), // This is consumed by the router-manifest, where the entry point is imported after the dev refresh runtime is resolved
```
SeanCassiere, line 141: An example of the value printed here is `/@fs/Users/seanc/code/tanstack/router/e2e/react-start/src/client.tsx`).

The `/@fs` prefix is something Vite lets you use to access the file-system. This is erased by Vite at build time.

Issue #2413, opened 2024-09-25
Vite asset import gives a type error on examples or guide


henriquekirchheck:
### Which project does this relate to?

Start

### Describe the bug

When using the examples or the guide provided in the TanStack Start documentation to base your own project, the types for asset imports are not defined like it is in a default Vite app.

This can be observed with the `app.css` import on all examples, where it has to be annotated with `// @ts-expect-error`.

This can be fixed by adding `/// <reference types="vinxi/types/client" />` and `/// <reference types="vinxi/types/server" />` to their respective files, like shown in `vinxi`s example.

```tsx
// app/ssr.tsx
/// <reference types="vinxi/types/server" />
import {
  createStartHandler,
  defaultStreamHandler,
} from "@tanstack/start/server";
```

```tsx
// app/client.tsx
/// <reference types="vinxi/types/client" />
import { hydrateRoot } from 'react-dom/client'
import { StartClient } from '@tanstack/start'
import { createRouter } from './router'
```

### Your Example Website or App

https://stackblitz.com/github/tanstack/router/tree/main/examples/react/start-basic

### Steps to Reproduce the Bug or Issue

1. Go to `app/routes/__root.tsx` in the example.
2. Remove the `// @ts-expect-error` comment from the following snippet:
```tsx
import { NotFound } from '~/components/NotFound'
// @ts-expect-error
import appCss from '~/styles/app.css?url'
import { seo } from '~/utils/seo'
```
3. See the type error

### Expected behavior

I'd much prefer if this was handled like Vite already does it, where it gives you the correct type for the URL that is passed in.

### Screenshots or Videos

Type error on the CSS URL import
![Type error on the url](https://github.com/user-attachments/assets/7ed5823f-a393-4744-be4a-4366b26ab7fe)

Expected behavior after fix:
![Expected behavior after fix](https://github.com/user-attachments/assets/c5117734-ecf5-43d8-a843-da34a94b86d2)

### Platform

This is irrelevant for this problem, as it does not affect the resulting website/api, but I'll include it anyway.

- OS: Windows, Linux
- Browser: Firefox
- Version: 130.0.1

SeanCassiere:
Feel free to open a PR fixing the examples.

PR #4175, opened 2025-05-15
fix(example): fix typescript error for "../styles/app.css?url"

ImSingee:
Like #2413 

If there's an import `../styles/app.css?url` (or any similar, that import something with a `?url` postfix), there'll be a typescript error without referencing `vinxi/types/client`


SeanCassiere:
We've solved this in the alpha branch, by writing to `src/tanstack-start.d.ts` that adds this `references` import.

@ImSingee thank you for contributing this, but I'm closing this to reduce the merge conflicts we'll see when it comes time to getting the `alpha` branch imported.

PR #4180, opened 2025-05-15
fix: resolve custom `server` entry path on windows fs

yjoer:
This is a patch following https://github.com/TanStack/router/issues/4138 and https://github.com/TanStack/router/pull/4172. The virtual path should be kept intact when there isn't a custom `server.ts` entry file. Excluding the root path resolves the following error:

```
Error: Cannot find module 'D:~startdefault-server-entry' imported from '/~start/server-entry.tsx'
```

On Windows, `path.resolve(...)` returns a path with backslashes, e.g., D:\\router\\examples\\react\\start-basic\\src\\server.ts. Injecting it directly into an import statement leads to the following error. I replaced all backslashes with forward slashes and got it working on start-basic.

```
Error: Cannot find module 'D:routerexamplesreactstart-basic_srcserver.ts' imported from '/~start/server-entry.tsx'
```

### DIFF
--- a/e2e/react-start/basic/src/server.ts
+++ b/e2e/react-start/basic/src/server.ts
@@ -1,0 +1,11 @@
+// DO NOT DELETE THIS FILE!!!
+// This file is a good smoke test to make sure the custom server entry is working
+import {
+  createStartHandler,
+  defaultStreamHandler,
+} from '@tanstack/react-start/server'
+import { createRouter } from './router'
+
+export default createStartHandler({
+  createRouter,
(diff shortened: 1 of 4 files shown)


SeanCassiere:
@yjoer could you try this on Windows with this `vite.normalizePath` changes I've made?

yjoer:
I logged the paths with and without the server.ts file, and it seemed accurate. Thanks!

image

D:/router/examples/react/start-basic/src/server.ts
/~start/default-server-entry

PR #4184, opened 2025-05-16
fix: prepend `/@fs` when a custom client entry is detected

yjoer:
This is related to https://github.com/TanStack/router/pull/4179. If there isn't `@fs` in the client entry path when a custom client.tsx file exists, the browser loads the script via file:///D:/router/examples/react/start-basic/src/client.tsx, which will fail to load. Prepending `/@fs/` results in the correct request URL: http://localhost:3000/@fs/D:/router/examples/react/start-basic/src/client.tsx.

A `/` is also needed after `/@fs`, otherwise the request URL becomes http://localhost:3000/@fsD:/router/examples/react/start-basic/src/client.tsx, leading to 404.

Since `getClientEntryPath(...)` is also used at [plugin.ts#L85](https://github.com/TanStack/router/blob/alpha/packages/start-plugin-core/src/plugin.ts#L85), I am unsure of the impact of `@fs` there. Nevertheless, I didn't find issues when using the dev server and building the app (start-basic).

### DIFF
--- a/packages/start-plugin-core/src/plugin.ts
+++ b/packages/start-plugin-core/src/plugin.ts
@@ -77,7 +77,10 @@ export function TanStackStartVitePluginCore(
           )
             ? startConfig.clientEntryPath
             : vite.normalizePath(
-                path.resolve(startConfig.root, startConfig.clientEntryPath),
+                path.join(
+                  '/@fs',
+                  path.resolve(startConfig.root, startConfig.clientEntryPath),
+                ),
               )
 
           return entry


yjoer:
Oh, wait. `path.resolve()` didn't work that way.

SeanCassiere:
Sorry I took so long to get to this one. It needed to test in the monorepo, blank project using the test packages, and against the documentation website.

Checking out these changes, I'm not seeing any issues with the `/@fs` prefix being there anymore. This is both in `dev` and the `prod` builds in the scenarios of both having and not having a custom client entry.

It looks to be that the fix from https://github.com/TanStack/router/pull/4179 was needed since `/@fs` was being prepended to the client entry, irrespective of whether it was an actual physical custom entry file or a virtual module. As such, when a custom client entry wasn't present (which'd be the default for most), it'd try to resolve the _virtual module_ for the client entry using the `/@fs` prefix which'd screw with Vite.

Now that we're conditionally selecting the entry path, I'm confident that the addition of the `/@fs` when a custom client entry is present shouldn't be an issue.

SeanCassiere (approved):

PR #4201, opened 2025-05-18
fix(examples): add the required peer deps of @tanstack/router-devtools-core and vite-tsconfig-paths

antitoxic:
`@tanstack/router-devtools-core` which is a dependency of `@tanstack/react-router-devtools` have 2 missing peer dependencies in the `start-basic` example:

1. `@tanstack/router-core` I used the version of other `@tanstack/*` deps
2. `tiny-invariant` I used the version that `@tanstack/router-devtools-core` specifies 

`vite-tsconfig-paths` has 1 missing peer dependency: `vite`. I copied the version from react basic example.

I found that while doing a tanstack start POC using the [breakproof repo template](https://github.com/YotpoLtd/breakproof-base-monorepo) where the automatic peer dependency installation is disabled.

I can apply the same/similar fixes for other examples if you OK this.

### DIFF
--- a/examples/react/start-basic/package.json
+++ b/examples/react/start-basic/package.json
@@ -12,6 +12,7 @@
     "@tanstack/react-router": "^1.120.5",
     "@tanstack/react-router-devtools": "^1.120.6",
     "@tanstack/react-start": "^1.120.5",
+    "@tanstack/router-core": "^1.120.5",
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "tailwind-merge": "^2.6.0",
@@ -24,7 +25,9 @@
     "autoprefixer": "^10.4.20",
     "postcss": "^8.5.1",
     "tailwindcss": "^3.4.17",
+    "tiny-invariant": "^1.3.3",
     "typescript": "^5.7.2",
+    "vite": "^6.1.0",
     "vite-tsconfig-paths": "^5.1.4"
   }
 }


examples/react/start-basic/package.json
```diff
@@ -12,6 +12,7 @@
     "@tanstack/react-router": "^1.120.5",
     "@tanstack/react-router-devtools": "^1.120.6",
     "@tanstack/react-start": "^1.120.5",
+    "@tanstack/router-core": "^1.120.5",
```
schiller-manuel, line 15: this should not be necessary as it is a dep of `@tanstack/react-router`

schiller-manuel:
appreciate the effort, we should hold this back until we merge alpha in main

examples/react/start-basic/package.json
antitoxic: You are right that, this is not needed in default `npm` setup, but it is necessary in the setup our company and others like it could be using: 

`@tanstack/react-router-devtools` requires `@tanstack/router-core` as a **peer** and if you use a stricter dependency manager like `pnpm` it means that the implicit dependencies of `@tanstack/react-router` are **not shared** with the rest of your dependencies, a.k.a. `@tanstack/react-router-devtools` doesn't have access to the dependencies of `@tanstack/router-core`.

This is meaningful and intentional for portability, so that developers need to declare what dependencies they use, and not be able to import some implicit dependency of a dependency.

Does it make sense? I can try to rephrase if i wasn't able to express very well.

LadyBluenotes:
Thanks for reporting this and putting together a fix! 

The dependency setup has changed since this PR was opened, and these packages are no longer required by the current start-basic example, so I’m closing this because of that :)

PR #4204, opened 2025-05-18
docs: add jsdoc comments to plugin/generator types

RMHonor:
This PR adds JSDoc comments to the plugin/bundle config options, as discussed in https://github.com/TanStack/router/discussions/3790, giving usage hints inline and to hopefully pave the way for auto generated docs too.

The content for these docs were largely copied directly from the [Router API docs](https://tanstack.com/router/latest/docs/framework/react/api/router) with some tweaks to make it logical in the code context.

No logic changes have been made, purely type annotations.

### Note about Zod

When adding JSDoc comments to fields with Zod, they are lost when using the `z.input` utility types ([issue](https://github.com/colinhacks/zod/issues/200)). To get around this, an interface is created, which is annotated with JSDoc, and then the accuracy of this is checked against again the schema with `satisfies`. This interface is then using for the type arguments in the plugin/generator.

A small drawback in this approach is that the compiler won't detect extraneous options being added to the schema.

There's some discrepancy with how input vs. output types are used here, but I believe that may be intentional? (see PR comments).

Example output:
Screenshot 2025-05-18 at 22 52 14


packages/router-plugin/src/core/router-generator-plugin.ts
```diff
@@ -15,9 +15,10 @@ const setLock = (bool: boolean) => {
 const PLUGIN_NAME = 'unplugin:router-generator'
 
 export const unpluginRouterGeneratorFactory: UnpluginFactory<
-  Partial<Config> | undefined
+  ConfigOptions | undefined
 > = (options = {}) => {
   let ROOT: string = process.cwd()
+   
   let userConfig = options as Config
```
RMHonor, line 22: As an unrelated point, this type casting here is risky. `Config` is the `z.output` type, which will contain defined properties due to the default option, but what's being passed it hasn't been parsed by Zod yet (and so is the `z.input` type).

For example, `userConfig.routesDirectory` may not have been defined yet, so the usage on L24-26 is unsafe.

There is some complexity here, because we can't immediately parse the provided schema, as there may be later config retrieved from a `tsr.config.ts` file ([here](https://github.com/TanStack/router/blob/main/packages/router-generator/src/config.ts#L60)), and the default values would override anything provided in the config file. This could be fixed by allowing the file config to override the inline config, if that is the desired hierarchy.

packages/router-generator/src/config.ts
```diff
@@ -2,6 +2,192 @@ import path from 'node:path'
 import { existsSync, readFileSync } from 'node:fs'
 import { z } from 'zod'
 import { virtualRootRouteSchema } from './filesystem/virtual/config'
+import type { VirtualRootRoute } from '@tanstack/virtual-file-routes'
+
+export interface ConfigOptions {
```
zrosenbauer, line 7: big fan of this, helps a lot when in middle of coding :)

zrosenbauer (approved):
IDK if tanstack team wants to add them but I'm all for it 👍

PR #4210, opened 2025-05-20
fix: buildLocation

tannerlinsley:


packages/router-core/src/router.ts
```diff
@@ -1403,75 +1399,64 @@ export class RouterCore<
...
+      if (dest.to && dest.from) {
+        fromId = dest.from
+      }
+
+      const existingFrom = [...allFromMatches].reverse().find((d) => {
+        return d.fullPath === fromId || d.fullPath === joinPaths([fromId, '/'])
+      })
+
+      if (!existingFrom) {
+        console.warn(`Could not find match for from: ${dest.from}`)
```
SeanCassiere, line 1427: Is something that'll commonly be triggered? (i.e. like on tanstack.com?)

If so, should be silence the warning in production?

packages/router-core/src/router.ts
```diff
@@ -1403,75 +1399,64 @@ export class RouterCore<
...
       }
 
-      const prevParams = { ...last(fromMatches)?.params }
+      // From search should always use the current location
+      const fromSearch = last(allFromMatches)!.search
```
SeanCassiere, line 1431: Re-use `lastMatch` from above?

packages/react-router/src/link.tsx
```diff
@@ -106,7 +106,7 @@ export function useLinkProps<
     structuralSharing: true as any,
   })
 
-  const isRelativeFromPath = options.relative === 'path'
+  const isRelativeFromPath = options.unsafeRelative === 'path'
```
SeanCassiere, line 109: @birkskyum reminder about porting this behaviour and its tests over to Solid.

It doesn't need to block this PR, but just something to be done after.

packages/router-core/src/router.ts
tannerlinsley: Shouldn't be triggered much, even for unsafeRelative links, since they don't have a "from" necessarily and will always find a match from the nearest route/path

Issue #4064, opened 2025-04-28
Devtools shadowDOMTarget regression when upgrading from 1.112.17 to 1.117.1


reihwald:
### Which project does this relate to?

Router

### Describe the bug

The devtools no longer render correctly when rendering inside of a shadow dom while passing the shadowDOMTarget prop after upgrading from 1.112.17 to 1.117.1. 

The devtools look like this after upgrading:

![Image](https://github.com/user-attachments/assets/1394a738-fae3-4bbb-9545-eb778fafbe1d)

They are therefore mostly unusable.

### Your Example Website or App

https://stackblitz.com/edit/github-wnl18pbd?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

1. Let the project launch
2. The broken devtools should be immediately visible

### Expected behavior

I'd expect the devtools to be rendered correctly as they were before this regression. 

The devtools looked like this before:

![Image](https://github.com/user-attachments/assets/21a39f80-cbd8-42d8-bfdb-7879faabdfae)

### Platform

- OS: Windows
- Browser: Firefox
- Version: 137

bastianplsfix:
We have the same issue. However, it seems to almost work using the `TanStackRouterDevtoolsPanel ` but is missing some styles.

schiller-manuel:
cc @birkskyum

maxx-coffee:
@bastianplsfix & @reihwald , I've added a fix for this thats waiting for review.  It looks like the provider was missing from the core component.

PR #4212, opened 2025-05-20
fix: Fix devtools when using shadowdom

maxx-coffee:
I noticed devtools not rendering correctly after a recent update.  This fix wraps the DevTools component in the shadowdom context.  

Fixes: #4064

### DIFF
--- a/packages/router-devtools-core/src/TanStackRouterDevtoolsCore.tsx
+++ b/packages/router-devtools-core/src/TanStackRouterDevtoolsCore.tsx
@@ -4,7 +4,6 @@ import { ShadowDomTargetContext } from './context'
 import type { AnyRouter } from '@tanstack/router-core'
 import type { Signal } from 'solid-js'
 
-
 interface DevtoolsOptions {
   /**
    * Set this true if you want the dev tools to default to being open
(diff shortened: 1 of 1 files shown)


schiller-manuel:
which situations would this affect?
we should add this to one of our examples if we dont have it already

maxx-coffee:
@schiller-manuel I'll try to find some time to add it to an example or create a new one possibly tomorrow.  The scenario where this is an issue is when you are rendering the dev tools within a shadow dom.  Styles are rendered on the document and the click handlers are not working correctly. 

This wraps the devtools in the appropriate provider with the shadowroot passed in.

schiller-manuel:
@reihwald  not necessarily. it's just that we try to have examples / tests for every feature so we have a higher change of noticing when it breaks.
i am not opposed to merge as is

reihwald:
@schiller-manuel Makes sense. Do you have an idea of how this could be tested well? Maybe asserting that a specific style is getting applied using playwright?

reihwald:
It this case that wouldn't have helped. Do you still think adding this makes sense?

schiller-manuel:
let's merge

Issue #3989, opened 2025-04-14
Random SSR invariant dehydration errors with bun


ci:
### Which project does this relate to?

Router

### Describe the bug

I had some random Invariant failures in https://github.com/TanStack/router/blob/6e183bae4f87373ddcff9f1e8f7ab7b8b3712dbf/packages/start-client-core/src/ssr-client.tsx#L83-L86 with a blank page.

I've tried to debug it a bit, unsure exactly why that happens yet - it seemed to re-render an additional time when this crashed.

Here's a fun one: this only seems to happen using bun, and happens on the examples too.

### Your Example Website or App

start-basic-react-query

### Steps to Reproduce the Bug or Issue

```bash
bunx gitpick TanStack/router/tree/main/examples/react/start-basic-react-query repro
cd repro
bun install
SERVER_PRESET=bun bun run build
SERVER_PRESET=bun bun start
```

now go to http://localhost:3000 with "fast 4g" throttling in the network tab (makes it a bit more consistent for me, not necessary though), refresh a few times with it open

### Expected behavior

Shouldn't run into the Invariant issues.

### Platform

OSX locally, Linux (inside Docker) in prod

abdelmalekkkkk:
Also having this issue with Bun 1.2.4 and `@tanstack/react-start` 1.116.1. However, the issue disappears when I building using the `node-server` preset and run using node. So I am not sure if this is an issue with TanStack or Bun.

schiller-manuel:
streaming was fixed. if anything new comes up, please create a new issue including a complete minimal reproducer

PR #4228, opened 2025-05-22
fix: SSR invariant dehydration errors with bun

Cyenoch:
Related: https://github.com/TanStack/router/issues/3989

The script assigned to `\_\_TSR_SSR\_\_.dehydrated` appears outside the `</html>` tag. When the entry script `/_build/assets/client-BJoNZ2zg.js` is cached locally by the browser, even if the entry script is async, the script outside the html may still be executed after the entry script.

https://github.com/TanStack/router/blob/72ffeb4d4ae23a30859b9c7b8a3164535d5f2b85/packages/start-server-core/src/transformStreamWithRouter.ts#L171

Problems occur when `text` is complete HTML

https://github.com/TanStack/router/blob/72ffeb4d4ae23a30859b9c7b8a3164535d5f2b85/packages/start-server-core/src/transformStreamWithRouter.ts#L174-L175

https://github.com/TanStack/router/blob/72ffeb4d4ae23a30859b9c7b8a3164535d5f2b85/packages/start-server-core/src/transformStreamWithRouter.ts#L190

`bodyEndMatch` is assigned before the `!headStarted` branch processes chunkString.

https://github.com/TanStack/router/blob/72ffeb4d4ae23a30859b9c7b8a3164535d5f2b85/packages/start-server-core/src/transformStreamWithRouter.ts#L216

The length of `bodyEndIndex` obtained here will exceed the length of the modified `chunkString`, resulting in `<script>` being appended after `</html>`

Image

The solution is to move `bodyEndMatch` and `htmlEndMatch` to after the `chunkString` change.

### DIFF
--- a/packages/start-server-core/src/transformStreamWithRouter.ts
+++ b/packages/start-server-core/src/transformStreamWithRouter.ts
@@ -171,8 +171,6 @@ export function transformStreamWithRouter(
       const text = decodeChunk(chunk.value)
 
       let chunkString = leftover + text
-      const bodyEndMatch = chunkString.match(patternBodyEnd)
-      const htmlEndMatch = chunkString.match(patternHtmlEnd)
 
       if (!bodyStarted) {
         const bodyStartMatch = chunkString.match(patternBodyStart)
@@ -201,6 +199,9 @@ export function transformStreamWithRouter(
         leftover = ''
         return
       }
+
+      const bodyEndMatch = chunkString.match(patternBodyEnd)
+      const htmlEndMatch = chunkString.match(patternHtmlEnd)
 
       // If either the body end or html end is in the chunk,
       // We need to get all of our data in asap


Cyenoch:
I think it would be better if ensure that all tsr-once is before the entry script

Cyenoch:
image

Even if tsr-once is in html, Invariant failed will still appear. So I think if this script is before client entry script, the problem should be solved. Or the other way to make sure dehydrate is executed after __TSR_SSR__.dehydrated is assigned a value by other means.

Cyenoch:
Change script loading attribute from async to defer seems to solve this problem

Cyenoch:
> i found the issue, a synchronization error when injecting the scripts. I'll fix this as part of a rewrite of this section.

very thankful

PR #4231, opened 2025-05-22
feat(start): type level serialization

chorobin:
WIP


packages/router-core/src/serializer.ts
```diff
@@ -1,32 +1,150 @@
...
+      : { [K in keyof T]: TypeSerializerParseBy<T[K], TSerializable> }
 
 export interface DefaultSerializerExtensions {
   ReadableStream: unknown
 }
 
 export interface SerializerExtensions extends DefaultSerializerExtensions {}
 
-export type Serializable = Date | undefined | Error | FormData | bigint
+export type DefaultTypeSerializable =
```
schiller-manuel, line 41: what makes it the default? that it is supported by seroval without any plugin?

chorobin: Its just the existing serialization code, we can probably remove it

Issue #4245, opened 2025-05-24
Navigating while refetch is in progress results in cache not being used


webertim:
### Which project does this relate to?

Router

### Describe the bug

When defining a loader with long loading time I see the stale data as expected when navigating to the route from another route. The navigation also correctly triggers the refetch of the stale data.

Now when I navigate to another route after triggering the refetch without waiting for the refetch to be done, and navigating back to the route immediately, I don't get the stale data, but instead have to wait for the loader to be done.

Debug Tools show the stale loaderData with the 'fetching' status, so it seems like cached data should be there for display.

I created a fork on Stackblitz: https://stackblitz.com/edit/tanstack-router-wczba6rg?file=src%2Froutes%2Findex.tsx

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-wczba6rg?file=src%2Froutes%2Findex.tsx

### Steps to Reproduce the Bug or Issue

1. Go to the index route
2. Click the link `Go to /about`
3. Immediately click the link `Go to /`
4. Immediately click the link `Go to /about`
5. Immediately click the link `Go to /`

Now instead of immediately showing the stale data, an empty page is displayed until the loader is done.

### Expected behavior

As a user I expected to see the initially fetched stale data again instead of waiting for the loader to be done, but I have to wait until the loader is done to see the page, even though we should have an old value cached already.

### Screenshots or Videos

https://github.com/user-attachments/assets/bc79bfa5-5c52-40a0-bf70-99714074810d

Quick description of the video:
When I first visit the page I see nothing as long as the loader is loading, which is normal.
Then I click the link to visit the `/about` page and immediately navigate back to the `/` page.
Now the loader is fired again, but during the loading time the stale data (`0.600...`) is displayed until the loader is done after which the new data (`0.942...`) is displayed which is also as expected.
Next I again navigate to the `/about` page and back which shows the same behavior (we see the stale data `0.942...`), except this time I navigate to `/about` again while the loader is still loading. Trying to navigate back this time doesn't show the stale data (`0.94...`,) but 'blocks' as if there was no data.

### Platform

- OS: Windows
- Browser: Firefox
- Version: 138.0.4 (64-bit)

### Additional context

I already wrote a unit test for reproducing the behavior: #4246

PR #4246, opened 2025-05-24
tests: reproducer for #4245

webertim:


packages/react-router/tests/loaders.test.tsx
```diff
@@ -325,3 +326,68 @@ test('throw error from beforeLoad when navigating to route', async () => {
...
+    getParentRoute: () => rootRoute,
+    path: '/foo',
+    component: () => <Link to="/">index</Link>,
+  })
+
+  const routeTree = rootRoute.addChildren([indexRoute, fooRoute])
+  const router = createRouter({ routeTree })
+
+  render(<RouterProvider router={router} />)
+
```
schiller-manuel, line 361: add 

```tsx
await router.load()
```

to ensure the test progresses only after the loading is done

Issue #3624, opened 2025-03-01
routeTree generation breaks when using different root in vite config


ensconced:
### Which project does this relate to?

Router

### Describe the bug

In my vite config, I have `root` set to `"./src"`. Initially  this resulted in an error from the `TanStackRouterVite` plugin due to it trying to read the routes from `./src/src/routes`, but I resolved this by adding `routesDirectory: "./src/routes"` and `generatedRouteTree: "./src/routeTree.gen.ts"` to the plugin config.

With this resulting setup, everything seems to work, except that when I create new route files they don't get filled with the usual boilerplate code. And for the `routeTree.gen.ts` to be updated I have to kill and restart the vite dev server.

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-mzkds9em?file=vite.config.js,src%2Findex.html&preset=node

### Steps to Reproduce the Bug or Issue

1. Go to the linked stackblitz
2. See that the app works fine
3. Try creating a new route file - it doesn't get automatically filled with the route boilerplate code
4. Manually add the code for the new route file - see that you get a TS error from createFileRoute (because the routeTree.gen.ts hasn't been properly updated) until you kill and restart the vite dev server

### Expected behavior

- automatic filling of route files with boilerplate code should work regardless of `root` setting in vite config
- routeTree.gen.ts should get properly updated when adding/updating route files regardless of `root` setting in vite config

### Platform

- OS: macOS

eliellis:
I'm having the same issue, for now I've worked around it (somewhat) with the following Vite config (using `vite-plugin-restart`): 

```typescript
import { defineConfig } from "vite";
import tailwindcss from "@tailwindcss/vite";
import react from "@vitejs/plugin-react";
import tsconfigPaths from "vite-tsconfig-paths";
import ViteRestart from "vite-plugin-restart";
import { TanStackRouterVite } from "@tanstack/router-plugin/vite";

export default defineConfig({
  root: "src/frontend",
  build: {
    outDir: "../../dist",
  },
  plugins: [
    // remove once https://github.com/TanStack/router/issues/3624 is fixed
    ViteRestart({
      restart: ["routes/**/*.ts(x)?"],
    }),
    tsconfigPaths(),
    TanStackRouterVite({
      target: "react",
      // autoCodeSplitting: true,
      routesDirectory: "src/frontend/routes",
      generatedRouteTree: "src/frontend/routeTree.gen.ts",
    }),
    tailwindcss(),
    react(),
  ],
});

[...]

PR #4249, opened 2025-05-24
fix(vite-plugin): don't modify ROOT

RMHonor:
Fixes https://github.com/TanStack/router/issues/3624

Stackblitz of the error: https://stackblitz.com/edit/tanstack-router-f8savdzt?file=README.md

The docs state that the `routesDirectory` should be relative to the cwd ([link](https://tanstack.com/router/latest/docs/api/file-based-routing#routesdirectory-required)), however the plugin was attempted to modify the ROOT to whatever was specified in the Vite config.

This caused the generator to look in the wrong directory for the routes and the generated route tree.

Simply removing this modification of ROOT so the behaviour matches the plugin.

**Is this a breaking change?** No, the current behaviour doesn't work, and the workarounds (using an absolute route directory path) would not be broken when pulling this change.


schiller-manuel:
we should use this in one of the examples to ensure this does not regress

SeanCassiere:
@RMHonor Small miscommunication on my part.

Don't modify any of the existing sandboxes. Rather, add a new one.

schiller-manuel:
i am currently rewriting the generator and the plugins. can you please provide the example in this PR and I will then pick it onto my branch once I am done? trying to avoid merge conflicts

RMHonor:
Fixed in 1.121

Issue #4224, opened 2025-05-21
(devinxi alpha) CSS stylesheet links missing from SSR response


ilarramendi:
### Which project does this relate to?

Start

### Describe the bug

If we have a component that is used in 2 routes, the css of that component is split into its own bundle, this makes sense, but if this is the case the css should be loaded in the SSR response, the same as the route css

![Image](https://github.com/user-attachments/assets/d05217ac-f82f-4d6c-ad24-cd1625d7374c)
Here we can see app.css being loaded as soon as the page load and Counter.css being lodaded after the JS code finishes loading, this causes a jump in the component styles since it finishes loading after the SSR response rendered

### Your Example Website or App

https://codesandbox.io/p/devbox/youthful-solomon-nv98fl

### Steps to Reproduce the Bug or Issue

From start-bare alpha example just use counter in the about page too, and test a production build

### Expected behavior

Both css files being requested as stylesheets in the SRR response

### Platform

- OS: linux, windows
- Browser: Any
- Version: alpha

### Additional context

![Image](https://github.com/user-attachments/assets/2cb2ea1f-250b-4b4d-b861-bc8766f69a70)

![Image](https://github.com/user-attachments/assets/bc236c9a-cbe8-447e-83bf-b8431753d1e8)

PR #4254, opened 2025-05-25
fix(alpha): Recursively import CSS assets

ilarramendi:
This PR resolves an issue where not all CSS files needed to render a route were included in the SSR response. The root cause was that only CSS directly imported by the main route was being requested, leaving out styles from nested components.

The fix ensures CSS is now recursively imported, so all necessary styles—regardless of where they’re required—are properly included.

Before
image
After
image

I also updated the start bare example importing the counter compontent from the about page to show the error and the fix, and ensure this is not a problem in the future.

This is my first PR in this repo, im not sure if this breaks anything else but for my testing this works as expected and solves the issue!

### DIFF
--- a/examples/react/start-bare/src/routes/__root.tsx
+++ b/examples/react/start-bare/src/routes/__root.tsx
@@ -1,13 +1,13 @@
+import * as React from 'react'
 import { TanStackRouterDevtools } from '@tanstack/react-router-devtools'
 import {
-  createRootRoute,
   HeadContent,
   Link,
   Outlet,
   Scripts,
+  createRootRoute,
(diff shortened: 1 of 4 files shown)


SeanCassiere (approved):
@ilarramendi thanks!

Just to confirm, this resolves https://github.com/TanStack/router/issues/4224 right?

If so, I'll do a test locally and merge in.

ilarramendi:
@SeanCassiere Yes, for my testing its fixed with that.
Great job with the library BTW! Im really happy with how it works!

SeanCassiere (approved):

PR #4274, opened 2025-05-30
🔥 feat(meta): use generic default metatags unless not overridden

nrjdalal:
- packages/react-router/src/HeadContent.tsx
- packages/react-router/tests/Scripts.test.tsx

### DIFF
--- a/examples/react/start-workos/src/tanstack-start.d.ts
+++ b/examples/react/start-workos/src/tanstack-start.d.ts
@@ -1,2 +1,2 @@
 /// <reference types="vite/client" />
-import '../.tanstack-start/server-routes/routeTree.gen';
+import '../.tanstack-start/server-routes/routeTree.gen'
--- a/examples/react/start-workos/src/tanstack-start.d.ts
+++ b/examples/react/start-workos/src/tanstack-start.d.ts
@@ -1,2 +1,2 @@
 /// <reference types="vite/client" />
-import '../.tanstack-start/server-routes/routeTree.gen'
+import '../.tanstack-start/server-routes/routeTree.gen';
--- a/packages/react-router/src/HeadContent.tsx
+++ b/packages/react-router/src/HeadContent.tsx
@@ -4,6 +4,29 @@ import { useRouter } from './useRouter'
 import { useRouterState } from './useRouterState'
 import type { RouterManagedTag } from '@tanstack/router-core'
 
+const defaultMeta: Array<{
+  key: 'charSet' | 'viewport'
+  tag: 'meta'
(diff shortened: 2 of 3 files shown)


nrjdalal:
will make the changes shortly

SeanCassiere:
@birkskyum would this be the "correct" way to do this for Solid?

SeanCassiere (approved):
Looks okay to me. Awaiting @birkskyum's thoughts on this.

packages/router-core/src/head-content.ts
```diff
@@ -0,0 +1,38 @@
...
+  {
+    key: 'viewport',
+    tag: 'meta',
+    matcher: (m) => m?.name === 'viewport',
+    attrs: {
+      name: 'viewport',
+      content: 'width=device-width, initial-scale=1',
+    },
+  },
+]
```
schiller-manuel, line 24: unless we have a way of overriding the defaultMeta, we cannot add this
this could be an option of `createRouter` similar to how we allow overriding other default values

nrjdalal: the current PR already includes ability to override

if the user specifies charset: ascii, it respects that and doesn't add default charset meta

---

you are asking bout that only or something else

SeanCassiere: Adding it to `createRouter` would be redundant right? since you could just add it to your root route. 

In terms of opting out of these defaults, that's what the `matcher` does here. As soon as the chooses to user renders anything that goes against the `defaultMeta`, it chooses the user's preference over these defaults.

nrjdalal: will check if it exists for all subroutes and also if the user does charset instead of charSet

nrjdalal: done work for cases and subroutes

packages/router-core/src/head-content.ts
```diff
@@ -0,0 +1,38 @@
+import type { RouterManagedTag } from './manifest'
+
+const defaultMeta: Array<{
+  key: string
+  tag: 'meta'
+  matcher: (m: any) => boolean
+  attrs: Record<string, string>
+}> = [
+  {
+    key: 'charSet',
```
birkskyum, line 10: this is the line specifically i'm thinking won't work with solid

nrjdalal:
> In solid it's called e.g. `charset` not `charSet` like react, so the differences in naming have to be accounted for if in core. Changing `charSet` -> `charset` is one of the standard things to do when porting from react to solid, same as `className` -> `class`

it is case insensitive I guess, could always make em small, be that react/solid

PR #4297, opened 2025-06-02
fix: export OptionalFetcher and RequiredFetcher types

IderAghbal:
I hit a similar issue to #4001 (ts(2742)) but on pnpm, which requires exporting the types all the way to the module that gets directly imported by the application according to https://github.com/microsoft/TypeScript/issues/47663#issuecomment-1519138189.
This fix resolves this issue.


schiller-manuel:
also needed in `solid-start-client` I guess?

IderAghbal:
Ah, indeed! Sorry for only focusing on React, as it's what I am using.
Added now.

PR #4305, opened 2025-06-03
docs(guide): integrate better auth

nrjdalal:
- docs/start/framework/react/integrate-better-auth.md
- preview: https://github.com/nrjdalal/better-start


ajxbit:
@nrjdalal can you check the above issue as well? The above example does not work for `tanstackStart({ target: "cloudflare-module" })`

Meliketoaste:
Supporting the better-auth cli in Start https://www.better-auth.com/docs/concepts/cli#init would be really cool if you want to work on that.

docs/start/framework/react/integrate-better-auth.md
```diff
@@ -0,0 +1,489 @@
...
+
+export const { signIn, signOut, useSession } = createAuthClient()
+```
+
+```ts
+// src/routes/api/auth/$.ts
+
+import { auth } from '@/lib/auth'
+
+export const ServerRoute = createServerFileRoute().methods({
```
max-programming, line 265: The `createServerFileRoute` function should have the server route path in it

```suggestion
export const ServerRoute = createServerFileRoute('/api/auth/$').methods({
```

stunaz:
Maybe show a /protected route

docs/start/framework/react/integrate-better-auth.md
nrjdalal: should be before, not in alpha I guess

nrjdalal:
> Maybe show a /protected route

http://localhost:3000/dashboard is protected route only, step 7

docs/start/framework/react/integrate-better-auth.md
max-programming: Currently this is how server routes (previously api routes) are defined. I don't know why the docs page for server routes does not have this but it acts the same way like normal routes as it updates the routeTree file

Issue #3578, opened 2025-02-26
Empty context with parent / child `beforeLoad`, `loader` and `validateSearch`


cyrilchapon:
### Which project does this relate to?

Router

### Describe the bug

Hello,

I'm facing a problem where I get an empty context passed down to a child route uppon `validateSearch` URL rewrite.

The setup looks like this
- *__root.tsx*
- *_parent.tsx*
  - `beforeLoad` (adds `{ session }` to `context`)
  -  `loader` (reads `context.session`)
- *_parent/child.tsx*
  - `validateSearch`
  - `beforeLoad` (reads `context.session`)
  - `loader` (reads `context.session`)

*_parent.tsx*
```tsx
export const Route = createFileRoute('/_some-parent')({
  component: () => <Outlet />,
  beforeLoad: async () => {
    // Fake session fetch
    await new Promise((resolve) => setTimeout(resolve, 1000));
    const session = 'fake-session';

    console.log('parent beforeLoad', session);

    return { session };
  },
  loader: async ({ context: { session } }) => {
    console.log('parent loader', session);
  },
});
```

*_parent/child.tsx*
```tsx
const applicationSearchOptionsSchema = z.object({
  page: fallback(z.number().positive(), 1).default(1),
});

export const Route = createFileRoute('/_some-parent/some-child')({
  component: () => <div>Hello world</div>,
  validateSearch: zodValidator(applicationSearchOptionsSchema),
  beforeLoad: ({ context: { session } }) => {
    console.log('child beforeload', session);
  },
  loader: ({ context: { session } }) => {
    console.log('child loader', session);
  },
})
```
(NB: I'm using `zodValidator` and `fallback` syntax straight from the documentation)

When navigating to `/child?page=1`, everything behaves correctly :
```
> parent beforeLoad fake-session
> child beforeload fake-session
> parent loader fake-session
> child loader fake-session
```

But when **hard navigating** (not client side redirect) to `/child`, I'm getting an empty `context` with undefined `session` on loaders **first**; then after the "redirection" (it's not a redirect but you get my point) to `?page=1`, it loads correctly :
```
> parent beforeLoad fake-session
> child beforeload fake-session
> parent loader undefined // <= HERE
> child loader undefined // <= HERE

> parent beforeLoad fake-session
> child beforeload fake-session
> parent loader fake-session
> child loader fake-session
```

What is even stranger; is on local I'm even also getting and `undefined` on the child `beforeLoad` and not on Stackblitz :
```
[...]

PR #4306, opened 2025-06-03
fix: ensure child routes receive context returned from parent routes' `beforeLoad` functions, even if `invalidate()` is called during another load

jamesknelson:
Possibly related: 
 
- #3403
- #3578

I've been struggling with an issue where a parent route's context, as returned by `beforeLoad`, is missing (as in completely empty) in a child route, in certain circumstances, particular immediately after logging in. The types suggest this should be impossible, but there are circumstances in where it happens.

This PR adds a failing e2e test that reproduces the issue, and a tentative fix to the router itself.

## Details

To add authentication to the router context, I'm passing a `context` prop into `<RouterProvider>`. This context is stored in React state, contains the current user, and is updated after login.

Also, in order to get ensure that the routes are re-loaded after log out, I've added a call to `router.invalidate()` in an effect within the same component.

```jsx
const [authContext] = useAuth()

useEffect(() => {
  if (notInitialRender) {
    router.invalidate()
  }
}, [authContext])

return (
  <RouterProvider router={router} context={authContext} />
)
```

This works as expected when logging out. However, after logging in, the router is calling the `loader()` function of the new page *without* the context injected by its parent route's `beforeLoad()` function. In my app, this causes an error to be logged to the console on the first run through, and causes the app to crash on the second. This behavior is repeated in the e2e test I've added here.

Adding `console.log()` statements in various places, I've noticed that the `beforeLoad()` function *is* running before the child loader is called, so it's not that the context isn't available, it's just not being passed in.

While debugging, I noticed that `invalidate()` calls `router.load()`, even when the router is already loading, e.g. because of a `navigate()` call being made after login.

https://github.com/TanStack/router/blob/9db424fb5b0c48a3ca87cf44677ad618778277a3/packages/router-core/src/router.ts#L2797-L2827

I noticed that adding a guard to this `router.load()`, so that it is only called if the router is not already loading, fixed my issue, but I worry it would introduce further bugs down the track.

So the fix I ended up arriving on is to use the existing `latestLoadPromise` to wait for previous `router.load()` calls to complete before starting a new one, which resolved the issue for me, and got the new e2e test that I've added passing.

### DIFF
--- a/e2e/react-router/basic-auth/.gitignore
+++ b/e2e/react-router/basic-auth/.gitignore
@@ -1,0 +1,11 @@
+node_modules
+.DS_Store
+dist
+dist-hash
+dist-ssr
+*.local
+
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
--- a/e2e/react-router/basic-auth/index.html
+++ b/e2e/react-router/basic-auth/index.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vite App</title>
+  </head>
+  <body>
+    <div id="app"></div>
+    <script type="module" src="/src/main.tsx"></script>
+  </body>
+</html>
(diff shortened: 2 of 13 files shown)


schiller-manuel:
can we test this in a unit test as well?

schiller-manuel:
https://github.com/TanStack/router/pull/4433 fixes this

Issue #4209, opened 2025-05-20
solid: `Link` gets children outside of createRoot


Huliiiiii:
### Which project does this relate to?

Router

### Describe the bug

Minimal reproduction:

```ts
const NumberContext = createContext<number>()
function Reproduction() {
  return (
    <NumberContext.Provider value={123}>
      <Link >
        <Button />
      </Link>
    </NumberContext.Provider>
  )
}

function Button() {
  const context = useContext(NumberContext)
  if (!context) {
    throw new Error("No context")
  }
  return <button onClick={() => console.log("Context value", context)}>Print</button>
}
```

When I click the button it shows:
```
Context value 123
computations created outside a `createRoot` or `render` will never be disposed
Uncaught Error: No context
```
Obviously, `Link` gets children outside of createRoot.

If I comment out useLinkProps, everything works fine. So I think the issue is in the `useLinkProps` function.

### Your Example Website or App

https://stackblitz.com/edit/github-1deax8jh?file=src%2Froutes%2Findex.tsx

### Steps to Reproduce the Bug or Issue

Click `Print`

### Expected behavior

No error throws

### Platform

- OS: Windows
- Browser: Chrome
- Version: 136.0.7103.93

Abdelrahmanaman:
Any luck?

schiller-manuel:
cc @birkskyum

Woyken:
Created pull request with some fixes.
Wanted to do more components (noticed a bunch of props spreading in other places), but that would require quite a bit of time to clean up

schiller-manuel:
this is released now

PR #4314, opened 2025-06-04
fix(solid-router):  Change `<Link/>, <Block/>` properties to be reactive

Woyken:
Change <Link/> and <Block/> elements to use `createMemo` and `splitProps` to preserve reactivity.
Render children only once / Remove props spreading

#4209


brenelz:
This definately seems reasonable from what I can see

birkskyum (approved):

birkskyum:
@Woyken thanks for contributing this fix. Can you rebase the PR and make sure the test pass?

PR #4317, opened 2025-06-05
refactor: jsdoc and docs for RouteOptions

KirillTregubov:
This is a draft based on [this conversation](https://x.com/schanuelmiller/status/1930283556078551340).

Currently only implements jsdocs for RouteOptionsType, which is used in the `createRoute` function.

Any feedback / review is highly appreciated! Feel free to leave some anytime.


schiller-manuel:
ideally we can eventually generate the docs from the code.
@crutchcorn  which format should we adhere to in the code to make this work?

KirillTregubov:
> ideally we can eventually generate the docs from the code. @crutchcorn which format should we adhere to in the code to make this work?

I tried to keep all the jsdocs as similar to the current docs website content as possible. I may have shortened some of them because they had multiple paragraphs of description, but if this is the route you want to take, I can amend those few instances to include everything.

It would be ideal if you can settle on a format that works best. We can test it with RouteOptions, and then I will implement it for the other APIs.

SeanCassiere:

Issue #3687, opened 2025-03-07
Scripts not loading properly (Document Head Management)


johnyvelho:
### Which project does this relate to?

Router

### Describe the bug

External scripts are not getting loaded when adding it through scripts/HeadContent/Scripts component.
`<HeadContent />` and `<Scripts />` components are rendered in the __root.tsx.

It only get loaded if the flag `sync: true` is added to the script. but doing it the script may not be ready when the page loads.

```
export const Route = createFileRoute('/onboarding/$participant_id/(nostepper)/signature/docusign')({
    component: DocusignSignaturePage,
    preload: false,
    loader: async ({ params, context }) => {
        const response = await signatureMutation(params.participant_id);

        if (response.error) {
            throw new Error('There was an error starting the signature process');
        }

        return {
            docusignUrl: response.data.url,
        };
    },
    head: () => {
        return {
            scripts: [
                {
                    src: 'https://js-d.docusign.com/bundle.js',
                },
            ],
        };
    },
});
```

minimal example: https://stackblitz.com/edit/tanstack-router-frejwjfz
discussion on discord: https://discord.com/channels/719702312431386674/1007702008448426065/threads/1347295057870454804

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-frejwjfz

### Steps to Reproduce the Bug or Issue

1. navigate to /about
2. observe in the network events that bundle.js script from docusign is not getting loaded

### Expected behavior

external scripts get loaded with/without async flag

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

onionhammer:
I have yet to figure out how to have scripts or head sections of a route/index have any effect

lo1tuma:
I’ve ran into the same issue.

dburles:
I'm not having any luck either. The docs also aren't super clear on how to manage inline scripts and the provided example isn't correct.

GraxMonzo:
Adding `async: true` resolves the problem for me
```tsx
// ...
component: RootComponent,
scripts: () => [{ src: '/js/LunaPass.2.0.9.js', async: true }],
// ...
```

onionhammer:
The scripts seem to get injected into the head and run, but they dont get loaded before the component loads, which is probably quite important.

If you specify a head or body scripts for a component on a route, they should finish loading before rendering the route's component.

brenelz:
[...]

PR #4323, opened 2025-06-05
fix(react-router): make head function scripts load properly in SPA environment

w00khyung:
#3687 

Scripts returned from the head function in createFileRoute only get properly inserted and loaded when they have an async attribute. Scripts without the async attribute are not inserted at all.

Fixed the issue by modifying the ScriptAsset component in Asset.tsx to properly handle script loading through direct DOM manipulation.

### DIFF
--- a/packages/react-router/src/Asset.tsx
+++ b/packages/react-router/src/Asset.tsx
@@ -1,3 +1,4 @@
+import * as React from 'react'
 import type { RouterManagedTag } from '@tanstack/router-core'
 
 export function Asset({ tag, attrs, children }: RouterManagedTag): any {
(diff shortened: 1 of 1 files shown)


schiller-manuel:
but now this wont work during SSR anymore. so we need two different paths, one for SSR and one for client side rendering.
we also need the same solution for solid-router
cc @birkskyum

w00khyung:
I’ve updated the PR based on your feedback.

During SSR, it renders a <script> tag as before. On the client side, it dynamically appends the script via useEffect.
I also applied the same fix to solid-router to keep both implementations consistent.

Let me know if there’s anything else I should improve.

hyungwookchoi:
@schiller-manuel I’ve updated the PR based on your feedback.

During SSR, it renders a <script> tag as before. On the client side, it dynamically appends the script via useEffect.
I also applied the same fix to solid-router to keep both implementations consistent.

Let me know if there’s anything else I should improve.

brenelz:
Do we have e2e tests for this at all?

schiller-manuel:
yeah we need e2e tests for this. not necessarily in this PR. @brenelz can you add some in a different PR (that would fail now)?

schiller-manuel:
@w00khyung  can you please add an e2e test for both react and solid?

could be a route next to `e2e/solid-start/basic/src/routes/scripts.tsx`
first check if the script executes when SSRed, then do a client side navigation to that route and also check if the script executed

brenelz:
So a couple notes on this pr. It looks like the Solid version still fails the e2e test when client side navigating. The react one passes but the e2e tests seem to be doing a full page refresh. I'm not 100% sure why.


We can very simply add a e2e test in `solid-start` and `react-start` as follows.

```
// e2e/solid-start/basic/tests/navigation.spec.ts
test('client side navigating to a route with scripts', async ({ page }) => {
  await page.goto('/')
  await page.getByRole('link', { name: 'Scripts' }).click()
  expect(await page.evaluate('window.SCRIPT_1')).toBe(true)
  expect(await page.evaluate('window.SCRIPT_2')).toBe(undefined)
})
```

w00khyung:
I’ll write the E2E tests and take a look into the issue.

hyungwookchoi:
I’ll add e2e tests and look into the issue in Solid.

birkskyum:
The " client side navigating to a route with scripts " appear to be failing for both solid and react

schiller-manuel:
should we use this for solid? https://primitives.solidjs.community/package/script-loader/

birkskyum (approved):

schiller-manuel:
thanks a lot for your contribution!

PR #4341, opened 2025-06-07
feat: full document hydration for solid

brenelz:
We've known about this for awhile but we basically hydrate the document.body instead of the entire document.

To be on par with React we would like to give the end user control of the full html document.

This is my attempt at modifying our existing code to support that but I am currently stuck.

> template2 is not a function

image

### DIFF
--- a/examples/solid/start-bare/src/routes/__root.tsx
+++ b/examples/solid/start-bare/src/routes/__root.tsx
@@ -1,24 +1,50 @@
-import { createRootRoute, Link, Outlet } from '@tanstack/solid-router'
+import {
+  createRootRoute,
+  HeadContent,
+  Link,
+  Outlet,
+  Scripts,
+} from '@tanstack/solid-router'
 import appCss from '~/styles/app.css?url'
 import * as Solid from 'solid-js'
 import { TanStackRouterDevtools } from '@tanstack/solid-router-devtools'
+import { Hydration, HydrationScript, NoHydration } from 'solid-js/web'
 
 export const Route = createRootRoute({
   head: () => ({
     links: [{ rel: 'stylesheet', href: appCss }],
   }),
-  component: RootDocument,
+  component: RootComponent,
 })
+
+function RootComponent() {
+  return (
+    <RootDocument>
+      <Outlet />
+    </RootDocument>
+  )
+}
 
 function RootDocument({ children }: { children: Solid.JSX.Element }) {
   return (
-    <>
-      <div class="p-2 flex gap-2 text-lg">
-        <Link to="/">Index</Link>
-        <Link to="/about">About</Link>
-      </div>
-      <Outlet />
-      <TanStackRouterDevtools position="bottom-right" />
-    </>
+    <NoHydration>
+      <html>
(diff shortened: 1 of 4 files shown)


examples/solid/start-bare/src/routes/__root.tsx
```diff
@@ -1,24 +1,50 @@
...
+function RootDocument({ children }: { children: Solid.JSX.Element }) {
+  return (
+    <NoHydration>
+      <html>
+        <head>
+          <HydrationScript />
+          <HeadContent />
+        </head>
+        <body>
+          <Hydration>
```
lxsmnsyc: This should probably be similar to SolidStart's approach. The current `<NoHydration>` placement is correct but not for `<Hydration>`. We should pick a pre-rendered root as a hydration root (instead of `document.body`), so something like

```tsx
return (
  <body>
    <div id="app">
      <Hydration>
        {content}
      </Hydration>
    </div>
  </body>
);
```

This should make it easier to pick up on the client. That leaves you the `TanStackRouterDevtools` to handle later (which should be easy, it could be its own hydration root too).

brenelz: made this change to wrap in an app div

packages/start-server-core/src/createStartHandler.ts
```diff
@@ -49,8 +49,10 @@ export function getStartResponseHeaders(opts: { router: AnyRouter }) {
 
 export function createStartHandler<TRouter extends AnyRouter>({
   createRouter,
+  RootDocument,
 }: {
   createRouter: () => TRouter
+  RootDocument?: any
```
brenelz, line 55: i wasn't sure how to type this since its a Solid component but its in the core package

packages/start-server-core/src/handlerCallback.ts
```diff
@@ -5,6 +5,7 @@ export interface HandlerCallback<TRouter extends AnyRouter> {
     request: Request
     router: TRouter
     responseHeaders: Headers
+    RootDocument?: any
```
brenelz, line 8: same here

examples/solid/start-bare/src/routes/__root.tsx
```diff
@@ -1,24 +1,30 @@
...
   return (
     <>
+      <HeadContent />
       <div class="p-2 flex gap-2 text-lg">
         <Link to="/">Index</Link>
         <Link to="/about">About</Link>
       </div>
       <Outlet />
-      <TanStackRouterDevtools position="bottom-right" />
+      <Scripts />
```
lxsmnsyc, line 27: `Scripts` should be fine in the server entrypoint.

brenelz: The problem was the Scripts component needs to be in the "Router" context so that doesn't work outside just in the entry.

lxsmnsyc: ah I see, thanks for the clarification

brenelz:
This is so out of date i will just close

PR #4347, opened 2025-06-08
Slight simplification to llm rule generation now vibe-tools supports esm properly

eastlondoner:
Thank you so much for giving vibe-rules a go and for your feedback.

We have converted vibe-rules to be esm-first and to tolerate a range of package.json `exports` configurations so you now only need to export esm modules for it to work correctly (it will now also work correctly if you only export cjs `require` or if you only have a `main` or `default` export of either cjs or esm!)

We have added testing for a 2x2 matrix of

end-user package: `esm` | `cjs`
                                             x
library package:       `esm` | `cjs`                                             

to ensure that all cases work and there are no regressions in future.

That allows you to go back to using a single export and tsc as you added originally - this PR basically puts the flow back to the state of the original PR, although keeping code changes added in the other 2. 

I have added a `default` export in llms - but this is only necessary for backwards compatibility to previous versions of `vibe-rules`, if you only want to support the latest version of vibe-rules and above the default export is not needed.

You can see the changes that we made in vibe-rules here: https://github.com/FutureExcited/vibe-rules/tree/main/examples

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5802,8 +5802,8 @@ importers:
         specifier: ^19.0.0
         version: 19.0.0(react@19.0.0)
       vibe-rules:
-        specifier: ^0.2.55
-        version: 0.2.55
+        specifier: ^0.2.57
+        version: 0.2.57
       zod:
         specifier: ^3.24.2
         version: 3.25.32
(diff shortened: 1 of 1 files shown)


eastlondoner:
@jherr hope this makes things a bit simpler for you

packages/react-router/package.json
```diff
@@ -60,12 +60,12 @@
...
       "import": {
-        "types": "./dist/llms/esm/index.d.ts",
-        "default": "./dist/llms/esm/index.js"
+        "types": "./dist/llms/index.d.ts",
+        "default": "./dist/llms/index.js"
       },
-      "require": {
-        "types": "./dist/llms/cjs/index.d.cts",
-        "default": "./dist/llms/cjs/index.cjs"
+      "default": {
```
schiller-manuel: do we still need this then if vibe-rules is esm-first?

eastlondoner: Only needed if you want to support "previous" versions of `vibe-rules`

> I have added a default export in llms - but this is only necessary for backwards compatibility to previous versions of vibe-rules, if you only want to support the latest version of vibe-rules and above the default export is not needed.

eastlondoner: I've removed it!

schiller-manuel:
please update the pnpm-lock.yaml

PR #4350, opened 2025-06-09
fix(router-plugin): respect root configuration when generating routes

### LINKED ISSUE #3624: routeTree generation breaks when using different root in vite config

LeCarbonator:
Closes #3624 

This PR fixes a line that reuses `process.cwd()` instead of the provided `ROOT` variable. The side effects are that there can be a desync between features of the router plugins.

### DIFF
--- a/packages/router-plugin/src/core/router-generator-plugin.ts
+++ b/packages/router-plugin/src/core/router-generator-plugin.ts
@@ -34,7 +34,7 @@ export const unpluginRouterGeneratorFactory: UnpluginFactory<
     setLock(true)
 
     try {
-      await generator(userConfig, process.cwd())
+      await generator(userConfig, ROOT)
     } catch (err) {
       console.error(err)
       console.info()


LeCarbonator:
Should've checked first. Closed in favor of #4249

PR #4356, opened 2025-06-10
fix: update state key references from 'key' to '__key'

naoya7076:
# Description

Currently, when using TanStack Router, if users define a property named `key` within the navigation state, its value is overridden by the router's internal logic, which assigns a random key.

## For example:

```tsx
navigate({ to: '/', state: { color: 'blue', key: 123 } });

// Defined type
interface HistoryState {
  key?: number;
  otherValue?: string;
}

// On navigation target
console.log(location.state.color); // correctly "blue"
console.log(location.state.key);   // incorrectly overridden by a random string like "p7axhg"
```

# Expected Behavior

When explicitly providing a `key` within the navigation state, it should preserve the user's defined value rather than being overridden by the router's internal random key.

### DIFF
--- a/docs/router/framework/react/guide/scroll-restoration.md
+++ b/docs/router/framework/react/guide/scroll-restoration.md
@@ -71,7 +71,7 @@ Falling in behind Remix's own Scroll Restoration APIs, you can also customize th
 
 The `getKey` option receives the relevant `Location` state from TanStack Router and expects you to return a string to uniquely identify the scrollable measurements for that state.
 
-The default `getKey` is `(location) => location.state.key!`, where `key` is the unique key generated for each entry in the history.
+The default `getKey` is `(location) => location.state.__key!`, where `__key` is the unique key generated for each entry in the history.
 
 ## Examples
 
@@ -95,7 +95,7 @@ const router = createRouter({
     const paths = ['/', '/chat']
     return paths.includes(location.pathname)
       ? location.pathname
-      : location.state.key!
+      : location.state.__key!
   },
 })
 ```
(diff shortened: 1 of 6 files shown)


naoya7076:
@SeanCassiere Thank you again for your feedback. I've changed `__key` to `__TSR_key`.  If merging it for v2 is not too much trouble, I would be very happy to see it included.

SeanCassiere:
> @SeanCassiere Thank you again for your feedback. I've changed `__key` to `__TSR_key`.  If merging it for v2 is not too much trouble, I would be very happy to see it included.

@naoya7076 we currently do not have a v2 development branch open. As I mentioned earlier, I have added it onto the [v2 list](https://github.com/TanStack/router/discussions/975) so that it gets done for its development.

The only path forward here, in having this merged into `main` at the moment, would be to have **both** `key` and `__TSR_key` be added in and have their functionality/values be synced during the v1 cycle. The `key` property would be marked as `TODO: Remove in v2` using a comment. Then, when it comes time to begin working on v2, we'd go ahead and act on the `TODO`.

naoya7076:
@SeanCassiere Thank you for the guidance! I've updated the implementation as requested:

- Added both `key` and `__TSR_key` properties with synced values
- Marked `key` with `TODO: Remove in v2` comments
- Updated all affected packages and documentation

Thanks for the clear direction! 🙏

SeanCassiere (approved):

SeanCassiere:
@naoya7076 the only thing left to resolve is the failing test.

SeanCassiere:
I've updated the description and added the removal into the v2 checklist.

@naoya7076 thank you for taking this up and working with us on this one!

PR #4358, opened 2025-06-10
feat(router-generator): add option to format with specified tab width

DASPRiD:
First of, I know that there is a declined pull request which, among other things, implemented this: https://github.com/TanStack/router/pull/1817

That being said, this is really more about the format of newly created route files from templates, rather than the route tree (that one Is just ignored in my Biome config).

Right now, route files are always created with two spaces, even if you set a custom scaffolding for the templates with the right number of spaces. This would eventually be fixed by Biome (which I do run in commit hooks), but until then, an IDE like WebStorm will detect that the file has two spaces and thus default to that for anything you write.

This is just a minor inconvenience, but one nonetheless, as after each file creation the first thing I have to do is to click on "2 spaces" and then "reformat as 4 spaces".

An alternative to this would be to be able to set an alternate formatter command in the plugin options. Though that is a lot more involved, it would allow the user to have full control over formatting of new files.

### DIFF
--- a/docs/router/api/file-based-routing.md
+++ b/docs/router/api/file-based-routing.md
@@ -17,6 +17,7 @@ The following options are available for configuring the file-based routing:
 - [`routeToken`](#routetoken)
 - [`quoteStyle`](#quotestyle)
 - [`semicolons`](#semicolons)
+- [`tabWidth`](#tabwidth)
 - [`apiBase`](#apibase)
 - [`autoCodeSplitting`](#autocodesplitting)
 - [`disableTypes`](#disabletypes)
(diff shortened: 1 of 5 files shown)


SeanCassiere:
@DASPRiD thank you for opening this, but we're not adding anymore options to control the formatting of the initial generation.

The goal of the formatter here is just to write the initial content onto the file and the user should then begin making changes. Prettier is only really being used so that the initial code isn't a complete mess.

We'll likely be removing the existing formatting options in v2, since having those toggles aren't what prettier is being used for.

DASPRiD:
@SeanCassiere Could at least custom templates be respected and not be re-formatted?

SeanCassiere:
@DASPRiD it is not the goal here for the formatter. As I mentioned earlier, the goal of the formatter, is to give you a "not-ugly" starting point.

In a real-world scenario, it is not expected for a user to just create a route, do not make _any_ changes to it and immediately commit it to git and send it up for deployment. This is irrespective of whether or not you have a custom template.

The normal workflow is, you create a file, make whatever changes and hit 'save' at which time your installed formatter will make changes as it sees fit.

PR #4376, opened 2025-06-11
fix: remove solid-js dependency from react-router-devtools

jlucaso1:
We don't need to depend on solid-js to import JSX types in react-router-devtools package

### DIFF
--- a/packages/react-router-devtools/package.json
+++ b/packages/react-router-devtools/package.json
@@ -62,8 +62,7 @@
     "node": ">=12"
   },
   "dependencies": {
-    "@tanstack/router-devtools-core": "workspace:^",
-    "solid-js": "^1.9.5"
+    "@tanstack/router-devtools-core": "workspace:^"
   },
   "devDependencies": {
     "@vitejs/plugin-react": "^4.3.4",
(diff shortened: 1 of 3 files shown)


brenelz (approved):

luixo:
I'm not using solid js and got the same error with `@tanstack/react-router-devtools` being `v1.121.5`.
Seem like `@tanstack/router-devtools-core` is still [dependant](https://github.com/TanStack/router/blob/main/packages/router-devtools-core/src/useStyles.tsx#L2) on `solid-js` package internally.

brenelz:
Do we need to revert this? @luixo

jlucaso1:
Maybe just move the solid-js dep to be an normal dependency instead an development?
https://github.com/TanStack/router/blob/main/packages/router-devtools-core/package.json

Gregory-Gerard:
On a completely fresh project using Yarn, the project fails to start because solid-js isn’t installed. With NPM, which installs peer dependencies, this problem doesn’t occur.

```text
✘ [ERROR] Could not resolve "solid-js"
```

Repro: https://stackblitz.com/edit/tanstack-router-ws9cx1ym

PR #4378, opened 2025-06-11
fix(router-plugin): initialise generator within webpack plugin

vkunz:
hi,

init of config and generator was changed in 25c1369726f8550f12a3f13556e42d9cd615e7eb part of #4337.

The webpack plugin needs a fix to incorporate init changes.

Use the same init method as other plugins for webpack.

Thanks,
Valentin


brenelz:
@SeanCassiere does this seem reasonable?

Issue #4331, opened 2025-06-06
Codesplitting stops working with files that reside outside the routesDirectory


jeroenpg:
### Which project does this relate to?

Start

### Describe the bug

This issue only happens with virtual routes (and potentially code based?), specifically when registering route that is NOT contained in the routesDirectory specified in app.config.ts.

I suppose we're not really supposed to load routes outside of the routesDirectory, although you would expect this not to matter with virtual routes. 

### Your Example Website or App

https://stackblitz.com/~/github.com/jeroenpg/tanstack-solid-start-load-all-modules

### Steps to Reproduce the Bug or Issue

Open network tab, see that icons defined in bar.tsx are loaded on all routes, but not the ones in foo.tsx

### Expected behavior

I either expect tanstack to warn me if I use routes that don't reside in my `routesDirectory` OR code splitting should keep working.

### Platform

- OS: Linux
- Browser: Brave
- Version: n/a

### Additional context

Discussed this in discord thread
https://discord.com/channels/719702312431386674/1379480149627768912

PR #4379, opened 2025-06-11
fix: Code-splitting for virtual routes outside routesDirectory

jiji-hoon96:
## Problem

#4331
TanStack Start currently does not support code splitting for Virtual Routes located outside
the configured routesDirectory. This causes all route modules to load simultaneously
during development, resulting in excessive HTTP requests (2500+) and slow initial page load
times (5-10 seconds).

## Solution

- Moving code-splitting logic from start-plugin-core to router-plugin for better architectural separation
- Automatic detection of virtual routes using existing route discovery mechanism
- Extending code-splitting to cover all discovered route files, regardless of location
- No configuration required - works automatically with existing virtual route setup

## Architecture Improvement
Before: start-plugin-core → handles route discovery + code-splitting
After:  start-plugin-core → route discovery + exports route info router-plugin → code-splitting for all discovered routes

### DIFF
--- a/packages/router-plugin/src/core/config.ts
+++ b/packages/router-plugin/src/core/config.ts
@@ -69,6 +69,7 @@ export const configSchema = generatorConfigSchema.extend({
       return codeSplittingOptionsSchema.parse(v)
     })
     .optional(),
+  virtualRouteDirectories: z.array(z.string()).optional(),
 })
 
 export const getConfig = (inlineConfig: Partial<Config>, root: string) => {
--- a/packages/router-plugin/src/core/router-code-splitter-plugin.ts
+++ b/packages/router-plugin/src/core/router-code-splitter-plugin.ts
@@ -17,10 +17,7 @@ import {
   tsrSplit,
 } from './constants'
 import { decodeIdentifier } from './code-splitter/path-ids'
-import {
-  debug,
-  fileIsInRoutesDirectory,
-} from './utils'
+import { debug, fileIsInRoutesDirectory } from './utils'
 import type { CodeSplitGroupings, SplitRouteIdentNodes } from './constants'
 
 import type { Config } from './config'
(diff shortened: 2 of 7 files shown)


SeanCassiere (commented):
I don't think this is where this should be happening. This should be happening in the code-splitting plugin (in the `router-plugin` package) - https://github.com/TanStack/router/blob/37e5269f3b15052a1f32a5dfdbd848459572bdbf/packages/router-plugin/src/core/router-code-splitter-plugin.ts#L64

@schiller-manuel what are your thoughts to here? I'm thinking, that this is probably something where we can extract out the route nodes (used by the generator) and use it as a filter list in the code-splitting plugin.

jiji-hoon96:
@SeanCassiere Excellent point! I've moved the code-splitting logic to the router-plugin package as suggested.

### Key Changes
- Moved shouldCodeSplit logic from start-plugin-core → router-plugin
- Implemented route nodes information sharing mechanism
- Added virtualRouteDirectories support in router-plugin

The architecture now properly separates concerns.
- start-plugin-core: Route discovery & export route nodes info
- router-plugin: Code-splitting with virtual routes support

Thanks for guiding me toward the right architectural direction! Let me know if you'd like any further adjustments.

SeanCassiere (changes requested):
@jiji-hoon96 could you trim this down to be without the new `virtualRouteDirectories` option? We should be able to do this without introducing a new option.

I'd prefer to have the "fix" for code-splitting of the virtual routes separate to the new "feature" of `virtualRouteDirectories`.

I don't fully understand the requirement for the new feature, plus I'd like to discuss it with the other maintainers and get their thoughts on its requirement. If needed, they may have other ideas as to how the API for this should be implemented,

jiji-hoon96:
@schiller-manuel You're absolutely right to separate the bug fix from the new feature.

### Removed `virtualRouteDirectories` configuration option

The PR now fixes code-splitting for virtual routes without requiring any user configuration - it automatically detects and handles virtual routes using the existing route discovery mechanism.
The configuration feature can be discussed separately with the team if there's a clear use case for it.
Thanks for keeping the scope focused! 🎯

schiller-manuel:
thanks for looking into this. we already have all the information necessary in the router generator. I implemented this in https://github.com/TanStack/router/pull/4407

PR #4394, opened 2025-06-11
chore(examples): start-styled-components

Maxtremee:
For those of us stuck with CSS-in-JS and planning on migrating old apps to Tanstack Start in the future

Now updated after devinxi


delambo:
~~Is the `CustomizeStartHandler` API still available?~~ It is!

LadyBluenotes:
Thanks for putting this together. The styled-components migration use case is still useful, but this example is now far behind the current Start server setup.

I’ve opened #7781 to track a current implementation with the updated entrypoints and SSR coverage. A fresh PR based on that task would be welcomed, if you'd like to do it!

PR #4413, opened 2025-06-13
feat(react-router): allow to disable formatting of custom scaffolding

DASPRiD:
As #4358 was denied, I'd like to be able to disable formatting, so that the formatting of my custom scaffolding templates stays intact.


LadyBluenotes:
Thanks for following up on this. We discussed the custom-template case in #4358 and are keeping the same direction: the generator formats the initial file to provide a readable starting point, without adding formatting controls.

I’m closing this to keep that configuration surface small.

PR #4419, opened 2025-06-13
fix: specify better minimum vite version

RMHonor:
v1.121 of the router plugin makes use of the environment API from Vite ([here](https://github.com/TanStack/router/blob/main/packages/router-plugin/src/core/router-generator-plugin.ts#L77)). This API is only available in vite v6 onwards.

Updating the peer dep to reflect this.

Alternatively, we could safely navigate to the environment API? (`this.environment?.config.consumer`)


schiller-manuel:
thanks for raising this, we still support vite 5 for the router-plugin
added the optional access in https://github.com/TanStack/router/pull/4420

PR #4426, opened 2025-06-14
chore(examples): Some packages are missing in package.json in start-supabase-basic project

benny856694:


schiller-manuel:
this needs updating the lockfile as well

PR #4429, opened 2025-06-14
fix react-basic-ssr-streaming-file-based example

nlynzaad:
The react basic-ssr-streaming-file-based example is a bit outdated and no longer working.

This fix updates the server.js and entry-server.tsx to match those of the basic-ssr-file-based example as these were using the new createRequestHandler and defaultStreamHandler.

This also addresses an issue with the index route where the deferred value is causing an error when accessing the getDate() function. The value when streamed in is not being transformed back into a date. I assume this will be addressed with the serializer re-write/seroval but until then this at least allows the example to run.

### DIFF
--- a/examples/react/basic-ssr-streaming-file-based/package.disabled.json
+++ b/examples/react/basic-ssr-streaming-file-based/package.disabled.json
@@ -11,10 +11,10 @@
     "debug": "node --inspect-brk server"
   },
   "dependencies": {
-    "@tanstack/react-router": "^1.120.20",
-    "@tanstack/react-router-devtools": "^1.120.20",
-    "@tanstack/router-plugin": "^1.120.20",
-    "@tanstack/react-start": "^1.120.20",
+    "@tanstack/react-router": "^1.121.12",
+    "@tanstack/react-router-devtools": "^1.121.12",
+    "@tanstack/router-plugin": "^1.121.12",
+    "@tanstack/react-start": "^1.121.12",
     "get-port": "^7.1.0",
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
--- a/examples/react/basic-ssr-streaming-file-based/server.js
+++ b/examples/react/basic-ssr-streaming-file-based/server.js
@@ -1,3 +1,4 @@
+import path from 'node:path'
 import express from 'express'
 import getPort, { portNumbers } from 'get-port'
 
@@ -44,19 +45,19 @@ export async function createServer(
     try {
       const url = req.originalUrl
 
-      if (url.includes('.')) {
+      if (path.extname(url) !== '') {
         console.warn(`${url} is not valid router path`)
         res.status(404)
         res.end(`${url} is not valid router path`)
         return
       }
 
-      // Extract the head from vite's index transformation hook
+      // Best effort extraction of the head from vite's index transformation hook
       let viteHead = !isProd
         ? await vite.transformIndexHtml(
-            url,
-            `<html><head></head><body></body></html>`,
-          )
+          url,
+          `<html><head></head><body></body></html>`,
+        )
         : ''
 
       viteHead = viteHead.substring(
@@ -73,7 +74,7 @@ export async function createServer(
       })()
 
       console.info('Rendering: ', url, '...')
-      entry.render({ req, res, url, head: viteHead })
+      entry.render({ req, res, head: viteHead })
     } catch (e) {
       !isProd && vite.ssrFixStacktrace(e)
       console.info(e.stack)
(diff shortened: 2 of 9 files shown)


schiller-manuel:
thanks! can you please rename package.disabled.json back to package.json?

nlynzaad:
Will do. 

Was wondering why it was named as such but didn't want to change it.

schiller-manuel:
the name was just a way to disable it from building (and failing) in CI

schiller-manuel:
needs updating the lockfile as well

nlynzaad:
I see it fails a the build test if '@tanstack/react-start' is not added to the ssr external array in vite for this example.

I've added it to the PR

SeanCassiere (approved):

schiller-manuel:
please don't merge this yet, we need to still figure out how to handle the ssr noExternal setting properly

SeanCassiere:
Trying this out locally, whilst this does work in `dev`, it doesn't seem to work once a production build has been made.

SeanCassiere (changes requested):
Blocking this till we figure out both Manuel's comment and mine.

nlynzaad:
Thanks for looking at this.

I've made a couple of updates to resolve the ssr build actions for both server and client builds as well as serving in production.

When building this outside of the mono-repo this builds perfectly without the external clause and runs fine in prod and dev modes.

However as soon as we try and build it in the mono repo the below appears:

image

Honestly not too sure what might be the cause other than its clearly not able to resolve the virtual import in the mono-repo. If you have any ideas where to start looking I would be more than happy to continue scratching around and resolving it.

nlynzaad:
Scratched a bit further.

From what I can see the dependencies flow from start-server-core => react-start-server => react-start/server.

The virtual modules is part of start-server-core but does not get imported/used in the functions needed for the router ssr (non-start) functionality. However, this is included in the barrel exports and also marked as external dependencies in start-server-core. My guess is this is where it falls over when building in the mono-repo but passes when building for standalone.

By specifically marking the VIRTUAL_MODULES as external the build passes in both the mono-repo and standalone. And since this is not used (from what I can see) for SSR to function in a pure SPA router app this should not be a problem. 

Also ran this through the test suite and it passes all tests.

Ideally this should not be making its way into the build process at all but I'm not familiar enough with nx mono-repos to understand how to exclude this specifically in the build and other than stripping it out into either separate exports or alternative package this seems to be the path of least resistance. 

Open to any suggestions you might have.

nlynzaad:
@schiller-manuel @SeanCassiere 

I was able to avoid the externalising of the VIRTUAL_MODULES by bubbling up an additional export for the core ssr functions from start-server-core all the way through to react-start.

This is in addition to current exports and does not alter those at all. 

If this method is acceptable then I'm wondering if this should not be split into separate PRs probably since this is now more than just fixing an example. 

Alternatively, these functions can possible be split into a separate package that can be re-exported by start-server-core and hence not affect the current packages but can be used by a pure router only project for SSR purposes.

Again, open to your thoughts on this

schiller-manuel:
interesting! why does the separate export solve this?

nlynzaad:
My understanding is it has to do with how the bundler traverses the barrel exports coming from the dependencies. (see blog post written by TKDodo: [Please Stop Using Barrel Files](https://tkdodo.eu/blog/please-stop-using-barrel-files))

When running standalone (not part of the router mono-repo) these packages (and therefore the virtual modules) are included in the NODE_MODULES and by default don't get traversed. 

When part of the mono-repo it seems it is traversing the packages, and it then hits the VIRTUAL_MODULES which it can't find and the error occurs. Since the examples is part of the mono-repo the dependencies are resolved from the mono-repo and not from node_modules. As per vite's documentation: "Vite automatically detects dependencies that are not resolved from node_modules and treats the linked dep as source code." [Monorepos and Linked Dependencies](https://vite.dev/guide/dep-pre-bundling#monorepos-and-linked-dependencies).

This is not a problem in pure start examples since that uses the start plugin that initializes the VIRTUAL_MODULES but for pure router examples these don't get initialized and it breaks the build.

When doing a separate export for only the essential modules that is required for SSR, the loadVirtualModules.js file is not traversed as it doesn't form part of the export and hence the problem is avoided.

nlynzaad:
The real problem here seems to be that we are using the start package to do the SSR for the router project. It blurs the already grey line separating start from router even more. 

I assume you would not be particularly keen on introducing a new package, but the cleanest and most modular solution is probably to move these SSR essentials to a separate package (for example router-ssr). It could even be moved to router-core but this would muddy the responsibilities that router-core has.

Start-server-core could depend on and re-export as it does currently for any project/package that depends on it while router projects would then depend on this "router-ssr" project and would not need to depend on start and hence the import tree is kept clean from start only functionality.

schiller-manuel:
i like the idea of having separate packages. we just need to find a clean separation. For example, where should the serialization be put?

nlynzaad:
The broad idea would be to separate anything that is generic ssr into this package.

This is probably a huge over simplification, but I will work on a proposal tonight and send it through.

nlynzaad:
closing in favour of #4495

PR #4442, opened 2025-06-16
Add support for Zod v4

### LINKED ISSUE #4322: Zod 4 support for `@tanstack/zod-adapter` (specifically `fallback`)
**Update 2025-08-10**: 
**tldr:** ✅ Everything works natively now via `.default('').catch('')`. No more `zod-adapter` required for Zod 4 as long as you use Zod >= [v4.0.6](https://github.com/colinhacks/zod/releases/tag/v4.0.6) 

**Update 2025-07-16**: 
Nice! With ~the newest update from zod~ a new zod update coming out (hopefully) soon, `fallback` ~is not~ will not be necessary anymore! 🎊 ([More infos](https://github.com/TanStack/router/pull/4442#issuecomment-3078284064))

____________________

Discussion: https://github.com/TanStack/router/discussions/4092

Router (`@tanstack/zod-adapter`)

### Describe the bug

It appears that `fallback` from `@tanstack/zod-adapter` isn't ready for Zod v4.
More information about the expected behavior is in my comment below: https://github.com/TanStack/router/issues/4322#issuecomment-2948514758

```ts
import { z } from "zod/v4";
import { fallback, zodValidator } from "@tanstack/zod-adapter";

const searchParamSchema = z.object({
  test: fallback(z.string().optional(), null).default(null),
//               ^ ❗ Type Error at z.string().optional()
});
```

**Type Error:**
```
Argument of type 'ZodOptional<ZodString>' is not assignable to parameter of type 'ZodTypeAny'.
  Type 'ZodOptional<ZodString>' is missing the following properties from type 'ZodType<any, any, any>': _type, _parse, _getType, _getOrReturnCtx, and 7 more.ts(2345)
```

### Expected behavior

Zod 4 support

niba:
fixes #4322 

In Zod v4, we still need a fallback trick to prevent type loss.
Added tests for Zod v4 and the fallback case

related discussion #4092

### DIFF
--- a/packages/zod-adapter/package.json
+++ b/packages/zod-adapter/package.json
@@ -67,7 +67,7 @@
     "@testing-library/jest-dom": "^6.6.3",
     "@testing-library/react": "^16.2.0",
     "@tanstack/react-router": "workspace:^",
-    "zod": "^3.24.2",
+    "zod": "^3.25.64",
     "react": "^19.0.0",
     "react-dom": "^19.0.0"
   },
--- a/packages/zod-adapter/src/index.ts
+++ b/packages/zod-adapter/src/index.ts
@@ -1,4 +1,5 @@
-import { z } from 'zod'
+import * as z3 from 'zod'
+import * as z4 from 'zod/v4'
 import type { ValidatorAdapter } from '@tanstack/react-router'
 
 export interface ZodTypeLike {
@@ -58,12 +59,32 @@ export const zodValidator = <
   }
 }
 
-export const fallback = <TSchema extends z.ZodTypeAny>(
+const isZod3Schema = (schema: any): schema is z3.ZodSchema => {
+  return (
+    '_def' in schema &&
+    typeof schema._def === 'object' &&
+    'typeName' in schema._def
+  )
+}
+
+export function fallback<TSchema extends z3.ZodTypeAny>(
   schema: TSchema,
   fallback: TSchema['_input'],
-): z.ZodPipeline<
-  z.ZodType<TSchema['_input'], z.ZodTypeDef, TSchema['_input']>,
-  z.ZodCatch<TSchema>
-> => {
-  return z.custom<TSchema['_input']>().pipe(schema.catch(fallback))
+): z3.ZodPipeline<
+  z3.ZodType<TSchema['_input'], z3.ZodTypeDef, TSchema['_input']>,
+  z3.ZodCatch<TSchema>
+>
+export function fallback<TSchema extends z4.ZodTypeAny>(
+  schema: TSchema,
+  fallback: z4.output<TSchema>,
+): z4.ZodPipe<
+  z4.ZodType<TSchema['_input'], TSchema['_input']>,
+  z4.ZodCatch<TSchema>
+>
+export function fallback(schema: any, fallback: any): any {
+  if (isZod3Schema(schema)) {
+    return z3.custom().pipe(schema.catch(fallback))
+  } else {
+    return z4.custom().pipe(schema.catch(fallback))
+  }
 }
--- a/packages/zod-adapter/tests/zod-v3.test-d.ts
+++ b/packages/zod-adapter/tests/zod-v3.test-d.ts
@@ -1,0 +1,286 @@
+import {
+  createRootRoute,
+  createRoute,
+  createRouter,
+  Link,
+} from '@tanstack/react-router'
+import { test, expectTypeOf } from 'vitest'
+import { fallback, zodValidator } from '../src'
+import { z } from 'zod'
+
+test('when creating a route with zod validation', () => {
+  const rootRoute = createRootRoute({
+    validateSearch: zodValidator(
+      z.object({
+        page: z.number().optional().default(0),
+      }),
+    ),
+  })
+
+  const indexRoute = createRoute({
+    getParentRoute: () => rootRoute,
+    path: '/',
+    validateSearch: zodValidator(
+      z.object({
+        indexPage: z.number().optional().default(0),
+      }),
+    ),
+  })
+
+  const routeTree = rootRoute.addChildren([indexRoute])
+
+  const router = createRouter({ routeTree })
+
+  expectTypeOf(Link<typeof router, string, '/'>)
+    .parameter(0)
+    .toHaveProperty('search')
+    .exclude<Function | true>()
+    .toEqualTypeOf<{ page?: number; indexPage?: number } | undefined>()
+
+  expectTypeOf(Link<typeof router, string, '/'>)
+    .parameter(0)
+    .toHaveProperty('search')
+    .returns.toEqualTypeOf<{ page?: number; indexPage?: number }>()
+
+  expectTypeOf(Link<typeof router, '/', '/'>)
+    .parameter(0)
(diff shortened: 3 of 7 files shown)


schiller-manuel:
thanks for starting this. however, we need to discuss first whether / how we want to continue with the whole adapter setup

schiller-manuel:
this fails to build, can you please have a look? 

https://cloud.nx.app/runs/q1gPhPfKu4/task/tanstack-search-validator-adapters%3Abuild

niba:
Build should be fixed

There are a few things we should be aware of before merging it:

1.  If your project enforces a different version of `Zod` than the one used by the `zod-adapter` package, you might encounter the error `Type instantiation is excessively deep and possibly infinite.` This issue affected the build (wrong peerDependency and devDependency in zod-adapter), and there's a long discussion about it here: https://github.com/colinhacks/zod/issues/2697
2.  `Zod 4` introduces a "mini" version with a tree-shakeable API. However, in our case, it's complicated to use because you cannot chain the result of a fallback with "default," etc.
3.  The bundle size increases by 20 kB after adding support for `Zod 4` (using the mini version can reduce this to 6 kB).

```
Before this PR:
dist/index.html                          0.39 kB │ gzip:   0.26 kB
dist/assets/index-DjYYQclb.css           6.23 kB │ gzip:   1.85 kB
dist/assets/Search-De6RLLiF.js           0.31 kB │ gzip:   0.24 kB
dist/assets/valibot.index-CY1Ddu6G.js    0.42 kB │ gzip:   0.28 kB
dist/assets/zod.index-DZLhYpWM.js        0.43 kB │ gzip:   0.29 kB
dist/assets/arktype.index-B-cgD59V.js    0.43 kB │ gzip:   0.29 kB
dist/assets/index-BhpiWg34.js          507.92 kB │ gzip: 153.78 kB

After this PR:
dist/index.html                          0.39 kB │ gzip:   0.26 kB
dist/assets/index-DjYYQclb.css           6.23 kB │ gzip:   1.85 kB
dist/assets/Search-Cy5gcvTr.js           0.31 kB │ gzip:   0.24 kB
dist/assets/valibot.index-CC4lPNzG.js    0.42 kB │ gzip:   0.28 kB
dist/assets/zod.index-C6nO_o6N.js        0.43 kB │ gzip:   0.29 kB
dist/assets/arktype.index-B0G5htEQ.js    0.43 kB │ gzip:   0.29 kB
dist/assets/index-BRnstef1.js          527.57 kB │ gzip: 159.04 kB
```


I believe that for the functionality we are adding, this code introduces too much complexity. Instead, I think we should document how to write a helper function for your Zod / validator.

schiller-manuel:
> Instead, I think we should document how to write a helper function for your Zod / validator.

yes! do you want to create docs for this?

michael-wolfenden:
@niba Is there a reason why we have to specify both a fallback and a default rather than just a fallback. I can't imagine a scenario where you would want a different value for your default vs your fallback.

For example

```ts
 sort: fallback(z.enum(['oldest', 'newest']), 'oldest').default('oldest')
 ```
vs

```ts
 sort: fallback(z.enum(['oldest', 'newest']), 'oldest')
 ```
 
 I did notice though that when a `default` is not provided then search parameters must be passed to the `Link` component

**Comment by niba** [comment]
@michael-wolfenden They handle different cases, but it's true that you almost always use a combination of them. I also don't know of a scenario where you wouldn't want to use `default`.

@schiller-manuel
Sure! I can try. First, I need to process everything. Basically, the problem only exists with Zod because of how the schema is inferred (using standard schema).
In Zod 3, the problem was that `catch` destroyed the entire type.
In Zod 4, the type is inferred correctly, but it introduces a "whatever" type for `catch` that allows you to use any value.

```typescript
const schema = z.object({
  page: z.number().default(1),
  filter: z.string().default(''),
  sort: z.enum(['newest', 'oldest', 'price']).default('newest').catch("newest"),
})

/*
Zod 3 output: {
  page?: number | undefined;
  filter?: string | undefined;
  sort?: unknown;
}
*/
type Zod3Type = StandardSchemaV1.InferInput<typeof schema>

/* 
Zod 4 output: {
  page?: number | undefined;
  filter?: string | undefined;
  sort?: z4.core.util.Whatever | "newest" | "oldest" | "price";
} */
type Zod4Type = StandardSchemaV1.InferInput<typeof schemaZ4>
```

I don't understand why Zod4 generates this `whatever` type for `catch`. I think that most people use `catch` as a safety mechanism and don't want to "relax" the types at all. I will try to ask on zod repo.

schiller-manuel:
> I will try to ask on zod repo.

absolutely!
as you wrote, this is only needed for zod, other libraries (e.g. arktype and valibot) just work with standard schema.
we can document those workarounds for zod and then deprecate the adapter packages.

packages/zod-adapter/package.json
```diff
@@ -67,12 +67,12 @@
...
     "@testing-library/react": "^16.2.0",
     "@tanstack/react-router": "workspace:^",
-    "zod": "^3.24.2",
+    "zod": "^3.25.0",
     "react": "^19.0.0",
     "react-dom": "^19.0.0"
   },
   "peerDependencies": {
-    "zod": "^3.23.8",
+    "zod": "^3.25.0",
```
hanneswidrig, line 75: https://zod.dev/library-authors

With zod 4.0.x released last week, this line needs to be modified slightly to include version four releases.

`"zod": "^3.25.0 || ^4.0.0"`

This value also needs to be set for devDependencies too.

packages/zod-adapter/src/index.ts
```diff
@@ -1,4 +1,5 @@
-import { z } from 'zod'
+import * as z3 from 'zod/v3'
+import * as z4 from 'zod/v4'
```
hanneswidrig, line 2: This import should be modified slightly to include the `/core` path.

`import * as z4 from "zod/v4/core";`

niba:
@hanneswidrig  thanks for the review!

Zod v4 almost supported TanStack Router without requiring an adapter. The only issue was related to type inference when using ⁠`.catch()`.

I forwarded our issue to the Zod team, and they've already fixed the problem we were facing (https://github.com/colinhacks/zod/issues/4851).

With this change, the `⁠zodAdapter` is no longer necessary to achieve type safety in TanStack Router. We can now use ⁠`zod` directly:

```typescript
export const Route = createFileRoute('/shop/products/')({
  validateSearch: z.object({
       page: z.number().default(1),
       filter: z.string().default(''),
       sort: z.enum(['newest', 'oldest', 'price']).default('newest').catch("newest"),
   }),
})
```

I'm waiting for the new release of zod before opening a PR to update the documentation in Tanstack Router. Should we close this one? @schiller-manuel

schiller-manuel:
@niba very cool! thanks for that!

yes, let's close this one and create a new one for the docs.

colinhacks:
@niba @schiller-manuel This has landed in Zod 4.0.6, apologies for the delay!

packages/zod-adapter/package.json
altanbgn: A bit annoying for npm users when this version is not 4. Is there any discussion or issue that is attempting to update this?

PR #4447, opened 2025-06-17
docs: added missing middleware method chaining in context example

berz8:
The example "[Providing context to the next middleware via next](https://tanstack.com/start/latest/docs/framework/react/middleware#providing-context-to-the-next-middleware-via-next)" was missing the middleware method chaining to provide the first middleware context to the second one


schiller-manuel:
thanks!

PR #4457, opened 2025-06-18
docs: Update head management instructions for SPA's

rhuanbarreto:
This is important in the case you are doing SPA's. If you have the title tag set in the `index.html` file from Vite the component won't work as intended.


schiller-manuel:
thanks!

PR #4472, opened 2025-06-19
fix: navigation respects basepath when no `to` was given

### LINKED ISSUE #4401: Wrong destination path calculated on navigation when "basepath" is set for Router

Router

### Describe the bug

The issue reproduces when creating Router instance with `basepath` router option specified.

When navigating via `<Link>` or using the function from `useNavigate` hook, it does not work properly if the `to` property/ option is not specified.
In this case, the core router logic takes `fullPath` value from the latest route match which does not include the base path part: https://github.com/TanStack/router/blob/bab29d0e88945fa72f55ca3284461bb47c550f50/packages/router-core/src/router.ts#L1438-L1441

The navigation works as expected if the `to` property/option is specified, as a full or relative path.

https://stackblitz.com/edit/github-cwqd7wap?file=src%2Fmain.tsx

Create a Router instance with `basepath` option specified.
```js
const router = createRouter({
  routeTree,
  basepath: '/base', 
  // ...
});
```

Navigate using `<Link>` or `useNavigate` hook. Do not specify `to` property/option.
```jsx
// Navigated to /base/some-page 

// Navigates to "/some-page?foo=bar" instead of "/base/some-page?foo-bar"
<Link search={{ foo: 'bar' }}>Update search params</Link>

// Also navigates to "/some-page?foo=bar" instead of "/base/some-page?foo-bar"
navigate({ search: { foo: 'bar'  } });
```

### Expected behavior

When the `to` property/option is not specified, the destination path is expected be calculated properly including base path.

### Screenshots or Videos

_No response_

### Platform

- OS: any
- Browser: any
- Version: "@tanstack/react-router": "^1.121.0"

### Additional context

_No response_

### LINKED ISSUE #4471: useNavigate() with absolute paths strips basepath during navigation in v1.121.21

Router

### Describe the bug

When using `useNavigate()` with absolute paths in components, the router basepath is stripped from the URL during navigation, particularly when URL parameters are involved. This breaks routing for applications deployed with a basepath (e.g., `/streaming`).
The issue appears to be similar to #2046, #1779, and #3436, but specifically affects `useNavigate()` usage with absolute paths.

```typescript
// router.ts
import { queryClient } from '@providers/QueryProvider/QueryProvider.util.ts';
import { createRouter } from '@tanstack/react-router';
import { routeTree } from './routeTree.gen.ts';

    routeTree,
    basepath: import.meta.env.PROD ? '/streaming' : import.meta.env.BASE_URL,
    defaultPreload: 'intent',
    scrollRestoration: true,
    defaultStructuralSharing: true,
    context: {
        auth: undefined!,
        queryClient: queryClient,
    },
});

// Component using useNavigate
import { useNavigate } from '@tanstack/react-router';

export const GenericFolder = ({ type }) => {

    const onClick = () => {
        navigate({
            to: `/saved/folder/generic/${type}`, // Absolute path
        });
    };

    return <div onClick={onClick}>Navigate</div>;
};
```

https://stackblitz.com/github/tanstack/router/tree/main/examples/react/quickstart-file-based?file=src%2Fmain.tsx

1. Set up TanStack Router with a basepath (e.g., `basepath: '/streaming'`)
2. Deploy application so it runs at `https://example.com/streaming`
3. Create a component that uses `useNavigate()` with absolute paths (like `/saved/folder/generic/${type}`)
4. Navigate to a page that renders this component (e.g., https://example.com/streaming/dashboard)
5. Trigger the navigation that adds URL parameters or changes the route
6. Observe that the URL becomes `https://example.com/saved/folder/generic/sometype` instead of `https://example.com/streaming/saved/folder/generic/sometype`

### Expected behavior

When using absolute paths with `useNavigate()`, the basepath should be preserved in the final URL. The navigation should result in `https://example.com/streaming/saved/folder/generic/sometype`.

Actual behavior
The basepath (`/streaming`) is stripped from the URL, resulting in `https://example.com/saved/folder/generic/sometype`, which breaks the application routing and leads to a 404 or "not found" page.

### Additional context

Environment
- TanStack Router version: 1.121.21
- Framework: React 19.1.0
- Build tool: Vite 6.3.5
- Deployment: Production environment with basepath

Package.json dependencies
```json
{
  "@tanstack/react-router": "1.121.21",
  "@tanstack/react-router-devtools": "1.121.21",
  "@tanstack/router-plugin": "1.121.22",
  "@tanstack/zod-adapter": "1.121.21"
}
```

Version Information

Working version: 1.120.20 (but has build errors due to generator export changes)
Broken version: 1.121.21+ (navigation works but basepath is stripped)

### Root Cause Analysis
Based on the TanStack Router documentation: "If a from route path isn't provided the router will assume you are navigating from the root / route and only auto-complete absolute paths".
This suggests that useNavigate() without a from property bypasses basepath handling when using absolute paths, while router.navigate() properly maintains the router context.

### Impact on Developer Experience
This issue particularly affects:
1. Reusable components that need to navigate but can't specify a from property (since they don't know where they'll be rendered)
2. Monorepo/microfrontend architectures where components are shared across different applications
3. Dynamic routing where the navigation target is computed at runtime

### Related Issues
- #2046: Basepaths being decoded during navigation
- #1779: Basepath breaks routing when found within path
- #3436: Relative routing issue with useNavigate()

jvllmr:
Simple fix by passing "." to `Router.resolvePathWithBase` instead of using `fromPath`. Parameterized relative URL tests with basepath to avoid future regressions. I also managed to reproduce the issue this way.

fixes #4401 
fixes #4471


schiller-manuel:
thanks for your PR!

I am wondering why this works, given that this user wrote:

https://github.com/TanStack/router/issues/4401#issuecomment-2976060849
Screenshot 2025-06-21 at 01 05 50

SeanCassiere (changes requested):
This does look OK to me.

@schiller-manuel’s point is interesting. Worth looking into.

@jvllmr please:

- Replicate the tests in the Solid package
- Use unique test names for each run: `test name xyz for $basepath`

jvllmr:
> This does look OK to me.
> 
> @schiller-manuel’s point is interesting. Worth looking into.
> 
> @jvllmr please:
> 
> * Replicate the tests in the Solid package
> * Use unique test names for each run: `test name xyz for $basepath`

I wondered about that case as well, but I focused on reproducing the problem with tests and started with the clue from the original issue. I could add the case from the issue to the tests as well just to be sure.

I made the test names unique and started on porting the tests to solid as well and pushed my progress for now because I ran out of time for today.

SeanCassiere:
> I could add the case from the issue to the tests as well just to be sure.

Yes please. Better to cover our bases with more testing to cover the case.

jvllmr:
All done ✅ 

- I have finished porting the tests to `@tanstack/solid-router`. When doing so I found an inconsistency with resolving the `from` option between `react-router` and `solid-router`. I updated `solid-router` to match the behavior of `react-router`
- I added a new testcase for the `Link` component that should reproduce the problem from the issue comment. However, when I reverted my patch the test still passed. Maybe there is more context to it we do not know or it is a false positive altogether?

SeanCassiere (approved):
> I have finished porting the tests to `@tanstack/solid-router`. When doing so I found an inconsistency with resolving the `from` option between `react-router` and `solid-router`. I updated `solid-router` to match the behavior of `react-router`.

👍🏼

> I added a new testcase for the `Link` component that should reproduce the problem from the issue comment. However, when I reverted my patch the test still passed. Maybe there is more context to it we do not know or it is a false positive altogether?

🤔... I'm starting to think that comment is a red herring.

Running the original sandbox (from the mentioned issue) against the pr.pkg.new packages from this PR, the runtime issue of navigation without `to` resulting in the omission of the configured basepath, does seem to be fixed.

---

Either ways, this was clearly an issue in the resolution logic not accounting for adding the basepath when `to` wasn't supplied.

If any further fixes are required, we'll fix forward.

SeanCassiere:
@jvllmr thank you for taking up this one and working with us on fixing this problem!

PR #4495, opened 2025-06-21
Refactor: Move core SSR functions into its own packages

nlynzaad:
This centralizes the SSR generic logic into its own package and enables pure Tanstack Router projects to not depend on Start imports for SSR functions.

This should also be non-breaking as all Start core packages either re-exports these or implements its own functionalities.

Going forward this should enable Start specific changes to the rendering functions to be isolated to Start without adversely affecting Router packages/projects that depends on the SSR functionality.

This PR currently adds 2 new packages: router-core-ssr and react-router-ssr. A similar package, to react-router-ssr, can be created for solid-router going forward if required.


nlynzaad:
@schiller-manuel as discussed in #4429. Herewith proposed refactor of the generic SSR functions into its own packages

packages/router-core-ssr/package.json
```diff
@@ -0,0 +1,76 @@
+{
+  "name": "@tanstack/router-core-ssr",
```
SeanCassiere: Change the package name from router-core-ssr to router-ssr-core.

nlynzaad: updated the name and added solid-router-ssr

nlynzaad:
Apologies for the pipeline failures. 

Hope the lock file was the last of the build issues. 

Tests and builds run fine locally.

schiller-manuel:
instead of adding new packages, can we move stuff into a separate `/ssr` entrypoint?

```
@tanstack/router-core/ssr
@tanstack/react-router/ssr
@tanstack/solid-router/ssr
```

nlynzaad:
Definitely, will make the change a bit later and send the update through.

nlynzaad:
I've merged router-ssr-core into router-core, react-router-ssr into react-router and solid-router-ssr into solid-router. each with a sperate entrypoint for ssr.

solid-router was proving to be difficult, probably my complete lack of having ever worked anything SolidJs before this, but I had to separate the ssr entry-point into a separate build and dist since it requires ssr to be generated. When generating ssr for all entry-points all current tests break for both solid-router and the e2e tests. Having seperate builds for the ssr entrypoints and the others resolves the problems.

Other than separating this into a different package and re-export that as a ssr entry-point in solid-router I can't see another way. If someone who has better understanding of SolidJS and the build configurations required for this can point me in the right direction I would much appreciate it

as it stands now this passes all tests and builds and the examples are running fine.

nlynzaad:
okay that seems to have been resolved.

all tests, builds and examples working as expected now.

packages/react-router/vite.config.ts
```diff
@@ -19,7 +19,7 @@ const config = defineConfig({
 export default mergeConfig(
   config,
   tanstackViteConfig({
-    entry: './src/index.tsx',
+    entry: ['./src/index.tsx', './src/ssr/client.ts', './src/ssr/server.ts'],
```
SeanCassiere, line 22: Please use `.tsx` extensions, like what's being used for the `solid-router` build, OR switch it all to use `.ts` instead.

SeanCassiere (changes requested):

packages/react-router/vite.config.ts
nlynzaad: updated solid-router to use .ts extension

SeanCassiere (approved):
👍🏼 from me

packages/react-router/src/ssr/RouterClient.tsx
```diff
@@ -0,0 +1,28 @@
...
+      hydrationPromise = hydrate(props.router)
+    } else {
+      hydrationPromise = Promise.resolve()
+    }
+  }
+  return (
+    <Await
+      promise={hydrationPromise}
+      children={() =>
+        props.children ?? <RouterProvider router={props.router} />
```
schiller-manuel: why do we need `props.children ??` ? is this used anywhere

schiller-manuel: removed it

nlynzaad: The original rationale for it was to allow for a bit more complex setup in the RouterClient i.e. wrapping the RouterProvider in others but this can be done directly in hydrateRoot, but I abandoned the idea as unnecessary, but seems I forgot to remove it completely. 

This can be removed

schiller-manuel:
thanks for that huge contribution!

PR #4501, opened 2025-06-23
docs(start): assign .ts extension to app.config file everywhere

ulrichstark:


ulrichstark:
Seems like app.config.ts isn't supported anymore but still mentioned everywhere in the docs. Is that correct?

```
app.config.ts:1:30 - error TS2307: Cannot find module '@tanstack/react-start/config' or its corresponding type declarations.

1 import { defineConfig } from "@tanstack/react-start/config";
                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in app.config.ts:1
```

CanRau:
yup [#2863](https://github.com/TanStack/router/discussions/2863#:~:text=app.config.ts%20to%20vite.config.ts)

SeanCassiere:
Our docs as a whole will be getting a revamp, as the current docs are somewhat a mash between the Vinxi and Devinxi versions.

Closing this for now, since `app.config.(j | t)s` is not supported anymore.

PR #4504, opened 2025-06-23
fix: move beforeLoad property before head property when creating route

ulrichstark:
As described here: https://tanstack.com/router/latest/docs/eslint/create-route-property-order


SeanCassiere (approved):

SeanCassiere:
Thanks!

PR #4513, opened 2025-06-25
Fix: Preserve cached data during concurrent navigation and refetch

### LINKED ISSUE #4245: Navigating while refetch is in progress results in cache not being used

jiji-hoon96:
Fixes #4245

## Problem
When a route refetch is in progress and the user navigates away then immediately returns, the cached stale data is lost, causing an empty page until the loader completes.

- Expected behavior: Show stale data immediately while refetch continues in background
- Actual behavior: Empty page until refetch completes

## Root Cause
In beforeLoad(), the cached matches filter was too aggressive - it removed all cached matches that had the same ID as pending matches, even when those cached matches contained valid stale data from an ongoing refetch.

## Solution
Modified the cache filtering logic to preserve successful cached matches that are currently fetching or have loader data available.


schiller-manuel:
thanks for the fix!

PR #4515, opened 2025-06-25
Fix: Improve TypeScript Type Safety for Search Params

### LINKED ISSUE #4225: TypeScript: `linkOptions` permits search params that the route does not accept

Router

### Describe the bug

When supplying extra search parameters on `linkOptions`, no typescript error is raised.
The same options raise a typescript error when supplied directly to `Link`.

https://github.com/jtannas/react-tooling-experimentation/blob/linkOptions-broken/src/routes/__root.tsx

1. Clone the linked repo
2. Go to the `src/routes/__root.tsx` file to confirm the same options are being passed to both links to `/about` and that both pass `notAllowed` as a search parameter.
3. Run `bun run tsc --build` to confirm that only 1 error is raised instead of the expected 2.

### Expected behavior

As a user, I expect `linkOptions` to raise a typescript error if I supply a search param the route does not accept.

### Screenshots or Videos

Image

Image

### Platform

- OS: Ubuntu 24.04
- TypeScript: 5.8.3
- TanStack Router: 1.120.3

### Additional context

_No response_

jiji-hoon96:
Fixes #4225

## Issue
Currently, `linkOptions.search` allows any search params at the type level, even those not permitted by the route.

## Solution
- Added `StrictSearchParamsForRoute` type to strictly validate search parameters against route definitions
- Replaced generic `any` type with strictly typed search params
- Updated related types `MakeOptionalSearchParams`, `MakeRequiredSearchParams` to use the new type checking
- Added appropriate type constraints to prevent type errors

## Benefits
- Compile-time validation for search params
- Better IDE autocompletion for valid search parameters
- Prevents usage of invalid search parameters for specific routes

### DIFF
--- a/packages/router-core/src/index.ts
+++ b/packages/router-core/src/index.ts
@@ -35,6 +35,7 @@ export type {
   MaskOptions,
   ToSubOptionsProps,
   RequiredToOptions,
+  StrictSearchParamsForRoute,
 } from './link'
 
 export type {
--- a/packages/router-core/src/link.ts
+++ b/packages/router-core/src/link.ts
@@ -529,7 +529,7 @@ export interface MakeOptionalSearchParams<
   in out TFrom extends string,
   in out TTo extends string | undefined,
 > {
-  search?: StrictSearchParamsForRoute<TRouter, TFrom, TTo>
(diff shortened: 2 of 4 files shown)


schiller-manuel:
this looks like it breaks the functional form for search params. probably best is to add failing type tests to show where the existing implementation fails. then we can look at how to solve it

PR #4519, opened 2025-06-25
fix: follow server fn redirects when submitting formdata

frenzzy:
### What’s the problem?
Previously, client-side redirects from `serverFn` handlers worked only when the request payload was a plain object. If the payload was a `FormData` instance, the redirect was ignored because the `'x-tsr-redirect': 'manual'` header wasn’t set.

### What’s changed?
This PR ensures that the `'x-tsr-redirect': 'manual'` header is always included, regardless of the request body type. As a result, redirects now work consistently, even when sending `FormData`.

### Example
```tsx
const fn = createServerFn({ method: 'POST' }).handler(() => redirect({ to: '/' }));
const action = useServerFn(fn);

// Previously: only this worked
<button onClick={() => action({ data: 'text' })}>Redirect works</button>

// Now: this works too 🎉
<button onClick={() => action({ data: new FormData() })}>Redirect also works</button>
```

### DIFF
--- a/e2e/react-start/server-functions/src/routes/index.tsx
+++ b/e2e/react-start/server-functions/src/routes/index.tsx
@@ -77,7 +77,7 @@ function Home() {
             server function redirects when FormData is submitted (via JS)
           </Link>
         </li>
-         <li>
+        <li>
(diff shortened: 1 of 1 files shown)


schiller-manuel:
thanks for the fix!

PR #4527, opened 2025-06-26
docs(guide) - update router ssr guides

nlynzaad:
This PR updates the React SSR guide to reflect the latest changes after #4495. It also adds a SolidJS SSR guide.


schiller-manuel:
we must add to the docs that these APIs are still experimental. until we hit stable in start, those APIs might need to change as well

nlynzaad:
Thanks, I've added a note at the top

PR #4528, opened 2025-06-26
feat: add optional shellComponent to root route

schiller-manuel:


tannerlinsley (approved):

Nandan1996:
Hi @schiller-manuel i was trying spa mode, after build the _shell.html in output doesn’t include html & body element rendered by component option given to createRootRoute.

is it not the expected behaviour or shellComponent is expected to be used for this?
The spa mode doc doesn’t have any reference to shellComponent instead component option is used throughout the doc.

schiller-manuel:
@Nandan1996 please create a new issue including a complete minimal example

PR #4531, opened 2025-06-27
fix(router-core): use `signal.reason` when aborting the match

SeanCassiere:


SeanCassiere:
This is not the correct way to go about fixing this.

PR #4552, opened 2025-06-30
feat(react-router): useSearchState

Sheraff:
This is a 1st draft proposal for `useSearchState`: a hook designed to be used like a react `useState`, but manipulates route search params instead.

Example usage:
```tsx
export const Route = createFileRoute('/foo')({
  validateSearch: z.object({
    bar: z.number()
  }),
  component: MyComponent,
})

const MyComponent = () => {
  const [state, setState] = useSearchState({
    from: '/foo',
    key: 'bar',
  })
  state
  // ^? number
  setState(2)
  setState(prev => prev + 1)
```

Comparison w/ and w/o this hook:
```tsx
const Without = () => {
  const navigate = useNavigate()
  const value = useSearch({
    from: '/foo',
    select: s => s.value
  })
  const onClick = () => navigate({
    to: '.',
    from: '/foo',
    search: prev => ({...prev, value: 2 })
  })
}
const With = () => {
  const [value, setValue] = useSearchState({
    from: '/foo',
    key: 'value',
  })
  const onClick = () => setValue(2)
}
```

Points to address in this RFC:
1. is it useful to have an `initial` argument? (we have it in our codebase at work, and never use it, i'd vote for removing it)
2. are there options it would make sense for the `setState` to support? 
     ```ts
    const [state, setState] = useSearchState({...})
    setState(value, {
      replace: false // defaults to true
      ignoreBlocker, 
      reloadDocument,
      viewTransition,
    })
    ```
3. types currently don't support route masking (I've never used it myself). Does it make sense to support it for this hook that never touches the path?
4. this hook calls `router.navigate`, should it call `buildAndCommitLocation` directly instead?
5. this hook batches calls to `setState` in the same tick (to behave like react's  `setState`) using `queueMicrotask` and accumulating values using symbol keys on the router. This might seem pretty alien, are we OK with this implementation?
6. with this implementation, calling `setState` *and* `navigate` in the same tick is *undefined behavior*. Should we be more strict on this? (To me it doesn't really make sense to call both in the same tick)

TODO:
- [ ] Add a `useSearchState.test-d.tsx` type test file
- [ ] Docs
- [ ] Expose hook on route object `Route.useSearchState({ key: 'foo' })`

Just in case, here's the support for `queueMicrotask`
developer mozilla org_en-US_docs_Web_API_Window_queueMicrotask


packages/react-router/src/useSearchState.tsx
```diff
@@ -0,0 +1,222 @@
...
+}: Params<TRouter, TFrom, TStrict, TKey>): readonly [
+  state: TValue,
+  setState: SetSearchState<TValue>,
+] {
+  // state
+  const state = useSearch(
+    React.useMemo(
+      () => ({
+        from,
+        select: getter.bind(null, key),
```
TkDodo, line 101: would be nice if users could pass in `select` in addition to `key` to further refine what to subscribe to

packages/react-router/src/useSearchState.tsx
```diff
@@ -0,0 +1,222 @@
...
+const routerStore = new WeakMap<
+  AnyRouter,
+  {
+    search: object | null
+    opts: SetSearchStateOptions | null
+    scheduled: boolean | null
+  }
+>()
+
+function setter<T>(
```
TkDodo, line 139: why not just call `navigate` in here with the newly computed state? all the optimizations like bail-out with `Object.is` or `queueMicrotask` should probably happen inside `navigate`, too

maelp:
Really cool! Will it prevent infinite loops if we now have both

User clicks -> Update search params -> update values

and

User clicks -> Update values -> Update search params 

routes "working" (I assume that's the main contribution of that PR, rather than the "classical Tanstack router way" which forces the User clicks -> Update search params -> update values" way?)

Sheraff:
> Really cool! Will it prevent infinite loops if we now have both
> 
> User clicks -> Update search params -> update values
> 
> and
> 
> User clicks -> Update values -> Update search params
> 
> routes "working" (I assume that's the main contribution of that PR, rather than the "classical Tanstack router way" which forces the User clicks -> Update search params -> update values" way?)

Sorry I didn't understand your comment.

maelp:
What I'm saying is that we have both "local (react) state" and "state handled by the URL searchParam",

and I think that before, Tanstack Router would "force you" to change state to do:

1. User does interaction
2. We update the corresponding stack in search params
3. In turn, this re-renders the page, and loads the corresponding state in the local react state

This prevents any "infinite loop" by having a clear, one-direction only, path: user -> search-params -> react state

I guess this feature would try to synchronize the local state with the search param, so that (as you explain above), we would get the more natural flow:

1. User does interaction
2. We update the local react state
3. useSearchState ensures that this is synchronized with the URL search params

so this path is: User -> React state -> Search params

Obviously, since we want a user to be able to just copy-and-paste a URL with some defined search-params in the browser, and the corresponding page to be loaded with the corresponding state, it means that even in your implementation, there is an "update path" existing somewhere which does: searchParams updated -> react state updated

So, in theory at least, there exists a way to shoot oneself in the foot by having some "loop" which does:

searchParams updated -> local state updated -> searchParams updated -> local state updated -> ...

because there is no longer a "one-direction update path" but a bi-directional one now

so I was wondering if there was anything that prevents / reduce the likelihood of such instances?

matthiasfeist:
I'm super excited about this PR. I didn't know about it until it was linked from a[ discussion that is discussion the same thing](https://github.com/TanStack/router/discussions/4140#discussioncomment-14331215).

khashvin:
Any updates on this PR? Would be great to have a nuqs-like feature in TSR. @Sheraff

LadyBluenotes:
@Sheraff Are you planning to keep working on this? I think it’s close, but it would need a rebase, the queued-state cleanup and route-mask behavior addressed, and the missing type and setter-option coverage.

#4973 covers the broader atomic update and masking semantics

PR #4555, opened 2025-07-01
fix(server-functions): prevent HMR race condition by regenerating virtual manifest module from `directiveFnsById` state

### LINKED ISSUE #4486: Server function module export not resolved for serverFn (probably some race condition)

Start

### Describe the bug

I experience a failure deep in tanstack code after starting dev mode server and running a specific page for the first time. 

I did manage to narrow down the problem to this place: [https://github.com/TanStack/router/blob/d672b229aefc0eed4a463485e29ef880f803622a/packages/start-server-core/src/server-functions-handler.ts#L66
](url)

but more importantly adding just before a 1ms delay 
```
  const delay = (duration) =>
    new Promise(resolve => setTimeout(resolve, duration));
  
  await delay(1)

```

solves the problem which suggests some race condition or similar type of issue. 

Unfortunatly, I did not manage to create a simple reproduction as it is time dependent and even the same source code placement (see observations below)

The error stack tarce:
```
Error: Server function module export not resolved for serverFn ID: src_server_api_routers_document-serverfn_ts--getYDocDiffs_createServerFn_handler
    at jsonPayloadAsString (/Users/pbuszka/projects/verosetta/webapp/node_modules/@tanstack/start-server-core/src/server-functions-handler.ts:79:5)
    at async eval (/Users/pbuszka/projects/verosetta/webapp/node_modules/@tanstack/start-server-core/src/createStartHandler.ts:143:20)
    at async startRequestResolver (/Users/pbuszka/projects/verosetta/webapp/node_modules/@tanstack/start-server-core/src/createStartHandler.ts:127:24)
    at async file:///Users/pbuszka/projects/verosetta/webapp/node_modules/@tanstack/start-plugin-core/dist/esm/dev-server-plugin/plugin.js:45:30

```

did not succeeded as it seems to be issue depending on time

The problem is not reproducible in a separate repo but what I found out in my own dev environment

1. start dev server: npm run dev
2. go to the offending page, there are 4 calls to a speciifc serverFn on this page 
3. one of the 4 calls (in most cases the 1st) to serverFn fails with error as in descirption, subsequent calls to the same serverFn work as expected
4. browser refresh of the page WITHOUT restart of the dev server doesn't trigger the problem again
5. restarting dev server triggers

Another path that triggers the problem:
1. start dev server
2. start with a different page - it works fine. This page uses a different serverFn from the same source file (so the problem is not linked to first usage of a file with serverfns) 
3. navigate to the offending page
4. the same problem is triggered on the same server fn as above

### Expected behavior

serverfn call not failing

### Screenshots or Videos

_No response_

### Platform

- OS: [ macOS 15.5, M1 chip]
- Browser: [Chrome]
- Version: [1.121.28]

### Additional context

The problem never appeared before 1.121. I found it when exploring migrating to 1.121

Observations:
- reducing the body of this server function to simple return did not solve the problem
- adding await with 1ms delay in handleServerAction "solves" the problem (see above)
- moving the problematic server function to **a separate source file** did "solve" the issue.
- moving the serverfn source code within the same source file did not help
- other functions from the same file execute fine on the same page
- the offending page makes 12 serverfn calls from 3 different source files, including 4 calls to the problematic fn 
- it is linked with the first call to the specific serverfn
- it is NOT linked to the first call of any serverfn in the same source file

My gustimate:
- there is some dynamic import process that in some circumstances is not finished before running the code in  handleServerAction
- it probably depends somehow on the size of the source file where server fn is located
- maybe it's somehow related with multiple calls when severfn is first used
- it is not related to the actual body of the server function
- it happens only once per dev server lifetime

Crsk:
This error occurred when triggering server functions during development

The idea basically is to avoid manual module invalidation (which is async and prone to race conditions), relying on Vite's built-in mechanisms

Fixes #4486

didn’t add e2e test because the race condition does not happen in the production build environment

to see if working, i setup a cookies loaded theme, the expected result is to hit the theme toggle, see the style update and no `Error: Server function module export not resolved for serverFn ID: xxx` error. 

---

Here's the theme setup i used

```ts
// theme-provider.tsx

import { useRouter } from '@tanstack/react-router'
import { createContext, PropsWithChildren, useContext, useState } from 'react'
import { setThemeServerFn } from './theme'

export type Theme = 'light' | 'dark' | 'system'

interface ThemeContextValue {
  currentTheme: Theme
  setTheme: (theme: Theme) => Promise<void>
}

const ThemeContext = createContext<ThemeContextValue | null>(null)

export const ThemeProvider = ({ children, initialTheme }: PropsWithChildren<{ initialTheme: Theme }>) => {
  const [currentTheme, setCurrentTheme] = useState<Theme>(initialTheme)
  const router = useRouter()

  const setTheme = async (theme: Theme) => {
    setThemeServerFn({ data: theme })
    setCurrentTheme(theme)
    router.invalidate()
  }

  return <ThemeContext.Provider value={{ currentTheme, setTheme }}>{children}</ThemeContext.Provider>
}

export const useTheme = () => {
  const context = useContext(ThemeContext)
  if (!context) throw new Error('useTheme must be used within a ThemeProvider')

  return context
}
```
```ts
// theme.ts

import { type Theme } from './theme-provider'
import { createServerFn } from '@tanstack/react-start'
import { getCookie, setCookie } from '@tanstack/react-start/server'

export const THEME_COOKIE_KEY = 'ui-theme'
export const THEME_VALUES = ['light', 'dark', 'system'] as const

export const getThemeServerFn = createServerFn().handler(async () => {
  return (getCookie(THEME_COOKIE_KEY) || 'light') as Theme
})

export const setThemeServerFn = createServerFn({ method: 'POST' })
  .validator((data: Theme) => {
    if (!THEME_VALUES.includes(data)) throw new Error('Invalid theme provided')

    return data
  })
  .handler(async ({ data }) => {
    setCookie(THEME_COOKIE_KEY, data)
  })
```

```ts
// theme-toggle.tsx

import { useTheme } from '~/theme-provider'


export const ThemeToggle = () => {
  const { currentTheme, setTheme } = useTheme()
  const toggle = () => setTheme(currentTheme === 'dark' ? 'light' : 'dark')

  return (
    <div className="p-2">
      <button onClick={toggle} className="bg-emerald-500 text-white px-2 py-1 rounded uppercase font-black text-sm">
        {currentTheme === 'light' ? 'Dark' : 'Light'}
      </button>
    </div>
  )
}
```
```ts
// __root.tsx

/// <reference types="vite/client" />
import {
  HeadContent,
  Link,
  Outlet,
  Scripts,
  createRootRoute,
} from '@tanstack/react-router'
import { TanStackRouterDevtools } from '@tanstack/react-router-devtools'
import { DefaultCatchBoundary } from '~/components/DefaultCatchBoundary'
import { NotFound } from '~/components/NotFound'
import appCss from '~/styles/app.css?url'
import { getThemeServerFn } from '~/theme'
import { seo } from '~/utils/seo'
import { ThemeProvider, useTheme } from '~/theme-provider'
import { ThemeToggle } from '~/theme-toggle'
import { useMemo } from 'react'

  head: () => ({
    meta: [
      {
        charSet: 'utf-8',
      },
      {
        name: 'viewport',
        content: 'width=device-width, initial-scale=1',
      },
      ...seo({
        title:
          'TanStack Start | Type-Safe, Client-First, Full-Stack React Framework',
        description: `TanStack Start is a type-safe, client-first, full-stack React framework. `,
      }),
    ],
    links: [
      { rel: 'stylesheet', href: appCss },
      {
        rel: 'apple-touch-icon',
        sizes: '180x180',
        href: '/apple-touch-icon.png',
      },
      {
        rel: 'icon',
        type: 'image/png',
        sizes: '32x32',
        href: '/favicon-32x32.png',
      },
      {
        rel: 'icon',
        type: 'image/png',
        sizes: '16x16',
        href: '/favicon-16x16.png',
      },
      { rel: 'manifest', href: '/site.webmanifest', color: '#fffff' },
      { rel: 'icon', href: '/favicon.ico' },
    ],
    scripts: [
      {
        src: '/customScript.js',
        type: 'text/javascript',
      },
    ],
  }),
  errorComponent: DefaultCatchBoundary,
  notFoundComponent: () => <NotFound />,
  shellComponent: RootComponent,
  loader: () => getThemeServerFn()
})

function RootComponent() {
  const initialTheme = Route.useLoaderData()

  return (
    <ThemeProvider initialTheme={initialTheme}>
      <RootDocument>
        <Outlet />
      </RootDocument>
    </ThemeProvider>
  )
}

function RootDocument({ children }: { children: React.ReactNode }) {
  const { currentTheme } = useTheme()
  const colorMap = useMemo(() => ({
    bg: currentTheme === 'dark' ? 'black' : 'white',
    color: currentTheme === 'dark' ? 'white' : 'black',
    hr: currentTheme === 'dark' ? '#161616' : '#ededed',
  }), [currentTheme])

  return (
    <html lang="en" style={{ backgroundColor: colorMap.bg }}>
      <head>
        <HeadContent />
      </head>
      <body style={{ backgroundColor: colorMap.bg, color: colorMap.color }}>
        <div className="p-2 flex gap-2 text-lg">
          <Link
            to="/"
            activeProps={{
              className: 'font-bold',
            }}
            activeOptions={{ exact: true }}
          >
            Home
          </Link>
        </div>
        <hr style={{ borderColor: colorMap.hr }} />
        <ThemeToggle />
        {children}
        <TanStackRouterDevtools position="bottom-right" />
        <Scripts />
      </body>
    </html>
  )
}
```

### DIFF
--- a/packages/server-functions-plugin/src/index.ts
+++ b/packages/server-functions-plugin/src/index.ts
@@ -43,21 +43,18 @@ export function createTanStackServerFnPlugin(opts: ServerFnPluginOpts): {
   const directiveFnsById: Record<string, DirectiveFn> = {}
   let viteDevServer: ViteDevServer | undefined
 
-  const onDirectiveFnsById = buildOnDirectiveFnsByIdCallback({
-    directiveFnsById,
-    manifestVirtualImportId: opts.manifestVirtualImportId,
-    invalidateModule: (id) => {
-      if (viteDevServer) {
-        const mod = viteDevServer.moduleGraph.getModuleById(id)
-        if (mod) {
-          if (debug) {
-            console.info(`invalidating module ${JSON.stringify(mod.id)}`)
-          }
-          viteDevServer.moduleGraph.invalidateModule(mod)
+  const onDirectiveFnsById = (d: Record<string, DirectiveFn>) => {
+    Object.assign(directiveFnsById, d)
+    if (viteDevServer) {
+      const mod = viteDevServer.moduleGraph.getModuleById(resolveViteId(opts.manifestVirtualImportId))
+      if (mod) {
+        if (debug) {
+          console.info(`invalidating module ${JSON.stringify(mod.id)}`)
         }
+        viteDevServer.moduleGraph.invalidateModule(mod)
       }
-    },
-  })
+    }
+  }
 
   const directive = 'use server'
   const directiveLabel = 'Server Function'
@@ -110,14 +107,14 @@ export function createTanStackServerFnPlugin(opts: ServerFnPluginOpts): {
 
           const manifestWithImports = `
           export default {${Object.entries(directiveFnsById)
-            .map(
-              ([id, fn]: any) =>
-                `'${id}': {
+              .map(
+                ([id, fn]: any) =>
+                  `'${id}': {
                   functionName: '${fn.functionName}',
                   importer: () => import(${JSON.stringify(fn.extractedFilename)})
                 }`,
-            )
-            .join(',')}}`
+              )
+              .join(',')}}`
 
           return manifestWithImports
         },
@@ -175,23 +172,9 @@ export function TanStackServerFnPluginEnv(
   const directiveFnsById: Record<string, DirectiveFn> = {}
   let serverDevEnv: DevEnvironment | undefined
 
-  const onDirectiveFnsById = buildOnDirectiveFnsByIdCallback({
-    directiveFnsById,
-    manifestVirtualImportId: opts.manifestVirtualImportId,
-    invalidateModule: (id) => {
-      if (serverDevEnv) {
-        const mod = serverDevEnv.moduleGraph.getModuleById(id)
-        if (mod) {
-          if (debug) {
-            console.info(
-              `invalidating module ${JSON.stringify(mod.id)} in server environment`,
-            )
-          }
-          serverDevEnv.moduleGraph.invalidateModule(mod)
-        }
-      }
-    },
-  })
+  const onDirectiveFnsById = (d: Record<string, DirectiveFn>) => {
+    Object.assign(directiveFnsById, d)
+  }
 
   const directive = 'use server'
   const directiveLabel = 'Server Function'
@@ -246,14 +229,14 @@ export function TanStackServerFnPluginEnv(
           }
           const manifestWithImports = `
           export default {${Object.entries(directiveFnsById)
-            .map(
-              ([id, fn]: any) =>
-                `'${id}': {
+              .map(
+                ([id, fn]: any) =>
+                  `'${id}': {
                   functionName: '${fn.functionName}',
                   importer: () => import(${JSON.stringify(fn.extractedFilename)})
                 }`,
-            )
-            .join(',')}}`
+              )
+              .join(',')}}`
 
           return manifestWithImports
         },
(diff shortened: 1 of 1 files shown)


schiller-manuel:
Thanks for taking a stab at this, but i don't understand how this would solve the problem (maybe i am not seeing it?). this still invalidates the virtual module (but only in the plugin we don't use in start), but does not do any invalidation the env-aware plugin we use in start. 

server functions are discovered incrementally in dev, so we need to at least invalidate the virtual module.
i had this (only invalidating the virtual module) in a previous version but this also had the same issue as described in #4486

Crsk:
@schiller-manuel 

Got it, i have no idea what i’m doing tbh, because reverting #4551 solves the issue for me but that does not explain why people is getting this same error before #4551 was introduced.

i’ll just share what i have learned about the issue and pray for a solution to come:
- from commit f0146e1f it seems like the simplifying (ironic) logic moving from static in-disk to incremental in-memory state is causing the race condition.
- i still believe the `buildOnDirectiveFnsByIdCallback` micromanager should be removed, it’s trying to do Vite’s job by manually invalidating the modules, that’s prone to race condition since Vite's reprocessing is async (after invalidation it takes some time to compile and load), meaning the manifest could reload before the function was ready, so should be a better deal to tell Vite hey bro this manifest is stale `viteDevServer.moduleGraph.invalidateModule(...)`, figure out a new one from the plugin’s loader callback and the `directiveFnsById` state, but before do your internal thing by analyzing the dependency chain solving them properly so that we have no manifest pointing to stale or non-existing references.
- ok that one was long, in essence probably not the best idea ever to tell Vite to do both things asynchronously: reprocess the file/module and update its index/manifest.
- invalidation ≠ readiness. Vite's `moduleGraph.invalidateModule()` is synchronous, but the actual module reprocessing is asynchronous.
- if i git check out before #4551 then i cannot reproduce the error. Probably we’re seeing two sources of the same issue, one of them solved by reverting 4551 and no idea about the other one reported in #4486.

PR #4559, opened 2025-07-01
feat: Add style to HeadContent

felipestanzani:
Currently Route HeadContent don't have support to inline style tags.

After a contribution proposition made by @schiller-manuel in the Discord Group I have decided to implement this.

I also have used this in the basic example for react-start and solid-start and added style and style w/loader test cases to both modules.

### DIFF
--- a/packages/react-router/src/HeadContent.tsx
+++ b/packages/react-router/src/HeadContent.tsx
@@ -120,6 +120,23 @@ export const useTags = () => {
     structuralSharing: true as any,
   })
 
+  const styles = useRouterState({
+    select: (state) =>
+      (
+        state.matches
+          .map((match) => match.styles!)
+          .flat(1)
+          .filter(Boolean) as Array<RouterManagedTag>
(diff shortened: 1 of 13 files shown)


packages/react-router/src/HeadContent.tsx
```diff
@@ -120,6 +120,23 @@ export const useTags = () => {
...
+      (
+        state.matches
+          .map((match) => match.styles!)
+          .flat(1)
+          .filter(Boolean) as Array<RouterManagedTag>
+      ).map(({ children, ...style }) => ({
+        tag: 'style',
+        attrs: {
+          ...style,
+        },
```
Sheraff: There is no need to make a shallow copy of `style` here, because it's already a *rest* object
```suggestion
      ).map(({ children, ...attrs }) => ({
        tag: 'style',
        attrs,
```

schiller-manuel:
very nice, thanks a lot!

PR #4561, opened 2025-07-02
fix: update import for server function in React basics guide

periakteon:
Replaced `useServerFunction` with `useServerFn` in the React basics documentation to reflect the latest API changes.


schiller-manuel:
thanks for spotting this, we never had `useServerFunction` IIRC, it was always `useServerFn`

PR #4564, opened 2025-07-02
fix(router-plugin): handle explicit undefined component in code splitter

wtlin1228:
Handle the edge case where `options.component` exists but the value is `undefined`.

```js
const Route = createFileRoute('/')({
  component: undefined,
})
```

Before: `SplitComponent` is not defined -> runtime error 🚨
After: `SplitComponent` is defined -> it works ✅


schiller-manuel:
nice! can you please rebase once we merged https://github.com/TanStack/router/pull/4565 since the snapshots will conflict?

wtlin1228:
@schiller-manuel done!

wtlin1228:
Oh, it hasn't been merged. I can rebase again later.

packages/router-plugin/src/core/code-splitter/compilers.ts
```diff
@@ -620,6 +620,17 @@ export function compileCodeSplitVirtualRoute(
               console.info('Unexpected splitNode type:', splitNode)
               throw new Error(`Unexpected splitNode type ☝️: ${splitNode.type}`)
             }
+          } else {
```
SeanCassiere: This escape seems bit too far down the tree and probably shouldn't be happening at an indentation after `510 - if (splitNode) {`. 

Use the `babelHandleVirtual` function to track the instance of `undefined` and then you can handle it at the same indentation level of `511 - if (t.isFunctionDeclaration(splitNode)) {` instead of at `510 - if (splitNode) {`.

SeanCassiere (changes requested):
Please make the changes listed. It probably shouldn't affect any of your tests or snapshots.

packages/router-plugin/src/core/code-splitter/compilers.ts
wtlin1228: I agree that the current handling is too far down the tree.

I actually considered adding the `undefined` check upfront, similar to what you suggested. The reason I didn’t go with that initially was to make the change smaller.

So, what do you think I change it from

```js
if (splitNode) {
  if (t.isFunctionDeclaration(splitNode)) {
    // ...
  } else if (t.isFunctionExpression(splitNode) || t.isArrowFunctionExpression(splitNode)) {
    // ...
  } else {
    console.info('Unexpected splitNode type:', splitNode)
    throw new Error(`Unexpected splitNode type ☝️: ${splitNode.type}`)
  }
} else {
  // splitNode === undefined
}
```

to

```js
if (splitNode === undefined) {
  // handle undefined explicitly
} else if (t.isFunctionDeclaration(splitNode)) {
  // ...
} else if (t.isFunctionExpression(splitNode) || t.isArrowFunctionExpression(splitNode)) {
  // ...
} else {
  console.info('Unexpected splitNode type:', splitNode)
  throw new Error(`Unexpected splitNode type ☝️: ${splitNode.type}`)
}
```

This approach keeps all type checks at the same indentation level and makes the logic easier to follow

SeanCassiere: I'd still rather have be handled above. 

This current approach, relies on the happen-stance of how the current node tracking data-structure works. 

For this AST-related transformations, I'd rather have it be explicitly handled.

Plus, it gives us the flexibility to make changes to the data structure knowing that the node handling wouldn't be coupled to it.

wtlin1228: @SeanCassiere I have one new commit to handle it explicitly, please help review, thanks!

SeanCassiere (approved):

SeanCassiere:
@wtlin1228 thank you for working with us and contributing this!

PR #4565, opened 2025-07-02
feat: ssr setting per route

schiller-manuel:


packages/router-core/src/recoverableError.ts
```diff
@@ -0,0 +1,13 @@
+export const RECOVERABLE_ERROR = 'TSR_RecoverableError'
+
+export function isRecoverableError(error: unknown): boolean {
+  if (error instanceof Error) {
+    return error.message.includes(RECOVERABLE_ERROR)
```
SeanCassiere: Maybe make this a more specific? 
```
startsWith(`${RECOVERABLE_ERROR}: `)
```

packages/router-core/src/ssr/ssr-client.ts
```diff
@@ -157,10 +177,9 @@ export async function hydrate(router: AnyRouter): Promise<any> {
     ;(match as unknown as SsrMatch).extracted?.forEach((ex) => {
       deepMutableSetByPath(match, ['loaderData', ...ex.path], ex.value)
     })
-
-    return match
   })
 
+  console.log(`Hydrated ${matches.length} matches from SSR.`, matches)
```
SeanCassiere: Does this get piped through to the user? Maybe hide it behind `DEBUG`?

packages/router-plugin/src/core/code-splitter/compilers.ts
```diff
@@ -151,14 +152,30 @@ export function compileCodeSplitReferenceRoute(
               }
 
               if (t.isObjectExpression(routeOptions)) {
+                if (opts.deleteNodes && opts.deleteNodes.length > 0) {
+                  routeOptions.properties = routeOptions.properties.filter(
+                    (prop) => {
+                      if (t.isObjectProperty(prop)) {
+                        if (t.isIdentifier(prop.key)) {
+                          if (
+                            opts.deleteNodes?.includes(prop.key.name as any)
```
SeanCassiere: Might we worth checking this out against a large amount of routes to see if there's much of perf impact of using an Array vs a Set.

Nothing blocking though since its currently just expected to `Array<'ssr'> | undefined`.

packages/start-plugin-core/src/start-router-plugin/plugin.ts
```diff
@@ -14,20 +14,33 @@ import {
...
 import type { PluginOption } from 'vite'
-import type { Config } from '@tanstack/router-generator'
+import type { Config } from '@tanstack/router-plugin'
 
 export function tanStackStartRouter(config: Config): Array<PluginOption> {
   return [
     tanstackRouterGenerator({
       ...config,
       plugins: [serverRoutesPlugin(), routesManifestPlugin()],
+      environmentName: VITE_ENVIRONMENT_NAMES.client,
```
SeanCassiere: Rename this to to something like `viteEnvironmentName` or something like `viteApplyEnvironment`?

Purely since it wouldn't have any effect on the other bundlers.

packages/router-core/src/router.ts
```diff
@@ -2163,12 +2217,72 @@ export class RouterCore<
...
+              const parentMatch = parentMatchId
+                ? this.getMatch(parentMatchId)!
+                : undefined
 
               const route = this.looseRoutesById[routeId]!
 
               const pendingMs =
                 route.options.pendingMs ?? this.options.defaultPendingMs
 
+              if (this.isServer) {
```
SeanCassiere, line 2238: Could use some comments here.

birkskyum (approved):
Works as I'd expect - I'd prefer to have tests for solid too, but I can see it's tricky to add because of the tsconfig.json, which I don't have a good solution for.

SeanCassiere (approved):

packages/router-plugin/tests/delete-nodes/snapshots/react/1-delete-nodes-undefined/development/createRootRoute-ssr-function.tsx
```diff
@@ -0,0 +1,20 @@
...
+    if (crypto.randomInt(0, 2) === 0) {
+      return 'data-only';
+    }
+    return false;
+  },
+  component: () => {
+    return <div className="p-2">hello world</div>;
+  }
+});
+if (import.meta.hot) {
```
SeanCassiere, line 14: Is it necessary to have both `development` and `production` snapshots related testing for the "delete-nodes" functionality?

Currently, the difference between 'development' and 'production', is the inclusion of the HMR setup using `import.meta.hot.accept`.

packages/router-plugin/tests/delete-nodes.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    name: '1-delete-nodes-undefined',
+  },
+  { deleteNodes: [], name: '2-delete-nodes-empty' },
+  {
+    deleteNodes: ['ssr'],
+    name: '3-delete-nodes-ssr',
+  },
+]
+
+const frameworks: Array<Config['target']> = ['react', 'solid']
```
SeanCassiere: This should probably now become a shared variable between `code-splitter.test.ts` and `delete-nodes.test.ts` so that any additions in the future (of other frameworks) don't cause drift.

packages/router-plugin/src/core/config.ts
```diff
@@ -55,12 +55,28 @@ export type CodeSplittingOptions = {
...
+  /**
+   * The nodes that shall be deleted from the route.
+   * @default undefined
+   */
+  deleteNodes?: Array<DeletableNodes>
+
+  /**
+   * @default true
+   */
+  addHmr?: boolean
```
SeanCassiere, line 68: What use-case would cause the user to set this to `false`?

packages/router-plugin/src/core/code-splitter/compilers.ts
```diff
@@ -138,27 +143,51 @@ export function compileCodeSplitReferenceRoute(
 
             if (
               !(
-                path.node.callee.name === 'createRoute' ||
-                path.node.callee.name === 'createFileRoute'
+                path.node.callee.name === 'createFileRoute' ||
```
SeanCassiere:
```suggestion
                ['createFileRoute', 'createRootRoute', 'createRootRouteWithContext'].includes(path.node.callee.name)
```

packages/router-plugin/src/core/code-splitter/compilers.ts
```diff
@@ -138,27 +143,51 @@ export function compileCodeSplitReferenceRoute(
...
+                          if (opts.deleteNodes?.has(prop.key.name as any)) {
+                            return false
+                          }
+                        }
+                      }
+                      return true
+                    },
+                  )
+                }
+                if (createRouteFn !== 'createFileRoute') {
```
SeanCassiere: I think its worth adding a comment here for future adventurers stating that the code-splitting we're exiting early here since the `rootRoute` is never code-split.

Perhaps the exit condition should just check for `createRootRoute` or `createRootRouteWithContext`.

e2e/react-start/selective-ssr/src/routes/posts.$postId.tsx
```diff
@@ -0,0 +1,90 @@
+import { createFileRoute } from '@tanstack/react-router'
+import z from 'zod'
+import { ssrSchema } from '~/search'
+
+export const Route = createFileRoute('/posts/$postId')({
+  validateSearch: z.object({ postId: ssrSchema }),
+  ssr: ({ search }) => {
+    if (typeof window !== 'undefined') {
```
SeanCassiere, line 8: Not sure how important this is, but we can probably share the content of the `ssr`, `beforeLoad`, and `component` between the routes yes?

SeanCassiere (commented):
.

packages/router-plugin/src/core/config.ts
schiller-manuel: see start-plugin-core. on the server there is no hmr added for example

PR #4571, opened 2025-07-03
Update vite.config.ts to use @vitejs/plugin-react-swc

spa5k:


spa5k:
We can later on move to https://www.npmjs.com/package/@vitejs/plugin-react-oxc when vite-rolldown becomes mainstream, but until then, SWC brings more speed with no downgrade.

schiller-manuel:
start will soon not wrap the react plugin anymore, so a user can choose which variant he wants to use then

GuskiS:
@schiller-manuel when is that happening? Interested in trying oxc + rolldown.

PR #4573, opened 2025-07-03
fix: Make buildLocation aware of non-changing routes

### LINKED ISSUE #4526: Regression: navigate({ to: "." }) redirects to base path after update to v1.121.34

Router

### Describe the bug

After updating ```@tanstack/react-router``` from version `1.120.13` to `1.121.34`, I encountered a regression when navigating with `navigate({ to: "." })` from ``useNavigate`

Previously, this would keep the user on the current route and only update the search params. Now it unexpectedly redirects to the root path (/), even though the current route is a child like /post.

Here is a test i prepare to reproduce this scenario 
[Fork Link](https://github.com/TanStack/router/compare/main...amargielewski:router:use-navigate-broken-test)

https://codesandbox.io/p/sandbox/tanstack-router-bug-example-j4qlpj

I prepared a minimal demo:
[codesandbox]( https://codesandbox.io/p/sandbox/tanstack-router-bug-example-j4qlpj)

To reproduce:

Go to /map by clicking the "Map" link.

Click the "Open Panel" button.

Instead of staying on /map, the router navigates back to /.

This breaks functionality that previously relied on navigate({ to: ".", search: {...} }) to preserve the current route while updating search state.

### Expected behavior

`navigate({ to: "." })` should stay on the current route (/post in this case) and update the search params without redirecting to the base path (/).

### Screenshots or Videos

_No response_

### Platform

- Bundler: [e.g. vite]

### Additional context

_No response_

nlynzaad:
This is an attempt to resolve #4526 by making buildLocation aware of the fact that even though there is a "to" path the route is not always changing. This is specifically the case with to="." navigations.

This also adds a test largely based on the test provided by [amargielewski](https://github.com/amargielewski)

fixes #4526


schiller-manuel:
looks good! can you please add a test that changes path params like this `navigate({to: '.', params: {foo : 'bar'}})`?

nlynzaad:
great. I've added the test. will send through momentarily

schiller-manuel:
an index route cannot have child routes  really. although this might somehow work it's better to use a different structure. 
in that case just move the component from the posts index route to posts route

schiller-manuel:
can you please use getByTestId instead of role /text based lookup? we write new tests based on this (and hopefully port the old ones eventually)

nlynzaad:
done and done

schiller-manuel:
thanks a lot!

checkerschaf:
@nlynzaad Thanks for looking into this! This code is still failing for me with the latest version `1.124.2`.

Example:
```ts
const shareRecipeId = useSearch({
  from: "/$lang/_app",
  select: (search) => search._shareRecipeId,
});

navigate({
  to: ".",
  search: (prev) => ({
    ...prev,
    _shareRecipeId: undefined,
  }),
  resetScroll: false,
});
```

Instead of remaining on the same page, it navigates to `/$lang`. The route file is located at `routes/$lang/route.ts`, which may be helpful.

Do you have any insights into why this is causing problems? If not, I can look into the test in the next few days. Thanks!

nlynzaad:
@checkerschaf from the looks of it, it should be covered by the update. However it differs slightly to the previous in the it rerouted to "/" in your case it's back to "/$lang" and might have small detail that was missed.

Please can you open a new issue with a reproducible example or test.

checkerschaf:
@nlynzaad Thank you. I've created an issue with a reproducible example: https://github.com/TanStack/router/issues/4580

PR #4574, opened 2025-07-04
fix(solid-router): loaders being called twice during SSR

brenelz:
Basically my understanding is that the `load` function gets called once during SSR but right now our `Transitioner` calls it again as it appears `createRenderEffect` run on the server too where as the `useEffect` in React doesn't.

The fix is just to bail out if its running on the server. Feels kinda hacky but it works.


Ingramz:
Pointing out that the React version is using  a `useLayoutEffect` utility function which corresponds to `React.useLayoutEffect` or `React.useEffect` depending on the environment.

https://github.com/TanStack/router/blob/7938ff4832835d57bf62daa32a3b10fc3e3ddfb2/packages/react-router/src/utils.ts#L13-L14

This is then used throughout the Transitioner.

https://github.com/TanStack/router/blob/7938ff4832835d57bf62daa32a3b10fc3e3ddfb2/packages/react-router/src/Transitioner.tsx#L7

There is also a rough equivalent for this in Solid version, but it isn't used anywhere.
https://github.com/TanStack/router/blob/9e35918038b455417272aec42f1d49aec220af01/packages/solid-router/src/utils.ts#L3-L4

Instead, all such occurrences are currently using `Solid.createRenderEffect` directly.

https://github.com/TanStack/router/blob/9e35918038b455417272aec42f1d49aec220af01/packages/solid-router/src/Transitioner.tsx#L68

Currently the pull request only changes the behavior of the first `Solid.createRenderEffect` to be not ran on server, but the 3 other occurrences still might. Shouldn't all of the `Solid.createRenderEffect` usages be uniformly treated to be not executed on the server and not just the first one?

brenelz:
Yeah i tried to touch as little as possible. I tried changing from a renderEffect to a createEffect but that broke some unit tests. Didn't look too much further after that

SeanCassiere (approved):

PR #4579, opened 2025-07-05
fix: clear pendingTimeout when match resolves

schiller-manuel:


hanneswidrig:
I can confirm that this fixes the issues I was having that were re-created with the current latest version.

Reproduction below is using this PR's preview release.

To demonstrate bug, change package.json to point to 1.125.0.

https://stackblitz.com/edit/tanstack-router-w94nsq5x?file=package.json,src%2Froutes%2F(protected)%2Fposts%2Froute.tsx

PR #4581, opened 2025-07-06
fix: buildLocation check for non changing routes - consider trailing slashes

### LINKED ISSUE #4580: router: buildLocation non-changing route with index route

Router

### Describe the bug

When using this code to do a search param change on the current page, the route is changing incorrectly:

```ts
navigate({
  to: ".",
  search: (prev) => ({
    ...prev,
    [`_${name}`]: open ? true : undefined,
  }),
  resetScroll: false,
});
```

https://github.com/checkerschaf/tanstack-start-buildlocation-bug

1. Go to http://localhost:3000/en/hello
2. Press the top `Add test` button
3. The url should stay the same, but it changes to `/en` instead

### Expected behavior

The url should stay `http://localhost:3000/en/hello` and append `?_test=true` as a search param

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.125.1
- OS: macOS
- Browser: Chrome
- Browser Version: latest
- Bundler: vite

### Additional context

_No response_

nlynzaad:
This PR is a follow-up to #4573. 

When checking if the paths have changed, we need to also take into account any trailing slashes. This resolves #4580.

when comparing the paths in this issue the resolved path for dest.to did not have a trailing slash while the fromPath did.

Setting up test for similar routing structures did not lead to the same failures so I'm not too sure how to recreate this with a test, to be honest not sure what would be the required actions to cause the from path to contain a trailing slash.

Tested this by patching the modules in the example and it succeeded then.


nlynzaad:
ran all tests locally multiple times, by resetting the local nx cache and workspaceData, and I'm not able to reproduce the failure reported here.

schiller-manuel:
that's a flaky test in ci

nlynzaad:
when running the test in isolation I was able to get the error.

I added a check for the "Client only Content" before navigation to ensure it has rendered prior to navigation. This seems to have resolved the error. Also updated it to find by testId and added mock for scrollTo. 

Hoping this resolves the issue in the CI

schiller-manuel:
very cool. 

coming back to the original issue, can we not build a unit test that resembles the reproducer?

nlynzaad:
I'll spend some time to try and recreate it now. 

Going to be a bit trial and error since I'm not too sure what navigation action caused a trailing slash on the fromPath that broke it. But once I have figured that out, I can add the test

nlynzaad:
test added. ended up just being the route path that had to be set accordingly

checkerschaf:
It works with the latest release 🥳 Thanks @nlynzaad & @schiller-manuel

PR #4587, opened 2025-07-07
fix: prevent duplicate script tags during hydration in Asset component

### LINKED ISSUE #4585: Using `createRootRoute`, `Scripts`, and `HeadContent` components adds same script multiple times

Start

### Describe the bug

When using the `Scripts` component and defining `scripts` array inside `createRootRoute`, the same script gets added multiple times in the head tag. This also results in double scripts for hot reload **inside the body**

**Removing the `<Scripts />` component removes the entire client side functionality**

https://codesandbox.io/p/devbox/github/tanstack/router/tree/main/examples/react/start-basic?embed=1&theme=dark&file=src/routes/__root.tsx

1. Go to the codesandbox url
2. Open the devtools using the button to the right of the URL bar (in codesandbox preview)
3. Open the head tag

### Expected behavior

The script tag should be added only once, because it is defined only once

### Screenshots or Videos

Image

Image

Image

### Platform

- Router / Start Version: 1.125.4
- OS: Linux (Codesandbox)
- Browser: Edge
- Browser Version: 139.0.3405.13
- Bundler: Vite
- Bundler Version: 6.3.5

### Additional context

I think this may be caused due to the script being added automatically by `createRootRoute`, `HeadContent` and `Scripts`. There must be a way to filter out duplicate scripts in order to fix this

max-programming:
Fixes #4585 

The issue was that during hydration, the `Asset` component adds the same `script` to the `head` twice. 

The fix applied is that if a script already exists, it will exit out of `useEffect` so that it doesn't get added again


max-programming:
Okay, so the problem is more complex than I thought.

1. If the page with scripts is directly loaded, 2 script tags get added, and both execute
2. If the page was navigated to, 2 script tags get added, and only one executes

schiller-manuel:
we should start with a failing unit test for router as well as a failing e2e test for both router and start 
can you please add this?

also, does this also get added twice without strict mode?

max-programming:
@schiller-manuel only the e2e test is failing [here](https://cloud.nx.app/runs/nLGokQOMcs/task/tanstack-react-start-e2e-basic%3Atest%3Ae2e), none of the unit tests

and yea i tried removing the StrictMode, still adds the scripts twice

schiller-manuel:
we need new tests that test that the scripts are not duplicated. those test should fail initially and then we can work on the solution

Christopher96u:
> we need new tests that test that the scripts are not duplicated. those test should fail initially and then we can work on the solution

Hi @schiller-manuel , I'd like to add the tests cases for this issue so we can get the fix in the following release. Just to confirm which test cases would be enough for this ? I'm planning just to check the number of scripts for a dummy script is 1.

AkhilLV:
I have this case where the script is added twice: once in the <head> and then in <div#app>. Never used the Scripts tag

```js
  component: App,
  head: () => ({
    scripts: [
      {
        type: 'application/ld+json',
        children: JSON.stringify({
          '@context': 'https://schema.org',
          '@type': 'WebPage',
          name: 'Index',
          description: 'Ducks are dumb',
        }),
      },
    ],
  }),
})
```

and called <HeadContent /> in __root.tsx.

[Minimal example on codesandbox](https://codesandbox.io/p/github/AkhilLV/tanstack-bug-reproduction/main?import=true)

image

schiller-manuel:
@Christopher96u 

we need to test:

SSR a route with some scripts, should not have duplicate scripts. 
client side navigation from some other route to that route, should also not result in duplicate scripts

Christopher96u:
> @Christopher96u
> 
> we need to test:
> 
> SSR a route with some scripts, should not have duplicate scripts. client side navigation from some other route to that route, should also not result in duplicate scripts

Thanks, I'll try to add this 2 cases this weekend

Christopher96u:
Hi @max-programming @schiller-manuel  I've tried to reproduce this issue cloning the repository and I can't any pointers? Maybe this bug has been fixed somehow in the current/latest versions of router? I've spent few hours to reproduce it without luck

max-programming:
@Christopher96u it stills seems to be there in the latest versions. try doing the following to reproduce

Clone the `start-basic` example

```sh
npx gitpick TanStack/router/tree/main/examples/react/start-basic start-basic
cd start-basic
npm install
npm run dev
```

Here's how the script is used - **inside `createRootRoute` > `head` > `scripts`**
https://github.com/TanStack/router/blob/9dec9b78b1c3d9e0f9f3cf8c8747ff7a2f4f1def/examples/react/start-basic/src/routes/__root.tsx#L53-L58

Upon running the app, open the devtools and check the `head` tag in Elements
You will see the below script
```html
<script src="/customScript.js" type="text/javascript"></script>
<script src="/customScript.js" type="text/javascript"></script> <!-- dupe -->
```

Also in the Console, you will see this console log twice
```
Hello from customScript.js!
Hello from customScript.js! // dupe
```

Christopher96u:
> @Christopher96u it stills seems to be there in the latest versions. try doing the following to reproduce
> 
> Clone the `start-basic` example
> 
> ```shell
> npx gitpick TanStack/router/tree/main/examples/react/start-basic start-basic
> cd start-basic
> npm install
> npm run dev
> ```
> 
> Here's how the script is used - **inside `createRootRoute` > `head` > `scripts`**
> 
> https://github.com/TanStack/router/blob/9dec9b78b1c3d9e0f9f3cf8c8747ff7a2f4f1def/examples/react/start-basic/src/routes/__root.tsx#L53-L58
> 
> Upon running the app, open the devtools and check the `head` tag in Elements You will see the below script
> 
> ```
> <script src="/customScript.js" type="text/javascript"></script>
> <script src="/customScript.js" type="text/javascript"></script> 
> ```
> 
> Also in the Console, you will see this console log twice
> 
> ```
> Hello from customScript.js!
> Hello from customScript.js! // dupe
> ```

Thanks mate. I'll try to do tonight and run the tests. Because I already wrote them, the thing is when I run the dev server I can't see duplicates. Maybe the trick is add them to the __root. If that's the case, I'll modify my tests, thank you!

max-programming:
Yeah adding them to __root might be the cause. But hey no problem, take your time ✌️

Christopher96u:
> Yeah adding them to __root might be the cause. But hey no problem, take your time ✌️

Ok, I could reproduce it, but just in the "example", here `router/e2e/react-start/basic` where actually is the setup with all the tests I can't reproduce it, so it's the setup the only difference, could you @max-programming help me to modify the react-start setup to reproduce the issue?

Christopher96u:
I've tried different approaches to reproduce the issue under `router/e2e/react-start/basic`, removing the custom server/client entry implementation or rolling back the shellComponent to component without luck 🥲

packages/react-router/src/Asset.tsx
```diff
@@ -46,6 +46,11 @@ function Script({
 }) {
   React.useEffect(() => {
     if (attrs?.src) {
+      const existingScript = document.querySelector(
```
brenelz, line 49: hmm not sure this is the ideal solution, but i guess it works

PR #4598, opened 2025-07-08
docs: replace reference to app folder with the correct src folder

JordanWinslow:
Incorrect reference to app folder in learn-the-basics.md comments may lead to confusion for students and break their project


schiller-manuel:
sorry misclicked. thanks for that!

PR #4600, opened 2025-07-08
feat: use seroval for SSR

schiller-manuel:
also stream all queries in react-router-with-query


packages/react-router-with-query/src/index.tsx
```diff
@@ -180,3 +163,31 @@ export function routerWithQueryClient<TRouter extends AnyRouter>(
...
+  let isClosed = false
+
+  return {
+    stream,
+    enqueue: (chunk) => controllerRef?.enqueue(chunk),
+    close: () => {
+      controllerRef?.close()
+      isClosed = true
+    },
+    isClosed,
```
SeanCassiere: Might be better to align and use a get function here (i.e `getIsClosedStatus()`)

packages/react-router-with-query/src/index.tsx
```diff
@@ -142,22 +85,62 @@ export function routerWithQueryClient<TRouter extends AnyRouter>(
...
-        dehydratedQueryClient: dehydrate(queryClient),
+        dehydratedQueryClient: queryDehydrate(queryClient),
+        // prepare the stream for queries coming up during rendering
+        queryStream: queryStream.stream,
       }
     },
-    hydrate: (dehydrated: any) => {
-      ogOptions.hydrate?.(dehydrated)
+    hydrate: async (dehydrated: DehydratedRouterQueryState) => {
+      console.log('dehydrated', dehydrated)
```
SeanCassiere: 💣

packages/router-core/src/ssr/ssr-client.ts
```diff
@@ -1,113 +1,77 @@
...
-  type: 'promise'
-  value?: ControllablePromise<any>
+  }
 }
-
-export interface ClientExtractedStream extends ClientExtractedBaseEntry {
-  type: 'stream'
-  value?: ReadableStream & { controller?: ReadableStreamDefaultController }
+export interface TsrSsrGlobal {
+  r?: DehydratedRouter
```
SeanCassiere: While it makes sense for the `matches` to have keys be shortened (i.e. `i`, `b`, `l`, etc.), I think these single callers may benefit from just being left with having descriptive names to represent their purpose (i.e. `router/dehydratedRouter`, `close`, `whatever_v_is`).

packages/start-client-core/src/serializer.ts
```diff
@@ -1,17 +1,12 @@
-import { isPlainObject } from './utils'
+import { isPlainObject } from '@tanstack/router-core'
 
-export interface TsrSerializer {
+export interface StartSerializer {
```
SeanCassiere, line 3: Why does the Serializer get moved to Start? why not keep it in `router-core`.

Wouldn't this same serializer benefit normal `basic-file-based-ssr-streaming` as well?

packages/router-core/src/ssr/ssr-client.ts
schiller-manuel: renamed r to router, v is gone
added a comment to c ( the cleanup function), this is called per script injection and thus benefits from a short name

packages/start-client-core/src/serializer.ts
schiller-manuel: this PR only uses seroval for SSR
for server functions, we still use the old serializer. in a followup PR, we'll move server functions over to seroval as well

packages/react-router-with-query/src/index.tsx
```diff
@@ -1,16 +1,15 @@
...
 import {
   QueryClientProvider,
-  dehydrate,
-  hashKey,
-  hydrate,
+  dehydrate as queryDehydrate,
+  hydrate as queryHydrate,
 } from '@tanstack/react-query'
 import { isRedirect } from '@tanstack/router-core'
+import '@tanstack/router-core/ssr/client'
```
isnifer, line 8: Looks like extra import

schiller-manuel: what's the issue here?

isnifer: Is this intended to import this module like side effect without import specific functions, etc?

schiller-manuel: the import is necessary to get the type import. is there any problem with that import?

isnifer: Sorry for disturbing.
I expected to see something like:

```ta
import { type SomeType } from '@tanstack/router-core/ssr/client'
```

instead of `import '@tanstack/router-core/ssr/client'`

gunnartorfis:
@schiller-manuel this breaks Clerk with Tanstack Start. 
Screenshot 2025-07-17 at 09 27 31

The Clerk starter isn't working in the latest start version. The getAuth() function throws: https://github.com/TanStack/router/tree/main/examples/react/start-clerk-basic

The reason I reach out here is that I want to point the Clerk team towards something if the fix lies no their end.

AhmedBaset:
Is there a way to configure a custom serializer? seems like firebase' Timestamp is broken

```
Error reading routerStream: g [Error]: The value [object Object] of type "object" cannot be parsed/serialized.
      
There are few workarounds for this problem:
- Transform the value in a way that it can be serialized.
- If the reference is present on multiple runtimes (isomorphic), you can use the Reference API to map the references.
    at G.parseObject (file:///path/to/seroval@1.3.2/node_modules/seroval/dist/esm/production/index.mjs:17:31059)
  value: Timestamp { seconds: 1739221386, nanoseconds: 843000000 }
}
Error reading appStream: TypeError [ERR_INVALID_STATE]: Invalid state: Controller is already closed
    at ReadableStreamDefaultController.enqueue (node:internal/webstreams/readablestream:1077:13)
    at Object.write (file:///path/to/@tanstack+router-core@1.128.4/node_modules/@tanstack/router-core/dist/esm/ssr/transformStreamWithRouter.js:29:18)
    at Object.onData (file:///path/to/@tanstack+router-core@1.128.4/node_modules/@tanstack/router-core/dist/esm/ssr/transformStreamWithRouter.js:158:26)
    at readStream (file:///home/ahmed/projects/59-haya-flower/node_modules/.pnpm/@tanstack+router-core@1.128.4/node_modules/@tanstack/router-core/dist/esm/ssr/transformStreamWithRouter.js:51:48) {
  code: 'ERR_INVALID_STATE'
}
node:internal/process/promises:394
    triggerUncaughtException(err, true /* fromPromise */);
    ^

Node.js v24.3.0
```

schiller-manuel:
@AhmedBaset not yet. there will be a way to add custom seroval plugins

AhmedBaset:
Any fixes at the moment except downgrading?

schiller-manuel:
don't return anything from loader / beforeLoad that contains custom classes

ttessarolo:
@schiller-manuel for instance "clerkInitialState" is a custom class (or maybe QueryClient)? 
That's my context that breaks the app:

```
context {
  queryClient: QueryClient {},
  clerkInitialState: {
    __internal_clerk_state: {
      __clerk_ssr_state: [Object],
      __publishableKey: 'pk_test_bW9yZS15ZXRpLTczLmNsZXJrLmFjY291bnRzLmRldiQ',
      __proxyUrl: '',
      __domain: '',
      __isSatellite: false,
      __signInUrl: '',
      __signUpUrl: '',
      __afterSignInUrl: '',
      __afterSignUpUrl: '',
      __clerk_debug: [Object],
      __clerkJSUrl: undefined,
      __clerkJSVersion: undefined,
      __telemetryDisabled: false,
      __telemetryDebug: false,
      __signInForceRedirectUrl: '',
      __signUpForceRedirectUrl: '',
      __signInFallbackRedirectUrl: '',
      __signUpFallbackRedirectUrl: ''
    }
  },
  user: { id: 'user_2zrMCOP6EyuJrJoiHwIrrc5qzAp' }
}
```

Issue #4010, opened 2025-04-17
Error "Invariant failed: Could not find match for from" when layout + file-based routing is used


Dharmang92:
### Which project does this relate to?

Router

### Describe the bug

When trying to use the layout routing file, the routes created are correct but when trying to open one of the route using <Link /> component, it fails to load or when trying to refresh the page which is already on the failed route.

### Your Example Website or App

https://stackblitz.com/edit/tanstack-invariant-bug

### Steps to Reproduce the Bug or Issue

1. Click on About or Dashboard link
2. Error is thrown

### Expected behavior

I should be able to properly add a layout routing file and the nested components should work correctly.

For example below routes:

app/home
app/about
app/dashboard

I have route.tsx (default pattern for layout file) file inside app folder which contains the nav bar, then /home, /about and /dashboard should have the nav bar present in them.

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/027f6a04-93c7-4074-9bd1-65122ecafee0)

### Platform

"@tanstack/react-router": "^1.116.0"
"@tanstack/router-plugin": "^1.116.1"

schiller-manuel:
your example has a basepath 'app' configured AND your routes are inside an `app` folder.
So the runtime URL would be /app/app/..
Probably not what you want?
Either remove the basepath or move the routes outside of the app folder

Dharmang92:
Is there any specific change in the latest version Cause the same code works for v1.102.5?
If I remove the basepath, is there any way to directly redirect to /app always?

schiller-manuel:
throwing a redirect to /app in __root's beforeLoad could work

Dharmang92:
Thank you for the quick response @schiller-manuel, I tested this out but even when I add beforeLoad and removing the app folder, it still doesn't redirect to app just like how basepath would work. Also renaming the app folder to app_ doesn't work...

Dharmang92:
+

I just want a way to do below, it's not very complex as far as the configuration goes...

1. I should be able to have an app directory (just to keep the structure of the routes in sync with the folder structure)
2. Start the application with default /app route - basically if someone lands on the root page for e.g. example.com/ then it should route to exmaple.com/app
[...]

PR #4610, opened 2025-07-09
fix: resolve false positive "could not find match from"

### LINKED ISSUE #4284: Route match warnings (main/devinxi)

Start

### Describe the bug

Under certain conditions, routes with dynamic parameters are logging match warnings when links are clicked.

This is reproduced in the start-basic example upgraded to the alpha release

```
posts.$postId.tsx:47 Could not find match for from: /posts/$postId
```

posts.$postId.tsx
```tsx
47:   <Link
        to="/posts/$postId/deep"
        params={{
          postId: post.id,
        }}
        activeProps={{ className: "text-black font-bold" }}
        className="inline-block py-1 text-blue-800 hover:text-blue-600"
      >
        Deep View
      </Link>
```

https://github.com/danecando/start-basic

1. Start the alpha/devinxi start-basic example
2. Navigate to `/posts`
3. Click a link for one of the posts in the list
4. Click the "Deep View" link in the post page

After clicking "Deep View" route match warning gets logged to the console

### Expected behavior

As a user, I expect to be able to link to full paths in my route tree without console warnings

### Screenshots or Videos

_No response_

### Platform

- OS: MacOS Sequoia 15.5
- Browser: Chrome
- Version: 136

### Additional context

Edit: Removed this note since it is no longer relevant. Other users are providing useful context and details in the comments

nlynzaad:
This PR modifies the check to not only check if from is not matched to routes related to currentLocation but also checks that currentLocation can't be matched to routes related to from.

It also adds a test for this.

fixes #4284

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -1444,26 +1444,28 @@ export class RouterCore<
       } else if (routeIsChanging && dest.from) {
         fromPath = dest.from
 
-        const allFromMatches = this.getMatchedRoutes(
-          dest.from,
-          undefined,
-        ).matchedRoutes
+        // do this check only on navigations during test or development
+        if (process.env.NODE_ENV !== 'production' && dest._isNavigate) {
+          const allFromMatches = this.getMatchedRoutes(
+            dest.from,
+            undefined,
+          ).matchedRoutes
 
-        const matchedFrom = [...allCurrentLocationMatches]
-          .reverse()
-          .find((d) => {
-            return this.comparePaths(d.fullPath, fromPath)
+          const matchedFrom = [...allCurrentLocationMatches]
+            .reverse()
+            .find((d) => {
+              return this.comparePaths(d.fullPath, fromPath)
+            })
+
+          const matchedCurrent = [...allFromMatches].reverse().find((d) => {
+            return this.comparePaths(d.fullPath, currentLocation.pathname)
           })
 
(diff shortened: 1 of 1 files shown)


nlynzaad:
looks like this might need more work. #4010 looks to be technically the same root cause but this doesn't resolve that. want to look a bit further.

packages/react-router/tests/link.test.tsx
```diff
@@ -1383,6 +1383,134 @@ describe('Link', () => {
...
+
+    const consoleWarnSpy = vi.spyOn(console, 'warn')
+
+    await act(() => fireEvent.click(homeLink))
+
+    expect(window.location.pathname).toBe('/')
+    const homeHeading = await screen.findByTestId('home-heading')
+    expect(homeHeading).toBeInTheDocument()
+
+    expect(consoleWarnSpy).not.toHaveBeenCalledWith(
```
schiller-manuel: can we just check that no warning at all occurred? otherwise changing the warning text would render this text useless

schiller-manuel (changes requested):

packages/react-router/tests/link.test.tsx
nlynzaad: saw this as well and updated accordingly.

nlynzaad:
updated the solution and added additional test for multi layer routes similar to what had problems in kitchen-sink-file based example. 

this should be good to go now.

PR #4619, opened 2025-07-10
fix(start-server-functions-server): add base path to the RPC URL

OnurGvnc:
Hi! This PR fixes a bug where server function urls are incorrect when the app uses a basePath.

## The Problem:

The `createServerRpc` function (on the server) was creating URLs without the base path. However, the `createClientRpc`
function (on the client) was including it.

https://github.com/TanStack/router/blob/b9f88c5426f3221ef5b43b11df35f8ba7d61042c/packages/start-server-functions-server/src/index.ts#L18

https://github.com/TanStack/router/blob/b9f88c5426f3221ef5b43b11df35f8ba7d61042c/packages/start-server-functions-client/src/index.ts#L9-L12

For example, with a base path of '/app':

- Server URL: /_serverFn/<function_id>
- Client URL: /app/_serverFn/<function_id>

`<form action={loginFn.url} method="POST">`
hydration error:
```
  <LoginPage>
    <div>
      <h1>
        <form
+         action="/app/_serverFn/src_routes_auth_login_tsx--loginFn_createServerFn_handler"
-         action="/_serverFn/src_routes_auth_login_tsx--loginFn_createServerFn_handler"
           method="POST"
        >
```

This difference caused two main problems:
- A hydration error
- Broken forms, because submitting a form would go to the wrong URL (without the /app prefix).

## The Solution:

I've updated `createServerRpc` to also use the base path when it creates the URL, just like the client version does.

File changed: packages/start-server-functions-server/src/index.ts

```ts
// Before
const url = `/${sanitizeBase(serverBase)}/${functionId}`

// After
const sanitizedAppBase = sanitizeBase(process.env.TSS_APP_BASE || '/')
const sanitizedServerBase = sanitizeBase(serverBase)

const url = `${sanitizedAppBase ? `/${sanitizedAppBase}` : ``}/${sanitizedServerBase}/${functionId}`
```

Now, both the client and server create the same correct URL.

### DIFF
--- a/e2e/react-start/custom-basepath/src/routeTree.gen.ts
+++ b/e2e/react-start/custom-basepath/src/routeTree.gen.ts
@@ -13,6 +13,7 @@ import { createServerRootRoute } from '@tanstack/react-start/server'
 import { Route as rootRouteImport } from './routes/__root'
 import { Route as UsersRouteImport } from './routes/users'
 import { Route as PostsRouteImport } from './routes/posts'
+import { Route as LogoutRouteImport } from './routes/logout'
 import { Route as DeferredRouteImport } from './routes/deferred'
 import { Route as IndexRouteImport } from './routes/index'
 import { Route as UsersIndexRouteImport } from './routes/users.index'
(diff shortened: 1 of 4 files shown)


schiller-manuel:
thanks for this!
can you please add an e2e test for this in the `e2e/react-start/custom-basepath` project?

OnurGvnc:
Thanks! I've added the e2e test.

  Without this PR's changes, the test would fail with:
```
  Expected: "/custom/basepath/_serverFn/src_routes_logout_tsx--logoutFn_createServerFn_handler"
  Received: "/_serverFn/src_routes_logout_tsx--logoutFn_createServerFn_handler"
```
  The test now passes with the PR's fix, confirming that server function URLs correctly include the app's basepath.

PR #4621, opened 2025-07-11
fix(solid-router): prevent client effects running on server

Ingramz:
This is exploration of the comment I posted on https://github.com/TanStack/router/pull/4574#issuecomment-3038862345.

Using `Solid.createRenderEffect` in SSR will execute the effects after the render takes place. To prevent server from doing meaningless work, we can use the `useLayoutEffect` utility function, which will make the server use `Solid.createEffect` instead, preventing such effects from being ran on server. Client side there are no differences as `Solid.createRenderEffect` is still used. This is closer to how it works in React version today.

To properly test this, we need true SSR with `environment: 'node'` and  `solidPlugin({ ssr: true })` in Vite, because:
1) `jsdom` environment provides `window`, which causes wrong branch in `useLayoutEffect` to be chosen. This is fixable by replacing `typeof window !== 'undefined'` check with `isServer` from `solid-js/web`, resulting in a marginally smaller client bundle, too (This was decided against in order to keep the code as similar to react-router). 
2) Without `ssr: true`, we do not have access to `renderToString`.

For now I added the SSR config to the same vite config via `--mode` switch. However I am unhappy with the way how now `package.json` script needs to launch vitest twice: `"test:unit": "vitest && vitest --mode server"`. Suggestions on better format or alternatives would be most welcome.

### DIFF
--- a/packages/react-router/src/Matches.tsx
+++ b/packages/react-router/src/Matches.tsx
@@ -54,7 +54,7 @@ export function Matches() {
 
   const inner = (
     <ResolvedSuspense fallback={pendingElement}>
-      <Transitioner />
+      {!router.isServer && <Transitioner />}
       <MatchesInner />
     </ResolvedSuspense>
   )
--- a/packages/react-router/src/Transitioner.tsx
+++ b/packages/react-router/src/Transitioner.tsx
@@ -30,14 +30,12 @@ export function Transitioner() {
   const isPagePending = isLoading || hasPendingMatches
   const previousIsPagePending = usePrevious(isPagePending)
 
-  if (!router.isServer) {
-    router.startTransition = (fn: () => void) => {
-      setIsTransitioning(true)
-      React.startTransition(() => {
-        fn()
-        setIsTransitioning(false)
-      })
-    }
+  router.startTransition = (fn: () => void) => {
+    setIsTransitioning(true)
+    React.startTransition(() => {
+      fn()
+      setIsTransitioning(false)
+    })
   }
 
   // Subscribe to location changes
--- a/packages/solid-router/package.json
+++ b/packages/solid-router/package.json
@@ -33,7 +33,7 @@
     "test:types:ts56": "node ../../node_modules/typescript56/lib/tsc.js -p tsconfig.legacy.json",
     "test:types:ts57": "node ../../node_modules/typescript57/lib/tsc.js -p tsconfig.legacy.json",
     "test:types:ts58": "tsc -p tsconfig.legacy.json",
-    "test:unit": "vitest",
+    "test:unit": "vitest && vitest --mode server",
     "test:unit:dev": "pnpm run test:unit --watch --hideSkippedTests",
     "test:perf": "vitest bench",
     "test:perf:dev": "pnpm run test:perf --watch --hideSkippedTests",
--- a/packages/solid-router/src/Matches.tsx
+++ b/packages/solid-router/src/Matches.tsx
@@ -50,7 +50,7 @@ export function Matches() {
 
   const inner = (
     <ResolvedSuspense fallback={pendingElement}>
-      <Transitioner />
+      {!router.isServer && <Transitioner />}
       <MatchesInner />
     </ResolvedSuspense>
   )
--- a/packages/solid-router/src/Transitioner.tsx
+++ b/packages/solid-router/src/Transitioner.tsx
@@ -30,12 +30,10 @@ export function Transitioner() {
   const isPagePending = () => isLoading() || hasPendingMatches()
   const previousIsPagePending = usePrevious(isPagePending)
 
-  if (!router.isServer) {
-    router.startTransition = async (fn: () => void | Promise<void>) => {
-      setIsTransitioning(true)
-      await fn()
-      setIsTransitioning(false)
-    }
+  router.startTransition = async (fn: () => void | Promise<void>) => {
+    setIsTransitioning(true)
+    await fn()
+    setIsTransitioning(false)
   }
 
   // Subscribe to location changes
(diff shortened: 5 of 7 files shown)


Ingramz:
The SSR test [passed on CI](https://github.com/TanStack/router/actions/runs/16211012012/job/45772082119#step:6:10573) so overall this approach seems to be working at least.

I did try Vitest's [test projects](https://vitest.dev/guide/projects.html) feature to combine client and server tests into same config, but I could not get it working, could very well be that it isn't the right tool for the job. Giving up on trying to improve it any further for now. Currently implemented approach is [borrowed from solid-primitives](https://github.com/solidjs-community/solid-primitives/blob/main/configs/vitest.config.ts).

---

As an added note to the main item of this pull request, `Solid.createEffect` on server is merely an [empty function](https://github.com/solidjs/solid/blob/3d3207dd3aeb84c2a38377cf9f3b895995c2d969/packages/solid/src/server/reactive.ts#L101), which makes me think that it's probably not even the best solution to use `useLayoutEffect` instead.

Transitioner largely seems to be consisting of client only behaviors and one might as well just short circuit it at some specific point with `if (isServer) return null` to reduce the amount of code built for and executed on the server, though the impact might be negligible as it is a micro optimization at best. I'll think about it a little in the meanwhile.

brenelz:
Im honestly not sure about this PR. Is there any specific issue this solves or just trying to run less on the server when its not needed?

packages/solid-router/src/utils.ts
```diff
@@ -1,7 +1,9 @@
 import * as Solid from 'solid-js'
+import { isServer } from 'solid-js/web'
 
-export const useLayoutEffect =
-  typeof window !== 'undefined' ? Solid.createRenderEffect : Solid.createEffect
+export const useLayoutEffect = isServer
```
brenelz: I'm not sure I like this indirection but i guess it was kinda there already

Ingramz:
> Im honestly not sure about this PR. Is there any specific issue this solves or just trying to run less on the server when its not needed?

The latter. Just rubs me the wrong way that there are effects queued and then ran after server has already rendered the output.

At closer inspection I'd say that the entire Transitioner is pretty much unnecessary on server side, the code that's before the client effects is just setup for the effects. So we could just introduce `if (isServer) return null` on the first line of the component and leave the rest as it was before. Optionally do away with `useLayoutEffect` as well as it seems unused and perhaps too misleading in solid-router. Let me know if this is acceptable and I will amend the PR accordingly.

schiller-manuel:
we should keep react and solid impl aligned when it comes to Transitioner. if there are changes applicable for both then please go ahead

Ingramz:
> we should keep react and solid impl aligned when it comes to Transitioner. if there are changes applicable for both then please go ahead

I reworked the commits such that both React and Solid Transistioners now bail out as early out as possible on the server. I may have misjudged the usefulness of Solid's native `isServer` in this particular instance as currently the code from `node_modules` isn't optimized in production builds for server, so it would have ended being a runtime check anyway with no dead code elimination.

~~On Solid's side I am not touching the `useLayoutEffect` util any longer. I tend to agree with @brenelz that the indirection is undesirable and since nothing is actively using it in `solid-router` currently, it shouldn't be landing into client bundles unless the developer imports it themselves. Perhaps it's a good candidate for removal in future, but that's out of scope for this PR.~~ Turns out we can't avoid it if the intent is to keep the react and solid implementations similar, a necessary evil.

For now I do not have any further changes to propose, the change set as it is fixes my concern of Solid doing more work than React and I guess there's a slight win for React as well unless somehow returning early on server is illegal.

schiller-manuel:
what would be needed to fully support isServer when building the server part?

Ingramz:
> what would be needed to fully support isServer when building the server part?

Nothing on the library side other than using it, it should be a bundle time optimization. I cannot find my test code for server output anymore, everywhere I look it's just a plain import for server code, so I must have been wrong. All the ones I could find are client code where it does evaluate to false and strip away dead branches.

---

The tests failed due going against hook rules in React 🤡. I am giving up. I can't make this optimization work in both without diverging the implementations, so I guess I'll use `useLayoutEffect` after all and call it a day. That way both will look similar and behave similar. It's ready for review now, for real.

schiller-manuel:
can we have separate ClientTransitioner and a noop ServerTransitioner impl that we render depending on the env?

Ingramz:
Yeah I suppose that can be done. It's effectively a client-only component, so we can probably avoid creating the no-op if we selectively render it.

We could simply check for server at the usage and then not render it. This would be my preferred choice as Transitioner cannot be used externally and there is already some [conditional rendering being done at the only usage](https://github.com/TanStack/router/blob/9fb55b88630b871bbb0d797cc442b66b29d62a60/packages/react-router/src/Matches.tsx#L49-L59).
```jsx
{!router.isServer && <Transitioner />}
```

Alternatively we could hide the isServer check inside Transitioner. This would only be an improvement in my opinion if it was part of public API.
```jsx
function Transitioner() {
  const router = useRouter()
  return router.isServer ? null :  <ClientTransitioner />
}
```

Or export a Transitioner impl based on some environment check which probably is the most literal interpretation of your suggestion:

```jsx
export const Transitioner = typeof window !== 'undefined' ? ClientTransitioner : (() => null)
```

Depending on bundler setup this can optimize away the unused branch for a given environment, but it's not entirely clear to me when these optimizations happen and don't happen, so I'd save this variant for a discussion in the future instead. When no optimizations take place, we'd end up shipping both branches to client, which is no better than just opting to do the check during runtime using options 1 or 2.

Lastly, I was thinking of just wrapping Transitioner with ClientOnly:
```jsx
<ClientOnly>
  <Transitioner />
</ClientOnly>
```

Unit tests seem to pass with this, but compared to the first option, I think this will delay the rendering of Transitioner in client by one render cycle, which might be fine, though I am unsure. Either way the ClientOnly usage will be some overhead and I'm more confident in option 1 working right than this.

If it wasn't obvious already, I'd just stick to option 1 for now.

schiller-manuel:
solution 1 looks good to me.

FYI i am adding a static `isServer` that is resolved by the bundler via exports conditions in https://github.com/TanStack/router/pull/4648

PR #4625, opened 2025-07-11
fix(router-core): type mismatch for navigate function

mkarajohn:
According to the docs this should only return a promise https://tanstack.com/router/latest/docs/framework/react/api/router/RouterType#navigate-method 
Makes sense otherwise you cannot .then() it


mkarajohn:
Not sure about those failing tests, the MouseEvent error seems to be there from `main` already
image

nlynzaad:
the errors are coming through since you are not handling the case where the if statements, within the handleClick handlers (both solid-router and react-router) ,are false and hence nothing is returned. with the updated type signature, void is no longer valid as a return for these handlers

mkarajohn:
> the errors are coming through since you are not handling the case where the if statements, within the handleClick handlers (both solid-router and react-router) ,are false and hence nothing is returned. with the updated type signature, void is no longer valid as a return for these handlers

Thanks, should have spotted it. (in my defense I was preparing to leave work yesterday :sweat_smile: )

mkarajohn:
...bump in case this has slipped through? (sorry, not sure how long it usually takes to merge or comment on PRs)
:peace_symbol:

nlynzaad:
Apologies for the delay in responding. Should have some time to look at this more closely again.

nlynzaad:
Thanks for this. it looks good

mkarajohn:
Sorry, just got home; thank you, cheers!

PR #4634, opened 2025-07-12
Add enableBeforeUnload option example in navigation-blocking.md

anthony2261:
Added a note on using the `enableBeforeUnload` option in `useBlocker` + example in the "Component-based blocking" section.

@schiller-manuel I wasn't sure if there should be an entire section for this or just a note with 1-2 examples in the existing sections (I did the latter).

Also thanks for pushing me to contribute even if it's a small one :)


schiller-manuel:
looking good, thanks!

PR #4635, opened 2025-07-12
chore: Document static prerendering options

radiosilence:
It's a shame that you can't add docblocks to the source schema, but they get removed by the time it gets to a compiled `.d.ts`.


SeanCassiere (approved):
What you talking is possible using JSDocs, but we’d just need to split the type from the actual zod schema definition (or have one feed into the other).

PR #4638, opened 2025-07-13
docs: Update tsr.config.json options link

RawToast:
The link at for tsr.config.json found [in this section](https://tanstack.com/router/latest/docs/framework/react/migrate-from-react-location#differences-between-react-location-and-tanstack-router) points to [this page about file routing styles in the docs](https://tanstack.com/router/latest/docs/framework/react/routing/file-based-routing#options)

That doesn't seem right, when the `tsr.config.json` options are [found here](https://tanstack.com/router/latest/docs/api/file-based-routing)

I've followed the docs in the `tanstack.com` page and checked this link works as expected 😌

### DIFF
--- a/docs/router/framework/react/migrate-from-react-location.md
+++ b/docs/router/framework/react/migrate-from-react-location.md
@@ -70,7 +70,7 @@ Create a `tsr.config.json` file in the root of your project with the following c
 }
 ```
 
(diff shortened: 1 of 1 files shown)


nlynzaad:
Thanks for fixing this

PR #4644, opened 2025-07-13
docs: add jsdoc comments to plugin/generator types

RMHonor:
This PR adds JSDoc comments to the plugin/bundle config options, as discussed in https://github.com/TanStack/router/discussions/3790, giving usage hints inline and to hopefully pave the way for auto generated docs too.

The content for these docs were largely copied directly from the [Router API docs](https://tanstack.com/router/latest/docs/framework/react/api/router) with some tweaks to make it logical in the code context.

No logic changes have been made, purely type annotations.

### Note about Zod

When adding JSDoc comments to fields with Zod, they are lost when using the `z.input` utility types ([issue](https://github.com/colinhacks/zod/issues/200)). To get around this, an interface is created, which is annotated with JSDoc, and then the accuracy of this is checked against again the schema with `satisfies`. This interface is then using for the type arguments in the plugin/generator.

A small drawback in this approach is that the compiler won't detect extraneous options being added to the schema.

Example output:
Screenshot 2025-05-18 at 22 52 14


schiller-manuel:
we could assert type equality using this

```tsx
// copied from https://github.com/sindresorhus/type-fest/blob/4d7cc50fbec6ae631fca98ae8d7e498f0d5f61ab/source/is-equal.d.ts#L27
type IsEqual<A, B> =
	(<G>() => G extends A & G | G ? 1 : 2) extends
	(<G>() => G extends B & G | G ? 1 : 2)
		? true
		: false;
type Expect<T extends true> = T;
// this would error if X and Y are not equal
type E = Expect<IsEqual<X,Y>>
```

packages/router-generator/src/config.ts
```diff
@@ -53,9 +254,9 @@ export const configSchema = baseConfigSchema.extend({
     .optional(),
   plugins: z.array(z.custom<GeneratorPlugin>()).optional(),
   tmpDir: z.string().optional().default(''),
-})
+}) satisfies z.ZodType<ConfigOptions>
 
-export type Config = z.infer<typeof configSchema>
+export type Config = z.output<typeof configSchema>
```
SeanCassiere, line 259:
```suggestion
export type ValidatedConfig = z.output<typeof configSchema>
```

packages/router-plugin/src/core/config.ts
```diff
@@ -94,9 +107,9 @@ export const configSchema = generatorConfigSchema.extend({
         .optional(),
     })
     .optional(),
-})
+}) satisfies z.ZodType<ConfigOptions>
 
-export const getConfig = (inlineConfig: Partial<Config>, root: string) => {
+export const getConfig = (inlineConfig: ConfigInput, root: string) => {
```
SeanCassiere, line 112: This could be `unknown` yeah?

```suggestion
export const getConfig = (inlineConfig: unknown, root: string) => {
```

SeanCassiere:
Also, please see the conflict with `packages/router-plugin/src/core/router-composed-plugin.ts`.

LadyBluenotes:
Thanks for putting this together. The editor-intellisense goal is still useful, but the config types and composed-plugin API have moved on enough that this needs a fresh pass against current main.

I’ve opened #7785 to track the current configuration type and JSDoc work. The issue should give this a cleaner starting point when it is picked up again!

PR #4647, opened 2025-07-13
fix(react-start-client): useServerFn returns a stable reference

timreinke:
Previously, `react-start-client`'s `useServerFn` would return a new handler function on every render. The handler is now wrapped in `useCallback`.

Returning a new handler function can lead to excessive re-renders. Developers usually expect hooks like this to return a stable value, so the behavior is surprising.

[Discussion in discord](https://discord.com/channels/719702312431386674/1393921652798001213/1393986769061675108) confirms we don't need to make an equivalent update to the solid client.

My first PR here, so I'm not sure if I followed conventions properly -- not sure if I need to make an issue to close here, or if this warrants an e2e or something. Just let me know :)


schiller-manuel:
thanks!

PR #4648, opened 2025-07-13
feat: add isServer

schiller-manuel:


nick-potts:
Is there a way to override this in the router config?

schiller-manuel:
@nick-potts no. what would you need this for?

packages/start-plugin-core/src/plugin.ts
```diff
@@ -243,6 +243,11 @@ export function TanStackStartVitePluginCore(
                 },
                 outDir: getClientOutputDirectory(viteConfig),
               },
+              define: {
+                // Replace __TSR_IS_SERVER__ with false for client builds
+                // This enables tree-shaking of server-only code branches
+                __TSR_IS_SERVER__: 'false',
+              },
```
Sheraff: I don't think I saw this variable anywhere, it is a remainder of a previous attempt?

PR #4653, opened 2025-07-14
fix: add port handler server for e2e tests

nlynzaad:
During e2e test ports are issued through hashing the package name and calculating a port from there. This is generally reliable, but collisions still occur.

This PR enhances this process by introducing a small node server at the start of the e2e tests. the tests query the server for a port. If the port has not been issued yet the server will generate a port using the get-ports package, this checks the underlying OS for available ports and returns them sequentially within the specified range. 

Once a port has been issued it is stored in a port registry for future requests against the package name. Any future requests for a port with the same package name will be retrieved from the registry to keep the ports that is used by the package constant.

In the case where the server is not started, for example on when running tests individually, the original port calculation implementation via derivePort is used.

Once all tests are completed the server is shutdown.


nlynzaad:
from the output its clear the server is started but the tests do not find the server, I'm assuming there is some form of isolation happening when running in the pipeline since this does not occur when running locally.

Ideally, we would want to start the server as part of the "Setup Tools" step and stop the server on "Post Setup Tools" instead of through package.json scripts as is currently implemented. This would probably resolve the issue mentioned above.

nlynzaad:
closing this in favour of #4676

PR #4662, opened 2025-07-15
fix: prevent infinite redirects with encoded URL parameters

### LINKED ISSUE #4514: [Start] Encoded URL params cause redirect loop on page refresh

Start

### Describe the bug

Is this supposed to happen?

When using encoded parameters in routes, client-side navigation works fine, but hard refresh/page reload causes `ERR_TOO_MANY_REDIRECTS.`

https://github.com/TanStack/router/tree/main/examples/react/start-basic

1. Use `start-basic` https://github.com/tanstack/router/tree/main/examples/react/start-basic example app
2. In `post.tsx`, change params for postId from `post.id` to `post.title`
```jsx
<Link
   to="/posts/$postId"
   params={{ postId: post.title}}
   ...
```
3. Uncomment `fetchPost({ data: postId })` in the loader and just return `{}` so you do not get a post not found error.
```jsx
export const Route = createFileRoute('/posts/$postId')({
  loader: ({ params: { postId } }) => {
    console.info(`Loading post with id ${postId}...`)
    // return fetchPost({ data: postId })
    return {}
  },
 ...
```
4. Navigate to a post with spaces in title
5. :white_check_mark: Client navigation works - URL shows encoded params, console shows decoded
6. :x: Hard refresh the page → infinite redirect loop

### Expected behavior

Page should load normally on refresh, just like client navigation
Actual: `ERR_TOO_MANY_REDIRECTS` error

### Screenshots or Videos

https://github.com/user-attachments/assets/7678a1c2-fb7b-41f7-a314-4a91ed450db8

### Platform

- OS: macOS
- Browser: Chrome
- Version: 1.121.35

### Additional context

_No response_

tannerlinsley:
Fixes infinite redirect loops when refreshing pages with encoded URL parameters.

Browser history stores encoded URLs while buildLocation produces decoded URLs, causing URL mismatch in beforeLoad comparison.

Solution: Added URL normalization during comparison to ensure consistent format.

Fixes #4514

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -1785,7 +1785,21 @@ export class RouterCore<
         state: true,
         _includeValidateSearch: true,
       })
-      if (trimPath(this.latestLocation.href) !== trimPath(nextLocation.href)) {
+
+      // Normalize URLs for comparison to handle encoding differences
+      // Browser history always stores encoded URLs while buildLocation may produce decoded URLs
+      const normalizeUrl = (url: string) => {
+        try {
+          return encodeURI(decodeURI(url))
+        } catch {
+          return url
+        }
+      }
+
+      if (
+        trimPath(normalizeUrl(this.latestLocation.href)) !==
+        trimPath(normalizeUrl(nextLocation.href))
+      ) {
         throw redirect({ href: nextLocation.href })
       }
     }


SeanCassiere (commented):
🚀

---

This may not be part of this fix, but should we be doing this normalization at these places as well?

LN#710 (`getLocationChangeInfo -> hrefChanged`).

```ts
// router-core/router.ts
// getLocationChangeInfo function
export function getLocationChangeInfo(routerState: {
  // ...
}) {
  // ...
  const hrefChanged = fromLocation?.href !== toLocation.href // here
  // ...
}
```

LN#1642 (`Router.commitLocation -> const isSameUrl`).

```ts
// router-core/router.ts
// Router class
  commitLocation: CommitLocationFn = ({
    ...next
  }) => {
    // ...
    const isSameUrl = this.latestLocation.href === next.href // here
    // ...
```

nlynzaad:
@SeanCassiere there is a new private function that was added in router.ts, as part of #4581, called comparePaths, that does some cleaning for trailing slashes before comparing them and returning either true or false depending on wether the paths match or not.

your comment had me thinking we should probably add this in there and start using that more broadly. 

I will look into that once this gets merged.

PR #4664, opened 2025-07-16
fix: reset statusCode to 200 on navigation start

### LINKED ISSUE #4663: Router Status Code does not change from `404` to `200`.

Router

### Describe the bug

I’m not sure if this is the intended behavior, but once the `statusCode` is set to `404`, it doesn’t change back to `200` even when navigating to a valid route afterward.

https://stackblitz.com/edit/tanstack-router-eekdmazt?file=src%2Froutes%2F__root.tsx

1.	Confirm that the status code is `200`.
2.	Click on `nowhere` and check that the status code changes to `404`.
3.	Then, click on `about` or `home` again.
4.	The status code does not change back to `200`.

### Expected behavior

When navigating to a valid route, the status code should be `200`.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.127.9
- OS: macOS
- Browser: Chrome
- Bundler: [e.g. vite]

### Additional context

_No response_

w00khyung:
fix #4663 

The problem appears to be that statusCode isn't reset to 200 at the beginning of navigation, causing 404 to persist across subsequent navigations.

### DIFF
--- a/e2e/react-router/basic/src/main.tsx
+++ b/e2e/react-router/basic/src/main.tsx
@@ -8,6 +8,7 @@ import {
   createRoute,
   createRouter,
   redirect,
+  useRouterState,
 } from '@tanstack/react-router'
 import { TanStackRouterDevtools } from '@tanstack/react-router-devtools'
 import { NotFoundError, fetchPost, fetchPosts } from './posts'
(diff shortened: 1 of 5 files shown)


schiller-manuel:
thanks for starting this.
i don't think we need an e2e test for this, at least not for router.
this should be testable in a unit test.
also, can you please add a test like that: 

1. start with /
2. go to a /non-existing-route (=404)
3. go to /existing-route (200)
4. go to /another-non-existing-route (=404)

hyungwookchoi:
removed the e2e tests and added unit tests for the statusCode reset behavior.

schiller-manuel:
very nice, thanks a lot!

PR #4669, opened 2025-07-16
feat(router-generator): implement AST-based export detection for routes

### LINKED ISSUE #4487: createRootRoute is not detected if exported via `export { Route }` instead of `export const Route`

Router

### Describe the bug

Hi team 👋

Thanks for all the great work on TanStack Router — it's been a pleasure using it.

I wanted to flag a subtle but important issue we ran into when using file-based routing with `createRootRoute`.

In our project, we follow a convention of placing all exports at the end of the file, so we had something like this:

```js
const Route = createRootRoute({...});
export { Route };
```

However, with this format, the root route was skipped. None of the logic or layout inside it was applied at all. After some debugging, we realized that simply changing it to this fixed the issue:

```js
export const Route = createRootRoute({...});
```

It would be great if both export styles could be supported. 

https://stackblitz.com/edit/github-dzbsynpg-x5ue7aau

Run the app.

You will see the index route mount, but the root layout is skipped — no red border, no heading, and no console log from the root.

Now update the export in `__root.tsx` to:

```tsx
  component: () => {
    console.log("Root route rendered");
    return (
      <div style={{ border: "2px solid red", padding: 20 }}>
        <h1>Root Layout</h1>
        <Outlet />
      </div>
    );
  },
});


### Expected behavior


The root route should be recognized and rendered whether it's exported as:

```tsx
export const Route = createRootRoute({...});
```

or 

```tsx
const Route = createRootRoute({...});
export { Route };
```

### Screenshots or Videos

_No response_

### Platform

- OS: Any
- Browser: Any
- Version: Not browser-related — export handling issue

### Additional context

_No response_

hyungwookchoi:
fix #4487

Enhanced the router generator with AST-based parsing logic to detect route exports. 
Added support for both default and named exports in route files.


schiller-manuel:
thanks for approaching this. we already have very similar code for detecting the normal route exports. we should generalize this so it can be reused for both normal routes and the root route.

hyungwookchoi:
separated the common logic into utility files and modified the normal routes and root route to enable sharing of that logic.

CC-Bell:
Small note, seems like the following is currently not supported (not sure if this PR addresses this):

```ts
export const { x } = y
```

schiller-manuel:
closing in favor of https://github.com/TanStack/router/pull/7167

PR #4674, opened 2025-07-16
fix(react-router): preserve ref type in createLink with required interface properties

### LINKED ISSUE #4673: `createLink` breaks ref type when using interface with required properties

Router

### Describe the bug

Currently, when creating a link using the `createLink` utility function and passing an interface with the required property, type of the `ref` will be broken (resolves as `never`)

https://www.typescriptlang.org/play/?ts=5.8.3#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQFCiSxwDeculyMWAMsAHYDWcAL5wiJcgAEYyXqmm5+Aenb4AtMQCunKLTq9tBPFjgAhNMFw8BABWJh0WAB6deAE3TY8MAHQBBXrgAFtAAEgAqALJcvjAwUMAARlpYqAA84VH+QdAAogA2WCBY+gB8zDQAkE5xyABccHLxvADmANw0QjQ0uBCy8GaoFlb8AMIkkLzF8AC8mFQ+BNAA7shQrtgE6ZHRAcFQ+YVTADSm5pZ8-LYQ9iUAFDRwcLdgdqgnlAQAlHDTZUwPjzglBgGigvCeAMBcFSyCBWAI0yYHxETG8aJe11QIlweTQqAAcsgioiyAk8hAFHAwI5VABmKkAT1UACY4JxnKoyRosKoAOwABn5ZBEihKkLgnwBQiONEl3V6cjgI3YnFcwzG4F6Ux+rBV3AutwGQwuGomUzlTgY8B6fSVCo0RSgOtu31+5QqNsVHx1nnw3g0qCwGy2mV2uQKRX0cAAPnBeBo8nk7vHE592pVgaDwallThVerxlqox9Eci2RBpgAiRSVuDVKDIKu10UdLowBlgYwbMId4yzX0+U1FmBXeypdudiAEJV6tUmwuTUoAbQo8LIAF12opFFCAHoAfiAA

1. Open Typescript Playground repro
2. See `ref` typing error in `Consumer` component
3. Also the type of `RefType` will be `React.Ref<never> | undefined`

### Expected behavior

It's possible to use ref of type `React.useRef<HTMLAnchorElement | null>(null)` and the type of `RefType` will be `React.Ref<HTMLAnchorElement> | undefined`

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.127.9
- OS: Windows
- Browser: Zen (Firefox based)
- Browser Version: 1.14.5b (140.0.4)
- Bundler: vite
- Bundler Version: 7.0.4

### Additional context

_No response_

kasmacioma:
Fixes #4673

I a bit simplified the `UseLinkReactProps` to make more use of typing from React (and this simplification just fixes the bug)
Also added a test to check that the bug is no longer reproduced

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -400,9 +400,12 @@ const composeHandlers =
     })
   }
 
-type UseLinkReactProps<TComp> = TComp extends React.ElementType
-  ? React.ComponentPropsWithRef<TComp>
-  : never
+type UseLinkReactProps<TComp> = TComp extends keyof React.JSX.IntrinsicElements
+  ? React.JSX.IntrinsicElements[TComp]
+  : TComp extends React.ComponentType<any>
+    ? React.ComponentPropsWithoutRef<TComp> &
+        React.RefAttributes<React.ComponentRef<TComp>>
+    : never
 
 export type UseLinkPropsOptions<
   TRouter extends AnyRouter = RegisteredRouter,
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -400,18 +400,9 @@ const composeHandlers =
     })
   }
 
-type UseLinkReactProps<TComp> = TComp extends keyof React.JSX.IntrinsicElements
-  ? React.JSX.IntrinsicElements[TComp]
-  : React.PropsWithoutRef<
-      TComp extends React.ComponentType<infer TProps> ? TProps : never
-    > &
-      React.RefAttributes<
-        TComp extends
-          | React.FC<{ ref: React.Ref<infer TRef> }>
-          | React.Component<{ ref: React.Ref<infer TRef> }>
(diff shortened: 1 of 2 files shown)


Sheraff:
I hope the type this PR deletes isn't covering some obscure edge-cases! But your changes make sense to me, thank you for the fix.

Would you mind fixing the 2 uses of `React.LegacyRef<never>` in `packages/react-router/tests/link.test-d.tsx`? I think this new type broke them.

kasmacioma:
Oops, my bad, forgot to run the tests

While fixing the tests, I noticed, that my solution doesn't allow to set `key` on `activeProps`. Fixed this point, in the new solution `key` can be passed in. Also, the new solution (which similar to the current functionality, but uses React's built-in types) passes the current tests without having to change them

PR #4676, opened 2025-07-16
fix: assign random ports for e2e tests

nlynzaad:
This is a follow on from #4653

This PR resolves flaky E2E tests caused by port collisions when running tests in parallel.

Instead of deriving ports from package names (which can collide), this new approach assigns a random, available port to each test server instance.

Key Changes:

- Random Port Allocation: Each test server (Playwright, dummy servers) gets a unique, random port at startup.
- File-based Sync: The server writes its assigned port to a port.txt file in its test project directory, this is kept unique through using the package name or another input string for the file name.
- Test Runner Sync: The test runner reads the port from that file to build the correct URL.
- Automatic Cleanup: All port files are deleted at the start of each test:e2e run to ensure fresh ports.

Local parallel runs (up to 10) have shown this to be stable and free of collisions.


nlynzaad:
> Does get `get-port-please` guarantee that the port it returns will be free?

yes get-port-please checks the host for available ports before returning it. [get-port-please](https://github.com/unjs/get-port-please?tab=readme-ov-file#host)

> * If so, why are we writing the `PORT` to disk? Wouldn't it be better of the tests do not occupy the same port?

it seems playwright reads its config file once for the webserver startup and then again when it starts each test to get the base url. by not fixing the port somewhere the ports are then different. for example, using the below for derive port:

```js
export async function derivePort(input: string): Promise<number> {
  const portNumber = await getRandomPort()
  console.info(`Mapped "${input}" to port ${portNumber}`)
  return portNumber
}
```
results in:

```bash
[WebServer] > tanstack-router-e2e-react-basic@ serve /home/nicol/Documents/programming/PRs/router/e2e/react-router/basic
[WebServer] > vite preview --port 39609
[WebServer] 
[WebServer]   ➜  Local:   http://localhost:39609/
[WebServer]   ➜  Network: use --host to expose


Running 23 tests using 1 worker
Mapped "tanstack-router-e2e-react-basic" to port 43335
  1) [chromium] › tests/app.spec.ts:7:1 › Navigating to a post page ────────────────────────────────

    Error: page.goto: net::ERR_CONNECTION_REFUSED at http://localhost:43335/

```

> * If not, then isn't this approach also prone to collisions?

From a theoretical standpoint the current implementation already brings enough uniqueness to each derived port and we should not be seeing any collisions at all. The inability to reliably recreate this makes it very difficult to ascertain with 100% certainty what the cause is. The only reasons I can think these might still be happening is one of the following:

1. We have clashing package names (I can't find any of these cases)
2. the derive port calculation is resulting in duplicated ports. from testing across 65'000 random input strings I got about 63% unique ports issued, when dropping this down to 30'000 we are sitting with about 90% uniqueness and testing on current package names (thanks @schiller-manuel) we did not find any collisions on the ports. So, this doesn't seem to be the cause
3. while moving from one test to another, in the same package, playwright is shutting down the one dev server and starting another within milliseconds while the port is still reserved at the OS level. This seems to be most likely scenario as it would explain the intermittent nature. This could probably be addressed adding a miniscule delay in derivePort of say 10ms.

This PR address all 3 of these by:

1. ensuring the port is unique per run per package. even if package names are duplicated the port saved within the package will be unique to the specific package.
2. ports are randomly selected from available ports and hence unless we saturate 65`000 we should not be running into any collisions.
3. the nature of accessing the file, reading the file and/or creating a new file with a random port that is queried from the underlying OS means that a) if a port is still reserved and a new port is requested we would get a new unused one, b) if we reuse the pre-allocated one the time delay (we really only need nanoseconds) is enough to allow the old server to spin down and the port to be released.

If nothing else this takes points 1 and 2 of the table and should this keep on happening, we can focus on point 3.

SeanCassiere (approved):
🚀 thank you!

PR #4677, opened 2025-07-17
docs: add more details for webpack config

situ2001:
I found that webpack dev server will response me with 404 status or blank screen, when:

1. use nest route, but without `output.publicPath="/"`
2. reload the page when url has router path, for example `foo.com/about`, if without configuring `historyApiFallback`

Thus, I add these tips to docs.

Also, this PR comes with a modified demo so you can try it locally.


SeanCassiere (changes requested):
Could you also copy one of the existing test workflows from the Vite quickstart in the e2e folder too please?

It'd be best if we tested this in our CI pipeline.

Edit: Also, you may want to setup your editor to pickup our prettier config from the root.

LadyBluenotes:
Thanks for documenting this. The Webpack quickstart already has the needed settings, but the guide and test coverage need a fresh update against current main.

I’ve opened #7786 to track that work, if you want to port/continue your work there :)

PR #4678, opened 2025-07-17
docs: custom Vite Plugin

max-programming:
The quickstart page in the docs did not have the updated way of handling react vite plugin


nlynzaad:
Could we also possibly make note of the fact that this change is currently only needed to opt in to the new and upcoming behaviour. It can be very similar to @schiller-manuel's twitter post:

Soon, TanStack Start will stop auto-configuring React/Solid Vite plugins.
You’ll get full control - choose @vitejs/plugin-react, @vitejs/plugin-react-oxc, etc.

Want that power now?
Set customViteReactPlugin: true and you're all set!

max-programming:
Yeah that sounds like a good idea @nlynzaad . I think I'll put up a blockquote for that

max-programming:
Updated solidjs docs @SeanCassiere

docs/start/framework/solid/build-from-scratch.md
```diff
@@ -85,17 +85,24 @@ We'll then update our `package.json` to use Vite's CLI and set `"type": "module"
 
 Then configure TanStack Start's Vite plugin in `vite.config.ts`:
 
+> TanStack Start will stop auto-configuring React/Solid Vite plugins. You’ll get full control - choose `vite-plugin-solid`. Set `customViteSolidPlugin: true` to opt in to this feature right now!
```
SeanCassiere, line 89: You may want to use GFM alert for better styling here

SeanCassiere:
```suggestion
> [!NOTE]
> TanStack Start will stop auto-configuring React/Solid Vite plugins. You’ll get full control - choose `vite-plugin-solid`. Set `customViteSolidPlugin: true` to opt in to this feature right now!
```

SeanCassiere (approved):

SeanCassiere:
🚀

shaundavin13:
Perhaps we should keep the warning in case people who update the library suddenly encounter "React is not defined" issue. After all, there is no changelog.

PR #4691, opened 2025-07-17
fix(router-core): change error type in ErrorComponentProps

### LINKED ISSUE #4686: ErrorComponentProps#error incorrectly models the error type

Router

### Describe the bug

The [`ErrorComponentProps`](https://github.com/TanStack/router/blob/v1.128.1/packages/router-core/src/route.ts#L1320) type, that replaces the deprecated [`ErrorRouteProps`](https://github.com/TanStack/router/blob/v1.128.1/packages/router-core/src/route.ts#L1314), incorrectly models the [`error`](https://github.com/TanStack/router/blob/v1.128.1/packages/router-core/src/route.ts#L1321) as an `Error` .

In our production app, our fetch behaviour does not throw an `Error`, but instead a detailed rich object that consumers can dig into.

In the simplest example, a route's `loader` could throw a primitive data type such as a `string` instead of an `Error`, however the props of the  associated `errorComponent` (modelled as `ErrorComponentProps`) presume that the `error` will always be an `Error`.

https://stackblitz.com/edit/github-gk2uhnxn?file=src%2Froutes%2Findex.tsx

Throw a non-`Error` data type in a `loader`, for example `throw 'hello'`.

Observe that the props of the  loader's `errorComponent` incorrectly model the `error` prop as an `Error` type, even though we called `throw` with a typeof `string`.

### Expected behavior

Preferably:

```diff
- export type ErrorComponentProps = {
-  error: Error
+ export type ErrorComponentProps<E> = {
+  error: E
  info?: { componentStack: string }
  reset: () => void
}
```

or at least:

```diff
- export type ErrorComponentProps = {
-  error: Error
+ export type ErrorComponentProps = {
+  error: unknown
  info?: { componentStack: string }
  reset: () => void
}
```

### Screenshots or Videos

Image

### Platform

@tanstack/react-router@1.127.9

leesb971204:
fixes #4686


packages/router-core/src/route.ts
```diff
@@ -1318,7 +1318,7 @@ export type ErrorRouteProps = {
 }
 
 export type ErrorComponentProps = {
-  error: Error
+  error: unknown
```
leesb971204: If we choose an approach similar to [AxiosError](https://github.com/axios/axios/blob/8517aa16f8d082fc1d5309c642220fa736159110/index.d.ts#L405C1-L405C60), which extends the `Error` class, it could help minimize changes to existing code.
```tsx
export type ErrorComponentProps<TError extends Error> = {
  error: TError
```

I’ll address the remaining type errors after this approach has been reviewed.

michaelbull: Given react-query parameterises the error type, which we are using, it would be nice to parameterise it with the solution here as well.

The alternative of making it `unknown` means that I would have to (a) narrow the type manually, (b) use the `declare module '@tanstack/router-core' { ... }` trick to override the type manually.

The generic approach seems much closer to the "react-query" way of doing things.

leesb971204: @michaelbull 

Thank you for the feedback!
Could you please share more details about the approach you mentioned from React Query? Perhaps an issue or PR?

From what I can see in the official [React Query documentation](https://tanstack.com/query/latest/docs/framework/react/typescript#registering-a-global-error), it still seems to be using `declare` for now.

michaelbull: @leesb971204 

Doing it globally in the way the docs recommend is not how we have approached it, as that seems unsafe - if you had two distinct `queryFn`s that had distinct error types, you could not model it in such a way that your call to `useQuery` would know exactly what the error type is based on what `queryFn` you called (just that its either one or the other).

Our approach is simple: we implement a wrapper that extends react-query's [`queryOptions`](https://github.com/TanStack/query/blob/3bcaa417ac302a93d17c4287d518121bb77d1b26/packages/vue-query/src/queryOptions.ts#L18), but explicitly sets the `TError` generic type to our internal company error type.

All of our queries use our custom `queryOptions` wrapper, for example:

```ts
interface MyBusinessErrorType {
    errorCode: number;
    errorMsg: string;
}

function exampleQueryOptions() {
     return myBusinessQueryOptions({
          queryKey: ...,
          queryFn: ....,
    })
}

function Example() {
     const { error } = useQuery(exampleBusinessQueryOptions);

     if (isError) {
         const { errorCode, errorMsg } = error;
        return <>error {errorCode} : {erorrMsg}</>
    } ....
}
```

I would like to do the exact same, such that we could have:

```ts
interface MyBusinessErrorType {
    errorCode: number;
    errorMsg: string;
}

function MyBusinessErrorComponent(props: ErrorComponentProps<MyBusinessErrorType>) {
     const { error } = props;
     const { errorCode, errorMsg } = error;

    return <>error {errorCode} : {erorrMsg}</>
}
```


Of course at that point is our responsibility to ensure that anything thrown in `beforeLoad` is a type of `MyBusinessErrorType` - however that is already abstracted in all of our query functions (such that they only ever throw one of those rich detailed objects, they never throw blanket `Error` instances).

packages/router-core/src/route.ts
```diff
@@ -1317,8 +1317,8 @@ export type ErrorRouteProps = {
   reset: () => void
 }
 
-export type ErrorComponentProps = {
-  error: Error
+export type ErrorComponentProps<TError = Error> = {
+  error: TError
```
leesb971204, line 1321: This reflects @michaelbull ’s suggestion.

The change respects the diversity of error objects returned from the backend while setting `Error` as the default type to ensure compatibility with existing code.

leesb971204, line 1321: This reflects @michaelbull ’s suggestion.

The change respects the diversity of error objects returned from the backend while setting `Error` as the default type to ensure compatibility with existing code.

schiller-manuel:
thanks for this PR!

can you please update the docs 

docs/router/framework/react/api/router/errorComponentComponent.md

leesb971204:
> thanks for this PR!
> 
> can you please update the docs
> 
> docs/router/framework/react/api/router/errorComponentComponent.md

I updated the type of the `error`, and also added a description for `info`, which seemed to be missing.

PR #4702, opened 2025-07-18
feat(router-core) path: pre-compile encodePathParam during router initialization

Sheraff:
When NOT using the `pathParamsAllowedCharacters` option, this implementation is identical to the previous one. Both simply use `encodeURIComponent`. The bench comparison is a coin flip.
```
 ✓  @tanstack/router-core  tests/temp.bench.ts > compare with no replacements 2892ms
     name            hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · old   8,305,805.98  0.0000  0.0291  0.0001  0.0001  0.0002  0.0002  0.0002  ±0.03%  4152903
   · new   8,310,709.30  0.0000  0.0286  0.0001  0.0001  0.0002  0.0002  0.0002  ±0.04%  4155355   fastest
```

When using `pathParamsAllowedCharacters`, this implementation "pre-compiles" a `Function` during the router initialization. This makes it consistently faster. Though it's only a ~10% speed-up, this method is called *a lot*, so any gain is beneficial.
```
 ✓  @tanstack/router-core  tests/temp.bench.ts > compare with many replacements 1611ms
     name            hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · old   1,365,878.18  0.0006  1.1122  0.0007  0.0007  0.0010  0.0010  0.0018  ±0.55%   682940
   · new   1,500,735.70  0.0005  0.2715  0.0007  0.0007  0.0009  0.0009  0.0013  ±0.11%   750368   fastest
```

This is technically a **breaking change** because `interpolatePath` is part of the public API surface.

### DIFF
--- a/packages/router-core/src/path.ts
+++ b/packages/router-core/src/path.ts
@@ -323,8 +323,13 @@ interface InterpolatePathOptions {
   params: Record<string, unknown>
   leaveWildcards?: boolean
   leaveParams?: boolean
-  // Map of encoded chars to decoded chars (e.g. '%40' -> '@') that should remain decoded in path params
-  decodeCharMap?: Map<string, string>
+  /**
+   * How to encode strings so they can be used inside a URL path. Defaults to `encodeURIComponent`.
+   *
+   * The router's implementation is exposed as `router.encodePathParam`. Using `pathParamsAllowedCharacters`
+   * will result in the router's `encodePathParam` keeping some characters un-encoded (e.g. '%40' -> '@').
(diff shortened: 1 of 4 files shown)


Sheraff (commented):
I'm not 100% convinced the small perf gain, only when using a specific option, is worth a breaking change.

schiller-manuel:
probably not worth the added complexity for this small gain.
let's try to find the big perf issues first

PR #4714, opened 2025-07-19
refactor(router-plugin): Build-time route matching

Sheraff:
Performance of `matchByPath` (`router-core` > `path`) can be improved by generating a "pre-compiled" route matching function at build-time


Sheraff:
We built something better instead https://github.com/TanStack/router/pull/5722

LadyBluenotes:
Closing this since I think it's superseded by #5722, which implemented the route-matching improvement with the segment-tree approach and has since merged into main :)

Issue #4696, opened 2025-07-18
Context is missing from parent loader when child's validateSearch fails


Dakkers:
### Which project does this relate to?

Router

### Describe the bug

When a parent route (in my case, the root route) has a `beforeLoad` parameter, the parent route's `loader`'s `context` is missing expected attributes when a child route's `validateSearch` fails.

[StackBlitz project](https://stackblitz.com/edit/vitejs-vite-pmq9xxrp?file=src%2Froutes%2F__root.tsx)

[StackBlitz container](https://vitejsvitepmq9xxrp-by02--5173--96435430.local-credentialless.webcontainer.io/)

### Your Example Website or App

https://vitejsvitepmq9xxrp-by02--5173--96435430.local-credentialless.webcontainer.io/

### Steps to Reproduce the Bug or Issue

1. Go to the example site
2. Go to the dashboard link
3. Ensure you can see the navbar (this is working)
4. Remove the query parameters from the URL and press enter (to navigate directly to `/dashboard` with no query params)
5. You will not see the Navbar anymore 🐛 

### Expected behavior

I expected the parent route's `loader` to have the values of the context that were returned by the same route's `beforeLoad` but the attributes are missing.

### Screenshots or Videos

https://github.com/user-attachments/assets/1d778b33-021c-459f-8b17-7807a35aba96

### Platform

- Router / Start Version: 1.128.3
- OS: MacOS
- Browser: Chrome
- Browser Version: 138.0.7204.157 (Official Build) (arm64)
- Bundler: Vite
- Bundler Version: 7.0.5

### Additional context

Related issues, PRs, discussions:
- https://github.com/TanStack/router/issues/3578
- https://github.com/TanStack/router/issues/3430
- https://github.com/TanStack/router/pull/4439
- https://github.com/TanStack/router/pull/4433
- https://github.com/TanStack/router/issues/4151
- https://github.com/TanStack/router/discussions/1737

Dakkers:
My super awesome coworker @danconnell found out that this behaviour doesn't occur in 1.121.14, but it does on the Stackblitz instance which is 1.128.3, so I will do a bit more work to figure out when it started happening again.

Dakkers:
It seems that the bug started occurring again in version **1.123.2**. (1.123.0 had no issues.) 

I would assume it's because of this PR because it makes modifications to `beforeLoad` / `loader` logic: https://github.com/TanStack/router/pull/4550

schiller-manuel:
[...]

PR #4726, opened 2025-07-20
test: added (currently failing) unit test to reproduce regression introduced in 1.123.2

Dakkers:
As requested by @schiller-manuel for this reported issue: https://github.com/TanStack/router/issues/4696

## Summary by CodeRabbit

- New Features
  - Added support for memory history–based routing, enabling in-memory navigation scenarios without touching the browser history.
- Tests
  - Added a comprehensive reproduction test covering loader-provided context, authenticated navigation visibility, suspense/pending states, and query handling across routes.


packages/react-router/tests/loaders.test.tsx
```diff
@@ -729,3 +730,89 @@ test('clears pendingTimeout when match resolves', async () => {
...
+      }
+    },
+    component: () => <h1>Dashboard</h1>,
+  })
+
+  const routeTree = rootRoute.addChildren([indexRoute, dashboardRoute])
+  const router = createRouter({ routeTree })
+
+  render(<RouterProvider router={router} />)
+  await act(async () => {
```
schiller-manuel: replace 

```tsx
await act(async () => {

    /* no-op */
  })
  ```
  
  by
  
  ```tsx
  await router.load()
```

packages/react-router/tests/loaders.test.tsx
```diff
@@ -729,3 +730,89 @@ test('clears pendingTimeout when match resolves', async () => {
...
+    /* no-op */
+  })
+  expect(screen.getByRole('heading', { name: 'Index' })).toBeInTheDocument()
```
schiller-manuel: please use getByTestId and use `data-testid="..."`
we try not to use getByRole in new tests anymore

packages/react-router/tests/loaders.test.tsx
schiller-manuel: this applies to all occurences

packages/react-router/tests/loaders.test.tsx
Dakkers: got it – why's that? (just curious)

schiller-manuel: it allows to change text, tags etc without having to update the getByRole calls
you just need to make sure the test-id is correctly specified

packages/react-router/tests/loaders.test.tsx
Dakkers: for some reason when I change it to `await router.load()` for the index route, the test fails there – it doesn't render the navbar as it did with the `act` call.

image

when I use `await router.load()` for the dashboard route part, it doesn't render anything at all:

image

Dakkers:
@schiller-manuel Is it possible for me to re-run the CI for this PR? It's possible the bug has been fixed

Dakkers:
@Sheraff do you think I can close this PR, based off the changes you made in this PR? https://github.com/TanStack/router/pull/4993

PR #4727, opened 2025-07-21
fix(router-core): update ErrorComponentProps to not support generic error type

### LINKED ISSUE #4686: ErrorComponentProps#error incorrectly models the error type

leesb971204:
fixes #4686 
more of #4691 

I added a generic to `ErrorComponentProps` to allow specifying the error type,
but since the generic wasn’t passed through in the `ErrorRouteComponent` type, I added explicit typing there as well.

### DIFF
--- a/packages/react-router/src/route.tsx
+++ b/packages/react-router/src/route.tsx
@@ -49,7 +49,7 @@ import type { LinkComponentRoute } from './link'
 declare module '@tanstack/router-core' {
   export interface UpdatableRouteOptionsExtensions {
     component?: RouteComponent
-    errorComponent?: false | null | ErrorRouteComponent<any>
+    errorComponent?: false | null | ErrorRouteComponent
     notFoundComponent?: NotFoundRouteComponent
     pendingComponent?: RouteComponent
   }
@@ -555,9 +555,7 @@ export type AsyncRouteComponent<TProps> = RouteTypes<TProps>['component'] & {
 
 export type RouteComponent = AsyncRouteComponent<{}>
 
-export type ErrorRouteComponent<TError = Error> = AsyncRouteComponent<
-  ErrorComponentProps<TError>
->
+export type ErrorRouteComponent = AsyncRouteComponent<ErrorComponentProps>
 
 export type NotFoundRouteComponent = RouteTypes<NotFoundRouteProps>['component']
 
--- a/packages/react-router/src/route.tsx
+++ b/packages/react-router/src/route.tsx
@@ -49,7 +49,7 @@ import type { LinkComponentRoute } from './link'
 declare module '@tanstack/router-core' {
   export interface UpdatableRouteOptionsExtensions {
     component?: RouteComponent
-    errorComponent?: false | null | ErrorRouteComponent
+    errorComponent?: false | null | ErrorRouteComponent<any>
     notFoundComponent?: NotFoundRouteComponent
     pendingComponent?: RouteComponent
   }
@@ -555,7 +555,9 @@ export type AsyncRouteComponent<TProps> = RouteTypes<TProps>['component'] & {
 
 export type RouteComponent = AsyncRouteComponent<{}>
 
-export type ErrorRouteComponent = AsyncRouteComponent<ErrorComponentProps>
+export type ErrorRouteComponent<TError = Error> = AsyncRouteComponent<
+  ErrorComponentProps<TError>
+>
 
 export type NotFoundRouteComponent = RouteTypes<NotFoundRouteProps>['component']
 
--- a/packages/router-core/src/route.ts
+++ b/packages/router-core/src/route.ts
@@ -1317,8 +1317,8 @@ export type ErrorRouteProps = {
   reset: () => void
 }
 
-export type ErrorComponentProps<TError = Error> = {
-  error: TError
+export type ErrorComponentProps = {
+  error: Error
   info?: { componentStack: string }
   reset: () => void
 }
--- a/docs/router/framework/react/api/router/errorComponentComponent.md
+++ b/docs/router/framework/react/api/router/errorComponentComponent.md
@@ -11,7 +11,7 @@ The `ErrorComponent` component accepts the following props:
 
 ### `props.error` prop
 
-- Type: `TError` (defaults to `Error`)
+- Type: `Error`
 - The error that was thrown by the component's children
 
 ### `props.info` prop
--- a/packages/solid-router/src/route.tsx
+++ b/packages/solid-router/src/route.tsx
@@ -49,7 +49,7 @@ import type { LinkComponentRoute } from './link'
 declare module '@tanstack/router-core' {
   export interface UpdatableRouteOptionsExtensions {
     component?: RouteComponent
-    errorComponent?: false | null | ErrorRouteComponent
+    errorComponent?: false | null | ErrorRouteComponent<any>
     notFoundComponent?: NotFoundRouteComponent
     pendingComponent?: RouteComponent
   }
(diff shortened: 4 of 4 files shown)


packages/react-router/src/route.tsx
```diff
@@ -49,7 +49,7 @@ import type { LinkComponentRoute } from './link'
 declare module '@tanstack/router-core' {
   export interface UpdatableRouteOptionsExtensions {
     component?: RouteComponent
-    errorComponent?: false | null | ErrorRouteComponent
+    errorComponent?: false | null | ErrorRouteComponent<any>
```
michaelbull: Is there a reason we wouldn't add the `TError` generic?

```ts
export interface UpdatableRouteOptionsExtensions<TError = Error> {
    errorComponent?: false | null | ErrorRouteComponent<TError>
```

leesb971204: > Is there a reason we wouldn't add the `TError` generic?
> 
> ```ts
> export interface UpdatableRouteOptionsExtensions<TError = Error> {
>     errorComponent?: false | null | ErrorRouteComponent<TError>
> ```

If we make this change, wouldn’t it cause a type error in the `errorComponent`?
스크린샷 2025-07-21 오후 9 18 39

leesb971204: @michaelbull 

Do you have any additional ideas regarding this?

michaelbull: Not really sure to be honest. Its unclear to me why that would cause a compiler error, and I explicitly use code-based routing (createRoute) not file based. I suspect you would have to chase the generic error type onto `UpdatableRouteOptionsExtensions` as well, but now I'm starting to second guess following this through.

I would defer to @schiller-manuel.

leesb971204: @michaelbull 

`UpdatableRouteOptionsExtensions` is not explicitly used in user code.
So if we declare it as `UpdatableRouteOptionsExtensions<TError = Error>`, the `ErrorRouteComponent` will always be inferred with `Error` as its generic type.
As a result, if the user specifies a type like `ErrorComponentProps<string>`, it will cause a type error—since `Error` is not compatible with `string`.

In conclusion, my personal opinion is that `ErrorRouteComponent<any>` is the appropriate way to declare it. What are your thoughts on this?

leesb971204: > @michaelbull
> 
> `UpdatableRouteOptionsExtensions` is not explicitly used in user code. So if we declare it as `UpdatableRouteOptionsExtensions<TError = Error>`, the `ErrorRouteComponent` will always be inferred with `Error` as its generic type. As a result, if the user specifies a type like `ErrorComponentProps<string>`, it will cause a type error—since `Error` is not compatible with `string`.
> 
> In conclusion, my personal opinion is that `ErrorRouteComponent<any>` is the appropriate way to declare it. What are your thoughts on this?

스크린샷 2025-07-30 오전 9 54 37

Additional note:
The same type error also occurs with `createRoute`.

leesb971204:
is there any review on this PR?

packages/react-router/src/route.tsx
leesb971204: I don’t think this approach can fundamentally solve the problem, as it changes the errorComponent’s props type to `any`.

leesb971204:
In my view, changing the type of `ErrorComponentProps.error` from `unknown` to `TError` was not the right choice. 
In fact, `ErrorRouteComponent` is defined via `UpdatableRouteOptionsExtensions`, so `ErrorComponentProps` has no way to know the concrete `error` type that the user defines inline.

leesb971204:
Even in TanStack Query, using `queryOptions` does not automatically infer the `error` type returned by `queryFn`.
You either specify the `error` type via `useQuery` generics or use a `declare` augmentation. Please refer to [this](https://tanstack.com/query/latest/docs/framework/react/typescript#typing-the-error-field).

If we follow the `useQuery` approach, it could be implemented like the example below—but realistically, no one is likely to use it this way.😅
```typescript
type CustomError = {
        status: number
}

export const Route = createRoute<{
	...
        CustomError,
	...
}>('/route')({
        ...
});

``` 

Conclusion: The TanStack Query approach cannot be applied here, so the user must either perform error type narrowing themselves or declare the `error` type via declare.

packages/router-core/src/route.ts
```diff
@@ -1317,8 +1317,8 @@ export type ErrorRouteProps = {
   reset: () => void
 }
 
-export type ErrorComponentProps<TError = Error> = {
-  error: TError
+export type ErrorComponentProps = {
+  error: Error
```
leesb971204, line 1321: Since the scope of the `errorComponent` covers both errors from the `loader` and errors from the component itself, using the `Error` type is appropriate

michaelbull: This seems to undo the work of making it handle throwing of non-errors. What if my loader calls some third-party API that internally does `throw "Some string"` or `throw 50`? Loaders can throw more than just `Error`s, and in some situations the user is not in control of what could be thrown.

leesb971204: @michaelbull 

That’s right. But since in most cases the error is returned as `Error`, I think it makes sense to set the type to `Error` by default, and if needed, the user can manually override the error type using something like declare. Please refer to the [link](https://github.com/TanStack/query/pull/9575#discussion_r2282056463).

binajmen: is that the definite point of view of tanstack? i'm as well throwing something different that an `Error` and it is sad to not have the possibility to properly type the error for the ErrorComponent.

LadyBluenotes:
Thanks for continuing to work through this. I opened #7788 to capture the error-boundary type contract.

I think this PR should move toward `unknown` as the safe baseline, with users narrowing inside their error component. The new issue covers the longer-term typed-error design if we can propagate that type through route configuration and all framework adapters.

michaelbull:
`unknown` is definitely a better option than presuming they are `Error`s, as that at least forces the user to check it. It was one of the two proposals I made in the original issue, so would appreciate that as a solution.

Issue #4695, opened 2025-07-18
Add proper safe types for return value of `Router#matchRoutes`


mutewinter:
### Which project does this relate to?

Router

### Describe the bug

`Router#matchRoutes` returns `AnyRouteMatch`. Instead, it should return typed routes for the user's application, e.g. `MakeRouteMatchUnion[]`.

See [Discord discussion](https://discord.com/channels/719702312431386674/1007702008448426065/threads/1394666372587458580).

### Your Example Website or App

https://stackblitz.com/github/tanstack/router/tree/main/examples/react/quickstart-file-based?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

None needed.

### Expected behavior

Proper return type.

### Platform

- Router / Start Version: [e.g. 1.121.0]
- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Browser Version: [e.g. 91.1]
- Bundler: [e.g. vite]
- Bundler Version: [e.g. 7.0.0]

PR #4730, opened 2025-07-21
fix(router-core): update MatchRoutesFn signature and proper deprecate old usage

### LINKED ISSUE #4694: Add proper deprecation for `Router#matchRoutes`

Router

### Describe the bug

https://github.com/TanStack/router/blob/9d9554c43ff9df605e41a8d48c4912abef02ebce/packages/router-core/src/router.ts#L1024-L1032

The way this deprecation is defined means that any usage of this API will show as deprecated.

Image

See [Discord discussion](https://discord.com/channels/719702312431386674/1007702008448426065/threads/1394666372587458580).

1. Use `Router#matchRoutes`
2. Get a deprecation warning.

### Expected behavior

Only show the deprecation warning for the invalid method signature.

leesb971204:
fixes #4694, #4695

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -625,9 +625,12 @@ export type StartTransitionFn = (fn: () => void) => void
 export interface MatchRoutesFn {
   (
     pathname: string,
-    locationSearch: AnySchema,
+    locationSearch?: AnySchema,
     opts?: MatchRoutesOpts,
-  ): Array<AnyRouteMatch>
+  ): Array<MakeRouteMatchUnion>
+  /**
+   * @deprecated use the following signature instead
+   */
   (next: ParsedLocation, opts?: MatchRoutesOpts): Array<AnyRouteMatch>
   (
     pathnameOrNext: string | ParsedLocation,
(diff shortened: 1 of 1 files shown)


packages/router-core/src/router.ts
```diff
@@ -625,9 +625,12 @@ export type StartTransitionFn = (fn: () => void) => void
 export interface MatchRoutesFn {
   (
     pathname: string,
-    locationSearch: AnySchema,
+    locationSearch?: AnySchema,
```
leesb971204, line 628: I’m not sure if the `locationSearch` property needs to be required.
The original `matchRoutes` function worked fine even without `locationSearch`.
I’d appreciate it if you could share your thoughts on this.

packages/router-core/src/router.ts
```diff
@@ -625,9 +625,12 @@ export type StartTransitionFn = (fn: () => void) => void
...
     opts?: MatchRoutesOpts,
-  ): Array<AnyRouteMatch>
+  ): Array<MakeRouteMatchUnion>
+  /**
+   * @deprecated use the following signature instead
+   */
   (next: ParsedLocation, opts?: MatchRoutesOpts): Array<AnyRouteMatch>
```
leesb971204, line 634: In router, adds a deprecation notice to usages of this format.

스크린샷 2025-07-21 오후 2 59 45

Otherwise, the deprecation notice will not be shown.

스크린샷 2025-07-21 오후 3 00 33

packages/router-core/src/router.ts
```diff
@@ -625,9 +625,12 @@ export type StartTransitionFn = (fn: () => void) => void
 export interface MatchRoutesFn {
   (
     pathname: string,
-    locationSearch: AnySchema,
+    locationSearch?: AnySchema,
     opts?: MatchRoutesOpts,
-  ): Array<AnyRouteMatch>
+  ): Array<MakeRouteMatchUnion>
```
leesb971204, line 630: In the user’s code, this type is correctly inferred without any deprecation warning.

스크린샷 2025-07-21 오후 3 02 03

Issue #4737, opened 2025-07-21
Cannot call `next` multiple times on server function middleware


alvarosevilla95:
### Which project does this relate to?

Start

### Describe the bug

Calling `next` multiple times on a server function middleware (both client and server) does not work as expected. The additional calls just return undefined.

This prevents adding things like retries on certain error types as generic middleware functionality.

### Your Example Website or App

https://github.com/TanStack/router/pull/4738

### Steps to Reproduce the Bug or Issue

Run e2e test

### Expected behavior

Calling `next` multiple times reruns the underlying middlewares and server function.

### Platform

- Router / Start Version: 1.129.1

schiller-manuel:
multiple next() calls are not supported

PR #4738, opened 2025-07-21
test(e2e): add failing e2e test for server function middleware retries

alvarosevilla95:
Adds failing e2e test for https://github.com/TanStack/router/issues/4737


schiller-manuel:
would it make sense to return something else per invocation? so that we can count how often the function was called. like increasing a counter?

alvarosevilla95:
> would it make sense to return something else per invocation? so that we can count how often the function was called. like increasing a counter?

I thought of this but I found it surprisingly not trivial to do elegantly, as I don't think I can access the return value from `next` in the middleware itself.

I can add a global variable like this? But that is going to be fragile when called multiple times (hence the added condition to reset the count when it reaches 4...), I assume this is not what you're looking for, if you have a suggestion happy to apply it

alvarosevilla95:
Or I guess I just ignore the type error and use the context like this... hard to check if the test is well formed without the ability to run it properly :D

PR #4745, opened 2025-07-22
docs: fix incorrect import path in automatic code splitting guide

JustinMissmahl:


nlynzaad:
thanks

JustinMissmahl:
> thanks

I tried playing around with it a bit. It looks interesting until now, but the next main thing I am really interested in and am trying to fully understand at the moment is the `createServerFileRoute` function, which seems to be internally linked and used in the better-Start template for TanStack and Better-Auth to have a stable API endpoint for auth. It looks like the `createServerFileRoute` is linked in the docs and internally available, but I can't seem to be able to use it in my code. So, I am currently wrapping it with a separate Hono server. My question now is: is it planned to allow it to expose stable API endpoints using TanStack directly, or should the separate server stay the solution? Or am I maybe even missing something here? (It could definitely be that I am missing something. I might have been awake too long when I tried to play around with it.🤣)

akhileshrangani4 (changes requested):
mistakenly typed

PR #4746, opened 2025-07-22
fix(react-router): update error handling in MatchInner to access error data directly

### LINKED ISSUE #4744: notFound data incorrectly typed as unknown and redundantly nested in data.data

Router

### Describe the bug

The props.data object passed to defaultNotFoundComponent is typed as unknown. At runtime, this object contains the user-provided data from notFound({ data: ... }) nested one level deeper under data.data, alongside metadata like isNotFound and routeId. This incorrect typing requires unsafe any casts, and the redundant nesting makes data access unintuitive

 https://stackblitz.com/edit/tanstack-router-zzqpfdfg?file=src%2Fmain.tsx

1. In a loader, return a notFound response with some data (e.g., in posts.$postId.tsx)

```ts
// posts.$postId.tsx
export const loader = async () => {
  return notFound({
    data: 'Hello from loader', // Data can be any type
  });
};
```

2. In your defaultNotFoundComponent, try to access props.data

```ts
// defaultNotFoundComponent defined in your router setup
defaultNotFoundComponent={(props) => {
  // TypeScript type of 'props.data' is 'unknown'
  // At runtime, 'props.data' is an object like:
  // {
  //   data: 'Hello from loader', // This is the user-provided data
  //   isNotFound: true,
  //   routeId: '/posts/$postId'
  // }

  return (
    <div>
      <p>
        Raw props.data:{" "}
        <pre>{JSON.stringify(props.data, null, 2)}</pre>
      </p>
      <p>
        To access the string, forced to use:{" "}
        {(props.data as any)?.data || 'N/A'}
      </p>
    </div>
  );
}}
```

### Expected behavior

1. Correct Type: The props.data parameter in defaultNotFoundComponent should be explicitly typed (e.g., { data: T, isNotFound: true, routeId: string }) so that T reflects the type of data provided to notFound
2. Intuitive Data Access: The user-provided data T should be directly accessible without the redundant .data nesting (e.g., via props.data.value or props.data.payload, or if isNotFound and routeId are top-level props, props.data itself).

### Screenshots or Videos

_No response_

### Platform

- Bundler: [e.g. vite]

### Additional context

_No response_

leesb971204:
fixes #4744


packages/react-router/src/renderRouteNotFound.tsx
```diff
@@ -10,7 +10,7 @@ export function renderRouteNotFound(
 ) {
   if (!route.options.notFoundComponent) {
     if (router.options.defaultNotFoundComponent) {
-      return <router.options.defaultNotFoundComponent data={data} />
+      return <router.options.defaultNotFoundComponent {...data} />
```
leesb971204, line 13: This part causes the structure to become nested like `data: { data: ... }`.

packages/router-core/src/route.ts
```diff
@@ -1322,10 +1322,7 @@ export type ErrorComponentProps<TError = Error> = {
   info?: { componentStack: string }
   reset: () => void
 }
-export type NotFoundRouteProps = {
-  // TODO: Make sure this is `| null | undefined` (this is for global not-founds)
-  data: unknown
-}
+export type NotFoundRouteProps = Partial<NotFoundError>
```
leesb971204, line 1325: The props can contain not only `data` but also any value of type `NotFoundError`.

Since the `global not found` case passes `undefined`, we use `Partial` to account for the possibility that required fields might be added to `NotFoundError` in the future.

leesb971204:
@schiller-manuel 
@SeanCassiere 

Is there anything else I should do to get a review on this?

schiller-manuel:
thanks for this PR!
can you please 
- add some runtime tests for the various scenarios
- add a typetest maybe? not sure if that makes sense / is possible
- update docs
- add usage to the kitchen sink example

leesb971204:
> thanks for this PR! can you please
> 
> * add some runtime tests for the various scenarios
> * add a typetest maybe? not sure if that makes sense / is possible
> * update docs
> * add usage to the kitchen sink example

Thank you for the review!

Could you explain a bit more about the docs update you mentioned?
I couldn’t find any documentation related to the props of `notFoundComponent` or about `notFoundRouteProps`.
Would you like me to add something to the docs?

docs/router/framework/react/guide/not-found-errors.md
```diff
@@ -240,7 +240,7 @@ export const Route = createFileRoute('/posts/$postId')({
       })
     return { post }
   },
-  // `data: unknown` is passed to the component via the `data` option when calling `notFound`
+  // `data: any` is passed to the component via the `data` option when calling `notFound`
```
schiller-manuel: why is this `any` instead of `unknown`?

leesb971204: This is because the type of `notFoundRouteProps` was changed from `unknown` to `Partial<NotFoundError>`,
and the `data` in `NotFoundError` is currently typed as `any`.

schiller-manuel: can we change it there to unknown? to force people to add some runtime type checks?

leesb971204: do you mean it should be reverted the change?

leesb971204: As I menthioned [here](https://github.com/TanStack/router/pull/4746#discussion_r2222315720), `notFoundRouteProps`  can contain not only `data` but also any value of type `NotFoundError` like `routeId`.

or is it appropriate to convert the `data` type of `NotFoundError` to `unknown`?

leesb971204:
> thanks for this PR! can you please
> 
> * add some runtime tests for the various scenarios
> * add a typetest maybe? not sure if that makes sense / is possible
> * update docs
> * add usage to the kitchen sink example

When you have some time, could you take another look at this PR?
I believe I’ve added everything you mentioned.

PR #4752, opened 2025-07-22
refactor(router-core): add LRU cache in front of parsePathname

Sheraff:
This PR proposes a basic least-recently-used cache implementation (the performance of that data structure is about the same as a `Map` according to benchmark, but it uses more memory).

We can add caching to `parsePathname` now that what it returns is readonly (https://github.com/TanStack/router/pull/4705), which should improve the performance of this bottleneck.

The cache is set to a size of 1000. 

When benchmarking `matchPathname` with and without the cache, we notice a ~7x increase in throughput with the cache.
```
  name             hz     min     max    mean     p75     p99    p995    p999     rme  samples
· cached   13,846.82  0.0645  0.2253  0.0722  0.0729  0.0951  0.1216  0.1635  ±0.24%     6924
· original  1,800.37  0.5097  0.6939  0.5554  0.5607  0.6182  0.6481  0.6939  ±0.20%      901

7.69x faster than original
```

### DIFF
--- a/packages/router-core/src/lru-cache.ts
+++ b/packages/router-core/src/lru-cache.ts
@@ -1,0 +1,71 @@
+export function createLRUCache<T>(max: number) {
+  const cache = new Map<string, T>()
+  type Node = { before?: Node; after?: Node; key: string }
+  const doublyLinkedList = new Map<string, Node>()
+  let oldest: Node | undefined
+  let newest: Node | undefined
+
+  const touch = (entry: Node) => {
+    if (!entry.before) {
+      if (entry.after) {
+        entry.after.before = undefined
+        oldest = entry.after
+        entry.after = undefined
+      }
+      if (newest) {
+        entry.before = newest
+        newest.after = entry
+      }
+    } else {
+      entry.before.after = entry.after
+      if (entry.after) {
+        entry.after.before = entry.before
+        entry.after = undefined
+      }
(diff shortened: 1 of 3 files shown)


schiller-manuel:
does this have any implications when running on dev? e.g. do we need to clear the cache upon hmr?

also we should make the size configurable 
size 0 would disable the cache

Sheraff:
> does this have any implications when running on dev? e.g. do we need to clear the cache upon hmr?

I don't think so. We're parsing strings, using strings as keys. If the string changes, it bypasses the cache. But if it's the same string, it can return the same result, even after a HMR, without any issue.

Sheraff:
Serving a local build (not exactly the same performance profile as a staging build in the flamegraph above) I can see the following, around the same nav event:

Without this PR

Screenshot 2025-07-23 at 09 39 18


With this PR

Screenshot 2025-07-23 at 09 39 25

Sheraff:
> also we should make the size configurable
> size 0 would disable the cache

I wasn't sure about this. I can't see anyone playing with this setting unless they have **many** routes and they actually want to try *increasing* its size. But for people w/ a small number of routes, I can't imagine why you would disable it. Measuring memory in the browser is kinda annoying but based on vibes I'd imagine this to have a quite minimal memory footprint.

Issue #3811, opened 2025-03-19
head properties documentation missing


lukedukeus:
### Which project does this relate to?

Router

### Describe the bug

[This](https://tanstack.com/router/latest/docs/framework/react/guide/document-head-management) document mentions how to set meta attributes when creating a route, but the `head` property is not documented [here](https://tanstack.com/router/latest/docs/framework/react/api/router/RouteOptionsType), which I would expect.

### Your Example Website or App

N/A

### Steps to Reproduce the Bug or Issue

N/A

### Expected behavior

N/A

### Platform

N/A

leesb971204:
@SeanCassiere 

Would it be okay if I added this information to the `RouteOptionsType.md` file (`headers`, `head` `scripts`)?
Also, I noticed that some properties seem to be missing from `RouteMatchType.md`, and I’d be happy to add those as well!

SeanCassiere:
> Would it be okay if I added this information to the `RouteOptionsType.md`...

@leesb971204 yup! please do!

PR #4754, opened 2025-07-23
docs: Update RouteOptionsType, RouteMatchType

leesb971204:
relates #3811

### DIFF
--- a/docs/router/framework/react/api/router/RouteMatchType.md
+++ b/docs/router/framework/react/api/router/RouteMatchType.md
@@ -8,22 +8,40 @@ The `RouteMatch` type represents a route match in TanStack Router.
 ```tsx
 interface RouteMatch {
   id: string
-  routeId: string
+  routeId: TRouteId
+  fullPath: TFullPath
+  index: number
   pathname: string
-  params: Route['allParams']
-  status: 'pending' | 'success' | 'error'
-  isFetching: boolean
-  showPending: boolean
+  params: TAllParams
+  status: 'pending' | 'success' | 'error' | 'redirected' | 'notFound'
+  isFetching: false | 'beforeLoad' | 'loader'
   error: unknown
   paramsError: unknown
   searchError: unknown
   updatedAt: number
   loadPromise?: Promise<void>
-  loaderData?: Route['loaderData']
-  context: Route['allContext']
-  search: Route['fullSearchSchema']
-  fetchedAt: number
+  beforeLoadPromise?: Promise<void>
+  loaderPromise?: Promise<void>
+  loaderData?: TLoaderData
+  context: TAllContext
+  search: TFullSearchSchema
+  fetchCount: number
   abortController: AbortController
-  cause: 'enter' | 'stay'
+  cause: 'preload' | 'enter' | 'stay'
+  loaderDeps: TLoaderDeps
+  preload: boolean
+  invalid: boolean
+  headers?: Record<string, string>
+  globalNotFound?: boolean
+  staticData: StaticDataRouteOption
+  minPendingPromise?: Promise<void>
+  pendingTimeout?: ReturnType<typeof setTimeout>
+  ssr?: boolean | 'data-only'
+  displayPendingPromise?: Promise<void>
+  scripts?: unknown
+  links?: unknown
+  headScripts?: unknown
+  meta?: unknown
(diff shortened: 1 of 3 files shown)


docs/router/framework/react/api/router/RouteMatchType.md
```diff
@@ -8,22 +8,40 @@ The `RouteMatch` type represents a route match in TanStack Router.
...
+  loaderData?: TLoaderData
+  context: TAllContext
+  search: TFullSearchSchema
+  fetchCount: number
   abortController: AbortController
-  cause: 'enter' | 'stay'
+  cause: 'preload' | 'enter' | 'stay'
+  loaderDeps: TLoaderDeps
+  preload: boolean
+  invalid: boolean
```
SeanCassiere: Don't make changes to the `RouterMatch` type, since that is more of an internal thing. We keep the public surface on the docs quite sparse on purpose.

At most, I'd keep the `status`, `isFetching`, and `ssr` changes.

leesb971204: > Don't make changes to the `RouterMatch` type, since that is more of an internal thing. We keep the public surface on the docs quite sparse on purpose.
> 
> At most, I'd keep the `status`, `isFetching`, and `ssr` changes.

Got it.
However, wouldn’t it be confusing for users if there’s no information here about the return value of the `head` method in `RouteOptionsType`?

SeanCassiere: You've already defined the `ReturnType` for `head` right?

What I'm talking about is the type for the `RouteMatch` which is being publicly shown. 

Ideally, the user _should_ have very few interactions with the actual route match itself.

docs/router/framework/react/api/router/RouteOptionsType.md
```diff
@@ -300,3 +300,69 @@ If you want to configure to remount a route component upon `params` change, use:
...
+
+- Optional
+- Returns additional elements to inject into the document `<head>` for this route. Use it to add route-level SEO metadata, preload links, inline styles, or custom scripts.
+
+### `scripts` method
+
+- Type:
+
+```tsx
+types scripts = (ctx: {
```
SeanCassiere:
```suggestion
type scripts = (ctx: {
```

docs/router/framework/react/api/router/RouteMatchType.md
leesb971204: > You've already defined the `ReturnType` for `head` right?
> 
> What I'm talking about is the type for the `RouteMatch` which is being publicly shown.
> 
> Ideally, the user _should_ have very few interactions with the actual route match itself.

Yes, exactly.
However, I mentioned it because the return type of `head` references the `RouteMatch` type like below,
so I assume users will look into `RouteMatch` to understand what type it is.

So I’m wondering if, at the very least, these should remain in `RouteMatch`.

```typescript
links?: RouteMatch['links']
scripts?: RouteMatch['headScripts']
meta?: RouteMatch['meta']
styles?: RouteMatch['styles']
```

SeanCassiere: The actual underlying type there is something similar to what we’ve got going on here https://tanstack.com/router/latest/docs/framework/react/api/router/LinkPropsType

or perhaps make up something like FrameworkMetaTagAttributes

Either way, when it comes to RouteMatch type, that isn’t some we want to commit to especially since we change that quite often.

leesb971204: > The actual underlying type there is something similar to what we’ve got going on here https://tanstack.com/router/latest/docs/framework/react/api/router/LinkPropsType
> 
> or perhaps make up something like FrameworkMetaTagAttributes
> 
> Either way, when it comes to RouteMatch type, that isn’t some we want to commit to especially since we change that quite often.

Oh, I see — that makes perfect sense now.
In that case, I’ll revert all the other changes except for `status`, `isFetching`, and `ssr`.

SeanCassiere (approved):

leesb971204:
@SeanCassiere 

Thank you so much for kindly explaining everything!

SeanCassiere:
> Thank you so much for kindly explaining everything!

Always happy to have good discussions and quality contributions!

PR #4757, opened 2025-07-24
docs(react-router): update custom-link guide for react-aria

devongovett:
As of [React Aria Components v1.11.0](https://react-spectrum.adobe.com/releases/2025-07-22.html), we now pass through DOM events, so it's no longer necessary to use hooks to create a custom link component. You can just use `createLink` with the React Aria component directly.

[Demo](https://stackblitz.com/edit/stackblitz-starters-qc66mmzn?file=src%2FLink.tsx)

### DIFF
--- a/docs/router/framework/react/guide/custom-link.md
+++ b/docs/router/framework/react/guide/custom-link.md
@@ -60,14 +60,14 @@ import { createLink } from '@tanstack/react-router'
 import { Link as RACLink, MenuItem } from 'react-aria-components'
 
 export const Link = createLink(RACLink)
-export const MenuItemLink = createLink(MenuItem);
+export const MenuItemLink = createLink(MenuItem)
 ```
 
 To use React Aria's render props, including the `className`, `style`, and `children` functions, create a wrapper component and pass that to `createLink`.
 
 ```tsx
-import { createLink } from '@tanstack/react-router';
-import { Link as RACLink, type LinkProps } from 'react-aria-components';
+import { createLink } from '@tanstack/react-router'
+import { Link as RACLink, type LinkProps } from 'react-aria-components'
 
 interface MyLinkProps extends LinkProps {
   // your props
@@ -81,10 +81,10 @@ function MyLink(props: MyLinkProps) {
         color: isHovered ? 'red' : 'blue',
       })}
     />
-  );
+  )
 }
 
-export const Link = createLink(MyLink);
+export const Link = createLink(MyLink)
 ```
 
 ### Chakra UI example
--- a/docs/router/framework/react/guide/custom-link.md
+++ b/docs/router/framework/react/guide/custom-link.md
@@ -53,17 +53,38 @@ Here are some examples of how you can use `createLink` with third-party librarie
 
 ### React Aria Components example
 
-React Aria Components v1.11.0 and later works with TanStack Router's `preload (intent)` prop.
+React Aria Components v1.11.0 and later works with TanStack Router's `preload (intent)` prop. Use `createLink` to wrap each React Aria component that you use as a link.
 
 ```tsx
-import { createLink, LinkComponent } from '@tanstack/react-router'
-import { Link as RACLink } from 'react-aria-components'
+import { createLink } from '@tanstack/react-router'
+import { Link as RACLink, MenuItem } from 'react-aria-components'
 
-const CreatedLinkComponent = createLink(RACLink)
+export const Link = createLink(RACLink)
+export const MenuItemLink = createLink(MenuItem);
+```
 
-export const CustomLink: LinkComponent<typeof RACLink> = (props) => {
-  return <CreatedLinkComponent preload={'intent'} {...props} />
+To use React Aria's render props, including the `className`, `style`, and `children` functions, create a wrapper component and pass that to `createLink`.
+
+```tsx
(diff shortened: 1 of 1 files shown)


GuskiS:
Is this still needed? https://react-spectrum.adobe.com/react-aria/routing.html#tanstack-router
If so, might be good idea to include in this PR?

devongovett:
No, don't think so. Going to update those docs on our side.

GuskiS:
Without mentioned provider `MenuItem` (maybe others too) change their `href` type and does native anchor link instead of tanstack router redirect.

Regarding PR, it might be good idea to tell how to use custom props for Link component? Best I came up with was this:
```ts
declare module "react-aria-components" {
  export interface LinkProps {
    variant?: "primary" | "secondary" | "unstyled";
  }
}
```

devongovett:
I think it would be more like the Chakra/MUI examples where you create a new interface that extends LinkProps, uses those in the CustomLink component.

One problem I just noticed is that React Aria's render props aren't working with `createLink`, because that has its own render props. The children function with `isActive` overrides our children function, and the `className` and `style` functions don't seem to get passed through (the types accept them though). I think maybe because of this where it will stringify the className function: https://github.com/TanStack/router/blob/6a89d3c048f3ee176a66d967ba398d893a4e043f/packages/react-router/src/link.tsx#L341-L347

Should `className` and `style` get passed through to the underlying component unmodified if they are functions/not a string? Other component libraries like Headless UI and Base UI also accept functions to these props.

For the children render function I'm not sure there's a way to detect which function signature is expected. But at least the types are correct in that case (TanStack's render props).

gregmsanderson:
I have been trying the new way to make a custom link with Tanstack Router and it is much neater 👍

From my brief experiments today, it seems the custom routing mentioned by @GuskiS _is_ still needed (or some variation of it, as you mention). If you _only_ use the custom link, it's not, but as soon as you use _other_ components (like a Menu) which have a `href`, those _also_ need to route client-side.

(Doing so _may_ need a more complicated router provider setup. Not sure. The docs say: `useHref={(href) => router.buildLocation(href).href}` ... but when I use that, I get `String.prototype.search called on null or undefined` presumably because I'm using components that default to a null/"" href 🤔. That's a different issue though).

As for render props, I'm not sure. I haven't done anything beyond a `className` string here and that seems to work fine.

devongovett:
Btw you should be able to use createLink with other components like MenuItem too.

YannBirba:
Oh men that's sooo cool, that was really needed !

gregmsanderson:
@devongovett Ah! I didn't know that.

It works! You can indeed. That enables Tanstack Router's prefetch magic on a RAC MenuItem. That's so, _so_ good.

That solves the blank/null href issue with the RouterProvider too, since it would use the `to` prop:

```ttsx
{/* <MenuItem href="/project/categories"> */}
<CustomMenuItem to="/project/categories">
```

Perfect.

GuskiS:
Oh, that is indeed cool 👌

schiller-manuel:
can we merge this? or do the docs need tweaking? i just saw this discussion going on and i am not sure what that means :)

devongovett:
I can add a bit more clarification based on the discussion here. I'll have time a bit later today.

devongovett:
Ok added some additional docs to mention that this works with other React Aria components like MenuItem, and show how to use the render prop functions.

SeanCassiere (approved):

ali-idrizi:
One issue I see when using the setup without react-aria's `<RouterProvider>` is that clicking a link with <kbd>Enter</kbd> triggers a hard reload.

damianstasik:
I also encountered a hard redirect when using the `createLink` with `GridListItem` (possibly the same issue with `ListBoxItem`). Is that an expected caveat in this new approach?

devongovett:
hmm yeah GridListItem and Table Rows are a bit weird because the HTML spec forbids them from rendering as `<a>` elements (links don't allow interactive children). `RouterProvider` works around that by handling the click event and triggering a client side navigation. Perhaps a noop RouterProvider whose navigate function does nothing would work (`createLink` should still handle the click event). But maybe we can somehow do something better, I'll have to think about that.

schiller-manuel:
would you want to merge this PR now or wait?

devongovett:
I think it's ok to merge. It's much better than before. We can make additional improvements on our side.

uncvrd:
Just in case anyone comes across this, the current solution is to do the following on your `createRouter` to get this to work for items like `ListBoxItem`. The temporary workaround comes from this comment https://github.com/adobe/react-spectrum/issues/8920#issuecomment-3383404322

`Wrap: ({ children }) => <RACRouterProvider navigate={() => {}}>{children}</RACRouterProvider>,`

schiller-manuel:
@uncvrd  should we document this in the router docs?

uncvrd:
@schiller-manuel It's not officially documented on the React Aria Components section for custom links on Tanstack Router https://react-spectrum.adobe.com/react-aria/routing.html#tanstack-router and described as a workaround in the Adobe ticket above, so I would say no (for now) unless you've been getting a lot of questions about it (or they come up with a more permanent solution)

sfc-gh-alisowski:
I found the same for Tab Triggers. unless I provides a no-op navigate it will trigger the wrong type of navigation and reload the whole page

PR #4765, opened 2025-07-24
feat(react-router): add disableGlobalCatchBoundary option

### LINKED ISSUE #4748: Forced global error boundary breaks smoke testing

Router

### Describe the bug

Currently, anything wrapped in a `RouterProvider` will always catch any errors thrown in React components.  
This seems to be caused by the [unconditional `<CatchBoundary>` inside `<Matches>`](https://github.com/TanStack/router/blob/dae8833a463448af695d65428bd649836988d074/packages/react-router/src/Matches.tsx#L82-L94).

This interferes badly with most "reliability tools", since the top-level uncaught error handler will not be triggered.

- Storybook Test Runner smoke test phase will never fail, even if any component throws an exception.
- Chromatic (snapshot testing) does not mark the story/test as failed.
- Custom tests require a hacky workarounds to verify exceptions.
- If you forget to add a custom `errorComponent`, error reporting tools such as Sentry will just ignore the exception.
- In debugging, the uncaught exception handler will not trigger. This also reduces the effectiveness of any agents & other such tools.

etc. etc.

It would be very good to be able to disable the global catch boundary with eg. a router option.

https://stackblitz.com/edit/github-hp5huvtj-wbbbpqgj?file=src%2Froutes%2Findex.tsx

Open the repro link.

### Expected behavior

Exception should not be caught inside router (when configured so).

### Screenshots or Videos

_No response_

### Platform

- Router:  ^1.129.2 (from the repro quickstart StackBlitz)

### Additional context

_No response_

jaens:
## Summary

Adds a new `disableGlobalCatchBoundary` router option that allows unhandled errors to bubble up to top-level error handlers.

## Motivation

Fixes #4748

The current unconditional `<CatchBoundary>` in `<Matches>` interferes with:
- Testing tools (Storybook Test Runner smoke tests never fail)
- Error reporting services (Sentry doesn't catch errors automatically)
- Debugging (browser's uncaught exception handlers don't trigger)

## Changes

- Added `disableGlobalCatchBoundary?: boolean` option to `RouterOptionsExtensions`
- Modified `MatchesInner` component to conditionally render the `CatchBoundary`
- Updated documentation in `RouterOptionsType.md`
- Defaults to `false` for backward compatibility

## Usage

```typescript
const router = createRouter({
  routeTree,
  disableGlobalCatchBoundary: true,
  // ... other options
})
```


jaens:
Actions failure seems to be due to cancellation, not actual failures...

schiller-manuel:
thanks for this!
we should add the same impl for solid
we also need some tests for this

jaens:
Tests & Solid support have been added. Not sure what's going on in CI.

Tests might contain some redundant "setup" code since not sure exactly what's required in them or not.

PR #4767, opened 2025-07-25
refactor(router-core): make parsePathnameCache internal

dsonet:
REF: https://github.com/TanStack/router/pull/4752

It introduced the LRU ache for parsePathname. However, it exposed the internal cache, making it the caller's responsibility to pass the cache. This could defeat the purpose if the parameter is missing.

I'd suggest it's better to keep the cache internal and transparent to the caller. This reduces the burden of passing the cache around and keeps the API cleaner, the bundle size would be smaller as well.

### DIFF
--- a/packages/router-core/src/path.ts
+++ b/packages/router-core/src/path.ts
@@ -1,5 +1,5 @@
 import { last } from './utils'
-import type { LRUCache } from './lru-cache'
+import { createLRUCache, type LRUCache } from './lru-cache'
 import type { MatchLocation } from './RouterProvider'
 import type { AnyPathParams } from './route'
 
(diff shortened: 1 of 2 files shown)


Sheraff:
We intentionally made it a thing that lives on the router, because depending on the setup, some people have several routers. The cache is passed to the appropriate methods internally, and no public APIs were affected. What is your use-case that is impacted by this change?

schiller-manuel:
we specifically chose to make the cache router instance specific. all the functions `parsePathname` etc. are not public API.
we'll leave the implementation as is.
in future, can you please create an issue so we can discuss before you spend the effort to create a PR?

PR #4771, opened 2025-07-25
fix: pin all workspace dependencies and devDependencies

alvarosevilla95:
I keep running into (mostly typing) issues when updating TSS packages to a version lower that latest, due to nested dependencies pulling a higher available version than the one I'm using.

I've updated all `"workspace:^"` to `"workspace:*"` for all dependencies and devDependencies within `packages/`. I did not change any `peerDependencies` to allow flexibility on those. 

I also didn't change the package.jsons in `e2e/` although I imagine those would be safe to pin exact as well.


alvarosevilla95:
Hmm not sure what's up with the failing check. It looks like all tests are passing (and so is `pnpm test` locally)..
I see on the NX site this failed:

```
Command
Install Node Modules
```

but it seems an issue with corepack installing pnpm due to a mismatched key... and it seems to occur on other (green) PRs too.

Can I ask for some guidance here?

SeanCassiere:
Just checked on Discord, the installation failure looks to be a NX specific issue. For the time being, we'll leave this PR open till the NX Cloud Team have managed to resolve the issue on their end.

Issue #4778, opened 2025-07-26
Generator rewrite broke backward compatibility for Classes


SeanCassiere:
### Which project does this relate to?

Router

### Describe the bug

The generator rewrite for detection using AST broke backwards compatibility for two usecases that should be still supported during the v1 release cycle.

### 1. Using component extensions.

Creating a virtual route using just the component. You'd use `.component`, `.pendingComponent`, `.errorComponent`, `.notFoundComponent`. This should create a virtual route (or update the route definition) in the route tree file to add a lazy import.

Like:
```
src/routes
  _auth/
    foo.component.tsx
```

Inside `foo.component.tsx`, you'd have it like this.

```ts
import { MyComponent } from '@/wherever'
export const component = MyComponent
```

Should result in.

```ts
// src/routeTree.gen.ts
const PostsComponentRoute = PostsComponentImport.update({
  path: '/posts,
  getParentRoute: () => rootRoute,
} as any).update({
  component: lazyRouteComponent(
    () => import('./routes/posts.component'),
    'component'
  ),
})
```

### 2. Not detecting class exports.

Now the export of `Route` using a constructor of `FileRoute` wasn't being detected.

```ts
// src/routes/index.tsx
import { FileRoute } from '@tanstack/react-router'

export const Route = new FileRoute('/').createRoute({ ... })
```

### Your Example Website or App

...

### Steps to Reproduce the Bug or Issue

.

### Expected behavior

.

### Platform

- Router / Start Version: 1.129.8
- OS: macOS
- Browser: Chrome
- Bundler: Vite
- Bundler Version: ^6.2.1

SeanCassiere:
I tried looking into this, but it looks like the plugin architecture will need some rewriting to support;

- Handling the other exports (component, loader, pendingComponent, errorComponent, notFoundComponent).
- Not require module augmentation.

I've added test files and opened it as a draft PR for now - https://github.com/TanStack/router/pull/4780

SeanCassiere:
Speaking with Manuel, we'll be closing this since this doesn't really seem to be an issue for the vast public userbase.

PR #4780, opened 2025-07-26
tests for generator breaking backward compatibility

SeanCassiere:
Related to #4778, this adds tests so development can be done to support two workflows which were lost in the generator rewrite.

1. Using extensions, like `src/routes/posts.component.tsx`. 
2. Using Class based Route inits, like `export const Route = new FileRoute('/posts').createRoute({ ... })`.

The test files for `1` lives `packages/router-generator/tests/generator/component-extensions`. These test files do not cause a fail condition, since the logic is outright missing for appending these extensions as lazy imports.

The test files for `2` lives `packages/router-generator/tests/generator/classes-mixed`. These test files mix both function and class based Route inits. The error message is as follows for a route using a class-based init.

```sh
tests/generator.test.ts > generator works > should wire-up the routes for a "classes-mixed" tree
Error: expected identifier to be present in /posts for export Route
 ❯ src/generator.ts:307:19
```

### DIFF
--- a/packages/router-generator/tests/generator/classes-mixed/routes/__root.tsx
+++ b/packages/router-generator/tests/generator/classes-mixed/routes/__root.tsx
@@ -1,0 +1,15 @@
+import * as React from 'react'
+import { Outlet, RootRoute } from '@tanstack/react-router'
+
+export const Route = new RootRoute({
+  component: RootComponent,
+})
+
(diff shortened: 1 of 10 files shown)


SeanCassiere:
Closing https://github.com/TanStack/router/issues/4778#issuecomment-3121246874

PR #4799, opened 2025-07-27
refactor(router-core): minor cleanup of loadMatches

Sheraff:
This PR is a minor cleanup of `router.ts`
- use `.some()` instead of `.find()` when we only need a boolean result
- the result of `updateMatch` is never used, so don't return anything
- `getMatch` doesn't need to create an array of all matches
- `loadMatches`
  - doesn't need to create a promise explicitly, itself is an async function already
  - can skip some code if there is no `route.options.beforeLoad`
  - can skip some code if there is no `route.options.loader`

And some minor "don't create an object if you don't need to".

---

A good goal would be to try and make `loadMatches` much lighter by updating the store less (fewer calls to `updateMatch`) as this can trigger a lot of synchronous code execution every time. But this PR only does minor cleanup because the big refactor is scary.

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -108,10 +108,10 @@ export type DefaultRemountDepsFn<TRouteTree extends AnyRoute> = (
   opts: MakeRemountDepsOptionsUnion<TRouteTree>,
 ) => any
 
-export interface DefaultRouterOptionsExtensions { }
+export interface DefaultRouterOptionsExtensions {}
 
 export interface RouterOptionsExtensions
-  extends DefaultRouterOptionsExtensions { }
+  extends DefaultRouterOptionsExtensions {}
 
 export interface RouterOptions<
   TRouteTree extends AnyRoute,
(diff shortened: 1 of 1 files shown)


packages/router-core/src/router.ts
```diff
@@ -2043,37 +2043,29 @@ export class RouterCore<
...
-    return updated
   }
 
   getMatch: GetMatchFn = (matchId: string) => {
-    return [
-      ...this.state.cachedMatches,
-      ...(this.state.pendingMatches ?? []),
-      ...this.state.matches,
-    ].find((d) => d.id === matchId)
+    const find = (d: { id: string }) => d.id === matchId
```
SeanCassiere, line 2063:
```suggestion
    const findFn = (d: { id: string }) => d.id === matchId
```

packages/router-core/src/router.ts
```diff
@@ -2858,7 +2813,7 @@ export class RouterCore<
           : (route.options.gcTime ?? this.options.defaultGcTime)) ??
         5 * 60 * 1000
 
-      return !(d.status !== 'error' && Date.now() - d.updatedAt < gcTime)
+      return d.status === 'error' && Date.now() - d.updatedAt >= gcTime
```
SeanCassiere, line 2816: Can we break up one-liner into a couple of booleans? with variables like `isError` and `isElapsed`.

packages/router-core/src/router.ts
Sheraff: I don't really know what I broke in this PR that I thought was basic, so I applied your proposed change in a separate PR that just touches *that* method https://github.com/TanStack/router/pull/4824

PR #4801, opened 2025-07-28
Fix code example typo in document-head-management.md

m-atthart:
RouteOptions.scripts is supposed to be a function, not an array. ask me how i know

### DIFF
--- a/docs/router/framework/react/guide/document-head-management.md
+++ b/docs/router/framework/react/guide/document-head-management.md
@@ -115,7 +115,7 @@ To do this, you must:
 
 ```tsx
 export const Route = createRootRoute({
-  scripts: [
+  scripts: () => [
     {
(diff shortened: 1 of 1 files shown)


SeanCassiere (changes requested):
There should be other places in the docs that have this mistake. Please change them as well.

m-atthart:
I looked through every occurrence of "scripts" in the docs folder, and every occurrence of "scripts:" in the entire codebase and confirmed there are no other places with this mistake.

Every other example in the docs that references adding scripts does it through the RouteOptions.head property, where scripts actually is an array, not a function, because it is a key in the object returned by the head method. That's where I assume the typo came from.

Thanks for reviewing my pr.

SeanCassiere (approved):

SeanCassiere:
@m-atthart thank you!

PR #4809, opened 2025-07-28
removes Tailwind

ZackBz:
The `start-bare` example includes Tailwind with a broken config (missing `tailwindcss()` in vite.config.ts plugins), which contradicts the 'bare' concept. Since `start-tailwind-v4` already provides a properly configured Tailwind example, this PR removes Tailwind from `start-bare` to make it actually bare.

### DIFF
--- a/examples/react/start-bare/package.json
+++ b/examples/react/start-bare/package.json
@@ -17,11 +17,9 @@
     "zod": "^3.24.2"
   },
   "devDependencies": {
-    "@tailwindcss/vite": "^4.0.8",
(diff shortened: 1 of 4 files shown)


schiller-manuel:
can you please apply the same to solid/start-bare ?

ZackBz:
I was wondering if this example is also something you guys wanna showcase on the TanStack Start React docs. Right now, there isn't any mention of it as far as I know. I am willing to get it added!

schiller-manuel:
please update lock file

PR #4823, opened 2025-07-29
docs: add info about zod 4

niba:
Add a note about zod 4. I think that at some point we can rewrite the entire doc about search-params because `adapter` becomes a special case only for zod 3

related: https://github.com/TanStack/router/pull/4442

### DIFF
--- a/docs/router/framework/react/guide/search-params.md
+++ b/docs/router/framework/react/guide/search-params.md
@@ -204,7 +204,7 @@ For validation libraries we recommend using adapters which infer the correct `in
 ### Zod 4
 
 > [!WARNING]
-> Zod v4.0.6 or higher is required for native TanStack Router support.
+> Zod v4.0.6 or higher is recommended for proper type inference.
 
 When using [Zod 4](https://zod.dev/) an adapter is not needed to ensure the correct `input` and `output` types are used for navigation and reading search params. This is possible thanks to `zod` enhanced type inference, which preserves complete type information when using [Standard Schema](https://github.com/standard-schema/standard-schema).
 
@@ -216,6 +216,26 @@ const productSearchSchema = z.object({
   page: z.number().default(1),
   filter: z.string().default(''),
   sort: z.enum(['newest', 'oldest', 'price']).default('newest').catch('newest'),
+})
+
+export const Route = createFileRoute('/shop/products/')({
+  validateSearch: productSearchSchema,
+})
+```
+
+Alternatively, you can use the [Zod Mini](https://zod.dev/packages/mini) variant
+
+```tsx
+import { createFileRoute } from '@tanstack/react-router'
+import * as z from 'zod/mini'
+
+const productSearchSchema = z.object({
+  page: z._default(z.number(), 1),
+  filter: z._default(z.string(), ''),
+  sort: z.catch(
+    z._default(z.enum(['newest', 'oldest', 'price']), 'newest'),
+    'newest',
+  ),
 })
 
 export const Route = createFileRoute('/shop/products/')({
--- a/docs/router/framework/react/guide/search-params.md
+++ b/docs/router/framework/react/guide/search-params.md
@@ -201,9 +201,31 @@ It might be surprising that when you try to navigate to this route, `search` is 
 
 For validation libraries we recommend using adapters which infer the correct `input` and `output` types.
 
-### Zod
+### Zod 4
 
-An adapter is provided for [Zod](https://zod.dev/) which will pipe through the correct `input` type and `output` type
+> [!WARNING]
+> Zod v4.0.6 or higher is required for native TanStack Router support.
+
+When using [Zod 4](https://zod.dev/) an adapter is not needed to ensure the correct `input` and `output` types are used for navigation and reading search params. This is possible thanks to `zod` enhanced type inference, which preserves complete type information when using [Standard Schema](https://github.com/standard-schema/standard-schema).
+
+```tsx
+import { createFileRoute } from '@tanstack/react-router'
+import { z } from 'zod'
+
+const productSearchSchema = z.object({
+  page: z.number().default(1),
+  filter: z.string().default(''),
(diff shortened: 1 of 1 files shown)


docs/router/framework/react/guide/search-params.md
```diff
@@ -201,9 +201,31 @@ It might be surprising that when you try to navigate to this route, `search` is
 
 For validation libraries we recommend using adapters which infer the correct `input` and `output` types.
 
-### Zod
+### Zod 4
 
-An adapter is provided for [Zod](https://zod.dev/) which will pipe through the correct `input` type and `output` type
+> [!WARNING]
+> Zod v4.0.6 or higher is required for native TanStack Router support.
```
schiller-manuel: "native TanStack Router support" sounds a bit weird.

docs/router/framework/react/guide/search-params.md
```diff
@@ -201,9 +201,31 @@ It might be surprising that when you try to navigate to this route, `search` is
...
+When using [Zod 4](https://zod.dev/) an adapter is not needed to ensure the correct `input` and `output` types are used for navigation and reading search params. This is possible thanks to `zod` enhanced type inference, which preserves complete type information when using [Standard Schema](https://github.com/standard-schema/standard-schema).
+
+```tsx
+import { createFileRoute } from '@tanstack/react-router'
+import { z } from 'zod'
+
+const productSearchSchema = z.object({
+  page: z.number().default(1),
+  filter: z.string().default(''),
+  sort: z.enum(['newest', 'oldest', 'price']).default('newest').catch('newest'),
```
KiwiKilian, line 218: I think it would be great to have a few words on the difference of `.default()` and `.default().catch()`.

Also, when the default is used already twice, should we recommend the pattern from the [search middlewares](https://tanstack.com/router/latest/docs/framework/react/guide/search-params#transforming-search-with-search-middlewares) of using an `defaultValues` object?

niba: the docs earlier on this page explain when to use ⁠default and when to use ⁠catch.

I follow the same style as used in the rest of the adapters. I think that the choice about using defaultValues belongs to the developer and is not important here, especially since the values that you put into default and catch are checked by TypeScript

docs/router/framework/react/guide/search-params.md
niba: true, maybe

"Zod v4.0.6 or higher is recommended for proper type inference"?

docs/router/framework/react/guide/search-params.md
slargat: in case you want to include any  zod/v4-mini examples :
catch:
```typescript
validateSearch:z.object({
    redirectTo: z.catch(z.string(), "")
})
```
default:
```typescript
validateSearch:z.object({
    redirectTo: z.catch(z._default(z.string(), ""), ""),
})
```

docs/router/framework/react/guide/search-params.md
```diff
@@ -201,9 +201,31 @@ It might be surprising that when you try to navigate to this route, `search` is
 
 For validation libraries we recommend using adapters which infer the correct `input` and `output` types.
```
KiwiKilian, line 202: This should be more relaxed, as Zod v4 is now the opposite? Like "For some validation libraries..."?

docs/router/framework/react/guide/search-params.md
KiwiKilian: > the docs earlier on this page explain when to use ⁠default and when to use ⁠catch.

But it doesn't explain why one would use `default` **and** `catch` on the some line which was quite confusing to me in the beginning.

KiwiKilian: While using the `defaultValues` is absolutely optional here, it's a required pattern for the search middleware default stripping to allow instance comparison. So I would think this would make the docs more aligned in recommending that patter subtly, but it's true other parts of the docs also don't use it.

niba: > > the docs earlier on this page explain when to use ⁠default and when to use ⁠catch.
> 
> But it doesn't explain why one would use `default` **and** `catch` on the some line which was quite confusing to me in the beginning.

You’re right. I’ll add explanation. In general, it’s tricky because ⁠`.catch()` normally behaves like ⁠`.default()`, but its behaviour changes when you combine it with ⁠`.optional()`. It also produces a different type, since ⁠`.default()` makes the field optional.

niba: > While using the `defaultValues` is absolutely optional here, it's a required pattern for the search middleware default stripping to allow instance comparison. So I would think this would make the docs more aligned in recommending that patter subtly, but it's true other parts of the docs also don't use it.

I can change all the examples to use `defaultValues` everywhere, I don't have a strong opinion about it

docs/router/framework/react/guide/search-params.md
niba: To be honest, I don’t understand the `Adapters` section. What are we trying to say here? I also tested the example in that section and it isn’t true. ⁠Link doesn’t require ⁠search to be defined, because all fields are optional (tested with Zod v3). Moreover, we already covered transforming search params with ⁠`.default()` in the previous section so I don't know why we write it like introduction to this topic.

So I think we should rewrite it or remove it, especially right now that we use the adapter only in Zod 3.

schiller-manuel: yes, a more aggressive rewrite is in order. are you up for it?

niba: I can try. If you have any requirements or wishes for a new doc page, you can send them over discord (my nickname is the same as here)

LadyBluenotes:
Thanks for the Zod 4 documentation work. The `search-params` guide has since been rewritten to cover the Zod 3 adapter path, direct Zod 4 schemas, and the `.default()` versus `.catch()` behavior.

Closing this as superseded by #6956.

PR #4830, opened 2025-07-29
refactor(router-core): router loadMatches doesn't need async IIFE

Sheraff:
This PR removes a pattern that was probably left behind from a time when it was more complicated, because it is not currently serving a purpose: "an awaited inline new `Promise` that executes an async IIFE to resolve at the end, or any error any" is just the same as directly running the code (with a few extra micro task loops)


packages/router-core/src/router.ts
```diff
@@ -2179,588 +2179,567 @@ export class RouterCore<
...
+            if (parentMatch?.ssr === false) {
+              ssr = false
+            } else {
+              let tempSsr: boolean | 'data-only'
+              if (route.options.ssr === undefined) {
+                tempSsr = defaultSsr
+              } else if (typeof route.options.ssr === 'function') {
+                const { search, params } = this.getMatch(matchId)!
+
+                function makeMaybe(value: any, error: any) {
```
SeanCassiere: Would the definition of `makeMaybe` outside of the for loop have any performance gains? (still being left within the scope of `loadMatches` though)

Sheraff: Hoisting that function should bring perf improvements indeed. More generally `loadMatches` should probably be split up into smaller functions. I'm making small "almost nothing changes" PRs right now, but I'll be attempting an actual perf PR just after.

PR #4837, opened 2025-07-30
docs:(start): add start mcp example

Balastrong:
A working example integrating the official `@modelcontextprotocol/sdk` to create a stateless http server on a start route.

The server is exposed on `http://localhost:3000/api` and has a resource (tells the counter value, from the counter example) and a tool (sums two numbers).


examples/react/start-mcp/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "tanstack-start-example-mcp",
+  "private": true,
+  "sideEffects": false,
+  "type": "module",
+  "scripts": {
+    "dev": "vite dev",
+    "build": "vite build && tsc --noEmit",
+    "start": "vite start"
```
schiller-manuel, line 9: `vite start` wont work I guess?

examples/react/start-mcp/src/routes/__root.tsx
```diff
@@ -0,0 +1,49 @@
...
+      {
+        name: 'viewport',
+        content: 'width=device-width, initial-scale=1',
+      },
+      {
+        title: 'TanStack Start MCP Starter',
+      },
+    ],
+  }),
+  component: RootComponent,
```
schiller-manuel: please use shellComponent

see https://github.com/TanStack/router/blob/main/examples/react/start-basic/src/routes/__root.tsx#L62

(yes I know i need to document this :D )

examples/react/start-mcp/src/utils/mcp-handler.ts
```diff
@@ -0,0 +1,61 @@
+import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js'
+import { getEvent } from '@tanstack/react-start/server'
+import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
+
+export async function handleMcpRequest(request: Request, server: McpServer) {
+  const body = await request.json()
+  const event = getEvent()
+  const res = event.node.res
```
schiller-manuel, line 8: this will only work in node then.... how about other runtimes?

Balastrong: With Bun looks like it's working. After all it's the official mcp sdk having a hard dependency on that, I don't think we can be much more flexible until they open their implementation a bit more

examples/react/start-mcp/package.json
Balastrong: I copied from the `start-counter` example but I haven't tried before. I did now and... yeah `vite start` doesn't seem to work.

Should we replace all `vite start` with `node .output/server/index.mjs` in the examples?

schiller-manuel: well it depends on the deployment target. by default, it's using node, so yes it would work. if someone changes to e.g. bun, the start command would need to be different ...

examples/react/start-mcp/src/routes/__root.tsx
```diff
@@ -22,7 +22,7 @@ export const Route = createRootRoute({
       },
     ],
   }),
-  component: RootComponent,
+  shellComponent: RootComponent,
```
schiller-manuel: you would need to use `RootDocument` as the `shellComponent`
and then move `<TanStackRouterDevtools />` into `RootDocument`, then delete `RootComponent`

Balastrong: Whoops sorry, I only looked at the linked line, I didn't notice more things were to be changed.

I copied from the start-basic example, is it ok now?

examples/react/start-mcp/package.json
Balastrong: Is it something related to this PR (adding the new mcp example) only or is it a global decision for all `start` commands in all examples/docs?

I'd say as long as we explain in the docs how to run in all environments should be fine that we show one in the feature-related examples

examples/react/start-mcp/src/routes/__root.tsx
schiller-manuel: yep!

LadyBluenotes:
Thanks for putting this together. The MCP use case is worth documenting, but the transport lifecycle, runtime support, and endpoint security need a fresh pass against current Start.

If you'd like, do you continuing this in a new PR?

Issue #4867, opened 2025-08-04
`history.go` does not respect `ignoreBlocker` option


codoodle:
### Which project does this relate to?

Router

#### Summary

When using TanStack Router with `BrowserHistory`, the `ignoreBlocker` option works as expected for `history.back` and `history.forward`, but **does not work for `history.go`**. This means that navigation blockers are still triggered even when `ignoreBlocker: true` is passed to `history.go`.

#### Details

- In the source code, both `back` and `forward` methods set the `skipBlockerNextPop` flag to `true` when `ignoreBlocker` is provided, allowing navigation to bypass blockers.
- However, the `go` method does **not** set this flag, so blockers are always executed, regardless of the `ignoreBlocker` option.
- This makes it impossible to programmatically skip blockers when navigating multiple steps in history using `go`.

### Your Example Website or App

https://github.com/codoodle/issues/tree/tanstack-router/go-does-not-skip-blocker

### Steps to Reproduce the Bug or Issue

1. Create a route (e.g., `/two`) that uses a blocker with `useBlocker`:

   ```tsx
   function Two() {
     useBlocker({
       shouldBlockFn: () => {
         return !confirm("Are you sure you want to leave?");
       },
     });

     const router = useRouter();

     const handleBack = () => {
       router.history.back({
         ignoreBlocker: true,
       });
     };

     const handleGo = () => {
       router.history.go(-1, {
         ignoreBlocker: true,
       });
     };

     return (
       <div>
         <h1>Hello "/two"!</h1>
         <button onClick={handleBack}>Back</button>
         <button onClick={handleGo}>Go(-1)</button>
       </div>
     );
   }
   ```

2. Add two buttons to the page, one that calls `router.history.back({ ignoreBlocker: true })` and another that calls `router.history.go(-1, { ignoreBlocker: true })`.
3. Navigate to the `/two` route in your application.
4. Click the "Back" button. The blocker is skipped as expected (no confirmation dialog).
5. Navigate again to `/two`.
6. Click the "Go(-1)" button. The blocker is **not** skipped and the confirmation dialog appears, even though `ignoreBlocker: true` was passed.

### Expected behavior

The `ignoreBlocker` option should work consistently across `back`, `forward`, and `go`. When `ignoreBlocker: true` is passed to `history.go`, blockers should be skipped, just like with `back` and `forward`.

### Platform

- Router / Start Version: 1.130.12
- OS: macOS
- Browser: Safari
[...]

PR #4872, opened 2025-08-06
fix(history): add ignoreBlocker support to go method (#4867)

KSJ27:
## Description

Fixes GitHub issue #4867 - `history.go` does not respect `ignoreBlocker` option.

Currently, when using TanStack Router with `BrowserHistory`, the `ignoreBlocker` option works as expected for `history.back` and `history.forward`, but does not work for `history.go`. This means that navigation blockers are still triggered even when `ignoreBlocker: true` is passed to `history.go`.

### Core Fix
- **`packages/history/src/index.ts`**: 
  - Add `ignoreBlocker` parameter to `go` method in `createBrowserHistory`
  - Update `createHistory` to pass `ignoreBlocker` to underlying `go` implementation
  - Ensure consistent behavior across `back`, `forward`, and `go` methods

### Testing
- **`packages/history/tests/ignoreBlocker.test.ts`**: 
  - Test all navigation methods (`back`, `forward`, `go`) with `ignoreBlocker` option in `createBrowserHistory`

### DIFF
--- a/packages/history/src/index.ts
+++ b/packages/history/src/index.ts
@@ -101,7 +101,7 @@ export function createHistory(opts: {
   getLength: () => number
   pushState: (path: string, state: any) => void
   replaceState: (path: string, state: any) => void
-  go: (n: number) => void
+  go: (n: number, ignoreBlocker: boolean) => void
   back: (ignoreBlocker: boolean) => void
   forward: (ignoreBlocker: boolean) => void
   createHref: (path: string) => string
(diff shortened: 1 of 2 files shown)


LadyBluenotes:
Thanks for the fix. The `go` path still needs this behaviour, but the current tests do not reach the browser `popstate` handler where blockers are actually skipped.

I’m closing this stale branch and keeping #4867 open for a fresh implementation with a behavioural `BrowserHistory` regression

PR #4879, opened 2025-08-07
chore(examples): remove unnecessary opts.deps to clear confusion

hamitaksln:
I think opts.deps is creating confusion here because there is no loader deps there.


SeanCassiere:
🚀

PR #4900, opened 2025-08-09
fix(example): work-os api callback to reflect documentation

nthonymiller:
The Workos example states that the callback URL should be `http://localhost:3000/api/auth/callback` however the example provided an incorrect API implementation.

Changed and moved file from `callback.tsx` to `api/auth.callback.tsx`


LadyBluenotes:
Thanks for catching the callback-path mismatch. This route move was merged later in #5048, and the WorkOS example has since moved to the current AuthKit integration, so this branch is now superseded.

I’ve opened #7790 for the remaining README redirect-URI correction.

PR #4911, opened 2025-08-10
feat(eslint): support the shorter syntax with verboseFileRoutes: false

Balastrong:
Flag `create-route-property-order` also when `verboseFileRoutes` is false and file routes are defined without explicit import, for example in:

```ts
export const Route = createFileRoute({
  component: RouteComponent,
  head: ({}) => ({}),
  loader: ({}) => {},
})
```


schiller-manuel:
this flag will not be supported in future so we'll close this

Issue #4359, opened 2025-06-10
`updateMatch` 10s of times on navigation, causing slowdowns


hadar759:
### Which project does this relate to?

Router

### Describe the bug

On navigation `loadMatches` runs twice, and inside it `updateMatch` runs 3 times for each route in the hierarchy of the navigated one. This causes `updateMatch` to run >50 times on a typical navigation in my app, slowing things down - especially when `updateMatch` itself is slow in some cases (for reasons unknown to me).

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-updatematch

### Steps to Reproduce the Bug or Issue

This is a reproduction of the route hierarchy in my app.
I couldn't get the manual setup stages to persist upon refresh, so the following needs to be ran:
1. Let dev servers load
2. Run `pnpm patch @tanstack/router-core --edit-dir 'pnpm-patches'`
3. Edit `pnpm-patches/dist/esm/router.js` - add `console.log('updateMatch', id);` after line 791
5. Run `pnpm patch-commit pnpm-patches`. The console will complain that no changes were found, ignore it.
6. Run `pnpm install && pnpm run dev`
7. Click on links and navigation buttons, and see lots of `updateMatch` for each one in the console

### Expected behavior

I expected `updateMatch` to run a lot less, and when it does to not take more than 1-2ms, it causes my app to completely freeze sometimes.

### Screenshots or Videos

Video + screenshots from a Chrome devtools Performance tab recording, where calling `navigate` froze my app for 5 seconds on `updateMatches` which took 50ms-200ms each

https://github.com/user-attachments/assets/5b2983ae-1002-47e1-8d47-01eec49ee019

<img width="612" alt="Image" src="https://github.com/user-attachments/assets/c54d3001-f5c2-4887-9a56-4c8bdf44901c" />

### Platform

- OS: macOS Sequoia 15.5 (24F74)
- Browser: Arc browser Version 1.97.1 (63934), Chromium Engine Version 137.0.7151.69

### Additional context

I'd be happy to provide more information as needed.
Besides the issue of `updateMatch` executing so many times, I couldn't understand why sometimes it takes very long - especially since that's the only thing running according to the Performance tab.

schiller-manuel:
are you using fine grained selectors? then i think those might be called once per updateMatch invocation

Sheraff:
I'm going to mark this issue as resolved following the latest performance improvements. Work on performance is still ongoing, but the specific issue mentioned here should have significantly improved already.

PR #4925, opened 2025-08-11
refactor(router-core): head, scripts, headers are executed in parallel and can skip store update

Sheraff:
Following https://github.com/TanStack/router/pull/4916, this PR keeps reducing the number of store updates (`__store.setState` through `updateMatch`).

We can skip `route.options.head()`, `route.options.scripts()`, and `route.options.headers()` calls, if they are not defined, and not even call `updateMatch` after.

Additionally, `head`, `scripts` and `headers` are now called in parallel. They don't depend on one another, and so there is no reason to delay their execution by calling them serially.

---

This PR also fixes 2 issues:
- in `cancelMatch`, we were resetting the timeout id to `undefined` **before** calling `clearTimeout()` with it
- `handleRedirectAndNotFound` is sometimes called w/ `match` being `undefined` (in case of a redirecting match during preload), which went undetected because of heavy `match!` assertions, and the many `try/catch` blocks

---

With this PR, the `redirection in preload` test in `store-updates-during-navigation` goes from **11 updates** during the preload to **8 updates**.

---

Should be a partial improvement of https://github.com/TanStack/router/issues/4359


packages/router-core/src/router.ts
```diff
@@ -2463,35 +2463,45 @@ export class RouterCore<
...
-                    return {
-                      meta,
-                      links,
-                      headScripts,
-                      headers,
-                      scripts,
-                      styles,
-                    }
+
+                    return Promise.all([
```
schiller-manuel, line 2494: why no `await Promise.all()`?
personally, i like async/await syntax more than "then".
unless there is a technical reason?

Sheraff: If we use `await` here, the function needs to be `async`, which means that it will always return a `Promise`, even if we returned early. I'd like to be able to do the `if('then' in foo)` thing where this function is called, but we can't do that if the function is `async`.

PR #4929, opened 2025-08-11
RFC: replace tsconfig paths with node native import subpaths

jansepke:
for discussion https://github.com/TanStack/router/discussions/4899


LadyBluenotes:
Thanks for putting together a concrete example for the discussion. I’m closing this draft for now because it's stale. We can keep #4899 open for if we need to reopen this.

PR #4943, opened 2025-08-13
chore: upgraded start-basic to Tailwind CSS 4

58bits:
We've just started testing Tanstack Start and began with the start-basic example. The first thing we did was upgrade to Tailwind CSS 4 - which is in this commit and PR. Probably just a drop in the bucket but hope this helps. Feel free to ignore if you have other plans.

### DIFF
--- a/examples/react/start-basic/package.json
+++ b/examples/react/start-basic/package.json
@@ -18,13 +18,12 @@
     "zod": "^3.24.2"
   },
   "devDependencies": {
+    "@tailwindcss/vite": "^4.1.11",
     "@types/node": "^22.5.4",
     "@types/react": "^19.0.8",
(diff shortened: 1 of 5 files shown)


schiller-manuel:
this needs to be applied here i guess

https://github.com/TanStack/router/pull/4606

58bits:
I've tested this - `npm install` and `npm dev` and it all seems to be working fine. I thought this one might be worth updating as it's one of the first examples that's mentioned on the getting started pages. Would be happy to take a look at the others if that would help.

KernelCode:
i think we need to update the docs  ..

brenelz (approved):

birkskyum:
All examples / e2e have been updated to tailwind 4

PR #4953, opened 2025-08-14
feat: enhance devtools to use event-bus

AlemTuzlak:
still WIP, need to:
- add detached mode version into examples and check what happens when you detach the devtools if the state is reflected (https://github.com/TanStack/devtools/pull/37)


LadyBluenotes:
Going to close this because its very stale :))

PR #4965, opened 2025-08-16
Make useBlocker exception on unknown routes optional

sescandell:
Adds a `throwOnUnknownRoute` option to the `useBlocker` hook for both React and Solid.

When set to `false`, this option prevents `useBlocker` from throwing an exception if the user navigates to a route that is not recognized by the router. Instead, the navigation is allowed to proceed. This is useful for applications that are progressively adopting the router and may have links that navigate from router-managed pages to non-router-managed pages.

The default value is `true` to maintain the existing behavior. 

I'm not sure the option is really necessary. We could imagine to completely reconsider throwing an exception from this function (I don't see anyway how the user could catch it. Let me know if we already can).

### DIFF
--- a/packages/react-router/src/useBlocker.tsx
+++ b/packages/react-router/src/useBlocker.tsx
@@ -57,8 +57,8 @@ type BlockerResolver<TRouter extends AnyRouter = RegisteredRouter> =
     }
 
 type ShouldBlockFnArgs<TRouter extends AnyRouter = RegisteredRouter> = {
-  current: MakeShouldBlockFnLocationUnion<TRouter>
-  next: MakeShouldBlockFnLocationUnion<TRouter>
+  current: MakeShouldBlockFnLocationUnion<TRouter> | null
+  next: MakeShouldBlockFnLocationUnion<TRouter> | null
   action: HistoryAction
 }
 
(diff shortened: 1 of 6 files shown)


schiller-manuel:
thanks for this PR.
we are in a feature freeze right now until the next start alpha is merged to main.
we will then revisit this PR, so please be a bit patient

sescandell:
Hi there 👋 

Just checking in for an update. I'm not sure when the next start alpha is scheduled to be merged.

On the surface, this PR doesn't actually change the default behavior. It just adds a way to customize a specific situation with useBlockers, which should help with adopting Tanstack Router in an existing application.

Let me know if you need anything.

Thanks for your time,

schiller-manuel:
sorry this got lost. we should still leave it open and review / merge when we find the time

sescandell:
Hi @schiller-manuel  no worries,

I closed it because the logic in this file has changed. There is no more exception thrown now since this specific change: https://github.com/TanStack/router/commit/c35f70ec3c90fb7e8534a7fbd547df81080bacba#diff-199e71f47ccda4760077ec63cee57ee1450854a1dde6be207401fb647f3cf354L182.

So, this PR is now kind of obsolete.

PR #4969, opened 2025-08-16
refactor(router-core): split loadMatches into its own file

Sheraff:
After the "flattening" PR https://github.com/TanStack/router/pull/4961, we now have a very clear *separate responsibility* the `Router` class is handling: loading matches.

This PR proposes we split the code that handles "loading matches" into its own `load-matches.ts` file, with the entry points being
- `loadMatches`
- `loadRouteChunk`

In future PRs, we might consider splitting other responsibilities out of the 3k LoC `router.ts` file


packages/router-core/src/router.ts
```diff
@@ -2095,872 +2081,7 @@ export class RouterCore<
...
-        throw err
-      }
-      if (isRedirect(err)) {
-        throw err
-      }
-    }
-
-    return innerLoadContext.matches
-  }
+  loadMatches = loadMatches.bind(this)
```
schiller-manuel: this is not a public API so we can just call the function without binding it

PR #4971, opened 2025-08-16
refactor(router-core): fewer getMatch calls

Sheraff:
some miscellaneous minor optimizations in the `loadMatches` pipeline.
- try and reduce the number of times we call `getMatch`

### DIFF
--- a/packages/router-core/src/load-matches.ts
+++ b/packages/router-core/src/load-matches.ts
@@ -5,7 +5,6 @@ import { isNotFound } from './not-found'
 import { rootRouteId } from './root'
 import { isRedirect } from './redirect'
 import type { NotFoundError } from './not-found'
-import type { ControlledPromise } from './utils'
 import type { ParsedLocation } from './location'
 import type {
   AnyRoute,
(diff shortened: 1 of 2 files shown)


packages/router-core/src/load-matches.ts
```diff
@@ -324,20 +317,20 @@ const shouldExecuteBeforeLoad = (
...
+      result = true
     } else if (
       match.preload &&
       (match.status === 'redirected' || match.status === 'notFound')
     ) {
       handleRedirectAndNotFound(inner, match, match.error)
     }
-    return shouldExecuteBeforeLoad
+    return result
   }
```
dsonet, line 334: the result is always `true` here.

Sheraff: Damn you're right. The PR that made this change is this one though: https://github.com/TanStack/router/pull/4961. I'll try and see if I can track down the original logic, but if you know what it's supposed to be already, I'm all ears!

Sheraff: Actually it looks like it's been always `true` since this older PR https://github.com/TanStack/router/pull/4550/files (2 months ago). So I think we can at least say that it's not causing things to break. But maybe it is sub-optimal and we're executing too many `beforeLoad` calls. @schiller-manuel do you remember this a little?

Sheraff: I have a PR for the cleanup https://github.com/TanStack/router/pull/4992, and a follow up (wip) for actually re-thinking what *should* happen there https://github.com/TanStack/router/pull/4993

PR #4980, opened 2025-08-17
refactor(router-core,react-router,solid-router): avoid copying arrays just to .reverse() them

Sheraff:
The pattern of copying an array just to `.reverse()` it and iterate it from the end ends up creating a transient object (a copy of the array) that needs to be garbage collected.

For example:
```ts
[...array].reverse().find(() => {/*...*/})
```
or
```ts
[...array].reverse().forEach(() => {/*...*/})
```

Instead we can use a regular `for` loop to iterate it from the end. This will be mostly beneficial in `router-core > router > buildLocation` that gets called pretty often.

## Summary by CodeRabbit

- **Refactor**
  - Improved head metadata processing in React and Solid routers to be more efficient and readable, reducing work for large route metadata sets.
  - Simplified route matching logic to consistently pick the last relevant match, improving clarity and maintainability.

- **Chores**
  - Added a utility to locate the last list item meeting a condition, supporting cleaner routing internals.

### DIFF
--- a/packages/router-core/src/utils.ts
+++ b/packages/router-core/src/utils.ts
@@ -485,13 +485,12 @@ export function isPromise<T>(
 }
 
 export function findLast<T>(
-  array: Array<T>,
+  array: ReadonlyArray<T>,
   predicate: (item: T) => boolean,
 ): T | undefined {
   for (let i = array.length - 1; i >= 0; i--) {
-    if (predicate(array[i]!)) {
-      return array[i]!
-    }
+    const item = array[i]!
+    if (predicate(item)) return item
   }
   return undefined
 }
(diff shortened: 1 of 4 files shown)


packages/router-core/src/utils.ts
```diff
@@ -483,3 +483,14 @@ export function isPromise<T>(
...
+export function findLast<T>(
+  array: ReadonlyArray<T>,
+  predicate: (item: T) => boolean,
+): T | undefined {
+  for (let i = array.length - 1; i >= 0; i--) {
+    const item = array[i]!
+    if (predicate(item)) return item
+  }
+  return undefined
+}
```
levrik, line 496: Is there anything that speaks against using the built-in?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast

Sheraff: `.findLast` is still a little bit recent (safari 15.4), and I didn't want this PR to be a breaking change. But I'd love to use the built-in array method instead.

PR #4985, opened 2025-08-17
refactor(router-core): simplify defaultParseSearch & defaultStringifySearch

Sheraff:
Wait for this PR before merging: https://github.com/TanStack/router/pull/4987. It adds unit tests that should help make sure we're not breaking anything here.

--- 

The functions used to serialize and deserialize search params are called *a lot*, but they are creating many unnecessary objects. This PR refactors
- `encode` to be only as generic as what we need => in practice we only need to handle 1 value per key (i.e. `.set()` and not `.append()`) so we don't need a special case for arrays
- `decode` can avoid creating many arrays
- `stringifySearchWith` (and `defaultStringifySearch`) don't need to pre-stringify the input into a `<string,string>` dictionary as `encode` now accepts a stringifier function

## Summary by CodeRabbit

- Breaking Changes
  - Removed encode and decode from React Router and Solid Router public exports.
  - Query string prefix support removed in encode/decode.
- Refactor
  - Updated query string handling: arrays encode as comma-separated values; duplicate keys decode into arrays.
  - Streamlined error handling and avoided input mutation; consistent leading “?” management.
- Tests
  - Updated tests to match new encoding/decoding behavior and removed prefix-related cases.


packages/router-core/tests/qss.test.ts
```diff
@@ -23,7 +17,7 @@ describe('encode function', () => {
   it('should handle encoding an object with array values', () => {
     const obj = { token: ['foo', 'bar'], key: 'value' }
     const queryString = encode(obj)
-    expect(queryString).toEqual('token=foo&token=bar&key=value')
+    expect(queryString).toEqual('token=foo%2Cbar&key=value')
```
SeanCassiere, line 20: In its current form, this change to how arrays are represented in stringified format, would break any existing bookmarks. This'd be a problem and a breaking change.

Plus, the general accepted spec for having the same key (i.e Array values) be repeated multiple times in the query parameters are as such.

```
?genre=Pop&genre=RNB&year=2005
```

packages/router-core/src/searchParams.ts
```diff
@@ -21,8 +21,8 @@ export function parseSearchWith(parser: (str: string) => any) {
       if (typeof value === 'string') {
         try {
           query[key] = parser(value)
-        } catch (err) {
-          //
+        } catch {
```
SeanCassiere: Better to underscore prefix it to silence any eslint warning, since having easy access to `console.log`-ing this value during dev would be nice.

```suggestion
        } catch (_err) {
```

packages/router-core/src/searchParams.ts
```diff
@@ -35,40 +35,29 @@ export function stringifySearchWith(
   stringify: (search: any) => string,
   parser?: (str: string) => any,
 ) {
+  const hasParser = typeof parser === 'function'
   function stringifyValue(val: any) {
     if (typeof val === 'object' && val !== null) {
       try {
         return stringify(val)
-      } catch (err) {
+      } catch {
```
SeanCassiere:
```suggestion
      } catch (_err) {
```

packages/router-core/src/searchParams.ts
```diff
@@ -35,40 +35,29 @@ export function stringifySearchWith(
...
       }
-    } else if (typeof val === 'string' && typeof parser === 'function') {
+    } else if (hasParser && typeof val === 'string') {
       try {
         // Check if it's a valid parseable string.
         // If it is, then stringify it again.
         parser(val)
         return stringify(val)
-      } catch (err) {
+      } catch {
```
SeanCassiere:
```suggestion
      } catch (_err) {
```

packages/router-core/tests/qss.test.ts
Sheraff: I removed `encode` support for array values because in practice we always provide a custom stringification. This should not represent a change in the public API.

For example, what we do looks like this:
```ts
encode({
  token: stringify(['foo', 'bar'])
})
```
So I don't think this is an issue

SeanCassiere: What concerns me, is what gets printed out to the end-user's URL bar and what our "sensible defaults" are.

So, at runtime, will inputting `search = { genres=[Pop,Rock] }` change from `?genres=Pop&genres=Rock` to something else (i.e `?genres=Pop%2CRock`)?

- **If so**, then this is a **breaking change**.
- **If not,** then my bad. The way I interpreted the test case, gave me the impression that this had an effect on the defaults and what get's output into the URL bar.

Sheraff: No, there should not be any change visible to the user. This is because `encode` is actually wrapped inside of `stringifySearchWith`.

The tests here show what `encode` would output without a custom `stringify` function. But `stringifySearchWith` provides a custom function.

The other PR i opened (https://github.com/TanStack/router/pull/4987) adds tests for what the user will actually see. That way we can make sure this PR doesn't change that.

SeanCassiere: Alrighty 👍🏼

As long as the user sees no changes to any of their existing bookmarks, then all good from my side!

PR #4987, opened 2025-08-17
test(router-core): add search param serialization / deserialization unit tests

Sheraff:
Just some tests for `defaultParseSearch` and `defaultStringifySearch`, because they're used a lot and are very public and can handle user inputs (as in final human users, not devs).

These tests are just documentation of the current behavior.

### DIFF
--- a/packages/router-core/tests/searchParams.test.ts
+++ b/packages/router-core/tests/searchParams.test.ts
@@ -1,0 +1,92 @@
+import { describe, expect, test } from "vitest"
+import { defaultParseSearch, defaultStringifySearch } from "../src"
+
+describe("Search Params serialization and deserialization", () => {
+	/*
+	 * JSON-compatible objects can be serialized into a string,
+	 * and then deserialized back into the original object.
+	 */
+	test.each([
+		[{}, ''],
+		[{ foo: '' }, '?foo='],
+		[{ foo: "bar" }, '?foo=bar'],
+		[{ foo: "bar baz" }, '?foo=bar+baz'],
+		[{ foo: 123 }, '?foo=123'],
+		[{ foo: '123' }, '?foo=%22123%22'],
+		[{ foo: true }, '?foo=true'],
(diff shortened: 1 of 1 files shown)


packages/router-core/tests/searchParams.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  test('[edge case] inputs that are not primitive objects', () => {
+    expect(defaultStringifySearch(new Number(99))).toEqual('')
+    expect(defaultStringifySearch({ foo: new Number(99) })).toEqual('?foo=99')
+    expect(defaultStringifySearch(new String('foo'))).toEqual('?0=f&1=o&2=o')
+    expect(defaultStringifySearch(new Promise(() => {}))).toEqual('')
+    expect(defaultStringifySearch({ foo: new Promise(() => {}) })).toEqual(
+      '?foo=%7B%7D',
+    )
+    expect(defaultStringifySearch([1])).toEqual('?0=1')
+    const date = new Date()
```
SeanCassiere: Please stabilize this date value.

Something where the same date is used over and over again.

```suggestion
    const date = new Date('2024-11-18')
```

packages/router-core/tests/searchParams.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    expect(defaultStringifySearch(date)).toEqual('')
+    expect(defaultStringifySearch({ foo: date })).toEqual(
+      `?foo=%22${encodeURIComponent(date.toISOString())}%22`,
```
SeanCassiere: Similarly, please have the expected value be listed 'in full' as a string. It'd be better for our tests on this to show us exactly what's expected for a certain transformation.

Sheraff:
@SeanCassiere thank you for the review. Can you think of any more weird cases that might be missing from these tests?

SeanCassiere (approved):
@Sheraff nope, these tests seems pretty good to me! Plus, they're laid out in such a way that adding more would be easy.

Issue #4572, opened 2025-07-03
Incorrect `preload` & `cause` for the root's `beforeLoad`' arguments when `defaultPreload: "intent"`


larrifax:
### Which project does this relate to?

Router

### Describe the bug

With `defaultPreload: "intent"` set on the router, the `preload` & `cause` arguments of the root's `beforeLoad` are incorrect when hovering `Link`.

There seems to be multiple things that are wrong here:
1. After initial load, hovering a `Link` causes `beforeLoad` to be triggered on the root with `cause: enter` - even on a link going to the currently active route. This behavior is inconsistent with the behavior after clicking a link, in which `cause: stay` is what's reported.
2. `preload: false` is always reported for the root's `beforeLoad` when hovering a `Link`. I would expect `preload: true` to be reported on a hover.
3. Going to a nested route, and then hovering on the link to the same route reports incorrect `cause: "enter", preload: false` on both the parent and on the nested route.

### Your Example Website or App

https://stackblitz.com/edit/github-9xzbqucd?file=src%2Fmain.tsx

## Reproduction for 1. above: 
1. Hover the `Home` and `About` links.
2. Observe that the console reports the following:
```
root beforeLoad enter false        (expected "stay" - or maybe "preload"?)
index beforeLoad enter false      (expected "stay" - or maybe "preload"?)
root beforeLoad enter false        (expected "stay" - or maybe "preload"?)
about beforeLoad preload true
```
3. Click the `Home` link.
4. Hover the `Home` link.
5. Observe that the console reports the following:
```
root beforeLoad stay false
index beforeLoad stay false
```

## Reproduction for 2. above:
1. Hover the `About` link.
2. Observe that the console reports `root beforeLoad enter false` (expected `preload: true`)

## Reproduction for 3. above:
1. Click the `Nested` link.
2. Hover the `Nested` link.
3. Observe that the console reports the following:
```
root beforeLoad stay false        (expected `cause: "stay", preload: true` - or maybe `cause: "preload", preload: true`?)
bag beforeLoad enter false       (expected `cause: "stay", preload: true` - or maybe `cause: "preload", preload: true`?)
nested beforeLoad enter false  (expected `cause: "stay", preload: true` - or maybe `cause: "preload", preload: true`?)
```

### Expected behavior

See notes in parentheses in the repro steps.

### Platform

- Router / Start Version: 1.124.0
- OS: Linux
- Browser: Chrome
- Browser Version: 138.0
- Bundler: vite
[...]

PR #4993, opened 2025-08-18
test(router-core): add initial tests for router behavior

Sheraff:
First PR to add framework agnostic unit tests covering router behavior. We need to add many more (and eventually lighten the tests in each framework).

This PR adds tests about when a beforeLoad and a loader should execute.

Next PR should add tests about how context is passed.

Look at these issues to see what should be tested: https://github.com/TanStack/router/issues/4696

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests validating preload behavior for route loading hooks (beforeLoad and loader), covering resolved, pending, and rejected outcomes including notFound and redirects.
  * Verifies navigation control flow, hook invocation counts, pending vs cached matches, final locations, and loader data handling.
  * Includes sibling-route "stay" scenarios and timing helpers to strengthen async/timing reliability.


larrifax:
Would it make sense to also include tests for the following?

1. the expected calls of a parent's `beforeLoad` in case of navigation to a nested route (url: `parent/child`)
2. calls to the rootRoute's `beforeLoad` when navigating to e.g. `/foo` (don't know if root has special handling or not)

Reason I'm asking is that I've filed https://github.com/TanStack/router/issues/4572, where it seems like `beforeLoad` calls down the route tree has differing internal logic for calculating what parameters they supply to the `beforeLoad` invocations - which leads me to think that this PR would benefit from having tests checking correct `beforeLoad`-invocations on more than just the leaf nodes in the "match-tree".

Sheraff:
> Would it make sense to also include tests for the following?
> 
> 1. the expected calls of a parent's `beforeLoad` in case of navigation to a nested route (url: `parent/child`)
> 2. calls to the rootRoute's `beforeLoad` when navigating to e.g. `/foo` (don't know if root has special handling or not)
> 
> Reason I'm asking is that I've filed #4572, where it seems like `beforeLoad` calls down the route tree has differing internal logic for calculating what parameters they supply to the `beforeLoad` invocations - which leads me to think that this PR would benefit from having tests checking correct `beforeLoad`-invocations on more than just the leaf nodes in the "match-tree".

Definitely, I think tests inside `router-core` should be consolidated to validate more behaviors. For now I only wrote tests for "should we call the beforeLoad or not" because that's what I'm working on.

Sheraff:
> Would it make sense to also include tests for the following?
> 
> 1. the expected calls of a parent's `beforeLoad` in case of navigation to a nested route (url: `parent/child`)
> 2. calls to the rootRoute's `beforeLoad` when navigating to e.g. `/foo` (don't know if root has special handling or not)
> 
> Reason I'm asking is that I've filed #4572, where it seems like `beforeLoad` calls down the route tree has differing internal logic for calculating what parameters they supply to the `beforeLoad` invocations - which leads me to think that this PR would benefit from having tests checking correct `beforeLoad`-invocations on more than just the leaf nodes in the "match-tree".

Is this what you had in mind? https://github.com/TanStack/router/pull/4993/commits/1bfb9f4feef692a7f1c7de4173346ef6d668d3c1

larrifax:
> > Would it make sense to also include tests for the following?
> > 
> > 1. the expected calls of a parent's `beforeLoad` in case of navigation to a nested route (url: `parent/child`)
> > 2. calls to the rootRoute's `beforeLoad` when navigating to e.g. `/foo` (don't know if root has special handling or not)
> > 
> > Reason I'm asking is that I've filed #4572, where it seems like `beforeLoad` calls down the route tree has differing internal logic for calculating what parameters they supply to the `beforeLoad` invocations - which leads me to think that this PR would benefit from having tests checking correct `beforeLoad`-invocations on more than just the leaf nodes in the "match-tree".
> 
> Is this what you had in mind? [1bfb9f4](https://github.com/TanStack/router/commit/1bfb9f4feef692a7f1c7de4173346ef6d668d3c1)

Yes! Looking good. Thanks 🙂

Issue #4998, opened 2025-08-19
Route component is rendered without route context from async beforeLoad with pendingComponent


longzheng:
### Which project does this relate to?

Router

### Describe the bug

When using a route with a pending component, the route's component is rendered without awaiting async `beforeLoad` leading to incomplete context data, leading to `undefined` errors.

For example with this route definition with a `beforeLoad` that has an awaited promise, a `pendingComponent` and a `component`

```tsx
export const Route = createFileRoute('/posts')({
  beforeLoad: async () => {
    console.log('beforeLoad');

    const test = await new Promise<string>((resolve) =>
      setTimeout(() => resolve('hello world'), 1000)
    );

    console.log('beforeLoad done');

    return {
      test,
    };
  },
  component: PostsLayoutComponent,
  pendingComponent: () => 'loading',
});
```

In the route component, when I try to use the context

```tsx
function PostsLayoutComponent() {
  console.log('route component render');
  const { test } = Route.useRouteContext();
  console.log('Route context test:', test);
```

the type of `test` is inferred as `string`, but it's rendered with `undefined` while the component is pending.

I'm not sure if the bug is either
- the route's `component` should not be rendered during pending state
- the `beforeLoad` should be awaited when there is a pending component

The **pending component is a significant factor** because if I remove the pending component, then this bug does not occur. The route component is always rendered with the correct context from awaited `beforeLoad`.

Note: the repro has `defaultPendingMs: 0` set to always show the pending component.

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-onocc5sg?file=src%2Froutes%2Fposts.route.tsx

### Steps to Reproduce the Bug or Issue

1. Open browser devtools console
2. In the preview, click on "Posts"
3. See console logs for route component rendering with missing context value for the property `test`

```
beforeLoad
route component render
Route context test: undefined
beforeLoad done
route component render
Route context test: hello world
````

### Expected behavior

I expect the route component to be rendered with the route context with `beforeLoad` awaited, as the types suggest.

### Platform

- Router / Start Version: 1.131.27
- OS: Windows
- Browser: Chrome
- Browser Version: 139.0.7258.67
- Bundler: vite
- Bundler Version: 6.3.6

longzheng:
@Sheraff I noticed you started working on a draft pr.

Do you know what is the reason behind the route component rendering while pending if there is a pending component?

Sheraff:
[...]

PR #4999, opened 2025-08-19
fix(router-core): await beforeLoad before rendering, even if preloaded

Sheraff:
fix https://github.com/TanStack/router/issues/4998

---

WIP

at the time of writing, we're on [v1.131.27](https://github.com/TanStack/router/commit/83ab17331ba465617c9cab3beac7b7bd3f8a7665), and as far as I can tell it's been like this since at least [v1.128.2](https://github.com/TanStack/router/commit/80e76377b568145e991ffab1d3c500c3924932e4). I didn't look further in the past, because maybe it's been like this forever.

### DIFF
--- a/packages/react-router/tests/loaders.test.tsx
+++ b/packages/react-router/tests/loaders.test.tsx
@@ -731,7 +731,7 @@ test('clears pendingTimeout when match resolves', async () => {
 })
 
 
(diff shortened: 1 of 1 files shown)


schiller-manuel:
closing in favor of https://github.com/TanStack/router/releases/tag/v1.140.1

PR #5002, opened 2025-08-20
fix(router-core): on navigate (with preload), route component isn't rendered with undefined context if beforeLoad is pending

### LINKED ISSUE #4998: Route component is rendered without route context from async beforeLoad with pendingComponent

longzheng:
Fixes #4998

### DIFF
--- a/packages/router-core/src/load-matches.ts
+++ b/packages/router-core/src/load-matches.ts
@@ -369,7 +369,7 @@ const executeBeforeLoad = (
   const parentMatchContext =
     parentMatch?.context ?? inner.router.options.context ?? undefined
 
-  const context = { ...parentMatchContext, ...match.__routeContext }
+  const context = { ...match.__beforeLoadContext, ...parentMatchContext, ...match.__routeContext }
 
   let isPending = false
   const pending = () => {
--- a/packages/react-router/tests/routeContext.test.tsx
+++ b/packages/react-router/tests/routeContext.test.tsx
@@ -3125,4 +3125,87 @@ describe('useRouteContext in the component', () => {
 
     expect(content).toBeInTheDocument()
   })
+
+  test("reproducer #4998 - on navigate (with preload), route component isn't rendered with undefined context if beforeLoad is pending", async () => {
+    const beforeLoad = vi.fn()
+    const select = vi.fn()
+    let resolved = 0
+    const rootRoute = createRootRoute()
+    const indexRoute = createRoute({
+      getParentRoute: () => rootRoute,
+      path: '/',
+      component: () => <Link to="/foo">To foo</Link>,
+    })
+    const fooRoute = createRoute({
+      getParentRoute: () => rootRoute,
+      path: '/foo',
+      beforeLoad: async (...args) => {
+        beforeLoad(...args)
+        await sleep(WAIT_TIME)
+        resolved += 1
+        return { foo: resolved }
+      },
+      component: () => {
+        fooRoute.useRouteContext({ select })
+        return <h1>Foo index page</h1>
+      },
+      pendingComponent: () => 'loading',
+    })
+    const routeTree = rootRoute.addChildren([
+      indexRoute,
+      fooRoute
+    ])
+    const router = createRouter({
+      routeTree,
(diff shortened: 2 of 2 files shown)


packages/router-core/src/load-matches.ts
```diff
@@ -369,7 +369,11 @@ const executeBeforeLoad = (
   const parentMatchContext =
     parentMatch?.context ?? inner.router.options.context ?? undefined
 
-  const context = { ...parentMatchContext, ...match.__routeContext }
+  const context = {
+    ...match.__beforeLoadContext,
```
schiller-manuel, line 373: this is not correct IMO as this will feed the previous beforeLoad context into beforeLoad again.

longzheng: @schiller-manuel yes that is the intention, as otherwise the route component will be rendered while the beforeLoad is resolving **without the beforeLoad context**, which leads to the runtime undefined issues in #4998 .

Or are you suggesting that the route component shouldn't be rendered at all while the beforeLoad is resolving? If so I couldn't pinpoint when that behaviour was changed since I went back to over a year ago and it seems to have been always the case.

paul-sachs: @schiller-manuel this also makes sense to me, FWIW. I'm not sure what the behaviour was before, but from a type perspective it definitely makes more sense to include the context generated by the route preload. Perhaps there's another approach that would fix this. Perhaps preload needs to load data into different context so it's unique compared to __beforeLoadContext? Feels a little clunky but this is blocking our usage of preload so i'll take any ideas. Alternatively, what if we have a flag to indicate that __beforeLoadContext was loaded by preload and only merge it then, would that satisfy?

paul-sachs: I've created https://github.com/TanStack/router/pull/5701 that I think solves the core issue by making sure we only merge when updating context for pending state after a preload. Ideally we could somehow not render component during pending but that's not how suspense works so at least this way we maintain the runtime typesafety of context.

schiller-manuel:
closed in favor of https://github.com/TanStack/router/releases/tag/v1.140.1

Issue #4973, opened 2025-08-16
Search Params as Actual State


Eliav2:
### Which project does this relate to?

Router

### Describe the bug

We tried to use TanStack Router search params as our primary global state. While the read story is great (`useSearch` + schema validation), the write and lifecycle aspects are too low-level. We ended up building a fairly involved layer to make search-params reliable, composable, and ergonomic. We’ve read ["Search Params Are State"](https://tanstack.com/blog/search-params-are-state) — we agree on the premise; the missing pieces are writes and cross-component coordination.

This issue proposes concrete primitives the router could provide to enable "search params as state" without userland workarounds.

### What exists today (and why it's not sufficient)

- Global encode/decode can be customized via `parseSearch`/`stringifySearch` (e.g., `parseSearchWith`, `stringifySearchWith`). This is helpful but applies globally, not per param.

```ts
const router = createRouter({
  routeTree,
  parseSearch: parseSearchWith((v) => v),
  stringifySearch: stringifySearchWith(String),
});
```

- Route-level `validateSearch` normalizes/validates incoming search values, but there’s no symmetric per-param encode on write.
- `navigate` supports a `mask` option to control visible URL segments, but masking is specified per navigation, not per param, and does not compose across components.

### Why the current API falls short

- **Last-write-wins within a render**: Multiple `navigate({ search: ... })` calls during the same render/commit will override each other; only the last one survives. Real apps routinely update different params from different components in the same render. We need deterministic, atomic batching across the tree.
  - Note: While `navigate({ search: (prev) => next })` exists, it does not provide atomic cross-component merging when several updates occur in the same commit; the final call still wins.

- **No per-param encode/decode contract**: `validateSearch` lets us normalize incoming values (decode), but there’s no symmetrical per-param encode path for writes. Apps often need compact encodings, readability encodings, or stable formats. Encode/decode must be defined per param and applied automatically on both read and write.
  - Note: We are aware of the global `parseSearch`/`stringifySearch` configuration (including the helpers `parseSearchWith`/`stringifySearchWith`). Those are useful, but they apply globally and cannot express heterogeneous per-param codecs that many apps require.

- **No persistence lifecycle**: There is no built-in way to persist param state to `localStorage`/`sessionStorage`, version it, and restore it on navigation/back/forward when the param re-enters scope. This is essential for UX continuity and offline-friendly flows.

[...]

PR #5004, opened 2025-08-20
feat: add search persistence middleware

NitsanCohen770:
## ✨ Summary

This PR adds a new middleware and supporting API for **search parameter persistence**:

- **`persistSearchParams` middleware**  
  Automatically saves and restores search params across navigations, with support for selective exclusion. Works seamlessly with SSR and remains framework-agnostic, with a React integration provided.

- **`getSearchPersistenceStore()` API**  
  Exposes a fully typed store for manual save, restore, and clear operations. Gives developers full control when they need to override the default middleware behavior.

- **Examples & Documentation**  
  Includes a new `search-persistence` React example (Users/Products routes) demonstrating how persistence works in practice. The example also shows how the store can be synchronized with `localStorage` for durability across reloads, while making it clear that other adapters (e.g. saving to a database in SSR environments) can be easily built on top of the same API.  

---

## 🚀 Why This Is Useful

- **Improved UX** – Users no longer lose their filters, pagination, or search state when navigating away and back.  
- **Flexibility** – Developers can choose which parameters persist, or clear them entirely, depending on the use case.  
- **Extensible** – The `getSearchPersistenceStore()` abstraction makes it trivial to integrate persistence into any storage layer:
  - Client-side (e.g. `localStorage`, session storage)  
  - Server-side (e.g. saving to a database or session store during SSR)  
- **Typed by default** – All persistence APIs automatically infer route-specific search schemas for safety and DX.  

---

## 📚 Additions

- `persistSearchParams` middleware in `@tanstack/router-core`  
- `getSearchPersistenceStore()` API with type inference  
- New React example: `examples/react/search-persistence`  
- Documentation for usage patterns, caveats, and extension points  

## Summary by CodeRabbit

- **New Features**
  - Search-parameter persistence: key-based middleware, typed per-route persistence store, router option to enable persistence, and APIs to save/restore/clear persisted searches.

- **Documentation**
  - New docs detailing middleware usage, props, exclusion/restore patterns, SSR guidance, manual store usage, and examples.

- **Chores**
  - New React examples (client and SSR) demonstrating persistence, localStorage sync, build/config files, and a type-inference test.


schiller-manuel:
hi there! we are right now in a feature freeze until the next alpha version is merged to main. we will revisit this PR after that then.

NitsanCohen770:
Noticed an issue: persisted search params are being saved to routes that weren’t intended.  
Cause: the middleware currently runs on both the `from` and `to` routes.  
Working on a fix ASAP.

schiller-manuel:
i am not sure what you are trying to do here is even possible with search middlewares as they run for links pointing to and from a route, not only when a route is visited

NitsanCohen770:
> i am not sure what you are trying to do here is even possible with search middlewares as they run for links pointing to and from a route, not only when a route is visited

Yes, this is exactly what I noticed.  

I’ve found a solution by making the search params you want to persist an explicit parameter:

```ts
export function persistSearchParams<TSearchSchema>(
  persistedSearchParams: Array<keyof TSearchSchema>,
  exclude?: Array<keyof TSearchSchema>,
): SearchMiddleware<TSearchSchema> {
```
Then you would use it like this and safely save only the intended params.

It’s not the best DX, but since we have both the persistedSearchParams param and the typed exclude list, it’s a reasonable balance.

Screenshot 2025-08-21 at 1 08 17

LadyBluenotes:
Thanks for the work on this. Search persistence is still a useful direction, but the current search-middleware behaviour and persistence contract have moved on enough that this branch is no longer safe to rebase.

We’re tracking the broader requirement in #4973, including persistence precedence, browser history, masking, and SSR lifecycle. I’m closing this implementation so that work can continue against the current search model.

PR #5016, opened 2025-08-23
Fix loader function for posts route example

coryhouse:
## Summary by CodeRabbit

* **Documentation**
  * Updated the React external data loading guide to include a concrete QueryClient instance in the TanStack Query example.
  * Ensures the loader and suspense query usage reference a real client and shared cache in the snippet.
  * Improves clarity and correctness of the example without altering any APIs or behavior.
  * No other content changes; examples remain functionally consistent with prior guidance.

### DIFF
--- a/docs/router/framework/react/guide/external-data-loading.md
+++ b/docs/router/framework/react/guide/external-data-loading.md
@@ -75,7 +75,7 @@ Let's take a look at a more realistic example using TanStack Query.
 
 ```tsx
 // src/routes/posts.tsx
-const queryClient = new QueryClient();
+const queryClient = new QueryClient()
 
 const postsQueryOptions = queryOptions({
   queryKey: ['posts'],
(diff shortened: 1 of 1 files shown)


docs/router/framework/react/guide/external-data-loading.md
```diff
@@ -75,6 +75,8 @@ Let's take a look at a more realistic example using TanStack Query.
 
 ```tsx
 // src/routes/posts.tsx
+const queryClient = new QueryClient()
```
schiller-manuel, line 78: you wouldnt create the queryClient instance in a specific route

coryhouse:
Yes, good point. I opened this because I was unclear how to access the `queryClient` in this location. Is the recommendation to export the queryClient when it's instantiated so it can be accessed elsewhere in spots where `useQueryClient` can't be called (because a hook isn't valid outside a React component)?

schiller-manuel:
typically you would add the queryClient to the router context, see for example here https://github.com/TanStack/router/blob/main/examples/react/basic-react-query-file-based/src/main.tsx#L14

PR #5018, opened 2025-08-23
refactor(router-core): flatten loadRouteMatch, can run synchronously

Sheraff:
This PR flattens the `loadRouteMatch` in `load-matches.ts`. This makes it easier to understand what code runs in which conditions, and makes it possible to not return a promise at all in some situations.

## Summary by CodeRabbit

- **Refactor**
  - Optimized route loading to handle synchronous and asynchronous results more efficiently, improving parallel loading and reducing blocking during navigation.
  - Centralized finalization of route loads for more consistent state updates (fetching, invalidation, timeouts) and smoother transitions.
  - Streamlined preload behavior to align with the unified loading flow, maintaining stale-while-revalidate semantics.
  - Internal performance and reliability improvements with no changes to the public API.

### DIFF
--- a/docs/router/framework/react/api/router/RouteOptionsType.md
+++ b/docs/router/framework/react/api/router/RouteOptionsType.md
@@ -131,9 +131,7 @@ type loader = (
     location: ParsedLocation
     params: TAllParams
     preload: boolean
-    parentMatchPromise:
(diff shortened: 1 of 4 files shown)


LadyBluenotes:
I’ve opened #7791  to preserve the optimization goal rather than try to rebase it through the current match-loading changes.

PR #5022, opened 2025-08-26
Explain how to use router context in react components

Floriferous:
The docs for router context explain how to set it up, and how to use it in loaders, but it doesn't explain how to simply use it in react components. I'm not even sure this is the right way to do it?

Happy to make it better if necessary.

## Summary by CodeRabbit

* **Documentation**
  * Added two practical examples to the React guide’s "Using the Router Context" section demonstrating how to access and render values from a route context (e.g., a user ID).
  * Examples are placed in two locations for easier discovery and reference.
  * No runtime behavior or public API changes.

### DIFF
--- a/docs/router/framework/react/guide/router-context.md
+++ b/docs/router/framework/react/guide/router-context.md
@@ -106,7 +106,7 @@ import { useRouteContext } from '@tanstack/react-router'
 export const Route = createFileRoute('/todos')({ component: Todos });
 
 function Todos() {
-  const routeContext = useRouteContext({ from: Route.id });
+  const routeContext = Route.useRouteContext();
   return <div>Todos from {routeContext.user.id}</div>
 }
 ```
--- a/docs/router/framework/react/guide/router-context.md
+++ b/docs/router/framework/react/guide/router-context.md
@@ -101,13 +101,13 @@ export const Route = createFileRoute('/todos')({
 You can use it in your components with the `useRouteContext` hook:
 
 ```tsx
-import { useRouteContext } from '@tanstack/react-router'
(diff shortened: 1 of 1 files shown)


Floriferous:
This is also a helpful pattern, but again maybe there is already something built-in?

```tsx
export const Route = createRootRouteWithContext<MyRouterContext>()({
  component: RootComponent, // Initialized router context in here
});

// Add this to easily access your root context across your app
export function useRootRouteContext() {
  return Route.useRouteContext();
}
```

Can add this in if it makes sense.

Floriferous:
Or maybe the best way is this:

```tsx
import { useRouteContext } from '@tanstack/react-router';

export function useRootRouteContext() {
  return useRouteContext({ from: "__root__" });
}
```

Tustar1759 (approved):

Tustar1759 (approved):

docs/router/framework/react/guide/router-context.md
```diff
@@ -98,6 +98,19 @@ export const Route = createFileRoute('/todos')({
 })
 ```
 
+You can use it in your components with the `useRouteContext` hook:
+
+```tsx
+import { createFileRoute, useRouteContext } from '@tanstack/react-router';
```
brenelz, line 104: I don't think you need this import since you are using `Route.useRouteContext`

LadyBluenotes:
Thanks for identifying this missing example. The docs structure has changed since this PR was opened, so the patch no longer applies directly.

The component usage gap still exists, and I’ve opened #7792  to track it against the current guide.

PR #5027, opened 2025-08-27
fix(react-router): ensure consistent disabled prop handling across Link variants

doandevhere:
Hi TanStack Router Team,

**Problem**
- Like Ant Design provide CSS styles that target disabled anchor elements using selectors like a[disabled]. 
- However, TanStack Router's current Link component implementation removes the disabled prop when rendering default <a> elements, breaking compatibility with these UI library styles
image

**Current behavior**:
- `Link` component: disabled prop gets deleted
- `createLink` with custom components: disabled prop is preserved
=> This inconsistency means developers must use createLink to get proper disabled styling from UI libraries, creating an inconsistent developer experience.

**Reproduce codebase**

```tsx
// Bug reproduction scenario  
function BugReproduction() {  
  // Case 1: Default Link (disabled prop gets deleted)  
  const DefaultLink = () => (  
    <Link to="/test" disabled={true}>  
      Default Link - disabled prop will be removed  
    </Link>  
  )  
    
  // Case 2: Custom Link (disabled prop is preserved)  
  const CustomButton = createLink('a')  
  const CustomLink = () => (  
    <CustomButton to="/test" disabled={true}>  
      Custom Link - disabled prop will be kept  
    </CustomButton>  
  )  
    
  // Expected: Both should handle disabled consistently  
  // Actual: Only CustomLink preserves the disabled prop  
}
```

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -536,12 +536,6 @@ export const Link: LinkComponent<'a'> = React.forwardRef<Element, any>(
           })
         : rest.children
 
-    if (_asChild === undefined) {
-      // the ReturnType of useLinkProps returns the correct type for a <a> element, not a general component that has a disabled prop
-      // @ts-expect-error
-      delete linkProps.disabled
-    }
-
     return React.createElement(
       _asChild ? _asChild : 'a',
       {


Sheraff:
`disabled` is not a valid attribute for an anchor.

> The `disabled` attribute is supported by `<button>`, `<fieldset>`, `<optgroup>`, `<option>`, `<select>`, `<textarea>` and `<input>`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled

I don't think it would be a good idea for tanstack/router to output invalid HTML by default. The "proper" way to disable an anchor is to remove the `href`. And the add some accessibility attributes.
```html
<a role="link" aria-disabled="true">
```

doandevhere:
> `disabled` is not a valid attribute for an anchor.
> 
> > The `disabled` attribute is supported by `<button>`, `<fieldset>`, `<optgroup>`, `<option>`, `<select>`, `<textarea>` and `<input>`.
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled
> 
> I don't think it would be a good idea for tanstack/router to output invalid HTML by default. The "proper" way to disable an anchor is to remove the `href`. And the add some accessibility attributes.
> 
> ```
> <a role="link" aria-disabled="true">
> ```
```tsx
  // Case 2: Custom Link (disabled prop is preserved)  
  const CustomButton = createLink('a')  
  const CustomLink = () => (  
    <CustomButton to="/test" disabled={true}>  
      Custom Link - disabled prop will be kept  
    </CustomButton>  
  )  
```
@Sheraff Agree but in case 2 we still have `disabled` prop. I think if we remove disabled in above element you listing. We should remove both. If we hold we should hold both. What do you think about this case?

btw, I think as a routing library. We should forward `disabled` to DOM element is fine. A user(developer using `react-router`) will make decision when should hold or move. Omit `disabled` in internal library is correct but not good to me

Thanks

a463:
In my opinion, the `disabled` prop should be removed from the `Link` component. Ideally, any router should provide a `Link` component that's as close as possible to a native `<a>` element. This would make the implementation more flexible, and a `disabled` prop can easily be handled in a higher level component, if that's a requirement.

I see there's some logic to disable preloading if the link is disabled, but that is probably there only because `Link` currently accepts the `disabled` prop. Wherever the user of `Link` implements the `disabled` logic, when it's `true`, they can easily pass `preload={false}`.

PR #5037, opened 2025-08-28
fix(react-router, solid-router): reduce useNavigate reactivity on path changes

nlynzaad:
This is a follow-up to #4978.

As picked up by @Sheraff the changes made caused useNavigate to negatively impact performance specifically with useNavigate being too reactive to path changes.

This reverts the use of useActiveLocation, that acted as a single implementation of the logic to determine the from location, and moves the determination of the from path back to each of Link and useNavigate. While the core logic is still the same across the implementations the implication of how it's used and the sources of the required inputs are different enough that it warrants separate implementations. 

This PR does not revert the full change set and still applies to core reasons and changes for the previous commit in keeping the logic for determining the from clause the same, i.e. the order of determining the from path for navigation remains: from clause => current location => rendered location. 

@Sheraff to advise if this implementation resolves the issues he picked up.

## Summary by CodeRabbit

* Breaking Changes
  * Removed active-location hooks and their associated types from React and Solid packages.
* Refactor
  * Link and navigation now accept an explicit "from" option instead of deriving it from active-location state.
* New Features
  * Added a programmatic Navigate helper for Solid to trigger navigation on mount.
* Bug Fixes
  * More predictable fallback behavior when no explicit "from" is provided.
* Performance
  * Reduced state subscriptions for minor efficiency gains.


packages/react-router/src/useNavigate.tsx
```diff
@@ -17,19 +18,32 @@ export function useNavigate<
...
+        _buildLocation: false,
+      })
+
+      const activeLocationMatch = last(activeLocationMatches)
+
+      const from =
+        options.from ??
+        _defaultOpts?.from ??
+        activeLocationMatch?.fullPath ??
+        router.state.matches[matchIndex]!.fullPath
```
Sheraff: This is tricky because `matchIndex` here might become stale because we're closing over the value the 1st time `useCallback` is called.
- Either we need to add it to the dependency array, but that means `navigate` won't be stable (which was the original issue)
- Or we need to obtain `matchIndex` from inside the `useCallback` based solely on `router` (which itself *is* stable).

If we want to obtain `matchIndex` from inside the `useCallback`, we need to replicate the logic from `packages/react-router/src/useMatch.tsx`:
```ts
const nearestMatchId = React.useContext(matchContext)
const match = router.state.matches.find((d: any) => d.id === nearestMatchId)
```

But now we need to unwrap the value from `matchContext`, so we have the same issue of a reactive value that needs to come from outside the `useCallback`. So we can loot at where this value comes from to see if we can obtain it another way. We find it in `packages/react-router/src/Match.tsx`:
```tsx
<matchContext.Provider value={matchId}>
```
But based on how it's used in there, it looks like it just refers to an element in the `state.matches` array:
```ts
  const parentRouteId = useRouterState({
    select: (s) => {
      const index = s.matches.findIndex((d) => d.id === matchId)
      return s.matches[index - 1]?.routeId as string
    },
  })
```

Based on your explanation of the issue:
> rendered in a layout route, and should be applied on a child route which may or may not have path

It makes me think that we don't actually want the `matchId` (which could refer to the layout route) but instead simply the *last* match:
```ts
last(router.state.matches)!.fullPath
```

--- 

If you agree with the above, now I'm wondering what the difference is between
```ts
last(router.matchRoutes(router.latestLocation)).fullPath
```
and 
```ts
last(router.state.matches)!.fullPath
```
Do you know? Because that would mean in the list of possibilities for `const from =`, we would have 2 alternatives that are the same thing, which could simplify things quite a bit.

nlynzaad: These serve 2 different purposes in here.

The matchIndex is used here to return the rendered route, hence why we don't include it in the dependency array, we actually want it to be stale. router.latestLocation is used inside the callback to return the latest location.

The usage of matchIndex should theoretically never be hit as activeLocation should always exist.

This was the original implementation that is left over as an extreme case fallback. However, thinking about it now, the matchIndex fallback can probably be left out all together as buildLocation defaults to current location should from be null.

Sheraff: Ok interesting. And do you know exactly when 
```ts
last(router.matchRoutes(router.latestLocation)).fullPath
```
would be different from
```ts
last(router.state.matches)!.fullPath
```
?

Because `matchRoutes` is quite expensive to run, whereas just getting the last match is basically free.

Just testing quickly, I can only find 1 time when they are different: while navigating, when the URL has already changed, but the mounted route is still the previous one (pending state), we observe
- `last(router.matchRoutes(router.latestLocation)).fullPath` => the next route
- `last(router.state.matches)!.fullPath` => the previous route

Is this the only difference? If it is, then I'd argue that `last(router.state.matches)!.fullPath` might be the better choice, as it's cheaper, and matches what's on screen.

nlynzaad: actually don't think any of these are necessary. further down stream in router.buildLocation we already default to the current location if no from is passed. this is how router.navigate runs. I removed all that activeLocation, matchIndex logic now and all tests passed. 

Just pushed an update for this, kept the callback in useNavigate to ensure its stable but this is now much simpler.

Sheraff: Oh I love that solution! The best performance is when you do nothing! ah ah

Sheraff (approved):

packages/react-router/src/useNavigate.tsx
nlynzaad: Haha great.

nlynzaad:
Todo: add some e2e tests before merging. Will merge later tonight

PR #5038, opened 2025-08-28
fix(router-ssr): ensure charset meta is placed before dehydration script

jsedlacek:
# Summary

When server-rendering with dehydration, the router currently injects its dehydration script immediately after the `<head>` tag, before any `<meta charSet="utf-8"/>` declaration. This causes non-ASCII characters in loader data (e.g., Czech, Chinese, emoji) to be misinterpreted in some browsers since the charset is not yet established.

# Fix

Updated the SSR stream transformation logic so that the dehydration script is injected only after the charset meta tag has been written. This guarantees compliance with the HTML5 requirement that the charset declaration appears before content requiring specific encoding:

* [[HTML Living Standard §4.2.5.3](https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-charset)](https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-charset)
* [MDN: `<meta charset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta#charset)

# Test

Added a regression test at:

* `router/e2e/react-start/basic/tests/charset-encoding.spec.ts`

The test verifies that the charset meta tag now appears before the dehydration script in prerendered output.

# Impact

This resolves potential encoding issues for any TanStack Start app with non-ASCII characters in loader data and ensures correct HTML behavior across browsers.

## Summary by CodeRabbit

- Bug Fixes
  - Ensures the router’s hydration script is injected in the head only once, after any charset meta tag (if present) and before the body, preventing encoding issues.

- Tests
  - Added end-to-end coverage for pages with and without a charset tag to verify correct script placement and HTML structure.

- Chores
  - Introduced configuration, build, and tooling setup for the new e2e app (Vite, Tailwind, PostCSS, Playwright).
  - Added ignore files to reduce noise in version control and formatting.


schiller-manuel:
what if the html does not contain that meta tag?

jsedlacek:
> what if the html does not contain that meta tag?

@schiller-manuel good point. I added a test for this scenario, and a small fix. I moved the test to separate e2e folder (e2e/react-start/encoding/tests/charset-encoding.spec.ts) since I did not find a way to remove the meta tag for a single route.

schiller-manuel:
i think the better solution to this is let react render the dehydration script. will look into implementing this once we are done with the next alpha branch

jsedlacek:
> i think the better solution to this is let react render the dehydration script. will look into implementing this once we are done with the next alpha branch

@schiller-manuel  If you're talking about the approach to testing, I have created unit tests instead of e2e tests in this branch: https://github.com/jsedlacek/router/commit/0dfe5b0c2371087a68a79f6d7da5b62dd8d8192d

I can remove the e2e test, and use this unit test instead.

schiller-manuel:
no i mean instead of injecting the script into the stream , let react render it as bootstrap script

schiller-manuel:
order is implemented in https://github.com/TanStack/router/pull/5870

PR #5049, opened 2025-08-30
Add support for setting <html lang> attribute in TanStack Router (React)

OutOfBoundCats:
Currently, TanStack Router provides a way to set head attributes such as <title> and <meta> tags. While building my website, I encountered a use case where I needed to set the lang attribute on the root <html> element.

This attribute is not currently supported by the router, so I’ve attempted to add support for it in the React adapter.

Key notes:

Introduces initial support for setting lang on <html>

The current code may or may not fully work as intended

Looking for feedback and guidance from maintainers/contributors to finish the job and align the approach with the project’s standards

I’m happy to refine the implementation based on any suggestions or alternative design ideas.

### DIFF
--- a/packages/react-router/src/ssr/renderRouterToString.tsx
+++ b/packages/react-router/src/ssr/renderRouterToString.tsx
@@ -13,6 +13,25 @@ export const renderRouterToString = async ({
 }) => {
   try {
     let html = ReactDOMServer.renderToString(children)
+
+    // Collect HTML attributes from all route matches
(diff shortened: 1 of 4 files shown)


schiller-manuel:
i would just set this on the `<html>` element in the root route. no need to hack around with the rendered HTML

OutOfBoundCats:
@schiller-manuel  You’re right — for some reason my mind sometimes gives up on the simpler solution 😅 Thanks a lot for the help!

PR #5083, opened 2025-09-04
Add `server-entry` export to `@tanstack/react-start`

jamesopstad:
Adds a `server-entry` export to experiment with.

### DIFF
--- a/e2e/e2e-utils/package.json
+++ b/e2e/e2e-utils/package.json
@@ -10,7 +10,7 @@
   "exports": {
     ".": {
       "import": "./src/index.ts",
-			"default": "./src/index.ts"
+      "default": "./src/index.ts"
     },
     "./package.json": "./package.json"
   },


packages/react-start/src/default-entry/server.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  createStartHandler,
+  defaultStreamHandler,
+} from '@tanstack/react-start/server'
+import type { RequestHandler } from '@tanstack/react-start/server'
+
+import { createRouter } from '#tanstack-start-router-entry'
+
+// Providing `RequestHandler` from `@tanstack/react-start/server` is required so that the output types don't import it from `@tanstack/start-server-core`
```
jamesopstad, line 9: Is there a better way to resolve this?

schiller-manuel: not aware of a better way

PR #5099, opened 2025-09-07
docs(Start): change cloudflare pages to workers on deployment target list

TomRadford:
Deployment target list was still referencing cloudflare pages instead of workers

## Summary by CodeRabbit

* **Documentation**
  * Updated React hosting guide to list Cloudflare Workers as the deployment target instead of Cloudflare Pages.
  * Adjusted in-page anchors and references to match the new deployment option.
  * Confirmed all other deployment targets remain unchanged.
  * Ensures readers are directed to the correct Cloudflare Workers guidance for deployment.


brenelz:
I think this PR is now out of date with the new release candidate as there is no more target property

PR #5108, opened 2025-09-09
fix(router-core): allow Route.to access before router initialization

### LINKED ISSUE #5090: `Route.to` may be undefined

Start

### Describe the bug

When `import {Route as SomeRoute} from "../some-route";` and then use `SomeRoute.to`, it may be undefined.

https://github.com/ImSingee/tanstack-start-bug-reproduce-for-route-to

1. Clone the repo, and start it, go to the main page http://localhost:6099/
2. See the navigation bar, the link doesn't work (because the `SomeRoute.to` is undefined)

### Expected behavior

The link works (the `SomeRoute.to` shouldn't be undefined)

### Screenshots or Videos

Image

### Platform

- Router / Start Version: 1.131.35
- OS: macOS
- Browser: Chrome
- Browser Version: 139.0.3405.125 (Official build) (arm64) 
- Bundler: Vite
- Bundler Version: 7.1.4

### Additional context

I think this may be a code split bug.

I found two ways to make my navigation links work.

First, in `src/routes/__root.tsx`, change `notFoundComponent` from `<Layout> <div>404 Not Found</div> </Layout>` to `<div>404 Not Found</div>` (remove the reference to `<Layout>` component). And refresh the page, the link will work correctly.

Another, in `src/routes/_app/-components/layout.tsx`, move the top-level `const menus = ...` into the function.

Change the following code:

```tsx
const menus = [
    {label: "Feedback", to: FeedbackRoute.to},
    {label: "Roadmap", to: RoadmapRoute.to},
] as const;

function Header() {
    return (
        <div style={{display: 'flex', gap: '1rem'}}>
            {menus.map((menu) => (
                <MenuItem key={menu.label} label={menu.label} to={menu.to}/>
            ))}
        </div>
    );
}
```

to 

```tsx
function Header() {
    const menus = [
        {label: "Feedback", to: FeedbackRoute.to},
        {label: "Roadmap", to: RoadmapRoute.to},
    ] as const;
    
    return (
        <div style={{display: 'flex', gap: '1rem'}}>
            {menus.map((menu) => (
                <MenuItem key={menu.label} label={menu.label} to={menu.to}/>
            ))}
        </div>
    );
}
```

Refresh the page, undefined gone.

joseph0926:
Fixes #5090 

### Problem

When accessing `Route.to` at module load time (e.g., in static arrays for navigation), it returns `undefined` instead of the expected path string. This occurs despite TypeScript correctly inferring the type (e.g., `AboutRoute.to: "/about"`).

```ts
// Module level
const navigation = [
  { label: 'About', to: AboutRoute.to }, // to is undefined
  { label: 'Contact', to: ContactRoute.to }, // to is undefined
]
```

### Root Cause

Route objects are created at module load time, but their path properties (`_to`, `_path`, etc.) are only initialized when the Router is created and calls `init()` on each route through `processRouteTree()`. This timing gap causes the issue.

The route objects are instantiated when the module is loaded, but their path-related fields (such as _to and _path) are not populated until the router is constructed. During module evaluation, any static arrays or constants that reference Route.to will therefore see it as undefined. Once the router is created, it invokes init() on each route via processRouteTree(), at which point the paths are computed. By the time the component renders, Route.to holds the correct value.

### Solution

The solution was to update the to getter so it computes the full path lazily when it is accessed before initialization. If _to has already been set by the initialization phase, the getter simply returns that value. Otherwise, it walks up the parent chain to assemble the complete path, stores the result, and thereafter serves it from the cache for minimal overhead.

### Changes

- Added lazy path computation to the `to` getter in `BaseRoute` class
- Added `_computedTo` field for caching pre-initialization computed paths
- Maintains backward compatibility - no breaking changes

### Tests

Added comprehensive test suite in `packages/react-router/tests/route-to-getter.test.tsx`
- Simple routes (`/about`)
- Nested routes (`/dashboard/settings/profile`)
- Index routes (`/dashboard/`)
- Dynamic parameters (`/users/$userId/posts/$postId`)
- Optional parameters (`/products/$category?/$productId?`)
- Wildcard routes (`/files/$`)
- Static array pattern (primary use case)

All existing tests pass without modification.

### Example

```ts
// Module level
const navigation = [
  { label: 'Home', to: homeRoute.to },      // '/'
  { label: 'About', to: aboutRoute.to },    // '/about'
  { label: 'Contact', to: contactRoute.to }, // '/contact'
]
```

### Notes

the documentation did not caution against this pattern, and the TypeScript definitions appeared to indicate that it should be supported. This approach is, in fact, a common way to declare navigation in React applications. The proposed solution imposes only a negligible performance cost because results are cached.

## Summary by CodeRabbit

- New Features
  - Route.to now lazily computes full route paths, yielding consistent results for nested, index, dynamic, optional, and wildcard routes with improved trailing-slash behavior.

- Bug Fixes
  - Route.to returns correct values both before and after router initialization; caches are invalidated when routes update so changes propagate.

- Tests
  - Added comprehensive tests validating Route.to resolution across many routing scenarios.

### DIFF
--- a/packages/react-router/tests/route-to-getter.test.tsx
+++ b/packages/react-router/tests/route-to-getter.test.tsx
@@ -190,4 +190,44 @@ describe('Route.to getter early access', () => {
     expect(navigation[1]?.to).toBe('/about')
     expect(navigation[2]?.to).toBe('/contact')
   })
+
+  test('Pathless parent routes should not affect child path', () => {
+    const rootRoute = createRootRoute({})
+
+    const layoutRoute = createRoute({
+      getParentRoute: () => rootRoute,
+      id: '_layout',
+    })
+
+    const childRoute = createRoute({
+      getParentRoute: () => layoutRoute,
+      path: '/child',
+    })
+
+    expect(childRoute.to).toBe('/child')
+
+    const routeTree = rootRoute.addChildren([
+      layoutRoute.addChildren([childRoute]),
+    ])
+
+    createRouter({
+      routeTree,
+      history: createMemoryHistory(),
+    })
+
+    expect(childRoute.to).toBe('/child')
+  })
+
+  test('Root route should always return "/" for to getter', () => {
+    const rootRoute = createRootRoute({})
+
+    expect(rootRoute.to).toBe('/')
+
+    createRouter({
+      routeTree: rootRoute,
+      history: createMemoryHistory(),
+    })
+
+    expect(rootRoute.to).toBe('/')
+  })
 })
--- a/packages/react-router/tests/route-to-getter.test.tsx
+++ b/packages/react-router/tests/route-to-getter.test.tsx
@@ -1,0 +1,193 @@
+import { describe, expect, test } from 'vitest'
+import {
+  createMemoryHistory,
+  createRootRoute,
+  createRoute,
+  createRouter,
+} from '../src'
+
+describe('Route.to getter early access', () => {
+  test('Route.to should be accessible before Router initialization', () => {
+    const rootRoute = createRootRoute({})
+    const aboutRoute = createRoute({
+      getParentRoute: () => rootRoute,
+      path: '/about',
+    })
+
+    const menuItems = [{ label: 'About', to: aboutRoute.to }]
+
+    expect(menuItems[0]?.to).toBe('/about')
+    expect(aboutRoute.to).toBe('/about')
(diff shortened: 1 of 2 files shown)


schiller-manuel:
importing Routes like this is not really a good idea. it can introduce circular type dependencies as well.
not sure we should add this

PR #5114, opened 2025-09-09
Update README

LadyBluenotes:
**Summary**:

- Updated the TanStack Router README to improve feature overview and descriptions.
- Added information for TanStack Start, including a card section consistent with Route.
- Refined wording for clarity and consistency across product blurbs and feature lists.
- Updated header images so they're consistent.


LadyBluenotes:
Just as a note for this: the image sources aren't hard-coded like previously so that will need to be updated.

PR #5115, opened 2025-09-10
fix(router): respect `target` attribute on inner component

### LINKED ISSUE #5011: `target` is not respected on link components created with `createLink`

Router

### Describe the bug

We have a generic link type in our project which accepts `openInNewTab` as a prop and sets `target` and `rel`:

```tsx
// Details omitted
const MyLink = ({ openInNewTab, href }) => (
	<a
		href={href}
		{...(openInNewTab ? { target: "_blank", rel: "noopener noreferrer" } : {})}
	/>
)

const CreatedMyLink = createLink(MyLink)
```

When we create a link with `openInNewTab` set, it will not open in a new tab and trigger `useBlocker`.

To actually make this work, we have to wrap the created link component again:

```tsx
const Wrapped: LinkComponent<typeof MyLink> = (props) =>
    <CreatedMyLink {...props} target="_blank" />
```

This does work, but since we aren't on react 19 yet, we have to forward the ref. This breaks types and we have to cast the resulting component:

```tsx
const Wrapped: LinkComponent<typeof MyLink> = forwardRef((props, ref) => 
    <CreatedLink ref={ref} {...props} target="_blank" />
) as LinkComponent<typeof MyLink>
```

https://stackblitz.com/edit/github-bjdswd1c?file=src%2Froutes%2Findex.tsx

The index page shows all cases I tried and has comments explaining them.

### Expected behavior

I think ideally, setting `target` on the inner component would be respected.

### Screenshots or Videos

_No response_

### Platform

- Router Version: 1.131.22
- OS: Linux
- Browser: Firefox
- Browser Version: 141.0
- Bundler: rsbuild
- Bundler Version: 1.4.11

### Additional context

_No response_

leesb971204:
fixes #5011 

## Summary by CodeRabbit

- Bug Fixes
  - Links now correctly respect the target attribute from rendered anchor elements when not explicitly provided as a prop. Clicks on links with target=_blank are no longer intercepted for client-side navigation, opening in a new tab as expected.
  - Explicit target props still override defaults from custom link components, ensuring predictable behavior.
  - Applies to both React Router and Solid Router link components, improving consistency with custom link implementations and browser defaults.

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -218,11 +218,15 @@ export function useLinkProps<
 
   // The click handler
   const handleClick = (e: React.MouseEvent) => {
+    // Check actual element's target attribute as fallback
+    const elementTarget = (e.currentTarget as HTMLAnchorElement).target
+    const effectiveTarget = target !== undefined ? target : elementTarget
+
     if (
       !disabled &&
       !isCtrlEvent(e) &&
       !e.defaultPrevented &&
-      (!target || target === '_self') &&
+      (!effectiveTarget || effectiveTarget === '_self') &&
       e.button === 0
     ) {
       e.preventDefault()
--- a/packages/react-router/tests/link.test.tsx
+++ b/packages/react-router/tests/link.test.tsx
@@ -4683,6 +4683,117 @@ describe('createLink', () => {
     const button3 = await screen.findByText('active: no - foo: no - Button3')
     expect(button3.getAttribute('overrideMeIfYouWant')).toBe('Button3')
   })
+
+  it('should respect target attribute set by custom component', async () => {
+    const CustomLinkWithTarget = React.forwardRef<
+      HTMLAnchorElement,
+      { href?: string; children?: React.ReactNode }
+    >((props, ref) => (
+      <a ref={ref} {...props} target="_blank" rel="noopener noreferrer" />
+    ))
+
+    const CreatedCustomLink = createLink(CustomLinkWithTarget)
+
+    const rootRoute = createRootRoute()
+    const indexRoute = createRoute({
+      getParentRoute: () => rootRoute,
+      path: '/',
+      component: () => (
+        <>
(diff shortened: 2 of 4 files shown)


packages/react-router/src/link.tsx
```diff
@@ -218,11 +218,15 @@ export function useLinkProps<
...
   const handleClick = (e: React.MouseEvent) => {
+    // Check actual element's target attribute as fallback
+    const elementTarget = (e.currentTarget as HTMLAnchorElement).target
+    const effectiveTarget = target !== undefined ? target : elementTarget
+
     if (
       !disabled &&
       !isCtrlEvent(e) &&
       !e.defaultPrevented &&
-      (!target || target === '_self') &&
```
leesb971204, line 225: If the `target` prop is `undefined`, the condition treats it as `_self`, causing the router to intercept the click and prevent `target="_blank"` links from opening in a new tab.

packages/react-router/src/link.tsx
```diff
@@ -218,11 +218,15 @@ export function useLinkProps<
 
   // The click handler
   const handleClick = (e: React.MouseEvent) => {
+    // Check actual element's target attribute as fallback
+    const elementTarget = (e.currentTarget as HTMLAnchorElement).target
+    const effectiveTarget = target !== undefined ? target : elementTarget
+
```
leesb971204, line 224: If no `target` prop is provided, check the rendered element’s `target` attribute to ensure correct behavior.

nlynzaad:
other than making use of test-ids for the tests this looks good to me

leesb971204:
> other than making use of test-ids for the tests this looks good to me

Thank you for the feedback.
I’ve updated it to perform the tests using `data-testid`.

nlynzaad:
one last thing please apply the same tests and update to solid-router

leesb971204:
> one last thing please apply the same tests and update to solid-router

I’ve added the related logic and tests to solid-router.

Issue #5059, opened 2025-08-31
SPA mode server build contains unnecessary SSR code


StevenPewsey:
### Which project does this relate to?

Start

### Describe the bug

When running tanstack start in SPA mode, the server build contains SSR code.

### Your Example Website or App

-

### Steps to Reproduce the Bug or Issue

Add

```
spa: {
  enabled: true,
}
```

to the `tanstackStart` vite plugin config

Run `npm build`

The built `.output/server` contains code that would be used to server side render the application. However this isn't actually used at runtime.

### Expected behavior

The built `.output/server` only contains the code needed to run server functions/api routes, so it is as slim as possible.

### Platform

- Start Version: 1.131.31
- OS: macOS
- Bundler: vite
- Bundler Version: 7.1.3

f1shy-dev:
can we make this configurable actually?
like it would be great if we can support creating a _shell file while still having SSR
it's perfect for my use-case of SSR -> SPA via a service worker

jk-gan:
I'm looking a way to disable the SSR for my SPA, it seems like it is not possible yet?

aarjithn:
@jk-gan did you try ssr: false at route?

zyojv:
is there any update on this?

andrewdoro:
yeah still happens

mattfysh:
may be related to https://github.com/TanStack/router/issues/5471

this became rather obvious while trying to integrate tanstack db, where i had this top-level code in a separate module:

```ts
export const list = createCollection(
  localStorageCollectionOptions({ ...
```

even after using `ssr: false` on routes that wanted to work with this collection in the client only, I noticed a lot of bundling logic that would include the collection import for server-side bundles

even when not using it, just the presence of the `createCollection` would leads to server-side errors, like:

```
LocalStorageCollectionError: [LocalStorageCollection] No storage available. Please provide a storage option or ensure window.localStorage is available
```

my gut feeling being that TSR doesn't quite deliver on the promise of selective SSR

mattrobrob:
Having this issue too - upgrading from tanstack router to tanstack start SPA mode is proving difficult.

jamesleeht:
The docs advertise that the SPA mode allows you to deploy on CDN: 

> Easier to deploy - A CDN that can serve static assets is all you need.

But unfortunately this only works with a project that is started purely with tanstack router and without start due to this issue.

jacobgad:
[...]

PR #5139, opened 2025-09-12
Add Pure SPA section to SPA Mode page

BennyHinrichs:
Let me know if you think this is an alright addition to the SPA Mode page. I spent like a whole day trying to get SPA mode to work without an FE server before asking in the Discord and learning that that is antithetical to the ethos of Start. Something akin to this would have saved me a lot of time by directing me to Router alone much sooner.

After the first sentence, there could also be another sentence added explaining

> Loader functions in your `__root.tsx` file will still run on the server on startup in SPA mode.

I left it out because I didn't know if it was overexplaining.

Another option is maybe adding a sentence in the [Why use Start without SSR?](https://tanstack.com/start/latest/docs/framework/react/spa-mode#why-use-start-without-ssr) section to clarify better. I guess what I'm looking for is something on the SPA docs specifying that Start is an obligate server environment and enabling SPA doesn't opt you out of that, it merely changes how you interact with said server. I.e., if you want to bring your own server, you *must* daisy chain it to the Start server.

My position is that I'm in a company with a large Java backend. They would like us to rewrite the FE, but they don't want to add infra for an FE server. Ergo, I saw SPA mode and thought that I could use Start, but such is not the case. 

## Summary by CodeRabbit

* **Documentation**
  * Added a new "Pure SPA" section to the SPA mode docs (placed after caveats).
  * Clarifies that SPA mode still involves a frontend server (TanStack Start).
  * Recommends TanStack Router for setups requiring no frontend server, with a link included.
  * Prerendering steps and configuration guidance remain unchanged.
  * No code or public API changes.

### DIFF
--- a/docs/start/framework/react/spa-mode.md
+++ b/docs/start/framework/react/spa-mode.md
@@ -22,6 +22,10 @@ For applications that do not require SSR for either SEO, crawlers, or performanc
 - **Slower time to full content** - Time to full content is longer since all JS must download and execute before anything below the shell can be rendered.
 - **Less SEO friendly** - Robots, crawlers and link unfurlers _may_ have a harder time indexing your application unless they are configured to execute JS and your application can render within a reasonable amount of time.
 
+## Pure SPA
+
+SPA mode does not mean there is no frontend server; it's just a different way for the client to interact with your frontend (i.e., TanSatck Start) server. If you need to do an SPA with no frontend server, then [TanStack Router](https://tanstack.com/router/latest) will be a better fit for your project.
+
 ## How does it work?
 
 After enabling the SPA mode, running a Start build will have an additional prerendering step afterwards to generate the shell. This is done by:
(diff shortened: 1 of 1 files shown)


sjoerdvanBommel:
Ah, I just ran into the exact same situation and [commented](https://github.com/TanStack/router/issues/5059#issuecomment-3650956757) about it. Good to see this clarification, I agree this is somewhat unclear.

Feedback:
- The [Benefits of SPA Mode](https://tanstack.com/start/latest/docs/framework/react/guide/spa-mode#benefits-of-spa-mode) section contains an incorrect statement as far as I understand:
  > Easier to deploy - A CDN that can serve static assets is all you need.

Potentially the other benefits could be rewritten slightly as well

LadyBluenotes:
Thanks for surfacing a real source of confusion here. I don't think what you mentioned is accurate now: SPA mode can be deployed as static client output, including the shell, without a running Start server when the app does not use Start server-side features.

Server functions and server routes are optional runtime capabilities and do need a server or edge handler. The root loader runs while the shell is prerendered at build time, rather than when a production server starts.

This PR also targets a documentation path that has since moved. The remaining deployment and server-build confusion is already tracked in #5059, so I’m closing this stale version and keeping the discussion there

Issue #4933, opened 2025-08-12
Router does not render index.tsx in a dynamic path parameter route that comes after an optional path parameter


kanasva:
### Which project does this relate to?

Router

### Describe the bug

Router does not render index.tsx in a dynamic path parameter route that comes after an optional path parameter.

### Your Example Website or App

https://github.com/kanasva/tanstack-router-example-basic-file-based-optional-path-param-issue

### Steps to Reproduce the Bug or Issue

<img width="323" height="265" alt="Image" src="https://github.com/user-attachments/assets/785ac8c7-9591-4f9f-bf32-a1566b4afd0d" />

**Case 1 Not use optional path param**: It works as expected.
`locale` is a dynamic path parameter.
`tenantId` is a dynamic path parameter.

- `/not-use-optional/locale`: Works
- `/not-use-optional/locale/tenantId`: Works
- `/not-use-optional/locale/tenantId/settings`: Works

**Case 2 Use optional path param**: It doesn't work as expected.
`locale` is an **optional** path parameter.
`tenantId` is a dynamic path parameter.

- `/use-optional/locale`: Works
- `/use-optional/locale/tenantId`: **Doesn't work, render blank page.**
- `/use-optional/locale/tenantId/settings`: Works

### Expected behavior

`/use-optional/locale/tenantId` should render `{-$locale}.$tenantId.index.tsx`

### Platform

{
  "dependencies": {
    "@tanstack/react-router": "^1.131.5",
    "@tanstack/react-router-devtools": "^1.131.5",
    "@tanstack/router-plugin": "^1.131.5",
    "react": "^19.0.0",
    "react-dom": "^19.0.0",
    "redaxios": "^0.5.1",
    "postcss": "^8.5.1",
    "autoprefixer": "^10.4.20",
    "tailwindcss": "^3.4.17",
    "zod": "^3.24.2"
  },
  "devDependencies": {
    "@types/react": "^19.0.8",
    "@types/react-dom": "^19.0.3",
    "@vitejs/plugin-react": "^4.3.4",
    "typescript": "^5.7.2",
    "vite": "^6.3.5"
  }
}

jiji-hoon96:
Would it be okay to do some simple work and submit a PR?

jiji-hoon96:
https://github.com/TanStack/router/pull/4917

It looks similar to the above PR. What do you think?

kanasva:
> Would it be okay to do some simple work and submit a PR?

[...]

PR #5176, opened 2025-09-21
fix(router-core): optional-path-param usage variations

nlynzaad:
#4933 highlighted a case where index routes would not resolve when an optional param is followed by a required path param.

apart from this specific issue, this PR unlocks some more variations in the usage of optional params including:

{-$optional}.$pathParam.index
{-$optional}.{-$optional}.$pathParam.path (with and without index)
{-$optional}.$pathParam.{-$optional}.path (with and without index)
{-$optional}.path.{-$optional}.$pathParam (with and without index)
{-$optional}.path.$pathParam.{-$optional} (with and without index)

bunch of unit tests added to match-path tests to test various variations and completions.
also added e2e tests on react and solid to test functionality of this. not all variations are covered here since they are covered in the unit test. this is just to ensure findings from unit tests carry over to e2e.

### DIFF
--- a/packages/router-core/src/path.ts
+++ b/packages/router-core/src/path.ts
@@ -838,6 +838,12 @@ function isMatch(
             // where consecutive optional params are used, we can break early.
             // preference is given to the first optional param
             if (futureRouteSegment?.type === SEGMENT_TYPE_OPTIONAL_PARAM) {
+              if (
+                routeSegmentLength - optionalCount + processedOptionals >=
+                baseSegments.length
+              ) {
+                shouldMatchOptional = false
+              }
               break
             }
 
@@ -858,8 +864,8 @@ function isMatch(
                 followingRouteSegment &&
                 (followingRouteSegment.type === SEGMENT_TYPE_PATHNAME ||
                   (followingRouteSegment.type === SEGMENT_TYPE_OPTIONAL_PARAM &&
-                    followingRouteSegment.prefixSegment) ||
-                  followingRouteSegment.suffixSegment)
+                    (followingRouteSegment.prefixSegment ||
+                      followingRouteSegment.suffixSegment)))
               ) {
                 const remainingRouteSegments = routeSegments.slice(
                   lookAhead + 1,
--- a/packages/router-core/src/path.ts
+++ b/packages/router-core/src/path.ts
@@ -611,6 +611,11 @@ function isMatch(
 ): boolean {
   let baseIndex = 0
   let routeIndex = 0
+  let processedOptionals = 0
+  let skippedOptionals = 0
+  const optionalCount = routeSegments.filter(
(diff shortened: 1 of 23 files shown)


packages/router-core/src/path.ts
```diff
@@ -796,33 +805,111 @@ function isMatch(
           // For optional params without prefix/suffix, we need to check if the current
           // base segment should match this optional param or a later route segment
 
-          // Look ahead to see if there's a later route segment that matches the current base segment
           let shouldMatchOptional = true
+          const remainingOptionals = optionalCount - processedOptionals - 1 > 0
+
+          // consider last route segment might be index route and any prior optionals that was not matched
+          const remainingRouteSegmentLength =
+            (routeSegments.slice(-1)[0]?.value === '/'
```
Sheraff, line 813: `isMatch` is a critical part of the router, we should be careful about performance. Maybe we can use the `last()` utility function to access the last item in an array instead of this?
```suggestion
            (last(routeSegments)?.value === '/'
```

packages/router-core/src/path.ts
```diff
@@ -796,33 +805,111 @@ function isMatch(
...
+
+                const remainingBaseSegments = baseSegments.slice(baseIndex + 1)
+
+                isMatchedFurtherDown =
+                  remainingRouteSegmentLength ===
+                    remainingBaseSegments.length &&
+                  isMatch(
+                    remainingBaseSegments,
+                    remainingRouteSegments,
+                    { ...params },
```
Sheraff, line 894: I don't think it needs to inherit params
```suggestion
                    {},
```

packages/router-core/src/path.ts
nlynzaad: thanks. applied this. agreed performance here is paramount hence added extra early exits from the for loop running the isMatch only in specific circumstances

packages/router-core/src/path.ts
nlynzaad: params gets modified in isMatch when the match is found, for this here we only want to confirm a match exists and don't want it to mutate the params, hence the spread.

Sheraff: but it's never reading from it either right? So if we're not gonna use it, and it doesn't need it, why spread at all? Why not just pass in an empty object?

amosbastian:
I also encountered this problem today

nlynzaad:
The added tests has highlighted additional issues with optional params, unrelated to this PR, that we need to address.

Unfortunately I have been busy dealing with other issues on non-nested paths that has taken my attention away from this PR. I hope to return to this as soon as we have resolved those.

neznayer:
Me too encountering this problem. It is quite critical for my case, I use optional i18n routes everywhere.

nlynzaad:
Closing thia infavour of #5722 that was merged today. This resolved the underlying issue. @neznayer, @amosbastian please see the feedback on #4933.

PR #5186, opened 2025-09-22
Fix: add (de)hydration options to tanstack query integration

luixo:
I tried to add [Tanstack Query integration](https://tanstack.com/router/v1/docs/integrations/query) to Tanstack Start and it failed with Seroval's error (also crashed the whole app!)
```
Error reading routerStream: g [Error]: The value [object Object] of type "object" cannot be parsed/serialized.
```

Before that I had my own implementation of Tanstack Query integration and it didn't fail because I (de)serialized data, but with integration that's not an option.
This PR adds such an option.

## Summary by CodeRabbit

* **New Features**
  * Added optional SSR hydration and dehydration configuration (server and client) for finer control over how query state is serialized and restored.
  * Streaming updates now include per-item hydration configuration to ensure consistent behavior during incremental rendering.
  * Client hydration respects these options for both initial load and streamed updates.
  * Defaults unchanged so existing setups continue to work.

### DIFF
--- a/packages/router-ssr-query-core/src/index.ts
+++ b/packages/router-ssr-query-core/src/index.ts
@@ -5,6 +5,8 @@ import {
 import { isRedirect } from '@tanstack/router-core'
 import type { AnyRouter } from '@tanstack/router-core'
 import type {
+  DehydrateOptions,
+  HydrateOptions,
   QueryClient,
   DehydratedState as QueryDehydratedState,
 } from '@tanstack/query-core'
@@ -20,6 +22,8 @@ export type RouterSsrQueryOptions<TRouter extends AnyRouter> = {
    * @link [Guide](https://tanstack.com/router/latest/docs/framework/react/api/router/redirectFunction)
    */
   handleRedirects?: boolean
+  hydrateOptions?: Omit<HydrateOptions, 'shouldDehydrateQuery'>
+  dehydrateOptions?: DehydrateOptions
 }
 
 type DehydratedRouterQueryState = {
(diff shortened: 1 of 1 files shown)


schiller-manuel:
can you please provide a full example where this is needed?

luixo:
> can you please provide a full example where this is needed?

Just use any non-serializable value in the query data (create a class that holds a value inside of itself).
The usage for hydration / dehydration options is described in [Tanstack Query docs](https://tanstack.com/query/latest/docs/framework/react/reference/hydration#limitations)

If you still need a full example - can you provide with a minimal boilerplate I should modify?

schiller-manuel:
setting up a serializationAdapter in start should solve this as well.

what are you putting into `hydrateOptions` for your use case?

luixo:
> setting up a serializationAdapter in start should solve this as well.

Is it implemented yet?

> what are you putting into hydrateOptions for your use case?

`{ serializeData: transformer.serialize }` for dehydrateOptions and `{ deserializeData: transformer.deserialize }` for hydrateOptions.

Transformer is a custom extended [SuperJSON](https://github.com/flightcontrolhq/superjson) instance.

luixo:
@schiller-manuel It's been a few months and I can't find any `serializationAdapter` mentions in the docs though I found this handle in the codebase (it doesn't seem it's usable though).

LadyBluenotes:
Thanks for reporting the Seroval problem and putting together a focused fix. This support has since landed in the SSR Query integration, including the streaming path and custom serialize/deserialize options.

The current implementation forwards `dehydrateOptions` and `hydrateOptions`, preserves caller filtering for streamed queries, and includes focused coverage and docs. I’m closing this older PR as superseded by #7246 and the follow-up streaming fix in #7497.

PR #5187, opened 2025-09-23
chore: get more e2e tests passing

brenelz:
Got most of the e2e tests passing except the solid start ones.

For some reason it imports react-router in the solid apps. When that's fixed i think those should pass.


e2e/react-router/basic-scroll-restoration/src/main.tsx
```diff
@@ -43,7 +43,7 @@ function RootComponent() {
 const indexRoute = createRoute({
   getParentRoute: () => rootRoute,
   path: '/',
-  loader: () => new Promise((r) => setTimeout(r, 500)),
+  loader: () => new Promise<any>((r) => setTimeout(r, 500)),
```
brenelz, line 46: it was complaining about Promise<unknown> so I typed as any for now

packages/react-router-devtools/package.json
```diff
@@ -62,7 +62,8 @@
     "node": ">=12"
   },
   "dependencies": {
-    "@tanstack/router-devtools-core": "workspace:*"
+    "@tanstack/router-devtools-core": "workspace:*",
+    "vite": "^7.1.1"
```
brenelz, line 66: not sure this is needed but for me it was complaining about not having vite

examples/solid/start-basic/src/routes/_pathlessLayout/_nested-layout/route-b.tsx
```diff
@@ -1,3 +1,4 @@
+import { createFileRoute } from '@tanstack/react-router'
```
brenelz, line 1: these automatically got a react-router import added

PR #5201, opened 2025-09-24
Fix clearExpiredCache/AgeTicker on removed routes

CC-Bell:
I am working on a application where we update the `routeTree` as new parts of the application is loaded (Module Federation architecture). When a new route is registered, these functions error out since the previous route no longer exists.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved router stability by safely handling missing routes during loader checks, preventing rare crashes when a route is undefined.
  * Updated devtools AgeTicker to gracefully handle absent routes, avoiding null-reference errors and returning no output when appropriate.
  * Overall, reduces runtime exceptions in edge cases involving unresolved or missing routes, enhancing reliability in both runtime and devtools views.

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -2259,9 +2259,9 @@ export class RouterCore<
   clearExpiredCache = () => {
     // This is where all of the garbage collection magic happens
     const filter = (d: MakeRouteMatch<TRouteTree>) => {
-      const route = this.looseRoutesById[d.routeId]!
+      const route = this.looseRoutesById[d.routeId]
 
-      if (!route.options.loader) {
+      if (!route?.options.loader) {
         return true
       }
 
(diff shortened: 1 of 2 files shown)


packages/router-core/src/router.ts
```diff
@@ -2259,9 +2259,9 @@ export class RouterCore<
   clearExpiredCache = () => {
     // This is where all of the garbage collection magic happens
     const filter = (d: MakeRouteMatch<TRouteTree>) => {
-      const route = this.looseRoutesById[d.routeId]!
+      const route = this.looseRoutesById[d.routeId]
```
schiller-manuel, line 2263: should we not just return here if route is undefined?

LadyBluenotes:
Thanks for digging into this. The missing-route guards in cache cleanup and Devtools are both needed, but route-tree replacement can also leave stale matches for lifecycle callbacks, so the current patch does not cover the whole crash path.

The plugin and router internals have changed substantially since this branch was opened, so I’m closing this version and tracking the complete dynamic-route-removal fix in #7797.

PR #5202, opened 2025-09-24
fix(router-core): ensure rewrite basepath applies consistently

### LINKED ISSUE #5200: v1.132.2 basePath is not working for browser url

Router

### Describe the bug

Since that version, basepath has been replaced with the rewrite method, and it seems that the rewrite logic has been applied to basepath for backward compatibility with previous versions.

When basepath is configured the same way as in earlier versions, accessing without a basepath (e.g., http://localhost:3000) does not add the basepath to the browser URL.

https://stackblitz.com/edit/github-lebazlff?file=src%2Froutes%2F__root.tsx

1. check window.location.pathname in home
2. click about
3. check window.location.pathname in about

### Expected behavior

window.location.pathname in home is /app

### Screenshots or Videos

_No response_

### Platform

- Bundler: [e.g. vite]

### Additional context

_No response_

leesb971204:
fixes #5200 

## Problem
When using the new rewrite system with basepath, redirects from `/` to `/basepath` were not working in client-side navigation.

## Root Cause
The `buildLocation` function was returning the original `fullPath` for the `href` property instead of applying the output rewrite. This caused the Transitioner component to compare non-rewritten URLs, preventing proper redirect detection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Links/href generation now uses the rewritten path (without origin) to prevent mixed-origin or malformed links while keeping the full rewritten URL for navigation and deep links.

* **Documentation**
  * Examples updated to demonstrate the new rewrite API shape and composing rewrites; public interfaces/signatures unchanged.

* **Tests**
  * Added tests validating basepath rewrite behavior and backward compatibility with basepath option.


schiller-manuel:
thanks for the PR!
any idea why the e2e tests fail?

also, we should add tests for this (ideally some unit tests)

leesb971204:
> thanks for the PR! any idea why the e2e tests fail?
> 
> also, we should add tests for this (ideally some unit tests)

I’m not really sure — honestly, I’m not very familiar with TanStack Start.
When I run the Start example on my branch, the app gets stuck due to excessive redirects caused by the basepath.
Do you have any idea what might be happening?

I’ll go ahead and add a unit test in the meantime!

packages/react-router/src/Transitioner.tsx
```diff
@@ -52,10 +52,10 @@ export function Transitioner() {
...
     })
 
-    if (
-      trimPathRight(router.latestLocation.href) !==
-      trimPathRight(nextLocation.href)
-    ) {
+    const latestPublicHref = trimPathRight(router.latestLocation.publicHref)
+    const nextPublicHref = trimPathRight(nextLocation.publicHref)
+
+    if (latestPublicHref !== nextPublicHref) {
```
leesb971204, line 58: publicHref carries rewrite/basePath changes that should reflect in the browser URL

packages/router-core/src/router.ts
```diff
@@ -1688,7 +1691,7 @@ export class RouterCore<
       return {
         publicHref:
           rewrittenUrl.pathname + rewrittenUrl.search + rewrittenUrl.hash,
-        href: fullPath,
+        href: rewrittenUrl.href.replace(rewrittenUrl.origin, ''),
```
schiller-manuel: i dont think this is right

`href` is the internal router's representation while `publicHref` contains the external one
we need to keep both worlds

cc @tannerlinsley

leesb971204: > i dont think this is right
> 
> `href` is the internal router's representation while `publicHref` contains the external one we need to keep both worlds
> 
> cc @tannerlinsley

Thank you for the review.
At first, I thought this change was the right way to fix the issue, but it turned out that switching the URL comparison target from `href` to `publicHref` resolved it, so this fix was no longer needed.
I’ve reverted the change.

tannerlinsley: Yeah, pretty sure we want to keep origin here, or at least in one of them.

packages/react-router/src/Transitioner.tsx
schiller-manuel: what about the case when the rewrite changes the origin?
we would need to do a hard navigation here i think

packages/react-router/tests/router.test.tsx
```diff
@@ -2836,6 +2836,72 @@ describe('rewriteBasepath utility', () => {
...
+    const routeTree = rootRoute.addChildren([indexRoute])
+
+    const history = createMemoryHistory({
+      initialEntries: ['/my-app/'],
+    })
+
+    const router = createRouter({
+      routeTree,
+      history,
+      rewrite: rewriteBasepath({ basepath: '/my-app' }),
```
schiller-manuel: please go back to `basepath` prop for testing this

we reverted the deprecation of that prop

packages/react-router/src/Transitioner.tsx
schiller-manuel: if we can get this solved, we can merge this

leesb971204: > what about the case when the rewrite changes the origin? we would need to do a hard navigation here i think

Sorry for the late reply — I was on vacation.
Are there any cases where the origin can be changed through `basepath` or `rewrite`?
I think a bit more explanation would be helpful.

schiller-manuel: yes you can for example map a path segment to a subdomain via a rewrite.

see this test for example

https://github.com/TanStack/router/blob/e0c63d47b4d6891bb186c16ece27bd452a37406b/packages/react-router/tests/router.test.tsx#L2143

packages/react-router/src/Transitioner.tsx
```diff
@@ -52,10 +52,20 @@ export function Transitioner() {
...
+      const latestOrigin = new URL(router.latestLocation.url).origin
+      const nextOrigin = new URL(nextLocation.url).origin
+
+      if (latestOrigin !== nextOrigin) {
+        if (typeof window !== 'undefined') {
+          window.location.href = nextLocation.url
```
leesb971204: Would it be more appropriate to use `router.navigate({ href: nextLocation.url })`?

schiller-manuel: we could just call `router.navigate()`  in any case (so regardless whether origin changed or not)

however, `router.navigate()` needs to be made origin aware (it currently isnt)

leesb971204: > we could just call `router.navigate()` in any case (so regardless whether origin changed or not)
> 
> however, `router.navigate()` needs to be made origin aware (it currently isnt)

Do we really need switch to using `router.navigate()`?
If it’s not necessary for `router.navigate()` to be aware of the origin, I think it’s better to leave it as is without changing it.

schiller-manuel: i would prefer having it navigate() since it also enables navigation to other origins (this works with link already, but it has special treatment for this)

packages/router-core/src/router.ts
```diff
@@ -1913,10 +1917,26 @@ export class RouterCore<
       } catch {}
     }
 
+    const nextLocation = this.buildLocation({
+      to,
+      ...rest,
+      _includeValidateSearch: true,
+      _isNavigate: true,
+    } as any)
```
leesb971204, line 2009: Call `buildLocation` only once and reuse the result.

packages/router-core/src/router.ts
```diff
@@ -1926,11 +1946,13 @@ export class RouterCore<
...
-      href,
-      to: to as string,
-      _isNavigate: true,
+    return this.commitLocation({
+      ...nextLocation,
+      replace: rest.replace,
+      resetScroll: rest.resetScroll,
+      hashScrollIntoView: rest.hashScrollIntoView,
+      viewTransition: rest.viewTransition,
+      ignoreBlocker: rest.ignoreBlocker,
```
leesb971204, line 2059: Since `buildLocation` is always called above, only `commitLocation` is executed here.

packages/router-core/src/router.ts
```diff
@@ -1913,10 +1917,26 @@ export class RouterCore<
...
+    } as any)
+
+    if (!reloadDocument) {
+      const currentOrigin = new URL(this.latestLocation.url).origin
+      const nextOrigin = new URL(nextLocation.url).origin
+
+      if (currentOrigin !== nextOrigin) {
+        reloadDocument = true
+        href = nextLocation.url
+      }
```
leesb971204, line 2018: Compare the origins to perform a hard navigation.

leesb971204:
@schiller-manuel 

Are there any other things we should consider?

reihwald:
@leesb971204 Do you need help pushing this over the finish line?

leesb971204:
> @leesb971204 Do you need help pushing this over the finish line?

I’ve handled the issue in what I believe is the appropriate way and am currently waiting for a review.
If there’s any additional way I can contribute, I’d be more than happy to help!

reihwald:
> > @leesb971204 Do you need help pushing this over the finish line?
> 
> I’ve handled the issue in what I believe is the appropriate way and am currently waiting for a review. If there’s any additional way I can contribute, I’d be more than happy to help!

Maybe Manuel lost track of this PR. Have you thought about asking him on Discord?

leesb971204:
> > > @leesb971204 Do you need help pushing this over the finish line?
> > 
> > 
> > I’ve handled the issue in what I believe is the appropriate way and am currently waiting for a review. If there’s any additional way I can contribute, I’d be more than happy to help!
> 
> Maybe Manuel lost track of this PR. Have you thought about asking him on Discord?

Thank you for the feedback.
I’ve sent a Discord message to Manuel and will wait for his response.

reihwald:
Heard something? @leesb971204

leesb971204:
> Heard something? @leesb971204

noting 🙁

nlynzaad:
Hi @leesb971204 there was quite a rewrite that was completed on router-core recently. 

Would you be able to merge this with the latest main?

I just ran the tests you included, and they passed on the latest builds without any changes so not too sure if the tests aren't built from a failing initially point or if the problem has been resolved? Also took the original issue's repro, set the base in vite (to ensure its initially mounted in the correct path during dev) and did not see any specific issues. 

I'm probably missing something here, so if you could merge with the latest main and ensure failing tests before your changes are applied and then obviously succeeding after your changes.

leesb971204:
@nlynzaad 

As you suggested, I just ran the tests on both the latest main branch and a PR branch that merges in the latest main. In both cases, the tests passed successfully.

However, when I merge main into the PR branch again, the issue reappears — even though it was previously resolved and working as expected.

I suspect that recent changes in the core are invalidating the fix I applied earlier.
I’ll need to look into it more closely, but I believe I’ll have to find a new solution and rewrite the test accordingly.

Also, when you said that setting Vite’s base solves the issue — does that mean this work is unnecessary?

To be honest, I don’t clearly remember whether setting only the router’s basePath (without Vite’s base) was enough to make things work.
This issue has been going on for quite some time, so I’m not sure anymore whether the problem occurred because the base wasn’t set at all.

nlynzaad:
@leesb971204 thanks for merging main into this.

When the app is served, through vite dev or another bundler or server in prod this will load at the route it is served at.

In vite dev's case this is by defualt at `/`, but can be set to `/custom` by configuring the base property in the vite config.

Router needs to know if the app is served at any path other than the default `/`. This is to ensure navigation happens correctly at that base path. 

While it can deduce this in vite dev's case, when served from another server or bundler there is no way for Router to know what the base path is. We therefore need to do this by configuration which is done by setting the basePath property in createRouter. In other words, these needs to match.

When updating the PR in the original issue and ensuring the app is served at the correct location I was no longer able to see a problem. 

So before any more work is done on this PR I think we need to ensure the issue this PR is trying to solve is in actual fact still occuring.

@reihwald seeing as you are experiencing a similar issue could you please add your own reproducer onto the original issue, ensure its updated to the latest main, the app is served at the correct custom path and the basePath in router is set to match this custom path.

leesb971204:
> @leesb971204 thanks for merging main into this.
> 
> When the app is served, through vite dev or another bundler or server in prod this will load at the route it is served at.
> 
> In vite dev's case this is by defualt at `/`, but can be set to `/custom` by configuring the base property in the vite config.
> 
> Router needs to know if the app is served at any path other than the default `/`. This is to ensure navigation happens correctly at that base path.
> 
> While it can deduce this in vite dev's case, when served from another server or bundler there is no way for Router to know what the base path is. We therefore need to do this by configuration which is done by setting the basePath property in createRouter. In other words, these needs to match.
> 
> When updating the PR in the original issue and ensuring the app is served at the correct location I was no longer able to see a problem.
> 
> So before any more work is done on this PR I think we need to ensure the issue this PR is trying to solve is in actual fact still occuring.
> 
> @reihwald seeing as you are experiencing a similar issue could you please add your own reproducer onto the original issue, ensure its updated to the latest main, the app is served at the correct custom path and the basePath in router is set to match this custom path.

I’m really sorry for asking again, but just to confirm —
If we don’t set the `base` in vite.config and only set `basePath` in the router,
is it expected for `useLocation().pathname` or `window.location.pathname` to return `'/'` when accessed from the root path?

nlynzaad:
On initial render yes. Router has not executed any navigations yet to actually update the url with the set basePath

leesb971204:
> On initial render yes. Router has not executed any navigations yet to actually update the url with the set basePath

In that case, this PR might not be necessary —
but from what I recall, before `rewrite` was added, the initial render would show the basePath in pathname regardless of the base setting in `vite.config`.

I tested it with version 1.129.2, and it does behave that way:
https://stackblitz.com/edit/github-lebazlff-9dj5638p?file=package.json,src%2Froutes%2F__root.tsx,src%2Fmain.tsx

Below is the test case I ran using the latest version:
https://stackblitz.com/edit/github-szphp94v?file=src%2Froutes%2Fabout.tsx,src%2Froutes%2F__root.tsx,src%2Froutes%2Findex.tsx

Shouldn't both versions behave the same way?

reihwald:
Thank you for looking into this @nlynzaad. I'm not facing the exact same issue as @leesb971204 it seems. Please take a look at https://github.com/TanStack/router/issues/5200#issuecomment-3627658496 for a reproducer for my issue. Would you agree that this is worth a different issue?

nlynzaad:
@reihwald yes agreed, please create a new issue. From the description it seems related but different enough for a seperate issue

leesb971204:
@nlynzaad 

How do you think this issue(#5200) should be handled in this PR?

nlynzaad:
@leesb971204 see the updated [stackblitz](https://stackblitz.com/edit/github-lebazlff-hhbgjj4v) apart from the bump in versions I added the base to the vite cconfig

nlynzaad:
@leesb971204 I'm going to close this in favour of #6157 which is updated with the latest main and addresses the issues in rewrite and buildLocation.

PR #5216, opened 2025-09-24
docs: update build-from-scratch replacements for SolidJS

jakst:
The docs make references to react in a lot of places. I updated the ones I could find in the build-from-scratch guide to use proper solid-js package names and other references.

## Summary by CodeRabbit

- **Documentation**
  - Expanded "build from scratch" guide for migrating from React to SolidJS.
  - Added comprehensive mapping of React libraries, tooling, and plugins to Solid equivalents.
  - Documents alternative React tooling options and Solid-compatible plugin recommendations.
  - Clarifies updating JSX runtime and project configuration for SolidJS.
  - Describes import/usage changes and fragment/JSX differences needed for Solid.
  - Notes removal of React-specific type artifacts during migration.


docs/start/framework/solid/build-from-scratch.md
```diff
@@ -5,5 +5,10 @@ replace:
     '@tanstack/react-start': '@tanstack/solid-start',
     'React': 'SolidJS',
     'react-router': 'solid-router',
+    'react react-dom': 'solid-js',
```
brenelz: so i guess it just string replaces the docs? I wonder how we would add the viteSolid({ssr: true}) part

jakst: Yeah, it's not the most flexible solution. I couldn't figure out a way to fix the children prop types in __root.tsx either

nlynzaad: Not pretty but could just add `'viteReact()': 'viteSolid({ssr:true})'` into the replace array as well

brenelz: I think there are some bigger differences like not hydrating the whole html document

nlynzaad: Problem seems to be in the `__root.tsx` file

```tsx
function RootDocument({ children }: Readonly<{ children: Solid.JSX.Element }>) {
	return (
		<html>
			<head>
				<HeadContent />
			</head>
			<body>
				{children}
				<Scripts />
			</body>
		</html>
	);
}
```

from looking at the start-bare and start-basic examples it seems like it should be:
```tsx
function RootDocument({ children }: Readonly<{ children: Solid.JSX.Element }>) {
	return (
		<>
			<HeadContent />
			{children}
			<Scripts />
		</>
	);
}
```

at least this got it as far running under development mode.

will see if I can push to this PR.

nlynzaad: I think this should all be resolved now. @jakst @brenelz please check if you are happy with the proposed changes.

brenelz (approved):
I guess can't really preview what this would output but overall looks good

nlynzaad:
> I guess can't really preview what this would output but overall looks good

used a local instance of tanstack.com to test, seems fine. will continue with the merge then. 

@jakst thanks for working on this.

PR #5236, opened 2025-09-25
Update environment-functions.md with correct function names

alexjamesgodfrey:
Since the RC, serverOnly and clientOnly no longer exist, and we use createServerOnlyFn and createClientOnlyFn. I updated the docs to reflect that.

### DIFF
--- a/docs/start/framework/react/environment-functions.md
+++ b/docs/start/framework/react/environment-functions.md
@@ -1,17 +1,18 @@
----
+di---
 id: environment-functions
 title: Environment Functions
+
 ---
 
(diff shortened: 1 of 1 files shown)


nlynzaad:
Thanks for this. Can you please check the comment made by code rabbit. Seems like it is relevant

alexjamesgodfrey:
> Thanks for this. Can you please check the comment made by code rabbit. Seems like it is relevant

Thx Nico, should be all set now! Thanks for updating the pr title as well, I'll use that format in the future.

nlynzaad:
Thanks for working on this

PR #5242, opened 2025-09-26
fix(router): hash history links with `target="_blank"` generate incorrect URLs

### LINKED ISSUE #5209: <Link target="_blank"> broken with hash history

Router

### Describe the bug

Hi 👋,

After upgrading to the latest TanStack Router (v1.132.2), I noticed that `<Link>` elements using `target="_blank"` no longer work correctly when using hash history.

Previously, the rendered `<a>` looked like this:
```
<a href="/#/about" target="_blank">About</a>
```

Now the rendered `<a>` looks like this:
```
<a href="/about" target="_blank">About</a>
```

This means opening the link in a new tab results in the URL:
```
http://localhost:5173/about
```

instead of the expected:
```
http://localhost:5173/#/about
```

So the hash is missing in the final URL when `target="_blank"` is used, breaking navigation in hash-history mode.

---

Expected behavior:

The `<Link>` should render with the hash prefix (`/#/about`) so that opening it in a new tab lands on the correct route in hash-history mode.

Actual behavior:

The `<Link>` renders as `/about`, causing the browser to open the wrong URL without the hash.

https://stackblitz.com/edit/github-gsxsnwbi?file=src%2Froutes%2F__root.tsx

1. Click on `About (target _blank)`.
2. It opens a new tab with the following URL: `<baseURL>/about` instead of `<baseURL>/#/about`

### Expected behavior

The `<Link>` should render with the hash prefix (`/#/about`) so that opening it in a new tab lands on the correct route in hash-history mode.

### Screenshots or Videos

_No response_

### Platform

- Router Version: v1.131.2
- History mode: `createHashHistory()`
- OS: macOS/Linux
- Browser: Chrome
- Browser Version: v138
- Bundler: vite
- Bundler Version:  v7.1.7

### Additional context

This only happens when using `target="_blank"`. Links without target navigate correctly within the same tab.

leesb971204:
fixes #5209

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -120,13 +120,13 @@ export function useLinkProps<
     let external = false
     if (router.origin) {
       if (href.startsWith(router.origin)) {
-        href = href.replace(router.origin, '') || '/'
+        href = router.history.createHref(href.replace(router.origin, '')) || '/'
       } else {
         external = true
       }
     }
     return { href, external }
-  }, [disabled, next.maskedLocation, next.url, router.origin])
+  }, [disabled, next.maskedLocation, next.url, router.origin, router.history])
 
   const externalLink = React.useMemo(() => {
     if (hrefOption?.external) {
--- a/packages/solid-router/src/link.tsx
+++ b/packages/solid-router/src/link.tsx
@@ -146,7 +146,7 @@ export function useLinkProps<
     let external = false
     if (router.origin) {
       if (href.startsWith(router.origin)) {
-        href = href.replace(router.origin, '')
+        href = router.history.createHref(href.replace(router.origin, ''))
       } else {
         external = true
       }


packages/react-router/src/link.tsx
```diff
@@ -120,13 +120,13 @@ export function useLinkProps<
     let external = false
     if (router.origin) {
       if (href.startsWith(router.origin)) {
-        href = href.replace(router.origin, '') || '/'
+        href = router.history.createHref(href.replace(router.origin, '')) || '/'
```
leesb971204, line 123: Since routes with `target="_blank"` don’t go through the `router.navigate` logic, the href needs to include `#`.
But the `hrefOption` generation logic wasn't applying `router.history.createHref()` transformation, which is essential for hash history to format URLs correctly (adding the `#` prefix).

schiller-manuel:
some tests would be nice

schiller-manuel:
thanks a lot!

PR #5299, opened 2025-09-29
fix: add `@tanstack/readt-devtools` to `optimizeDeps.exclude` for react-start vite pluging option

dario-piotrowicz:
The [Cloudflare Vite plugin](https://www.npmjs.com/package/@cloudflare/vite-plugin) doesn't support optimising the `@tanstack/readt-devtools` package causing Tanstack start app to error during development (here's a simple example: https://github.com/dario-piotrowicz/tanstack-start-cf-test where `@tanstack/readt-devtools` has been directly excluded from the deps optimization, including it causes `vite dev` not to work anymore: https://github.com/dario-piotrowicz/tanstack-start-cf-test).

The changes here are therefore adding the package to the `optimizeDeps.exclude` array, making sure that Tanstack apps can compatible with the Cloudflare plugin.


dario-piotrowicz:
Closing since the proper fix seems to instead be https://github.com/TanStack/devtools/pull/192

Issue #5261, opened 2025-09-26
Non-default Vite `base` parameter causes prerender to fail in SPA mode


andriygm:
### Which project does this relate to?

Start

### Describe the bug

(creating issue from discord discussion here https://discord.com/channels/719702312431386674/1421224773895192627/1421224773895192627)

When setting vite's `base` option in `vite.config.ts`:
```ts
export default defineConfig({
  base: '/custombasepath',
  ...
})
```

I get the following error:
```
[prerender] Prerendering pages...
[prerender] Concurrency: 8
[prerender] Crawling: /
[prerender] Prerendered 1 pages:
[prerender] - /
[sitemap] Hint: Pages found, but no sitemap host has been set. To enable sitemap generation, set the `sitemap.host` option.
file:///Users/andriymassimilla/Projects/a2-ai/portkey/frontend/node_modules/@tanstack/start-plugin-core/dist/esm/prerender.js:109
            throw new Error(`Failed to fetch ${page.path}: ${res.statusText}`, {
                  ^

Error: Failed to fetch /: 
    at file:///Users/andriymassimilla/Projects/a2-ai/portkey/frontend/node_modules/@tanstack/start-plugin-core/dist/esm/prerender.js:109:19
    at async file:///Users/andriymassimilla/Projects/a2-ai/portkey/frontend/node_modules/@tanstack/start-plugin-core/dist/esm/queue.js:38:17 {
  [cause]: Response {}
}

Node.js v24.9.0
```

note that `createRouter` includes the `rewriteBasepath` rewrite function (the other `basepath` parameter is marked as deprecated):
```ts
  const router = createRouter({
    routeTree,
    scrollRestoration: true,
    rewrite: rewriteBasepath({
      basepath: "/custombasepath",
    }),
  });
```

### Your Example Website or App

https://stackblitz.com/edit/vitejs-vite-jvynrgn2?file=package.json

### Steps to Reproduce the Bug or Issue

1. `pnpm i`
2. `pnpm run build`

### Expected behavior

As a user, I expected the build to complete without error even when specifying a custom base path, but instead an error is thrown.

### Platform

- Router / Start Version: 1.132.7/1.132.10
- OS: macOS
- Browser: Arc (chromium)
- Browser Version:
    Version 1.114.0 (69006)
    Chromium Engine Version 140.0.7339.208
- Bundler: vite
- Bundler Version: 7.1.7

krishna-santosh:
[...]

PR #5302, opened 2025-09-30
fix: basepath rewrite to match with or without trailing slash

andorep:
## Summary

This PR fixes the basepath rewrite logic to correctly match routes with or without a trailing slash.

## Problem

The rewrite logic didn’t consider basepaths without a trailing slash. For example, if you define a basepath like `/myapp` and try to access the root page at `/myapp`, it would not work unless you added a trailing slash (i.e., `/myapp/`). This caused the root page to be inaccessible without the slash, breaking prerendering in Tanstack Start

## Related Issue

See #5261 for more details

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved route matching for base paths to handle URLs with or without a trailing slash, reducing false 404s and ensuring correct resolution.
  * Enhances compatibility across environments where base paths may be configured inconsistently, resulting in more reliable navigation and redirects.
  * Preserves existing case sensitivity behavior with no changes to the public API or configuration.

### DIFF
--- a/packages/router-core/src/rewrite.ts
+++ b/packages/router-core/src/rewrite.ts
@@ -24,7 +24,7 @@ export function rewriteBasepath(opts: {
 }) {
   const trimmedBasepath = trimPath(opts.basepath)
   const regex = new RegExp(
-    `^/${trimmedBasepath}/`,
+    `^/${trimmedBasepath}(?:/|$)`,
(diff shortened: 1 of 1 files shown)


schiller-manuel:
we already have this PR open for that: https://github.com/TanStack/router/pull/5244

maybe help out there to figure out why the tests work without the fix applied?

PR #5331, opened 2025-10-01
Fix `getResponse()`

frenzzy:
It seems like the API is over-engineered with all these methods — `setResponseHeader`, `getResponseHeader`, `getResponseHeaders`, `removeResponseHeader`, `clearResponseHeaders` — and still doesn’t cover some cases.

For example:
- When I call two server functions in a row, each of them should be able to append a header with the same name. Currently, that isn’t possible.
- `getResponseHeader` does not use `getAll`, while `setResponseHeader` does support an array of values. This inconsistency makes things more confusing.
- The current `getResponseHeaders` implementation does not support returning multiple values for the same header.
- The current `getRequestHeaders` returns `Headers` but `getResponseHeaders` returns `Record<string, string>`.

Instead of maintaining all these methods, it would be simpler to just document how to manage response headers using the standard [Headers API](https://developer.mozilla.org/en-US/docs/Web/API/Headers):
```ts
const headers = getResponse().headers // or getResponseHeaders()
headers.set(name, value) // set
headers.append(name, value) // append
// etc.
```

### DIFF
--- a/packages/start-server-core/src/request-response.ts
+++ b/packages/start-server-core/src/request-response.ts
@@ -327,7 +327,7 @@ export function clearSession(config: Partial<SessionConfig>): Promise<void> {
 // not public API
 export function getResponse() {
   const event = getH3Event()
-  return event._res
+  return event.res
 }
 
 // not public API (yet)


schiller-manuel:
probably better to discuss this in our discord. as you might have seen this is not a public API, which there are reasons for.

frenzzy:
Unfortunately, the reasons are not described in the comment.
For me, this is a regression bug in the RC version that blocks my codebase from upgrading.

Previously, there was an appendHeader method that I relied on:
https://github.com/TanStack/router/blob/4a6dcd09cf16f7586228d0f68fe134a520a10033/packages/start-server-core/src/h3.ts#L449

schiller-manuel:
come on https://discord.com/invite/WrRKjPJ to discuss in detail!

frenzzy:
Closed in favor of https://github.com/TanStack/router/pull/5333

PR #5337, opened 2025-10-02
fix(router-core): set default value for TTo in UseNavigateResult type

### LINKED ISSUE #5203: The type of the "search" option in useNavigate is incorrect.

Router

### Describe the bug

```ts
import { NavigateOptions, useNavigate } from "@tanstack/react-router";

function Page() {

	const option: NavigateOptions = { search: (prev: any) => prev }; // no type error

	navigate(option); // no type error
	navigate({ search: (prev: any) => prev }); //type error
}

```

https://stackblitz.com/edit/github-9byjdyh7-rl6x81qz?file=src%2Froutes%2F__root.tsx

Paste the code above in VSCode

### Expected behavior

No type errors

### Screenshots or Videos

Image

### Platform

- Router / Start Version: [e.g. 1.132.2]
- Bundler: [e.g. vite]

### Additional context

_No response_

leesb971204:
fixes #5203 

## Summary by CodeRabbit

- Refactor
  - Updated navigation API typings to include a default destination when none is specified, improving TypeScript inference and reducing the need for explicit generics. This results in fewer editor warnings and a smoother development experience. Existing code continues to work without changes.
- Chores
  - Internal type cleanup only; no runtime or user-facing behavior changes.

### DIFF
--- a/packages/router-core/src/useNavigate.ts
+++ b/packages/router-core/src/useNavigate.ts
@@ -3,7 +3,7 @@ import type { RegisteredRouter } from './router'
 
 export type UseNavigateResult<TDefaultFrom extends string> = <
   TRouter extends RegisteredRouter,
-  TTo extends string | undefined,
+  TTo extends string | undefined = '.',
   TFrom extends string = TDefaultFrom,
   TMaskFrom extends string = TFrom,
   TMaskTo extends string = '',


packages/router-core/src/useNavigate.ts
```diff
@@ -3,7 +3,7 @@ import type { RegisteredRouter } from './router'
 
 export type UseNavigateResult<TDefaultFrom extends string> = <
   TRouter extends RegisteredRouter,
-  TTo extends string | undefined,
+  TTo extends string | undefined = '.',
```
leesb971204, line 6: The reason the code shown in the issue is inferred as `never` is because the `to` value is missing.

If `navigateOptions.to` is not set, it will navigate to current route from [here](https://github.com/TanStack/router/blob/4b0004cfbb1881199bf122becc5d5c179e033894/packages/router-core/src/router.ts#L1576).

lucerattow:
If this PR is rejected, it would be a good idea to update the documentation: [Search Params](https://tanstack.com/router/latest/docs/framework/react/guide/search-params#usenavigate-navigate-search-)

leesb971204:
@chorobin 
can you please review this?

LadyBluenotes:
This looks like the right direction. Omitting `to` already means “navigate relative to the current/default from route” at runtime, so defaulting `TTo` to `'.'` brings the shared type in line with that behaviour.

Could you add a small type regression for `navigate({ search: (prev) => ... })` without `to`, using a route with validated search? That will make sure `prev` stays inferred as the current route’s search shape rather than falling back to never.

And please rebase it on main, then we can get another review and see about getting this merged :)

schiller-manuel:
we still need @chorobin 's review here

leesb971204:
> we still need @chorobin 's review here

It's been 6 months since my last review request, should we still keep waiting for his review?

PR #5367, opened 2025-10-05
fixed infered type from useServerFn + added type test

### LINKED ISSUE #5352: Broken type for useServerFn inside useMutation

Router

### Describe the bug

Hi, I think the type is broken for the useServerFn.
I've reproduced and noticed that if I need to use the onSuccess callback I loose the types

https://github.com/Christopher96u/ts-issue-demo

1. Go to the index/home route

### Expected behavior

I expect as user, to have the onSuccess types back or improve the docs if this is the right behaviour

### Screenshots or Videos

If I comment out the onSuccess callback there is no TS issues 

Image

If I leave the onSuccess callback there is a TS issue

Image

If we use the useServerFn hook at top level (not inside the useMutation hook, we get the types back

Image

### Platform

- Bundler: [e.g. vite]

### Additional context

_No response_

Christopher96u:
## Summary
- tighten `useServerFn` typing in React/Solid so inline `useMutation({ mutationFn: useServerFn(fn) })`
  resolves to the awaited payload instead of `unknown`
- add a focused `.test-d.tsx` regression exercising the React Query reproduction
- restore the basic React Query example with a simple server function + mutation flow
- Fixes #5352 
- Updated 1 example to use useServerFn (I can remove it if I should?)

## Summary by CodeRabbit

* **New Features**
  * Example app adds a server-side mutation flow with a “Say hi” button and vertical layout that displays the greeting on success.

* **Refactor**
  * Revised useServerFn return shape and behavior for React and Solid to provide more consistent, ergonomic return types.

* **Tests**
  * Added integration/regression tests for server-function usage with mutations in React and Solid.

### DIFF
--- a/examples/react/start-basic-react-query/src/routes/index.tsx
+++ b/examples/react/start-basic-react-query/src/routes/index.tsx
@@ -1,12 +1,36 @@
+import * as React from 'react'
+import { useMutation } from '@tanstack/react-query'
 import { createFileRoute } from '@tanstack/react-router'
+import { createServerFn, useServerFn } from '@tanstack/react-start'
+
+const greetUser = createServerFn({ method: 'POST' })
+  .inputValidator((input: { name: string }) => input)
+  .handler(async ({ data }) => {
+    return Promise.resolve({ message: `Hello ${data.name}!` })
+  })
+
 export const Route = createFileRoute('/')({
   component: Home,
 })
 
 function Home() {
+  const greetingMutation = useMutation({
+    mutationFn: useServerFn(greetUser),
+    onSuccess: (data) => data,
+  })
+
   return (
-    <div className="p-2">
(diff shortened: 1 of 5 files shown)


Christopher96u:
It looks there are few tweaks I need to do

Christopher96u:
I think I finished on this, but one test is failing, it doesn't look related to my changes, could someone have a look?

LadyBluenotes:
Thanks for digging into the React Query inference problem and adding React and Solid type coverage. This branch now conflicts with main, and its last CI run is still failing, so the patch is no longer reviewable as-is. We’re going to close the PR, but keep #5352 open for a fresh reproduction against the current types and either a smaller type fix or documentation update. If you're interested in continuing it, please open another PR :)

PR #5373, opened 2025-10-05
Add support for customizable `functionId` generation.

zenios:
If not provided, the default (basename_functionName) will be used in development and sha256 of the default for production

### DIFF
--- a/packages/directive-functions-plugin/src/compilers.ts
+++ b/packages/directive-functions-plugin/src/compilers.ts
@@ -5,6 +5,7 @@ import {
   deadCodeElimination,
   findReferencedIdentifiers,
 } from 'babel-dead-code-elimination'
+import path from 'pathe'
 import { generateFromAst, parseAst } from '@tanstack/router-utils'
 import type { GeneratorResult, ParseAstOptions } from '@tanstack/router-utils'
 
@@ -22,6 +23,8 @@ export type SupportedFunctionPath =
   | babel.NodePath<babel.types.FunctionExpression>
   | babel.NodePath<babel.types.ArrowFunctionExpression>
 
+export type FunctionIdFn = (opts: { currentId: string }) => string
+
 export type ReplacerFn = (opts: {
   fn: string
   extractedFilename: string
@@ -38,6 +41,7 @@ export type CompileDirectivesOpts = ParseAstOptions & {
   getRuntimeCode?: (opts: {
     directiveFnsById: Record<string, DirectiveFn>
   }) => string
+  functionId?: FunctionIdFn
   replacer: ReplacerFn
   // devSplitImporter: string
   filename: string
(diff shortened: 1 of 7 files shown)


packages/start-plugin-core/src/schema.ts
```diff
@@ -159,6 +159,15 @@ const tanstackStartOptionsSchema = z
     serverFns: z
       .object({
         base: z.string().optional().default('/_serverFn'),
+        functionId: z
```
schiller-manuel: we need a better, more descriptive name here.
`functionId` makes it seem this IS the function id, but it is actually a function that generated functionIds

how about:

- `generateFunctionId`
- `createFunctionId`

?

i think we should call  it with the raw file name and raw function name as props instead of that single string argument

packages/directive-functions-plugin/src/compilers.ts
```diff
@@ -460,15 +465,33 @@ export function findDirectives(
...
+      if (functionId in directiveFnsById) {
+        let deduplicatedId = functionId
+        let iteration = 0
+        do {
+          deduplicatedId = `${deduplicatedId}_${++iteration}`
+        } while (deduplicatedId in directiveFnsById)
+        functionId = deduplicatedId
+      }
+    }
+    functionId = makeFunctionIdUrlSafe(functionId)
```
schiller-manuel: i wouldnt postprocess the function id anymore.
the function id that is returned by the function id generator function must be url safe. we need to document this requirement.

then the invocation of `makeFunctionIdUrlSafe` can simply go into the default function id generator function for dev

packages/server-functions-plugin/src/index.ts
```diff
@@ -262,6 +272,24 @@ function resolveViteId(id: string) {
...
+  functionId: FunctionIdFn
+  directiveFnsById: Record<string, DirectiveFn>
+}): FunctionIdFn {
+  return (functionIdOps) => {
+    let generatedId = opts.functionId(functionIdOps)
+    if (generatedId in opts.directiveFnsById) {
+      let deduplicatedId = generatedId
+      let iteration = 0
+      do {
+        deduplicatedId = `${deduplicatedId}_${++iteration}`
```
schiller-manuel: i am wondering whether we should call the function id generator function a single time at the end of the build with all of the functions?

then the function id generator function is responsible to provide unique function ids.

this ensures that we dont modfiy any of the function ids supplied by the user.

packages/start-plugin-core/src/schema.ts
zenios: Done

packages/directive-functions-plugin/src/compilers.ts
zenios: Done. ID generation and deduplicatoin is now the responsibility of server-functions-plugin

packages/server-functions-plugin/src/index.ts
zenios: We need the functionId at generation time for the replacers. We cant defer this up to the end

PR #5394, opened 2025-10-07
examples: Update WorkOS example to use new AuthKit SDK

nicknisi:
This PR updates the `start-workos` example to use the new [`@workos/authkit-tanstack-react-start`](https://github.com/workos/authkit-tanstack-start) SDK instead of the custom implementation.

## What changed

- Replaced custom AuthKit implementation with the official SDK package
- Removed all the manual session management, cookie handling, and JWKS verification code
- Updated routes to use the SDK's built-in `getAuth()` and `getSignInUrl()` server functions
- Added `AuthKitProvider` for client-side auth state management

## Why

We built an official SDK that handles all the AuthKit integration details, so this example should demonstrate the recommended approach rather than a custom implementation. Makes the example cleaner and easier to follow.

## Summary by CodeRabbit

* **New Features**
  * Added a Client Demo page showing client-side auth hooks, tokens, org switching, and sign-out methods; linked in navigation.
  * Integrated an AuthKit provider so loaders use new auth APIs for user and sign-in URLs.

* **Refactor**
  * Replaced local SSR auth implementation with the external AuthKit integration.

* **Chores**
  * Added AuthKit dependency, Radix theme CSS, Vite SSR tweak, and CSS URL TypeScript declarations.

### DIFF
--- a/examples/react/start-workos/package.json
+++ b/examples/react/start-workos/package.json
@@ -16,7 +16,7 @@
     "@radix-ui/themes": "^3.3.0",
     "@tanstack/react-router": "^1.158.1",
     "@tanstack/react-router-devtools": "^1.158.1",
-    "@tanstack/react-start": "^1.158.1",
+    "@tanstack/react-start": "^1.158.2",
     "@workos/authkit-tanstack-react-start": "^0.5.0",
     "react": "^19.0.0",
     "react-dom": "^19.0.0"
--- a/examples/react/start-bare/package.json
+++ b/examples/react/start-bare/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@tanstack/react-router": "^1.158.1",
     "@tanstack/react-router-devtools": "^1.158.1",
-    "@tanstack/react-start": "^1.158.1",
+    "@tanstack/react-start": "^1.158.2",
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "zod": "^3.24.2"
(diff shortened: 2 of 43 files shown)


examples/react/start-workos/package.json
```diff
@@ -18,6 +18,7 @@
     "@tanstack/react-router-devtools": "^1.132.47",
     "@tanstack/react-start": "^1.132.47",
     "@workos-inc/node": "^7.45.0",
+    "@workos/authkit-tanstack-react-start": "^0.1.0",
```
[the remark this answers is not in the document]
nicknisi: **note:** we have both the `@workos` and `@workos-inc` namespaces on npm. We've been working on a way to transition from `@workos-inc` to `@workos` because it's simpler and since this SDK was brand new, I figured I'd just start there. I can accelerate my idea to dual-publish if this is an issue.

packages/start-server-core/package.json
```diff
@@ -75,6 +75,7 @@
   "devDependencies": {
     "@standard-schema/spec": "^1.0.0",
     "cookie-es": "^2.0.0",
-    "fetchdts": "^0.1.6"
+    "fetchdts": "^0.1.6",
+    "vite": "^7.1.7"
```
schiller-manuel: why do we need `vite` here?

nicknisi: I think at the time I was having build issues without that - I rebased and tried again  and was able to remove it. Thanks!

DavidSchargel:
@schiller-manuel It would be great to see this merged sooner rather than later, so developers have clarity on current best practices and are not confused by outdated docs.

This is now the preferred way of using WorkOS and is being actively developed by @nicknisi and the WorkOS team. On Dec 15 2025, related templates and examples repos from WorkOS were archived in favor of this SDK, and it also aligns with @convex-dev/workos-authkit.

PR #5413, opened 2025-10-09
fix: export missing rewriteBasepath function

andorep:
`rewriteBasepath` is missing from the exports of react router

## Summary by CodeRabbit

* **New Features**
  * Expanded public API with a new router helper for base-path rewrites, making URL base-path customization easier.
  * Existing rewrite composition capability remains available and unchanged.
  * No breaking changes; apps continue to work as before and can adopt the new helper incrementally without config updates.

### DIFF
--- a/packages/router-core/src/index.ts
+++ b/packages/router-core/src/index.ts
@@ -437,6 +437,7 @@ export { defaultSerovalPlugins } from './ssr/serializer/seroval-plugins'
 
 export {
   composeRewrites,
+  rewriteBasepath,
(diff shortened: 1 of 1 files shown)


schiller-manuel:
no this function is not meant for external consumption.

andorep:
@schiller-manuel Sorry about that, I thought this was still the way to do the base_url configuration from vite.

Is there a doc where I can check that? 

If not, I probably could help with more documentation about that, if you just point me to the right direction to start, I can go from there myself

Issue #5420, opened 2025-10-09
Failing to resolve solid-js/web


murilobd:
### Which project does this relate to?

Router

### Describe the bug

I cannot build my project anymore after version 1.132.51 as I'm receiving the following error:
```
error during build:
[vite]: Rollup failed to resolve import "solid-js/web" from ".../node_modules/@tanstack/router-devtools-core/dist/index-BiZF0hsR.js".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
```

I'm not sure if this is because you moved solid-js from dependencies to devDependencies, but it doesn't make sense for me to install solid-js in my project.

### Your Example Website or App

https://github.com/TanStack/router/releases/tag/v1.132.51

### Steps to Reproduce the Bug or Issue

1. Update [@tanstack/react-router-devtools](https://github.com/TanStack/router/tree/HEAD/packages/react-router-devtools) to v1.132.51 and build a project with vite
2. It will output an error message like 
```
error during build:
[vite]: Rollup failed to resolve import "solid-js/web" from ".../node_modules/@tanstack/router-devtools-core/dist/index-BiZF0hsR.js".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
```

### Expected behavior

As a user I expect not to have to install solid-js library to use tanstack router devtools

### Platform

- Router: 1.132.51
- OS: macOS
- Bundler: vite
- Bundler Version: 7.1.9

benmccann:
I suppose I caused this in https://github.com/TanStack/router/pull/5406, but I can't reproduce this because I'm getting a different error. I tried to create a new app following [the docs](https://tanstack.com/router/v1/docs/framework/react/quick-start) and then build it:
```
pnpx create-tsrouter-app@latest my-test-tanstack-app --template file-router
cd my-test-tanstack-app
pnpm build
```

I got the error:
```
error during build:
rollupOptions.input should not be an html file when building for SSR. Please specify a dedicated SSR entry.
```

When I upgraded an existing project I have to the latest version and then rebuilt it, it succeeded without error.

benmccann:
Okay, not sure about the other issue I hit trying to reproduce this, but I've sent a fix for the issue originally described here

alexmakdis97:
I am having the same issue:
Module not found: Error: Can't resolve 'solid-js/web' for tanstack router devtools.
[...]

PR #5422, opened 2025-10-09
fix(router-devtools-core): actually bundle solid

benmccann:
closes https://github.com/TanStack/router/issues/5420

I'm not sure why upgrading `@tanstack/config` causes the build to fail. I could use some help with that as it seems a bit unrelated. I believe this fix should work once we get past that

## Summary by CodeRabbit

* **Chores**
  * Build configuration updated to bundle an additional UI library for more reliable builds across environments.
  * Development tooling dependency upgraded to a newer version for improved build compatibility and maintenance.
  * Linting configuration cleaned up to remove an unnecessary directive, simplifying developer tooling.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9317,8 +9317,8 @@ packages:
     engines: {node: '>=18'}
     hasBin: true
 
-  '@microsoft/api-extractor-model@7.29.6':
-    resolution: {integrity: sha512-gC0KGtrZvxzf/Rt9oMYD2dHvtN/1KPEYsrQPyMKhLHnlVuO/f4AFN3E4toqZzD2pt4LhkKoYmL2H9tX3yCOyRw==}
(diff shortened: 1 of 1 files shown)


schiller-manuel:
why do we need to upgrade config? what does this version do that we need?

birkskyum:
@schiller-manuel , I can see that the latest tanstack config has this change:
- https://github.com/TanStack/config/pull/302

birkskyum:
The issues that show up in the CI here are indeed unrelated to the PR. The errors show when bumping tanstack/config from 0.16.1 to [0.17.0](https://github.com/TanStack/config/blob/main/packages/config/CHANGELOG.md#0170) or higher on the main branch.   @lachlancollins , I can see 0.17.1 move to TypeScirpt, but it appear much stricter already from 0.17.0 - do you know how that can be?

PR #5430, opened 2025-10-10
tests: add ValidateLinkOptions in tanstack-router-react-example-large-file-based

beaussan:
Reproduction steps:
- pnpm pnpm --filter tanstack-router-react-example-large-file-based gen
- pnpm pnpm --filter tanstack-router-react-example-large-file-based dev
- pnpm pnpm --filter tanstack-router-react-example-large-file-based test:types


LadyBluenotes:
this is a tad stale so im going to close it :))

PR #5435, opened 2025-10-10
refactor: export devtools options type

lachieh:
When working with the TanStack Router Devtools, I was getting errors because the `DevtoolsOptions` type wasn't being exported from the main package. This isn't a problem when using the devtools directly from imports, but it does cause issues when trying to use the options to integrate tightly with my application.

```
Exported variable 'AppDevOptions' has or is using name 'DevtoolsOptions' from external module "/tmp/app/node_modules/@tanstack/react-router-devtools/dist/esm/TanStackRouterDevtools" but cannot be named. (ts4023)
```
In this instance, `AppDevOptions` is the type from my application.

I tried inferring the type, as well as creating a new type from the arguments, but it gives the same error:
```ts
import {TanStackRouterDevtools} from '@tanstack/react-router-devtools'

type DevtoolsOptions = Parameters<TanStackRouterDevtools>[0]
```

> [!NOTE]
> Despite what CodeRabbit says, these aren't actually breaking changes since the types were never exported in the first place. This is a minor change at best.

## Summary by CodeRabbit

* **Breaking Changes**
  * Public option type names were renamed — update imports and prop types accordingly.
  * Several devtools classes/components are now exported differently — adjust import syntax.

* **New Features**
  * New exported option interfaces and expanded public exports for core and panel devtools, offering richer typing and more configurable options for consumers.

### DIFF
--- a/packages/react-router-devtools/src/TanStackRouterDevtools.tsx
+++ b/packages/react-router-devtools/src/TanStackRouterDevtools.tsx
@@ -5,7 +5,7 @@ import type { ButtonHTMLAttributes, HTMLAttributes } from 'react'
 import type { AnyRouter } from '@tanstack/react-router'
 import type React from 'react'
 
-interface DevtoolsOptions {
+export interface TanStackRouterDevtoolsOptions {
   /**
    * Set this true if you want the dev tools to default to being open
    */
@@ -44,7 +44,7 @@ interface DevtoolsOptions {
 }
 
 export function TanStackRouterDevtools(
-  props: DevtoolsOptions,
+  props: TanStackRouterDevtoolsOptions,
 ): React.ReactElement | null {
   const {
     initialIsOpen,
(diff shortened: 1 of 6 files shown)


lachieh:
@coderabbitai generate docstrings

packages/react-router-devtools/src/TanStackRouterDevtools.tsx
SeanCassiere: I think these changes may also need to be applied for 'solid-router-devtools'. Please confirm, and if necessary apply them there as well.

lachieh: Added! Thanks.

SeanCassiere:
LGTM, merging!

Thank you @lachieh!

lachieh:
Thanks, @SeanCassiere! Would you mind adding the `hacktoberfest-accepted` label to this PR? No worries if not!

SeanCassiere:
> Thanks, @SeanCassiere! Would you mind adding the `hacktoberfest-accepted` label to this PR? No worries if not!

I don't think the TanStack libraries normally participate in Hacktoberfest, mostly due to the quality of PRs that get accepted, but I don't mind adding it here.

Fair warning though, we do ever so often clean up the "in-use" labels, which may see it disappearing from this PR in the future.

lachieh:
I totally understand the hacktoberfest burden. No problems if the label gets cleaned up! I think it only needs to be on at the end of the month to count. Thanks, again!

PR #5484, opened 2025-10-15
docs: Add route-level head explanation and example

LuiGeeDev:
I needed to set different titles for same page with different data and found that this can be done using the `head(ctx)` function.
It wasn’t clearly mentioned in the current docs, but my editor suggested the `head(ctx)` syntax, which worked as expected.
Since most other frameworks include a short explanation or example for this pattern, I thought it would be helpful to add a brief note here as well.

English is not my first language, so the added sentence and this description was refined with AI assistance — please feel free to adjust the wording if needed.

## Summary by CodeRabbit

* **Documentation**
  * Added a Route-level Document Head section to the React guide.
  * Shows how to use the route context inside the head function to customize titles and meta per route.
  * Includes an example of dynamic title/meta based on route data and illustrates RouteOptionsType head usage.
  * Clarifies required placement of the HeadContent component and links to head-method docs.


LadyBluenotes:
Thanks for highlighting dynamic titles and metadata from route data. The docs have since been reorganized into the shared document-head guide, so this branch now conflicts with main and targets an old path. The example also needs a parameterized route with a matching param name. I'm going to close this PR but if you're interested, we welcome you updating these docs with a new PR :)

PR #5485, opened 2025-10-15
fix(router-core): `onStay` callback should use 'current' matches not 'previous'

### LINKED ISSUE #1901: onStay property should use current matches, not previous
### Describe the bug

When I change url params on page via navigate like this
```ts
const router = useRouter()

const handleButtonClick = (pageNumber: number) => {
  router.navigate({ search: { page: pageNumber }, replace: true })
}
```
In onStay callback i got page number from params before navigate call
```ts
onStay: (match) => {
  console.log('onStay page number', match.search.page)
},
```
Maybe the problem is [here](https://github.com/TanStack/router/blob/f9ecd0e7f180b7998887e49332b9f5ec4aab5a2a/packages/react-router/src/router.ts#L1562C1-L1570C20) - `stayingMatches` should filter `newMatches` instead of `previousMatches`?

https://github.com/finnan444/tanstack-router-on-stay-error

1. Go to about page
2. Click "Go to next page"
3. Url will be updated, but in console you will see that the page number equals the one in url -1

### Expected behavior

I expect the match to be equal to current route search params

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 126.0.6478.127

### Additional context

_No response_

TeChn4K:
This PR fixes #1901 

`onStay` route hook currently gives the previous `match` instead of the current one.
Now, same as `onEnter` hook, you get the current `match`

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate detection of which views remain active during navigation, reducing unnecessary unmounts/reloads and preserving component state for smoother transitions.
  * Fewer edge cases when navigating between similar paths or parameter-only updates, improving perceived performance.

* **Tests**
  * Added lifecycle callback tests to ensure onEnter/onLeave/onStay behaviors are consistent across navigations.

### DIFF
--- a/packages/router-core/tests/callbacks.test.ts
+++ b/packages/router-core/tests/callbacks.test.ts
@@ -1,0 +1,106 @@
+import { describe, expect, it, vi } from 'vitest'
+import { createMemoryHistory } from '@tanstack/history'
+import { BaseRootRoute, BaseRoute, RouterCore } from '../src'
+
+describe('callbacks', () => {
+  const setup = ({
+    onEnter,
+    onLeave,
+    onStay,
+  }: {
+    onEnter?: () => void
+    onLeave?: () => void
+    onStay?: () => void
+  }) => {
+    const rootRoute = new BaseRootRoute({})
+
+    const fooRoute = new BaseRoute({
+      getParentRoute: () => rootRoute,
+      path: '/foo',
+      onLeave,
+      onEnter,
+      onStay,
+    })
+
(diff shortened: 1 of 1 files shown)


TeChn4K:
> Could you please add a test to make sure this behaviour is being enforced?

Added! @SeanCassiere
I was not sure where to write them, so I created a new callbacks.test.ts. Feel free to update to match your styles and conventions

PR #5491, opened 2025-10-15
fix(router-generator): remove trailing slash from pathless route group paths

### LINKED ISSUE #5486: Index in dynamic grouped layout does not render

Router

### Describe the bug

The index.tsx never renders in my dynamic grouped layout.

See the provided route tree in the image. In that case, the index.tsx never renders.

Image

Here's my route.tsx:

```tsx
import { createFileRoute, Outlet } from "@tanstack/react-router";
import { OrganizationNav } from "@/components/layout/organization-nav";

export const Route = createFileRoute("/_authed/$organizationSlug/(orglayout)")({
  component: RouteComponent,
});

function RouteComponent() {
  return (
    <>
      <OrganizationNav />
      <Outlet />
    </>
  );
}
```

It properly includes the outlet, and the /~/integrations route renders properly. But not the index.tsx, can someone tell me why this happens, is this intended and how should I get around this if it's not a bug?

https://stackblitz.com/edit/github-3jdwo6hv?file=src%2Froutes%2F_authed%2F%24organizationSlug%2F(orglayout)%2Ftest.tsx

If you go to something like /aaa you should see nothing which is incorrect since you should see what's inside index.tsx (should show « Hello "/_authed/$organizationSlug/(orglayout)/"! », from the index.tsx) and then go to /aaa/test you will then see the test.tsx page which is correct

### Expected behavior

Index page should render

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.132.47
- OS: Windows
- Browser: Chrome & Firefox
- Browser Version: Does not matter
- Bundler: Vite
- Bundler Version: 7.1.7

### Additional context

_No response_

nlynzaad:
when a route group is defined, the generator strips the group from the path but keeps the "/" preceding this group in the path resulting in these paths being treated as routes by the generator.

For example `/abc/(group)` is resolved to `/abc/` instead of `/abc`. This results in a few issues. One being that when both route and index are defined in a group, the route is determined to have a higher rank when sorting the routes and hence is matched prior to the index.

This PR resolves this issue by checking if the route is a layout (both pathless and non-pathless) and if so removes any trailing slash that remains after the group has been split.

Due to this error paths like `/(group)/_layout` resulted in `/(group)` to be defined as a virtual route to load `_layout` in this virtual parent route. With this fix the `_layout` route is correctly loaded with the rootRoute being its parent. Due to this, the generated route trees for basic-file-based e2e tests and the snapshots for the route-group unit tests in router-generator had to be updated as well.

This PR also adds a test to the `path-above-route-in-group` to test for the index path being correctly handled when used together with a route path in the same group.

This resolves #5486

## Summary by CodeRabbit

- New Features
  - Group routes are now exposed directly under the root, simplifying navigation and type mappings.
  - Added support for index routes within grouped paths, improving addressability of “path above route in group” scenarios.
- Bug Fixes
  - Normalized trailing slashes for layout and pathless layout routes to prevent duplicate or inconsistent paths.
- Tests
  - Updated e2e and snapshot tests to reflect the new routing structure and type surfaces.

### DIFF
--- a/e2e/react-router/basic-file-based/src/routeTree.gen.ts
+++ b/e2e/react-router/basic-file-based/src/routeTree.gen.ts
@@ -8,8 +8,6 @@
 // You should NOT make any changes in this file as it will be overwritten.
 // Additionally, you should also exclude this file from your linter and/or formatter to prevent it from being checked or modified.
 
-import { createFileRoute } from '@tanstack/react-router'
-
 import { Route as rootRouteImport } from './routes/__root'
 import { Route as Char45824Char54620Char48124Char44397RouteImport } from './routes/대한민국'
 import { Route as RemountDepsRouteImport } from './routes/remountDeps'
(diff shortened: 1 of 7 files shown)


xKesvaL:
Hey! Thanks for the quick PR. Do you think this can fix #5498 too ?

PR #5495, opened 2025-10-16
fix(react-router,solid-router): top-level `Matches` should use `pendingComponent` on root route

### LINKED ISSUE #2317: Pending component doesn't work on root route

Router

### Describe the bug

I've noticed the pending component doesn't stay mounted if it's on the __root route. It seems like it does quickly render but doesn't stay until the __root `loader` method is finished, instead it instantly unmounts

https://codesandbox.io/p/devbox/epic-shape-z676nh

1. Verify that you see a green box on first load to verify that non __root pending component works
2. Go to `src\routes\index.tsx` and comment out the pending component and loader
3. Go to `src\routes\__root.tsx` and uncomment the same code from previous file
4. Verify that the green box does not appear after refreshing window

### Expected behavior

I expect to see the green box upon loading after the 4th repro step. I've left a red box as the router's default pending component to show that does seem to be working normally in all relevant cases

### Screenshots or Videos

_No response_

### Platform

- Version: 1.57.8

### Additional context

_No response_

TkDodo:
fixes #2317

## Summary by CodeRabbit

* **New Features**
  * Route-level pending UI: root routes can now specify their own loading placeholder.

* **Bug Fixes**
  * Pending-component selection now prefers a route's pending UI and falls back to the router default for consistent loading visuals.

* **Tests**
  * Added tests verifying route/root pending behavior and router-level fallback during async route loading.

### DIFF
--- a/packages/react-router/tests/Matches.test.tsx
+++ b/packages/react-router/tests/Matches.test.tsx
@@ -124,11 +124,11 @@ test('when filtering useMatches by loaderData', async () => {
 
 test('should show pendingComponent of root route', async () => {
   const root = createRootRoute({
-    pendingComponent: () => <div>root pending...</div>,
+    pendingComponent: () => <div data-testId="root-pending" />,
     loader: async () => {
       await new Promise((r) => setTimeout(r, 50))
     },
-    component: () => 'root content',
+    component: () => <div data-testId="root-content" />,
   })
   const router = createRouter({
     routeTree: root,
@@ -138,6 +138,6 @@ test('should show pendingComponent of root route', async () => {
 
   const rendered = render(<RouterProvider router={router} />)
 
-  expect(await rendered.findByText('root pending...')).toBeInTheDocument()
-  expect(await rendered.findByText('root content')).toBeInTheDocument()
+  expect(await rendered.findByTestId('root-pending')).toBeInTheDocument()
+  expect(await rendered.findByTestId('root-content')).toBeInTheDocument()
 })
--- a/packages/solid-router/tests/Matches.test.tsx
+++ b/packages/solid-router/tests/Matches.test.tsx
@@ -224,11 +224,11 @@ test('Matches provides InnerWrap context to defaultPendingComponent', async () =
 
 test('should show pendingComponent of root route', async () => {
   const root = createRootRoute({
-    pendingComponent: () => <div>root pending...</div>,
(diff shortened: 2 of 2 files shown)


packages/react-router/src/Matches.tsx
```diff
@@ -41,10 +42,12 @@ declare module '@tanstack/router-core' {
 
 export function Matches() {
   const router = useRouter()
+  const rootRoute: AnyRoute = router.routesById['__root__']
```
TkDodo: do we need to be defensive here and type this as `AnyRoute | undefined`. I think a `__root__` route must always exist ...

schiller-manuel: please use `rootRouteId` from `@tanstack/router-core` instead of `__root__`

packages/react-router/tests/Matches.test.tsx
```diff
@@ -121,3 +121,23 @@ test('when filtering useMatches by loaderData', async () => {
 
   expect(await screen.findByText('Incorrect Matches -')).toBeInTheDocument()
 })
+
+test('should show pendingComponent of root route', async () => {
+  const root = createRootRoute({
+    pendingComponent: () => <div>root pending...</div>,
```
schiller-manuel: please use `data-testid` and getByTestId or similar methods instead of looking for textual matches

packages/react-router/tests/Matches.test.tsx
```diff
@@ -121,3 +121,23 @@ test('when filtering useMatches by loaderData', async () => {
...
+    loader: async () => {
+      await new Promise((r) => setTimeout(r, 50))
+    },
+    component: () => 'root content',
+  })
+  const router = createRouter({
```
schiller-manuel, line 133: same here

packages/react-router/src/Matches.tsx
SeanCassiere: and to answer your question, the answer is probably not, since a root route must always be present, but if it helps silence typescript then a `!` assertion or a `throw Error` should be fine.

TkDodo: done: 74a38350

packages/react-router/tests/Matches.test.tsx
TkDodo: done: 6fc6e2b6

Issue #5496, opened 2025-10-16
Start: Server-Fn Not Correctly Detected in Some Cases


yuta-ike:
### Which project does this relate to?

Start

### Describe the bug

Thank you so much for Tanstack Start.

When a type argument is passed to the handler function, as shown below, the server-fn code transformation doesn't occur. As a result, calling getNum from the server environment returns undefined.

```ts
const getNum = createServerFn({
  method: "GET",
}).handler<number>(() => 1);
```
Removing `<number>` from the handler call allows the code to be transformed correctly, and getNum will return 1.

While I haven't investigated in detail, I think this is due to the server-fn plugin's regular expression not matching. The regex /\.\s*handler\(/ doesn't account for type arguments.

https://github.com/TanStack/router/blob/4303a8a716c7f3d7ca292a6793b0cc4896394b8b/packages/start-plugin-core/src/create-server-fn-plugin/plugin.ts#L88

### Your Example Website or App

https://github.com/yuta-ike/tanstack-start-example

### Steps to Reproduce the Bug or Issue

1. npm install
2. npm run dev
3. Open `http://localhost:3000/`

You will see `num: (type: undefined)`.

### Expected behavior

`num: 1 (type: number)` is shown.

### Platform

- Router / Start Version: 1.132.0
- OS: macOS
- Browser: Chrome
- Browser Version: 140
- Bundler: vite
- Bundler Version: 7.1.7

numToStr:
I believe I'm facing the same issue, when I remove the type argument from the `handler` the app runs without any issue otherwise I get this error 

```sh
Query data cannot be undefined. Please make sure to return a value other than undefined from your query function. Affected query key: ["cart"]
Error: ["cart"] data is undefined
    at Query.fetch (/app/node_modules/.pnpm/@tanstack+query-core@5.90.5/node_modules/@tanstack/query-core/src/query.ts:556:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
Error: ["cart"] data is undefined
    at Query.fetch (/app/node_modules/.pnpm/@tanstack+query-core@5.90.5/node_modules/@tanstack/query-core/src/query.ts:556:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  cause: Error: ["cart"] data is undefined
      at Query.fetch (/app/node_modules/.pnpm/@tanstack+query-core@5.90.5/node_modules/@tanstack/query-core/src/query.ts:556:15)
[...]

PR #5497, opened 2025-10-16
fix: update regex to include angle brackets in handler filter

yuta-ike:
# Summary

This resolves https://github.com/TanStack/router/issues/5496.

When a type argument is passed to the handler function, as shown below, the server-fn code transformation doesn't occur. 

This PR changes the behavior so that the server-fn is detected correctly even when a type argument is specified

```ts
const getNum = createServerFn({
  method: "GET",
}).handler<number>(() => 1);
```

I struggled with how to write the test. Could you give me some guidance if needed?

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handler detection in the server plugin to recognize additional handler syntax variations, ensuring handlers are properly identified across different usage patterns.


schiller-manuel:
types are inferred. this syntax is not supported

PR #5510, opened 2025-10-17
fix(router-core): add support for handling any kind of empty _splat consistently

### LINKED ISSUE #5385: Trailing slash options isn't enforced with splat route

Router

### Describe the bug

When you have a route defined with a splat such as `some-testing.route.$.tsx` and have a link like so:

```tsx
<Link to="/some-testing/route/$" params={{ _splat: "" }} />
```

The generated path is always `/some-testing/route/` regardless of `trailingSlash` option. If we try to instead of `{{ _splat: undefined }}` the route is `/some-testing/route/undefined` which is also incorrect.

https://stackblitz.com/edit/github-oehqi8ag-yf2ynpjd?file=src%2Froutes%2F__root.tsx

1. Open stackblitz
2. Hover over Trailing Slash link and notice the href ends in slash

### Expected behavior

Because the default `trailingSlash` option is set to `never`, it should always remove the trailing slash. Also, an undefined `_splat` should probably not render as `undefined`.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.132.33
- OS: macOS
- Browser: Chrome
- Browser Version: 140.0.7339
- Bundler: Vite
- Bundler Version: 7.1.17

### Additional context

_No response_

paul-sachs:
When expanding a splat, `undefined` should be handled exactly like if the `_splat` param didn't exist at all. This allows consistent handling of _splat routes to ensure that trailing slashes are kept or dropped according to the spec.

Fixes #5385

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistent handling of empty, undefined, or omitted "splat" segments so routes, links, and navigation resolve predictably across trailing-slash settings.

* **New Features**
  * Trailing-slash behavior is now exposed and configurable, making URL trailing-slash handling explicit.

* **Tests**
  * Added end-to-end tests covering splat-route link generation and navigation for empty, undefined, and missing splat values across trailing-slash modes.

### DIFF
--- a/packages/solid-router/tests/navigate.test.tsx
+++ b/packages/solid-router/tests/navigate.test.tsx
@@ -1,6 +1,7 @@
 import { afterEach, describe, expect, it, vi } from 'vitest'
 
 import { trailingSlashOptions } from '@tanstack/router-core'
+import { waitFor } from '@solidjs/testing-library'
 import {
   createMemoryHistory,
   createRootRoute,
@@ -8,7 +9,6 @@ import {
   createRouter,
 } from '../src'
 import type { RouterHistory } from '../src'
-import { waitFor } from '@solidjs/testing-library'
 
 afterEach(() => {
   vi.clearAllMocks()


schiller-manuel:
for which scenarios is this relevant? navigating an setting `_splat` to e.g. `undefined`?
we should add tests for those scenarios as well

nlynzaad:
thanks for the unit tests on path and Link, as Manuel mentioned please add tests for navigate as well, both useNavigate and navigate, just to ensure consistency. 

Apart from that it's a go from me.

paul-sachs:
@schiller-manuel Added additional test coverage. I followed the suggestions of CodeRabbit as well, to better cover `trailingSlash` potential options even though I know `preserve` doesn't really make that much of a difference but it was easy enough to modify so might as well test it.

nlynzaad:
@paul-sachs thanks a lot for this.

I made some minor updates:
1) exported the constant from router-core so it can be reused more globally
2) replicated the tests for solid js

PR #5533, opened 2025-10-19
docs: Update migrate-from-next-js.md

HirotoShioi:
Fix how `__root.tsx` should be modified when migrating from Nextjs so that it loads css files. 

I had a issue when I've tried to migrate from Nextjs to Tanstack Start where my css file was not being loaded which caused my app to briefly show un-styled content.
This modification fixed it so I thought it would be cool to share it with other people who will most likely face this issue when migration from nextjs


HirotoShioi:
I've copied from this template https://github.com/TanStack/router/blob/main/examples/react/start-basic-react-query/src/routes/__root.tsx#L15-L37

PR #5534, opened 2025-10-19
fix: Backslash incorrectly decoded (#3962)

DreierF:
## Summary by CodeRabbit

* **Bug Fixes**
  * Pathname parsing now preserves percent-encoded sequences (e.g., encoded backslashes) within path segments instead of decoding them, preventing unintended changes to parameters.

* **Tests**
  * Added coverage to verify encoded characters in path parameters remain intact during parsing.

### DIFF
--- a/packages/router-core/src/path.ts
+++ b/packages/router-core/src/path.ts
@@ -348,12 +348,7 @@ function baseParsePathname(
       // Handle regular pathname segment
       return {
         type: SEGMENT_TYPE_PATHNAME,
-        value: partToMatch.includes('%25')
-          ? partToMatch
-              .split('%25')
-              .map((segment) => decodeURI(segment))
-              .join('%25')
-          : decodeURI(partToMatch),
+        value: partToMatch,
       }
     }),
   )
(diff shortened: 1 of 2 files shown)


nlynzaad:
Thanks for the this.

When replicating the test on its own into main, without the accompanied fix, I'm expecting the test to fail. Unfortunately, this passes already without the above changes.

We should have a failing test to which the applied changes resolve the issue. This is to ensure that we both understand the need for the change and to ensure future updates don't inadvertently break the functionality without the test giving a false positive.

DreierF:
🤦‍♂️ Good catch @nlynzaad! The new test fails without the change.

nlynzaad:
@DreierF thanks again for this. it helped a lot to understand the problem and the road to a solution. unfortunately, we need to decode the path segments and should just exclude the backslash similar to how percentage signs are currently handled.

I've been able to wire up a decoding function that is able to handle these requirements as well as any further additions we might need to specifically ignore.

I'm closing this PR in favour of #5663

PR #5537, opened 2025-10-19
fix(start-server-core): use headers.entries() to iterate headers correctly

rhyek:
## Summary

This PR fixes a critical bug in setResponseHeaders() where no headers were being set at all.

## The Bug

The previous code used Object.entries(headers) to iterate over a Headers object:

```typescript
for (const [name, value] of Object.entries(headers)) {
  event.res.headers.set(name, value)
}
```

**Problem:** Object.entries() doesn't work on Headers objects (Web API interface). It returns an empty array, so **no headers were being set**.

## The Fix

Changed to use the correct Web API method headers.entries():

```typescript
for (const [name, value] of headers.entries()) {
  ...
}
```

**Benefits:**
- ✅ Headers are now actually set (fixes the critical bug)
- ✅ First occurrence uses .set() to replace existing headers
- ✅ Subsequent occurrences use .append() to preserve multiple headers with the same name (e.g., Set-Cookie)

## Test Coverage

Added comprehensive tests in packages/start-server-core/tests/request-response.test.ts:
- Test for setting single header
- Test for setting multiple headers
- Test for iterating through Headers using .entries()
- Test for handling empty Headers object
- Test for multiple headers with the same name (e.g., multiple Set-Cookie headers)

All tests pass ✓

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved header handling to support multiple identical header names in a single response.
  * Ensures the first occurrence replaces existing values while subsequent duplicates are appended, preventing unintended overwrites.

* **Tests**
  * Added tests covering single/multiple headers, header replacement, empty headers, and multiple Set-Cookie entries.

### DIFF
--- a/packages/start-server-core/src/request-response.ts
+++ b/packages/start-server-core/src/request-response.ts
@@ -136,8 +136,16 @@ export function setResponseHeaders(
   headers: TypedHeaders<ResponseHeaderMap>,
 ): void {
   const event = getH3Event()
-  for (const [name, value] of Object.entries(headers)) {
-    event.res.headers.set(name, value)
+  const addedHeaderNames: Record<string, true> = {}
+  for (const [name, value] of headers.entries()) {
+    const found = addedHeaderNames[name] ?? false
+    if (!found) {
+      addedHeaderNames[name] = true
+    }
+    // If header already existed in h3 event headers, it will be replaced.
+    // However, headers object in this invocation might have multiple instances of the same header name (.append() was used), let's allow the duplicates.
+    const method = found ? 'append' : 'set'
+    event.res.headers[method](name, value)
   }
 }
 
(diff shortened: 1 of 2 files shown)


schiller-manuel:
thanks for the PR, we are currently reworking header handling, so either we will merge this PR or apply similar changes. let's keep this open until we are ready

rhyek:
> thanks for the PR, we are currently reworking header handling, so either we will merge this PR or apply similar changes. let's keep this open until we are ready

Ah, that makes sense. Was kind of hoping to get my contributor badge, but np! 😆

rhyek:
FYI, just getting the current headers object using `getResponseHeaders()` and updating headers through that works and you have the full Headers Fetch API available.

Maybe tanstack should just rely on the user doing that.

PR #5568, opened 2025-10-21
feat: support lazy error component on beforeLoad failures

beaussan:
Current behaviour: if you .lazy() an error component on a beforeLoad, then the error component won't be use and the default one will be instead

Reproduction of current error:  https://stackblitz.com/edit/tanstack-router-wiqzauc7?file=src%2Fmain.tsx

### Open questions:

I have the following test failing (most likely due to the loading of the lazy component):

```
 FAIL   @tanstack/react-router  tests/store-updates-during-navigation.test.tsx > Store doesn't update *too many* times during navigation > not found in beforeLoad
AssertionError: expected 8 to be 7 // Object.is equality

- Expected
+ Received

- 7
+ 8
```

Is that ok to bump to 8 ?

## Release Notes

* **Bug Fixes**
  * Enhanced error handling flow to ensure route resources are properly loaded before error processing
  * Improved support for lazy-loaded error components across router implementations


packages/react-router/src/useNavigate.tsx
```diff
@@ -7,6 +7,7 @@ import type {
   RegisteredRouter,
   UseNavigateResult,
 } from '@tanstack/router-core'
+
```
beaussan, line 11: This was causing the lint to fail, is that expected that it failed on main too ?

Sheraff:
> Is that ok to bump to 8 ?

Yes that seems good

beaussan:
> Yes that seems good

Great, thanks! Then I guess the pr is good to go from my point of view

Sheraff:
> Then I guess the pr is good to go from my point of view

I do think we should at least make sure not to load the components if it's throwing a redirect.

beaussan:
> I do think we should at least make sure not to load the components if it's throwing a redirect.

Done

packages/router-core/src/load-matches.ts
```diff
@@ -466,14 +466,22 @@ const executeBeforeLoad = (
...
     }
   } catch (err) {
     pending()
-    handleSerialError(inner, index, err, 'BEFORE_LOAD')
+    if (!isRedirect(err)) {
+      return loadLazyRouteChunks(route).then(() => {
+        handleSerialError(inner, index, err, 'BEFORE_LOAD')
+      })
+    }
+    return handleSerialError(inner, index, err, 'BEFORE_LOAD')
```
Sheraff: This part didn't use to return. Are we sure we need to return now? We're skipping the `updateContext()` call below

beaussan: Yeah that's what I was also wondering, let me change that, and introduce also the extra `updateContext()` in the `.then()` right above

beaussan:
https://github.com/TanStack/router/pull/7068 takes over this pr

PR #5571, opened 2025-10-21
fix: move solid-js to prod dep on devtools-core

beaussan:
Given vite only dedupe when a dependancy is an explicit dependancy, and not a peer dependancy, this lead to multiple solidjs version if you used other tanstack devtools

## Summary by CodeRabbit

* **Chores**
  * Refined dependency management to remove duplicate entries and streamline package installation.
  * Adjusted build/bundling configuration so certain libraries are no longer explicitly bundled, reducing package size and simplifying distribution.


AlemTuzlak:
This is a fix for:
https://github.com/TanStack/devtools/issues/183

birkskyum:
Does this effectively revert the changes here?
- https://github.com/TanStack/router/pull/5422

@benmccann , what do you think of this change?

benmccann:
Yes, this reverts the bundling. I do agree that not bundling is better for solid users. However, that's less than 1% of users. And not bundling is worse for everyone else. So I think on balance that we're better off with the bundling. I included some stats about that in https://github.com/TanStack/router/pull/5374

Is the concern with multiple copies of solid related to size or something else?

beaussan:
I don't have the full context on the changes, @AlemTuzlak doses have way more info around it. 

From https://github.com/TanStack/devtools/issues/183, it seems like the issue around multiple solid js version is not related to bundle size, but to have multiple version of solid-js within warnings in the console and could (_from my limited solid-js architecture knowledge_) cause issues.

image


I have doubts however if this is actually causing issues given we mount each devtools into it's own solid-js render tree

> By bundling solid we end up with a much smaller package since we keep only the parts of it that are used (~40kb) and tree-shake away the rest. This prevents deduping for solid users, so it's larger for those users. However, there are 3,700 weekly downloads from solid and 1 million from react, so this is a huge savings overall.

Are we talking only about npm install size? It shouldn't be in production (unless you use the devtools in production)

LadyBluenotes:
Going to close this since it's pretty stale

PR #5517, opened 2025-10-17
fix: refactor middleware

tannerlinsley:
fixes #5372
fixes #5266
fixes #4460
fixes #5239

## Summary by CodeRabbit

* **New Features**
  * Added a new page and navigation link for "Server Functions Middleware Unhandled Exception" E2E tests.

* **Bug Fixes**
  * Improved middleware execution and client/server error/redirect propagation for more consistent handling.
  * Enhanced server response handling to support streaming and structured server-function payloads.

* **Chores**
  * Removed a legacy middleware export from the public API.
  * Added a context header constant and updated client/server fetch plumbing to support the new server-function flow.

PR #5583, opened 2025-10-22
feat: enable async serverFn validation

canac:
The `inputValidator` in `createServerFn` doesn't currently support async validate. However, the `execValidator` result is awaited, even though it doesn't return a promise. Thus, I don't see a technical reason why `inputValidator` couldn't support async validation. I'm proposing this change because I'm hoping to be able to use an async zod refiner in my application.

## Summary by CodeRabbit

* **New Features**
  * Added support for async validation in server functions, allowing validators to perform asynchronous operations.
  * Introduced a new async validation demo page.

* **Bug Fixes**
  * Fixed a typo in navigation text.

* **Tests**
  * Added comprehensive test coverage for async validation scenarios with valid and invalid inputs.
  * Expanded type tests for various validator configurations.

### DIFF
--- a/e2e/react-start/server-functions/src/routeTree.gen.ts
+++ b/e2e/react-start/server-functions/src/routeTree.gen.ts
@@ -23,6 +23,7 @@ import { Route as FormdataContextRouteImport } from './routes/formdata-context'
 import { Route as EnvOnlyRouteImport } from './routes/env-only'
 import { Route as DeadCodePreserveRouteImport } from './routes/dead-code-preserve'
 import { Route as ConsistentRouteImport } from './routes/consistent'
+import { Route as AsyncValidationRouteImport } from './routes/async-validation'
 import { Route as IndexRouteImport } from './routes/index'
 import { Route as RedirectTestIndexRouteImport } from './routes/redirect-test/index'
 import { Route as RedirectTestSsrIndexRouteImport } from './routes/redirect-test-ssr/index'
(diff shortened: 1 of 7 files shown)


schiller-manuel:
interesting!
we should also make sure any kind of async function works, in terms of types and runtime

we definitely need e2e tests in e2e/react-start/server-functions

canac:
@schiller-manuel I tried to get the type-level test to work with a Standard validator like this:

```ts
const validator = {
  ['~standard']: {
     validate: (input: { input: string }) => ({ value: { a: input } }),
  },
}
```

~~But the type inference wasn't working. That's why I added zod. If you don't want to add zod to that package, I may need help with the "createServerFn with standard validator" test. With a working baseline, I'm sure I can get a working "createServerFn with async standard validator".~~

**EDIT**: I got it working without zod using the standard schema validator interface. 🎉

schiller-manuel:
can we also just to `inputValidator(async(foo:string) => foo)` ?

schiller-manuel:
the types for this are not yet correct:

```tsx
  const someFunction   = createServerFn()
  .inputValidator(async(data: string) => data)
  .handler(({ data }) => data)
```

Screenshot 2025-10-26 at 19 55 54

schiller-manuel:
we will need to wait with this until https://github.com/TanStack/router/pull/5517 is merged

canac:
> we will need to wait with this until #5517 is merged

@schiller-manuel Got it. I'll take a look to see if any changes are needed after that PR merges. Thanks for the code review so far and for pointing out improvements!

canac:
@schiller-manuel I rebased on `main` and force-pushed. There were some merge conflicts but no real code changes required.

PR #5584, opened 2025-10-22
docs: fixed example of env variable validation with zod

max-programming:
This PR changes the example provided in the docs for **environment variables validation with `zod`**

The current example in the docs is error-prone because it validates both client and server variables in the same file and when imported in any client-side code, the server side validation fails because `process.env` does not exist on client-side

So I make the following changes

1. Make use of the `createIsomorphicFn` to validate the variables in their respective environments
2. Intersect client side variables with server schema using zod's `.and` method because client-side variables are available in the server environment
3. Export a singular `env` object and typecast it to the server env schema type so that all variables are detected by TypeScript, just that in runtime if a server variable is used on client side, it'll fail. TypeScript cannot know which file is client-side code hence the typecast

## Summary by CodeRabbit

* **New Features**
  * Unified isomorphic environment handling: single, consistent env access across server and client via a new isomorphic getter.

* **Documentation**
  * Updated environment variable guides for React and Solid to reflect the unified approach and new usage.

### DIFF
--- a/docs/start/framework/react/guide/environment-variables.md
+++ b/docs/start/framework/react/guide/environment-variables.md
@@ -278,13 +278,8 @@ Use Zod for runtime validation of environment variables:
 
 ```typescript
 // src/config/env.ts
+import { createIsomorphicFn } from "@tanstack/react-start";
 import { z } from 'zod'
-
-const envSchema = z.object({
-  DATABASE_URL: z.string().url(),
-  JWT_SECRET: z.string().min(32),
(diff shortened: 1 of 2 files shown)


LadyBluenotes:
Since this was opened, the environment variable guide has been updated to cover request-time env handling on Cloudflare and other edge runtimes. This approach still calls getEnv() at module scope and also types server-only variables as available in client code, so I’m going to close this PR. The underlying client import issue is still valid, but it needs a solution that preserves the server/client boundary.

PR #5590, opened 2025-10-23
Fix React router Link component in SVG elements

dsdeur:
## Problem
When using the React Link component in an SVG it causes a full refresh, unless  the `target` prop is explicitly set. 

## Cause
This is caused by retrieving the element target with `.target`:
https://github.com/TanStack/router/blob/91d994046a2add5bdc25cea91092634b615fdeb8/packages/react-router/src/link.tsx#L246 

In an SVG, `element.target` returns an [SVGAnimatedString](https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString) instead of a string, making the target check evaluate to false:
https://github.com/TanStack/router/blob/91d994046a2add5bdc25cea91092634b615fdeb8/packages/react-router/src/link.tsx#L253

Resulting in the navigate function to not be triggered and the default link behavior to cause a full refresh. 

## Example/demo
https://stackblitz.com/edit/github-9fsuvtph?file=src%2Froutes%2Findex.tsx
- The red rectangle causes a full refresh as the target check fails
- The blue rectangle works as it should, as the target prop is set, which overrules the element target

## Solution
The easiest solution seems to be to use `.getAttribute("target")` instead of `.target` as that returns null or the value instead of the SVGAnimatedString.

(This is the same approach as [Next.js link uses](https://github.com/vercel/next.js/blob/7e41c122c3767325e5e5db4dbdc8733ee2144b1c/packages/next/src/client/link.tsx#L192)) 

## Summary by CodeRabbit

* **New Features**
  * Added accessible, clickable SVG links in app examples that navigate to the Posts page.

* **Bug Fixes**
  * Improved link click handling so SVG-based links respect link targets and no longer cause full page reloads.

* **Tests**
  * Added end-to-end tests verifying SVG links navigate to /posts without triggering a full page reload.

### DIFF
--- a/e2e/react-router/basic/src/main.tsx
+++ b/e2e/react-router/basic/src/main.tsx
@@ -69,7 +69,23 @@ function RootComponent() {
           }}
         >
           This Route Does Not Exist
-        </Link>
+        </Link>{' '}
+        <div className="flex items-center">
+          <svg width="20" height="20" viewBox="0 0 20 20" role="img">
+            <title id="rectTitle">Link in SVG</title>
+            <Link to="/posts" aria-label="Open posts from SVG">
+              <rect
+                x="0"
+                y="0"
+                width="20"
+                height="20"
+                rx="4"
+                fill="blue"
+                strokeWidth="2"
+              />
+            </Link>
+          </svg>
+        </div>
       </div>
       <Outlet />
       <TanStackRouterDevtools position="bottom-right" />
--- a/e2e/react-router/basic/tests/app.spec.ts
+++ b/e2e/react-router/basic/tests/app.spec.ts
@@ -1,4 +1,8 @@
 import { expect, test } from '@playwright/test'
+import { getTestServerPort } from '@tanstack/router-e2e-utils'
+import packageJson from '../package.json' with { type: 'json' }
+
+const PORT = await getTestServerPort(packageJson.name)
 
 test.beforeEach(async ({ page }) => {
   await page.goto('/')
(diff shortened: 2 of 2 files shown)


SeanCassiere (commented):
Could you please apply this on the Solid side please? The implementation should mostly be the same.

schiller-manuel:
can you please also add an e2e test?

dsdeur:
## React
I’ve added an e2e test for the React version to the basic example. For detecting the page reload I used the same method as in [redirect.spec.ts](https://github.com/TanStack/router/blob/main/e2e/react-start/basic/tests/redirect.spec.ts#L63).

## Solid

I’ve also looked into Solid and this turned out to be more tricky than the React version, mainly because of two quirks in Solid:
- `a` elements in components are rendered as `HTMLAnchorElement` instead of `SVGAElement` unless it's directly in an `svg`, [see example](https://playground.solidjs.com/anonymous/a79a4377-d297-4690-8563-a81d110c89c9). The blue rect is not rendered, and in the console you can see the types. 
  - Edit: For context https://github.com/solidjs/solid/issues/2048
- A dynamic class attribute on an `SVGAElement` fails, [see example](https://playground.solidjs.com/anonymous/2465ef70-cf0b-4239-8837-21b6613f395c). In the console you can see `Uncaught TypeError: Cannot set property className of #<SVGElement> which has only a getter` (seems like a similar attribute accessing issue as we are solving here).
  - Edit: https://github.com/solidjs/solid/issues/258
 
As you can see in the examples, these are not specific to Tanstack router.

I have prepared a fix for the original issue, but it required a bit of a workaround in the e2e test because of the mentioned issues. For now I have kept this separate so you can decide if you want to include it or not: https://github.com/dsdeur/tanstack-router/pull/1/files 

Just let me know, and I can include it in this PR or open a new one.

schiller-manuel:
@brenelz @birkskyum  can you please have a look at the solid part here?

birkskyum (approved):

birkskyum:
This is apparently much harder to do in solid than react - I've opened a [ticket](https://github.com/TanStack/router/issues/5632) for it, but I think we can fix it for react now, we should just go ahead, and we'll add solid tests when we have a solution for it.

PR #5609, opened 2025-10-24
docs: add note to set CLOUDFLARE_ENV for wrangler to load correct env…

lotap:
This PR adds a note to the environment variable docs to add `CLOUDFLARE_ENV` to `.env.<environment>` files.

Without specifying a Cloudflare environment, it will default to the top-level (production), so vars specified in .env.development would never be loaded

[Cloudflare Docs about .env on local](https://developers.cloudflare.com/workers/configuration/environment-variables/#local-development-with-secrets)

The solution comes from [Combining Cloudflare environments and Vite modes](https://developers.cloudflare.com/workers/vite-plugin/reference/cloudflare-environments/#combining-cloudflare-environments-and-vite-modes).

Vite loads the correct `.env.<environment>` file and that file specifies `CLOUDFLARE_ENV`. Then the Cloudflare Vite plugin will select the given Cloudflare environment, which means Wrangler will load the corresponding `.env.<environment>` (even if it's the same file where `CLOUDFLARE_ENV` was specified)

Setting `CLOUDFLARE_ENV=development` in `.env.development` should be enough to get most people started

## Summary by CodeRabbit

* **Documentation**
  * Added guidance to set CLOUDFLARE_ENV when using the Cloudflare Vite plugin so Wrangler loads the correct environment and environment-specific variables are injected into process.env. Tip added within the Environment Variables guides for both React and Solid, including an example for .env.development.

### DIFF
--- a/docs/start/framework/react/guide/environment-variables.md
+++ b/docs/start/framework/react/guide/environment-variables.md
@@ -102,8 +102,8 @@ TanStack Start automatically loads environment files in this order:
 > When using the Cloudflare Vite plugin, set `CLOUDFLARE_ENV` to ensure Wrangler loads the correct environment
 >
 > ```bash
->  # .env.development
->  CLOUDFLARE_ENV=development
+> # .env.development
+> CLOUDFLARE_ENV=development
 > ```
 >
 > Otherwise, the variables defined in your `.env.<environment>` files may not be read or injected into `process.env`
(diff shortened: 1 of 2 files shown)


LadyBluenotes:
Sorry this sat. This guidance is still useful, but the docs structure has changed. Could you open a new PR with the current Cloudflare Vite plugin behavior, and update only the React guide? The Solid guide is generated from it now

PR #5643, opened 2025-10-26
test(start): fix search-params tests

birkskyum:
@schiller-manuel , i found that these tests had `toBeTruthy` that weren't called, and they break if they are enabled, which is demonstrated here:
- https://github.com/TanStack/router/pull/5656

I think they should just be checking `expect(request.redirectedFrom()).toBeNull()`

## Summary by CodeRabbit

* **Tests**
  * Simplified redirect verification assertions in end-to-end tests across framework implementations to improve test clarity and maintainability.


birkskyum:
@schiller-manuel , i found that these tests had `toBeTruthy` that weren't called, and they break if they are enabled. I think they should just be checking `expect(request.redirectedFrom()).toBeNull()`

PR #5652, opened 2025-10-27
feat: add `$withType<TFn>()` to restrict isomorphicFn

juliusmarminge:
allows restricting the type on both "branches" in an isomorphic fn.

## Notes

- Use special naming convention `$withType` to signal that it's a special thing, does not have any affect at runtime
- When called with `$withType()`, both `.server()` and `.client()` must be chained
  - This is not (currently) enforced by the compiler, just typescript.
  - An argument could be made that when the return type of the type parameter passed to `$withType` extends `undefined`, then it should be fine that one of server/client can be omitted? idk...

## Summary by CodeRabbit

* **New Features**
  * Added $withType() to isomorphic functions to declare explicit server/client signatures and expose typed server-only, client-only, and both-side variants.

* **Tests**
  * Added type-checking tests validating typed isomorphic-function chains and expected type errors.

* **Documentation**
  * Added docs demonstrating how to restrict and enforce server/client implementation signatures with examples.

### DIFF
--- a/docs/start/framework/react/guide/environment-functions.md
+++ b/docs/start/framework/react/guide/environment-functions.md
@@ -83,6 +83,28 @@ A no-op (short for "no operation") is a function that does nothing when executed
 function noop() {}
 ```
 
+### Restricting the type of the implementations
+
+You can use `.$withType<TFn>()` to enforce the type of both the server and client implementations matches the same function signature.
+
+```tsx
+import { createIsomorphicFn } from '@tanstack/react-start'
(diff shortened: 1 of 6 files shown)


LadyBluenotes:
Thanks for exploring a typed createIsomorphicFn builder and including type tests and documentation. This introduces a new public $withType API and changes the expected client/server chaining contract, which needs fresh maintainer agreement before moving forward. Since the branch now conflicts with main and the proposal has gone stale, we’re going to close it rather than carry the API forward in its current form

PR #5663, opened 2025-10-28
refactor(router-core): more stable url decoding

nlynzaad:
This PR addresses the issue filed in #3962 and follows on from #5534 and #3987. As mentioned in #3987 we should ideally find a better solution to deal with this, however these will bring changes to users and ideally, we should leave these for the next major, this solution attempts to deal with this problem without breaking current use cases.

when backslashes are used in a path param the decoded value `\` is removed from the pathname. we should not be decoding `%5C` to `\` when decoding the various path segments. We already do this for `%25` (percentage signs) and a more complete solution is needed for these special characters that should not be decoded.

This PR adds a new utility function decodePathSegment that will decode the path segment ignoring a defined array of encoded values. As a default we are not decoding `\`, `%` and `/`. In addition to these, the defined list of special characters that can be [allowed](https://tanstack.com/router/latest/docs/framework/react/api/router/RouterOptionsType#pathparamsallowedcharacters-property) to not be encoded has also been added here.

While working on this PR it also came to fore that when navigating to a path that contains special characters these are shown in the URL in encoded form, however when refreshing the page or browsing to the path directly these are shown in decoded form. In addition to the above we are also now standardizing this to show the decoded form on navigation similar as to when browsing directly to the path.

This PR also adds tests for the new utility function as well updates the solid-router and react-router e2e tests to test the functionality.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved decoding of escaped characters in URL segments (backslashes, percent-encodings, emoji) for more accurate route matching and displayed params.

* **New Features**
  * Added example links and new wildcard route variants demonstrating complex escaped parameter values, including non‑ASCII tokens and escaped splats.

* **Tests**
  * Expanded test coverage for encoded, escaped, and Unicode path scenarios (wildcard, prefix, suffix, and link navigation).

### DIFF
--- a/e2e/react-start/server-routes/package.json
+++ b/e2e/react-start/server-routes/package.json
@@ -11,7 +11,7 @@
     "test:e2e": "playwright test --project=chromium"
   },
   "dependencies": {
-    "@tanstack/react-query": "5.90.0",
+    "@tanstack/react-query": "^5.90.0",
     "@tanstack/react-router": "workspace:^",
     "@tanstack/react-router-devtools": "workspace:^",
     "@tanstack/react-router-ssr-query": "workspace:^",
--- a/e2e/solid-start/basic-solid-query/src/router.tsx
+++ b/e2e/solid-start/basic-solid-query/src/router.tsx
@@ -9,8 +9,7 @@ export function getRouter() {
   const queryClient = new QueryClient({
     defaultOptions: {
       queries: {
-        // Enable only on client to avoid hydration mismatches
-        experimental_prefetchInRender: typeof window !== 'undefined',
+        experimental_prefetchInRender: true,
       },
     },
   })
--- a/e2e/solid-start/query-integration/src/router.tsx
+++ b/e2e/solid-start/query-integration/src/router.tsx
@@ -4,7 +4,13 @@ import { setupRouterSsrQueryIntegration } from '@tanstack/solid-router-ssr-query
 import { routeTree } from './routeTree.gen'
 
 export function getRouter() {
-  const queryClient = new QueryClient()
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: {
+        experimental_prefetchInRender: true,
+      },
+    },
+  })
   const router = createRouter({
     routeTree,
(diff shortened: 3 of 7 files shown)


packages/router-core/src/utils.ts
```diff
@@ -488,3 +488,57 @@ export function findLast<T>(
   }
   return undefined
 }
+
+const DECODE_IGNORE_LIST = [
```
SeanCassiere: Although it'll be negligible, could we precompute this into a `Set`? Perhaps even have them in their encoded form already?

packages/router-core/src/utils.ts
```diff
@@ -488,3 +488,57 @@ export function findLast<T>(
...
+        try {
+          return decodeURIComponent(match)
+        } catch {
+          return match
+        }
+      })
+    }
+  }
+
+  if (part === '' || !part.match(/%../g)) return part
```
SeanCassiere: Comment from here to end of the block please, for future adventurers.

packages/router-core/src/router.ts
```diff
@@ -1670,16 +1671,18 @@ export class RouterCore<
...
-        // Use the original template path for interpolation
-        // This preserves the original parameter syntax including optional parameters
-        path: nextTo,
-        params: nextParams,
-        leaveWildcards: false,
-        leaveParams: opts.leaveParams,
-        decodeCharMap: this.pathParamsDecodeCharMap,
-        parseCache: this.parsePathnameCache,
-      }).interpolatedPath
+      const nextPathname = decodePathSegment(
```
SeanCassiere, line 1674: How does this play with the prefix and suffix groups?

For the most part I think it should be fine, since the common characters are in the `DECODE_IGNORE_LIST`.

Rather I just want to make sure our testing covers it too.

nlynzaad: I'll add some test cases to the e2e test similar to the one for splats

nlynzaad: okay think we are testing all of the impacted areas now.

PR #5676, opened 2025-10-28
Implement meta tag deduplication logic

### LINKED ISSUE #5146: Set multiple 'theme-color' in head

Router

### Describe the bug

Following the doc of MDN : https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/theme-color

I would like to add 2 meta tags with the same name.

```html
<meta
  name="theme-color"
  content="cornflowerblue"
  media="(prefers-color-scheme: light)" />
<meta
  name="theme-color"
  content="dimgray"
  media="(prefers-color-scheme: dark)" />
```

But TanStack router says : _"meta tags with the same **name** or property will be overridden by the last occurrence of that tag found in nested routes"_ 
Src : https://tanstack.com/router/v1/docs/framework/react/guide/document-head-management

So how to handle my use case ? 

https://codesandbox.io/p/devbox/pedantic-poitras-7hrmrd?workspaceId=ws_VUPqXCcptK4abDwy6H7mEq

1. Go to https://codesandbox.io/p/devbox/pedantic-poitras-7hrmrd?workspaceId=ws_VUPqXCcptK4abDwy6H7mEq
2. Check head tag
3. Check that there is only 1 "theme-color" tag instead of 2.

### Expected behavior

As a user, I expect to see 2 theme color meta tags

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: Last version

### Additional context

_No response_

juliusmarminge:
Add logic to determine if meta tags should be deduplicated.

Closes #5146

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved meta tag handling to allow multiple theme-color entries while preserving automatic deduplication for other meta attributes.

### DIFF
--- a/packages/react-router/src/HeadContent.tsx
+++ b/packages/react-router/src/HeadContent.tsx
@@ -4,6 +4,12 @@ import { useRouter } from './useRouter'
 import { useRouterState } from './useRouterState'
 import type { RouterManagedTag } from '@tanstack/router-core'
 
+const METAS_TO_ALLOW_MULTIPLE = new Set(['theme-color'])
+
+function shouldDeduplicateMetaTag(attribute: string) {
+  return METAS_TO_ALLOW_MULTIPLE.has(attribute) === false
+}
+
 /**
  * Build the list of head/link/meta/script tags to render for active matches.
  * Used internally by `HeadContent`.
@@ -37,7 +43,7 @@ export const useTags = () => {
         } else {
           const attribute = m.name ?? m.property
           if (attribute) {
-            if (metaByAttribute[attribute]) {
+            if (metaByAttribute[attribute] && shouldDeduplicateMetaTag(attribute)) {
               continue
             } else {
               metaByAttribute[attribute] = true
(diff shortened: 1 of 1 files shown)


packages/react-router/src/HeadContent.tsx
```diff
@@ -4,6 +4,12 @@ import { useRouter } from './useRouter'
 import { useRouterState } from './useRouterState'
 import type { RouterManagedTag } from '@tanstack/router-core'
 
+const METAS_TO_ALLOW_MULTIPLE = new Set(['theme-color'])
```
morinokami, line 7: I believe there are other meta tags that can validly appear multiple times in the same document, such as `article:author` and `og:image` (https://ogp.me/#array). It might be worth considering expanding the allowed list to cover a broader range of repeatable meta tags.

LadyBluenotes:
The HeadContent logic has since moved into headContentUtils.tsx, where deduplication now happens in two code paths and needs regression coverage. I’m going to close this PR and leave #5146 open for an updated implementation

PR #5694, opened 2025-10-30
docs: add note that prerendering is not yet supported on cf workers

anyuj:
- Why
As noted on Discord: “right now prerendering does not run using wrangler and hence fails when you reference cloudflare env during prerendering”.
- What changed
Documentation update: added a note in the Static Prerendering section.
- Scope/Impact
Docs only.

## Summary by CodeRabbit

* **Documentation**
  * Updated static-prerendering guides for React and Solid frameworks with informational notes clarifying that prerendering on Cloudflare Workers is not yet supported and currently under development. Notes have been added to inform developers of current limitations.

### DIFF
--- a/docs/start/framework/react/guide/static-prerendering.md
+++ b/docs/start/framework/react/guide/static-prerendering.md
@@ -68,6 +68,7 @@ export default defineConfig({
   ],
 })
 ```
+Note: Prerendering on Cloudflare Workers is not yet supported, but is currently under development.
 
 ## Automatic Static Route Discovery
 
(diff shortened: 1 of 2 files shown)


daveycodez:
There is a workaround. Simply remove the Cloudflare plugin and use this wrangler -

```tsx
{
    "$schema": "node_modules/wrangler/config-schema.json",
    "name": "start-intl-example",
    "compatibility_date": "2025-09-02",
    "compatibility_flags": ["nodejs_compat"],
    "main": "./dist/server/server.js",
    "assets": {
        "directory": "dist/client",
        "html_handling": "drop-trailing-slash"
    }
}
```

Bindings won't work, but Environment variables do. Which suffices for most use cases.

PR #5700, opened 2025-10-30
Update the authentication guide to use useQuery within the AuthProvider context.

Callumk7:
The guide on authentication implies that `useServerFn` has a similar api to `useQuery`:

```tsx
export function AuthProvider({ children }: { children: ReactNode }) {
  const { data: user, isLoading, refetch } = useServerFn(getCurrentUserFn)

  return (
    <AuthContext.Provider value={{ user, isLoading, refetch }}>
      {children}
    </AuthContext.Provider>
  )
}
```

But this is not the correct return shape if the guide is followed; instead useServerFn will be returning a promise here. As such I have updated this component by wrapping the `userServerFn` with `useQuery`, to avoid any confusion. 

I am not sure if this is deliberate sudo-code here - but I think this suggestion is correct for the example provided, and also aligns with the useage guide on using server functions in components: https://tanstack.com/start/latest/docs/framework/react/guide/server-functions#where-to-call-server-functions

## Summary by CodeRabbit

* **Documentation**
  * Updated authentication guide to reflect improved data loading with enhanced support for caching, refetching, and loading state management in authentication flows.

### DIFF
--- a/docs/start/framework/react/guide/authentication.md
+++ b/docs/start/framework/react/guide/authentication.md
@@ -143,8 +143,10 @@ type AuthContextType = {
 const AuthContext = createContext<AuthContextType | undefined>(undefined)
 
 export function AuthProvider({ children }: { children: ReactNode }) {
-  const { data: user, isLoading, refetch } = useServerFn(getCurrentUserFn)
-
+  const { data: user, isLoading, refetch } = useQuery({
+    queryKey: ["auth"],
+    queryFn: useServerFn(getCurrentUserFn),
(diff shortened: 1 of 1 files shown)


CanRau:
Found this while on my way to open a PR for this 🤓

LadyBluenotes:
You’re right that useServerFn returns a callable function rather than query state. Could you add the missing `useQuery` import and default data to `null` so it matches `AuthContextType` while loading? After that and rebasing against main this should be good.

PR #5701, opened 2025-10-30
Merge __beforeLoadContext for pending loading state

paul-sachs:
Based off of #5002, this takes it a step further and only merges context from previous load (when preloaded is successful) and send that to the component. This ensures the context maintains it's proper type even when pending.

fixes https://github.com/TanStack/router/issues/4998

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed context behavior during route preloading and navigation to prevent stale context from being incorrectly carried over to subsequent route transitions.

* **Tests**
  * Added comprehensive tests validating proper context isolation during preload operations and subsequent navigation sequences, ensuring beforeLoad receives fresh context on each invocation.

### DIFF
--- a/packages/react-router/tests/routeContext.test.tsx
+++ b/packages/react-router/tests/routeContext.test.tsx
@@ -3128,4 +3128,176 @@ describe('useRouteContext in the component', () => {
 
     expect(content).toBeInTheDocument()
   })
+
+  test("reproducer #4998 - on navigate (with preload), route component isn't rendered with undefined context if beforeLoad is pending", async () => {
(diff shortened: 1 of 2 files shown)


paul-sachs:
@coderabbitai generate docstrings

paul-sachs:
Actually, this still doesn't fully solve the problem. The problem is that my component depends on context to such a degree that it cannot render at all until beforeLoad runs, and it seems that setting pendingComponent forces it to render to trigger suspense. This is gonna need some more work to figure out 🤔

PR #5708, opened 2025-10-31
feat(server-functions): add rawHandler for streaming file uploads

### LINKED ISSUE #5704: Support streaming file uploads in server functions

Start

### Describe the bug

Currently, TanStack Start automatically calls `await request.formData()` in server functions when the content-type is `multipart/form-data`, which loads the entire file into memory **before** the handler is invoked. This prevents implementing true streaming file uploads with size limit enforcement.

**Reference:** [`server-functions-handler.js:49`](https://github.com/TanStack/router/blob/main/packages/start-server-core/src/server-functions-handler.ts)

```javascript
if (formDataContentTypes.some(
  (type) => contentType && contentType.includes(type)
)) {
  // ...
  const formData = await request.formData();  // <-- File fully loaded into memory here
  // ...
  return await action(params, signal);  // Handler called after file is in memory
}
```

### Current Behavior

When a user uploads a large file (e.g., 100MB, 1GB, or even 10GB):
1. The entire file is loaded into server memory via `request.formData()`
2. Only then is the server function handler called
3. File size limits cannot be enforced before memory consumption
4. Large uploads can crash the server or cause OOM errors
5. True streaming from network → storage is impossible

### Expected Behavior

Server functions should be able to access the raw request body stream via `getRequest()` **before** TanStack Start parses the FormData, allowing developers to:

1. Enforce file size limits **during** upload (reject at 25MB instead of buffering 10GB)
2. Stream files directly from network to cloud storage (S3, Azure Blob, etc.)
3. Implement proper backpressure handling
4. Minimize memory footprint for file uploads

### Use Case

Streaming file uploads to Azure Blob Storage with size enforcement:

```typescript
export const uploadFileRpc = createServerFn({ method: 'POST' })
  .middleware([authMiddleware])
  .handler(async (ctx) => {
    const request = getRequest();
    
    // Would like to access raw body stream here
    const busboy = new Busboy({
      headers: {
        'content-type': request.headers.get('content-type') || '',
      },
      limits: {
        fileSize: 25 * 1024 * 1024, // Enforce 25MB limit DURING upload
      },
    });

    busboy.on('file', (fieldname, fileStream) => {
      // Stream directly to Azure without buffering in memory
      const webStream = Readable.toWeb(fileStream);
      await objectStorage.upload({ stream: webStream });
    });

    // Currently doesn't work because request.body is already consumed
    Readable.fromWeb(request.body).pipe(busboy);
  });
```

### Current Workaround

The only workaround is to bypass TanStack Start server functions entirely and create separate Hono/Express routes for file uploads, which defeats the purpose of having a unified server function API.

### Impact

This affects any TanStack Start application that needs to:
- Handle large file uploads (videos, backups, datasets)
- Implement proper file size validation
- Stream files to cloud storage
- Build production file upload services

Try to stream the file from user to other source, for example blob storage

### Expected behavior

Add a configuration option or special handler type for server functions that need raw stream access:

**Option 1:** Skip FormData parsing when a flag is set
```typescript
export const uploadFileRpc = createServerFn({ 
  method: 'POST',
  parseBody: false, // Skip automatic FormData parsing
})
  .handler(async (ctx) => {
    const request = getRequest();
    // request.body stream is still available
  });
```

**Option 2:** Provide a separate handler for raw body access
```typescript
export const uploadFileRpc = createServerFn({ method: 'POST' })
  .rawBodyHandler(async (ctx, request) => {
    // Access raw request before any parsing
    // request.body stream is available here
  });
```

### Screenshots or Videos

_No response_

### Platform

- **@tanstack/react-start**: `1.133.20`
- **Runtime**: Node.js with Hono

### Additional context

_No response_

gzaripov:
Adds .rawHandler() method to createServerFn() that provides direct access to the raw Request object without automatic body parsing. This enables:

- Streaming large file uploads directly to cloud storage without buffering in memory
- Enforcing file size limits during upload rather than after
- Implementing custom body parsing with libraries like busboy
- Minimizing memory footprint for file operations

The raw handler receives { request, signal, context } where:
- request: Raw Request object with unread body stream
- signal: AbortSignal for cancellation support
- context: Full middleware context (auth, user info, etc.)

Closes #5704

## Summary by CodeRabbit

* **New Features**
  * Added `rawHandler()` method for server functions, enabling streaming file uploads and direct access to the raw Request object.
  * Provides granular control over request handling with built-in backpressure and memory efficiency.

* **Documentation**
  * Added comprehensive guide covering `rawHandler()` usage, streaming patterns, and comparison with standard `handler()` approach.

### DIFF
--- a/docs/start/framework/react/guide/server-functions.md
+++ b/docs/start/framework/react/guide/server-functions.md
@@ -127,6 +127,123 @@ export const submitForm = createServerFn({ method: 'POST' })
   })
 ```
 
+## Raw Handler for Streaming Uploads
+
+For advanced use cases like streaming file uploads or accessing the raw request body, use `.rawHandler()` instead of `.handler()`. This gives you direct access to the Request object without automatic body parsing.
+
+### Why Use Raw Handler?
+
+The raw handler is essential when you need to:
+
+- **Stream large file uploads** directly to cloud storage without loading them into memory
+- **Enforce size limits during upload** rather than after the entire file is loaded
+- **Access raw request streams** for custom body parsing (e.g., multipart/form-data with busboy)
+- **Implement proper backpressure** for large data transfers
+- **Minimize memory footprint** for file uploads
+
+### Basic Raw Handler
+
+```tsx
+import { createServerFn } from '@tanstack/react-start'
+
+export const uploadFile = createServerFn({ method: 'POST' }).rawHandler(
+  async ({ request, signal }) => {
+    // Access the raw request object
+    const contentType = request.headers.get('content-type')
+    const body = await request.text()
+
+    return new Response(
+      JSON.stringify({
+        contentType,
+        size: body.length,
+      }),
+    )
+  },
+)
+```
+
+### Streaming File Upload Example
+
+With raw handlers, you can stream files directly to cloud storage without buffering them in memory:
+
+```tsx
+import { createServerFn } from '@tanstack/react-start'
+
+export const uploadFile = createServerFn({ method: 'POST' })
+  .middleware([authMiddleware]) // Middleware context is available!
+  .rawHandler(async ({ request, signal, context }) => {
+    // Access middleware context (user, auth, etc.)
+    const userId = context.user.id
+
+    // Access the raw request body stream
+    const body = request.body
+
+    if (!body) {
+      return new Response('No file provided', { status: 400 })
+    }
+
+    // Stream directly to your storage (S3, Azure, etc.)
+    // You can use libraries like busboy to parse multipart/form-data
+    // and enforce size limits DURING upload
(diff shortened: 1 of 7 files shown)


schiller-manuel:
would rather prefer the flag in createServerFn, however we first need to find out how to handle those raw request handlers when it comes to (global) server function middlewares

LadyBluenotes:
Thanks for the detailed investigation and the cross-file regression cases. This fix changes how the compiler crawls modules and builds the server-function manifest, which overlaps with broader compiler work planned by the maintainers. We aren’t accepting compiler-related PRs right now, so we’re going to close this one, but we’ll keep #7213 open to track the underlying issue

Issue #5674, opened 2025-10-28
useMatchRoute does not match routes with prefix.


danielknell:
### Which project does this relate to?

Router

### Describe the bug

if a route uses a path with a prefix like `/page-{$page}` then `useMatchRoute` cannot seem to match against it. 

```ts
import { createFileRoute, useMatchRoute } from '@tanstack/react-router';
import * as React from 'react';

export const Route = createFileRoute('/page-{$page}')({
  component: AboutComponent,
});

function AboutComponent() {
  const matchRoute = useMatchRoute();

  const match = matchRoute({ to: '/page-{$page}' });

  return (
    <div className="p-2">
      <h3>{JSON.stringify(match)}</h3>
    </div>
  );
}

```

will always show `false`.

### Your Example Website or App

https://stackblitz.com/edit/github-s7nqcjqt?file=src%2Froutes%2Fpage-%7B%24page%7D.tsx

### Steps to Reproduce the Bug or Issue

vist `/page-a`
notice matchRoute result is `false`

### Expected behavior

`matchRoute({to: '/page-{$page}'})` to return `{"page": "a"}` when visiting `/page-a` instead of `false`.

### Platform

- Router Version: 1.133.32
- OS: macOS / Linux
- Browser: Safari / Chrome
- Browser Version: 18.6 / 141.0.7390.123
- Bundler: vite
- Bundler Version: 7.1.14

### Additional context

Code works fine with routes like `/page/$page` but seems to specifically fail with the custom routes.

nlynzaad:
@danielknell I've been working on this and have reached near finality. we are however in the process of a rather large refactor that's nearly completed on the same sections so holding off a bit till that is done.

danielknell:
@nlynzaad No worries, do you have a non-committal rough feel on how long the blocking refactoring might take? (days/weeks/months/years) so I can decide how hacky a work around can be?

Issue #5635, opened 2025-10-26
Randomly not matching index.tsx routes where we have index.tsx + route.tsx per route.


shmuel-levinson-z:
### Which project does this relate to?

Router

### Describe the bug

Important: PROBLEM REPRODUCES ON CHROME BUT NOT ON FIREFOX. Verified by Tanstack maintainer.

I'm encountering a mind-bending issue with file-based routing: in our setup, each folder has route.tsx and index.tsx as siblings, both calling createFileRoute, and we expect two matches (one for each). 
For certain routes, the index.tsx match (the one rendering the component) completely fails to register, leaving only the route.tsx match. 
To make it worse, the issue seems interdependent—renaming or removing unrelated folders sometimes resolves or breaks it for others. Deleting route.tsx in any problematic case SOLVES the problem for that folder.

The folder names appears to be the ONLY determining factor from testing we'd done -- certain 'lucky' names work., but changing any of these had no effect at all: properties in createFileRoute, rendering Outlet explicitly in the route.tsx file, rendering components inline or imported, switching autoCodeSplitting to false (though we need it to be true...).

NOTE: if no change to code is introduced then the exact same pages will fail consistently across page visits.

Optional params or lack thereof make no difference. This issue existed on a page without optional param. Solved it there by removing route.tsx.

We also tried looking at routeTree.gen to see maybe the problematic pages had some duplicate or missing entries but couldn't spot any anomalies.

On the face of it problematic pages are implemented exactly like any other working pages in any regard.

The reproduction repo is a completely stripped down version with nothing but the routes structure and an example of a working page and a non-working page.

### Your Example Website or App

https://github.com/Shmuel-Levinson/random-non-render

### Steps to Reproduce the Bug or Issue

1) Open up the matches tab
2) Go to the working link
3) Observe there are 2 matches (one for the route.tsx file and the other for index.tsx)
4) Go to the not-working link and see the path for index.tsx is NOT matched and hence nothing is displayed.

### Expected behavior

I expect the index.tsx to be matched

### Screenshots or Videos

https://github.com/user-attachments/assets/f3cddf0e-223f-467c-97da-49ed72321221

### Platform

- Router Version: ^1.133.36 or latest
- OS: macOS
- Browser: Chrome
- Browser Version: 142.0.7444.60 (latest as of oct-29-2025)
- Bundler: vite
- Bundler Version: 7.1.11 (latest)

PR #5722, opened 2025-11-01
refactor(router-core): Process routeTree into segment tree instead of flatRoutes

Sheraff:
This PR rewrites `processRouteTree` to yield a trie-like tree of segments. This replaces
- `packages/router-core/src/process-route-tree.ts`
- parts of `packages/router-core/src/path.ts`

Matching complexity now grows linearly with the number of segments (i.e. `/a/b/c` => 3 segments), as opposed to the previous algorithm that was correlated with the number of routes.

Goals:
- faster pre-processing at start-time
- faster matching at run-time
- more consistent priority matching

Non-goals:
- low RAM usage

Fixes https://github.com/TanStack/router/issues/5635
Fixes https://github.com/TanStack/router/issues/5674
Fixes https://github.com/TanStack/router/issues/5799

## to-do:

- [x] fix "off-by-1" (leading/trailing `/`)
- [x] implement sorting of same-type siblings at the node level (case sensitivity, prefix/suffix length)
- [x] implement single-match ("does A match B")
- [x] implement all-route-match ("which route does A match")
- [x] bench pre-processing against current implementation
- [x] bench single-match against current implementation
- [x] bench all-route-match against current implementation
- [x] fix not found page
- [x] fix URL param encoding (see failing tests)
- [ ] write more tests and clean up existing tests
- [ ] document priorities and "matching logic quirks"

## Bench:

- pre-processing 
  - consistently >2x faster
- needle-in-a-haystack route matching
  - for very small route trees: 3.5x faster
  - for very large route trees: 950x faster (430 routes)
  - non-matching: 90x faster (430 routes, random path "that almost matches" requested, this is a worst-case-scenario simulation)
- 1-to-1 route matching (`useMatchRoute`)
  - for very simple routes: 1.5x **slower**
  - for very complex routes: 1x
- resolvePath
  - 2.7x faster (added a LRU cache to this, it's called a lot and is easy to cache, but function itself could be improved)
- interpolatePath
  - 1.4x faster

## Notes:

- in its current implementation, 1-to-1 route matching (`useMatchRoute`) does not care about the `routeTree`. It will match 2 arbitrary strings. This can lead to unexpected results, and is probably not the behavior we want in the long term. 
  For now, this PR implements `findSingleMatch` to preserve that behavior, and this method is slower than the current implementation. However, when we switch to matching against the routeTree (using `findRouteMatch` instead), it will be faster.
- We use an int32 bitmask to keep track of skipped optional segments, which means that segments beyond the 32nd segment cannot be optional
- We use a UInt16Array to keep track of some indices, which also has a theoretical limit (65536 chars) but it is greater than the maximum length of URLs supported by most browsers (AFAIK, Chrome has the lowest limit, Firefox is exactly 65536, and Safari supports longer)
- the segment tree can be built in several passes (like we do for route masks, provided as a list instead of a tree), so it is compatible with a "fog of war" approach. However, the tree must be walked to re-sort siblings after a batch of modifications.

## Future improvements:

- we could consider flattening the tree branches that contain a single child at each node
- removing route-by-route case sensitivity configuration (keep global option) could yield some performance improvements
- we could take inspiration from `find-my-way` (router for `fastify`) and use nodes that are smaller than a full pathname segment. This yields a structure close to a radix tree, which could perform better.

## Summary by CodeRabbit

* **Deprecations**
  * caseSensitive on match-route options is deprecated; use route- or router-level case sensitivity.

* **Breaking Changes**
  * Legacy path parsing/matching public APIs removed; integrations must adopt the new route-tree pipeline.
  * DevTools explorer no longer exposes flatRoutes.

* **New Features**
  * New processed route-tree matcher with revamped path resolution and interpolation for more accurate matching and params.

* **Tests**
  * Added and updated tests covering the new route-tree matcher and parsing semantics.

### DIFF
--- a/packages/router-core/src/new-process-route-tree.ts
+++ b/packages/router-core/src/new-process-route-tree.ts
@@ -368,6 +368,9 @@ function sortDynamic(
   if (!a.suffix && b.suffix) return 1
   if (a.caseSensitive && !b.caseSensitive) return -1
   if (!a.caseSensitive && b.caseSensitive) return 1
+
+  // we don't need a tiebreaker here
+  // at this point the 2 nodes cannot conflict during matching
   return 0
 }
 
@@ -506,20 +509,6 @@ type SegmentNode<T extends RouteLike> = {
   /** Same as `route`, but only present if both an "index route" and a "layout route" exist at this path */
   notFound: T | null
 }
-
-// function intoRouteLike(routeTree, parent) {
-// 	const route = {
-// 		id: routeTree.id,
-// 		fullPath: routeTree.fullPath,
-// 		path: routeTree.path,
-// 		isRoot: routeTree.isRoot,
-// 		options: routeTree.options && 'caseSensitive' in routeTree.options ? { caseSensitive: routeTree.options.caseSensitive } : undefined,
-// 	}
-// 	if (routeTree.children) {
-// 		route.children = routeTree.children.map(child => intoRouteLike(child, route))
-// 	}
-// 	return route
-// }
 
 type RouteLike = {
   path?: string // relative path from the parent,
@@ -774,7 +763,7 @@ function extractParams<T extends RouteLike>(
         node.suffix || node.prefix
           ? part!.substring(preLength, part!.length - sufLength)
           : part
-      if (value?.length) params[name] = decodeURIComponent(value)
+      if (value) params[name] = decodeURIComponent(value)
     } else if (node.kind === SEGMENT_TYPE_WILDCARD) {
       const n = node
       const value = path.substring(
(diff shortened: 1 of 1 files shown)


packages/router-core/src/new-process-route-tree.ts
```diff
@@ -0,0 +1,614 @@
...
+		// suffix counts single
+		+ (a.suffix ? a.suffix.length : 0)
+	const bScore =
+		// bonus for having both prefix and suffix
+		(b.prefix && b.suffix ? 500 : 0)
+		// prefix counts double
+		+ (b.prefix ? b.prefix.length * 2 : 0)
+		// suffix counts single
+		+ (b.suffix ? b.suffix.length : 0)
+	return bScore - aScore
```
Sheraff: fallback to alphabetical if same score

packages/router-core/src/new-process-route-tree.ts
```diff
@@ -0,0 +1,614 @@
...
+					}
+					if (suffix) {
+						const part = parts[parts.length - 1]!
+						const casePart = segment.caseSensitive ? part : part.toLowerCase()
+						if (!casePart.endsWith(suffix)) continue
+					}
+					// a wildcard match terminates the loop, but we need to continue searching in case there's a longer match
+					if (!wildcardMatch || (wildcardMatch.index < index)) {
+						wildcardMatch = { node: segment, index, skipped, depth }
+					}
```
Sheraff: score should be on `DynamicNode`, used in sort above, and in here if both indices are equal.

packages/router-core/src/new-process-route-tree.ts
```diff
@@ -0,0 +1,638 @@
...
+	if (a.suffix && b.suffix && a.suffix !== b.suffix) {
+		if (a.suffix.endsWith(b.suffix)) return -1
+		if (b.suffix.endsWith(a.suffix)) return 1
+	}
+	if (a.prefix && !b.prefix) return -1
+	if (!a.prefix && b.prefix) return 1
+	if (a.suffix && !b.suffix) return -1
+	if (!a.suffix && b.suffix) return 1
+	return 0
+}
```
Sheraff: should rank case-sensitive routes w/ prefix/suffix higher than case-insensitive ones

Sheraff:
@coderabbitai review

Sheraff:
@coderabbitai review

PR #5723, opened 2025-11-01
(docs:fix) fixed cloudflare workers anchor link

kulterryan:
Previously the anchor link was set to "#cloudflare-workers" which was not working due to change in title, so I replaced it with "#cloudflare-workers--official-partner".

## Summary by CodeRabbit

* **Documentation**
  * Updated the Cloudflare Workers hosting documentation link to reference the official partner resource.
  * Clarified the hosting deployment list so users are directed to the verified partner page, improving link accuracy and trustworthiness.


birkskyum:
Thanks!

Issue #5008, opened 2025-08-21
Special character in URL (ex: ê) with a <Link /> component for the same URL causes a SSR error


olivier-deschenes:
### Which project does this relate to?

Start

### Describe the bug

When you reload a route with a param and the param has a special character like ê, the link component in the route component seems to bug when SSR is activated.

I assume that it is because the special character is encoded/decoded at a different time in the server / client (one is encoded and the other not).

### Your Example Website or App

https://github.com/olivier-deschenes/tanstack-start-special-char-ssr-bug

### Steps to Reproduce the Bug or Issue

1. install the deps and run the app
2. Open the app (it should redirect you to the right url)
3. reload the page
4. Check the console

### Expected behavior

Nothing should happens, no error in the console

### Screenshots or Videos

<img width="3488" height="1152" alt="Image" src="https://github.com/user-attachments/assets/a6930b99-0e6c-4edb-b919-52c1d171b7f6" />

### Platform

- Start version : ^1.131.26

### Additional context

It seems to only happen when their is a special character in the url.

schiller-manuel:
the issue here is that when the link is rendered on the client, the comparison here fails:

https://github.com/TanStack/router/blob/3f05c0b16bb2d4084c90c2eebe9077a8a89b1c7e/packages/react-router/src/link.tsx#L147-L153

`s.location.pathname` has the path in the form of `'/test/ê'` while `next.pathname` is `'/test/%C3%AA'`

olivier-deschenes:
Right now the fix that I use is to simply decode/encode and normalize the params.

```ts
export const Route = createFileRoute("/test/$id")({
  component: RouteComponent,
  params: {
    parse: (raw) => ({ id: decodeURIComponent(raw.id).normalize("NFC") }),
    stringify: (v) => ({ id: encodeURIComponent(v.id.normalize("NFC")) }),
  },
});
```

But I assume that it is a bug right ?
It happens with spaces also.

schiller-manuel:
yes it is a bug

olivier-deschenes:
@schiller-manuel is there any plan to fix this in a future release ?

schiller-manuel:
absolutely! contributions welcome as well

nlynzaad:
think this might be resolved with #5663 available from 1.133.35. please can you check and confirm.

olivier-deschenes:
Hi, 

[...]

PR #5736, opened 2025-11-03
refactor(router-core): consistent Unicode handling

nlynzaad:
This PR follows on from #5663 and ensures that the path names are always in a consistent state by decoding the pathname when parsing the location. This ensures that the pathname is always in a decoded state and any further encoding/decoding that is required for navigation and matching is done in a consistent manner. While the pathname is decoded the URL is not ensuring the href and urls are always in a valid state.

One case where this currently fails currently is when matching a link as active or not when the params contain special characters, see #5008. While the resolved pathname are in encoded form, the params are on the link is not, and this can cause mismatching to occur.  

As an added bonus, by decoding the path upfront we do not need to decode each path segment individually when parsing the paths for matching.

This PR adds a bunch of tests to ensure this is treated consistently throughout router at the core, code-based routing in React and Solid as well as in the browser environment with file-based routing in React and Solid e2e tests.

## Release Notes

* **New Features**
  * Added support for Unicode characters (including emoji and non-Latin scripts) in route paths and parameters
  * Enhanced handling of complex URL encoding and special characters in navigation
  * New route examples demonstrating international path support

* **Performance**
  * Reduced page load delay for improved responsiveness

* **Testing**
  * Expanded test coverage for Unicode paths and encoded parameter handling

### DIFF
--- a/packages/react-router/tests/router.test.tsx
+++ b/packages/react-router/tests/router.test.tsx
@@ -48,6 +48,22 @@ afterEach(() => {
 
 const mockFn1 = vi.fn()
 
+const URISyntaxCharacters = [
+  [';', '%3B'],
+  [',', '%2C'],
+  ['/', '%2F'],
+  ['?', '%3F'],
+  [':', '%3A'],
+  ['@', '%40'],
+  ['&', '%26'],
+  ['=', '%3D'],
+  ['+', '%2B'],
+  ['$', '%24'],
+  ['#', '%23'],
+  ['\\', '%5C'],
+  ['%', '%25'],
+] as const
+
 export function validateSearchParams<
   TExpected extends Partial<Record<string, string>>,
 >(expected: TExpected, router: AnyRouter) {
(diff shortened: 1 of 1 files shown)


packages/solid-router/tests/router.test.tsx
```diff
@@ -723,63 +736,75 @@ describe('encoding: URL path segment', () => {
...
       history: createMemoryHistory({ initialEntries: [input] }),
     })
 
     render(() => <RouterProvider router={router} />)
     await router.load()
 
-    expect(router.state.location.pathname).toBe(output)
+    expect(router.state.location.pathname).toBe(path)
+    expect(new URL(router.state.location.url).pathname).toBe(url)
   })
```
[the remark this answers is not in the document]
nlynzaad: this is not the case. `new URL('http://x.com/path-segment/%C3%A9').pathname` returns the encoded path `/path-segment/%C3%A9` hence why the tests are not failing.

packages/react-router/tests/router.test.tsx
```diff
@@ -374,7 +374,7 @@ describe('encoding: URL param segment for /posts/$slug', () => {
 
     await act(() => router.load())
 
-    expect(router.state.location.pathname).toBe('/posts/%F0%9F%9A%80')
+    expect(router.state.location.pathname).toBe('/posts/🚀')
```
SeanCassiere, line 460: Not sure what needs to be changed here.

- Expectation of the test.
- Or the test name itself (i.e `it('should ...')`).

packages/react-router/tests/router.test.tsx
```diff
@@ -623,7 +628,8 @@ describe('encoding: URL splat segment for /$', () => {
 
     await router.load()
 
-    expect(router.state.location.pathname).toBe('/%F0%9F%9A%80')
+    expect(router.state.location.href).toBe('/%F0%9F%9A%80')
+    expect(router.state.location.pathname).toBe('/🚀')
```
SeanCassiere, line 669: Just to confirm, this difference is "expected" and should have differing values.

packages/router-core/src/router.ts
```diff
@@ -1173,7 +1173,7 @@ export class RouterCore<
         href: fullPath,
         publicHref: href,
         url: url.href,
-        pathname,
+        pathname: decodePathSegment(pathname),
```
SeanCassiere: Maybe a function rename here, since its no longer `decodePathSegment` (which was using `decodeURIComponent`) rather its more along the lines of `decodePath` (using `decodeURI`).

packages/react-router/tests/router.test.tsx
nlynzaad: yes, the url and href is to remain encoded. the pathname not

packages/react-router/tests/router.test.tsx
nlynzaad: will rename the test. we are expecting the pathname to be decoded

nlynzaad:
requested changes made. 

quick note on the actionable comment from code rabbit. in the specific test `@` is not added to the pathParamsAllowedCharacters and hence is encoded. This is specifically left as such and hence the use of decodeURI in decodePath.

nlynzaad:
> I still think that the test names need to be changed to match what we're now testing for.
> 
> Other than that, I'm OK with the changes themselves.

reworked the test names and cleaned up the structure a bit

PR #5756, opened 2025-11-04
fix(router-core): correctly resolve custom params from declarative masks when building the masked url

### LINKED ISSUE #5755: Declarative route masks are ignoring params

Router

### Describe the bug

When using custom params with declarative route masks, the mask params are ignored and the router falls back to the original URL params. This is not noticeable unless the mask params differ from the original ones.

In my case, users can have multiple accounts with access to multiple workspaces. Internally, I include the account ID in the URL but mask it so only the workspace ID is shown. Routing still works correctly, but since the masked params are ignored, the masked URL displays `undefined` for the dynamic segment that doesn’t exist in the original params.

https://stackblitz.com/edit/github-qkndtvzh?file=src%2Fmain.tsx

An illustration of the issue

```tsx
// this is the route which is being used internally
const workspacePrivateRoute = createRoute({
  path: '/workspace/$privateId',
  component: function About() {
    const { privateId } = workspacePrivateRoute.useParams();
    const workspace = workspaces.find((w) => w.privateId === privateId);
    if (!workspace) {
      return <div>Workspace not found</div>;
    }

    return (
      <div className="p-2">
        Workspace {workspace.name} | Private ID: {workspace.privateId} | Public
        ID: {workspace.publicId}
      </div>
    );
  },
});

//this is the masked route which is shown to the user
const workspacePublicRoute = createRoute({
  path: '/$publicId',
  beforeLoad: (ctx) => {
    const workspace = workspaces.find(
      (w) => w.publicId === ctx.params.publicId
    );

    if (!workspace) {
      throw notFound();
    }

    throw redirect({
      to: '/workspace/$privateId',
      params: {
        privateId: workspace.privateId,
      },
    });
  },
});

const routeTree = rootRoute.addChildren([
  workspacePrivateRoute,
  workspacePublicRoute,
]);


//here is the declarative mask definition
export const workspaceRouteMask = createRouteMask({
  routeTree: routeTree,
  from: '/workspace/$privateId',
  to: '/$publicId',
  params: (ctx) => {
    const workspace = workspaces.find((w) => w.privateId === ctx.privateId);
    if (!workspace) {
      throw notFound();
    }

    return {
      publicId: workspace.publicId,
    };
  },
});
```

The routing works as expected, but the URL shows 'undefined' instead of the public id of the workspace.

### Expected behavior

The final route should take into account the custom params in the declarative route definition.

### Screenshots or Videos

_No response_

### Platform

- Router Version: 1.134.12
- OS: Mac
- Browser: Chrome
- Browser Version: 142
- Bundler: vite
- Bundler Version: 7.1.10

### Additional context

I will propose a PR that tries to solve this issue.

hakanshehu:
This PR fixes #5755 , where custom params passed to declarative route masks are ignored. When the masked params differ from the original URL params, the masked URL can display `undefined` instead of the expected values.

I don't have full context on all internal router behaviors, so please treat this as a proposal. If there's a better approach, I’m happy to adjust the implementation.

## Summary by CodeRabbit

* **New Features**
  * Route masking supports per-mask parameter customization (merge/override/transform) and preserves search, hash, state, and original navigation info.
  * UI examples and links added to demonstrate original vs. masked pathnames.

* **Tests**
  * Added extensive unit and end-to-end tests validating mask discovery, param transformations (boolean/object/function), multiple masks, href/pathname integrity, and masking edge cases.

### DIFF
--- a/e2e/react-router/basic-file-based/src/main.tsx
+++ b/e2e/react-router/basic-file-based/src/main.tsx
@@ -1,8 +1,21 @@
 import React from 'react'
 import ReactDOM from 'react-dom/client'
-import { RouterProvider, createRouter } from '@tanstack/react-router'
+import {
+  RouterProvider,
+  createRouteMask,
+  createRouter,
+} from '@tanstack/react-router'
 import { routeTree } from './routeTree.gen'
 import './styles.css'
+
+const mask = createRouteMask({
+  routeTree,
+  from: '/masks/admin/$userId',
+  to: '/masks/public/$username',
+  params: (prev) => ({
+    username: `user-${prev.userId}`,
+  }),
+})
 
 // Set up a Router instance
 const router = createRouter({
@@ -10,16 +23,7 @@ const router = createRouter({
   defaultPreload: 'intent',
   defaultStaleTime: 5000,
   scrollRestoration: true,
-  routeMasks: [
-    {
-      routeTree: null as any,
-      from: '/masks/admin/$userId',
-      to: '/masks/public/$username',
-      params: (prev: any) => ({
-        username: `user-${prev.userId}`,
-      }),
-    },
-  ],
+  routeMasks: [mask],
 })
 
 // Register things for typesafety
--- a/e2e/react-router/basic-file-based/src/routes/__root.tsx
+++ b/e2e/react-router/basic-file-based/src/routes/__root.tsx
@@ -139,6 +139,7 @@ function RootComponent() {
           unicode path
         </Link>{' '}
         <Link
+          // @ts-expect-error
           to="/this-route-does-not-exist"
           activeProps={{
             className: 'font-bold',
--- a/e2e/solid-router/basic-file-based/src/main.tsx
+++ b/e2e/solid-router/basic-file-based/src/main.tsx
@@ -1,7 +1,20 @@
-import { RouterProvider, createRouter } from '@tanstack/solid-router'
+import {
+  RouterProvider,
+  createRouteMask,
+  createRouter,
+} from '@tanstack/solid-router'
 import { render } from 'solid-js/web'
 import { routeTree } from './routeTree.gen'
 import './styles.css'
+
+const mask = createRouteMask({
+  routeTree,
+  from: '/masks/admin/$userId',
+  to: '/masks/public/$username',
+  params: (prev) => ({
+    username: `user-${prev.userId}`,
+  }),
+})
 
 // Set up a Router instance
 const router = createRouter({
@@ -9,16 +22,7 @@ const router = createRouter({
   defaultPreload: 'intent',
   defaultStaleTime: 5000,
   scrollRestoration: true,
-  routeMasks: [
-    {
-      routeTree: null as any,
-      from: '/masks/admin/$userId',
-      to: '/masks/public/$username',
-      params: (prev: any) => ({
-        username: `user-${prev.userId}`,
-      }),
-    },
-  ],
+  routeMasks: [mask],
 })
 
 // Register things for typesafety


schiller-manuel:
we definitely need tests here. ideally start with the tests first and then the solution

hakanshehu:
@schiller-manuel thanks for the reply. Since I couldn't find any tests for masks specifically in the router-core I added a new file and tried to cover different cases with masks and params. Let me know what you think!

hakanshehu:
Hi @schiller-manuel, just checking in on this PR. Any sense of whether it might be merged or if I should work around it? Thanks!

nlynzaad:
Hi @hakanshehu thanks for this, I had a quick look through and it seems good. 

I want to have thorough look at it tomorrow to make sure I understand the problem and proposed solution fully and will advise.

packages/router-core/tests/mask.test.ts
```diff
@@ -0,0 +1,729 @@
...
+    expect(location.maskedLocation).toBeDefined()
+    expect(location.maskedLocation!.pathname).toBe('/photos/123')
+    expect(location.maskedLocation!.search).toEqual({ filter: 'recent' })
+    // Hash property stores the value without #, but href includes it
+    expect(location.maskedLocation!.hash).toBe('section1')
+    expect(location.maskedLocation!.href).toContain('#section1')
+    expect(location.maskedLocation!.state).toEqual({ modal: true })
+  })
+
+  test('should preserve opts.from when building masked destination', () => {
```
nlynzaad: what are we testing here

packages/router-core/tests/mask.test.ts
```diff
@@ -0,0 +1,729 @@
...
+      from: '/photos/456',
+      params: { photoId: '123' },
+    })
+
+    expect(location.maskedLocation).toBeDefined()
+    // The from option should be preserved in the masked destination
+    expect(location.maskedLocation!.pathname).toBe('/photos/123')
+  })
+
+  test('should handle unmaskOnReload property in mask', () => {
```
nlynzaad: not sure why we need this test. it doesn't test any actual functionality. maybe I'm missing something? 

Might be better suited in an e2e test?

packages/router-core/tests/mask.test.ts
```diff
@@ -0,0 +1,729 @@
...
+    const location = router.buildLocation({
+      to: '/photos/abc123/details',
+      params: { privateId: 'abc123' },
+    })
+
+    expect(location.maskedLocation).toBeDefined()
+    expect(location.maskedLocation!.pathname).toBe('/photos/abc123')
+    expect(location.pathname).toBe('/photos/abc123/details')
+    // Ensure no undefined values
+    expect(location.maskedLocation!.pathname).not.toContain('undefined')
```
nlynzaad: this assertion is already covered above. if it is '/photos/abc123' then we already know it doesnt contain undefined

packages/router-core/tests/mask.test.ts
```diff
@@ -0,0 +1,729 @@
...
+    })
+
+    const location = router.buildLocation({
+      to: '/photos/secret123/details',
+      params: { privateId: 'secret123' },
+    })
+
+    expect(location.maskedLocation).toBeDefined()
+    expect(location.maskedLocation!.pathname).toBe('/photos/secret123')
+    expect(location.maskedLocation!.pathname).not.toContain('undefined')
```
nlynzaad: this assertion is already covered above. if it is '/photos/secret123' then we already know it doesnt contain undefined

packages/router-core/tests/mask.test.ts
```diff
@@ -0,0 +1,729 @@
...
+      to: '/users/john/posts/my-first-post',
+      params: { userId: 'john', postSlug: 'my-first-post' },
+    })
+
+    expect(location.maskedLocation).toBeDefined()
+    expect(location.maskedLocation!.pathname).toBe(
+      '/profiles/john/articles/my-first-post',
+    )
+    expect(location.pathname).toBe('/users/john/posts/my-first-post')
+    expect(location.maskedLocation!.pathname).not.toContain('undefined')
```
nlynzaad: this assertion is already covered above. if it is '/profiles/john/articles/my-first-post' then we already know it doesnt contain undefined

packages/router-core/tests/mask.test.ts
```diff
@@ -0,0 +1,729 @@
...
+
+    const location = router.buildLocation({
+      to: '/admin/users/42',
+      params: { userId: '42' },
+    })
+
+    expect(location.maskedLocation).toBeDefined()
+    expect(location.maskedLocation!.pathname).toBe('/users/user-42')
+    expect(location.pathname).toBe('/admin/users/42')
+    expect(location.maskedLocation!.pathname).not.toContain('undefined')
```
nlynzaad: this assertion is already covered above. if it is '/users/user-42' then we already know it doesnt contain undefined

packages/router-core/tests/mask.test.ts
```diff
@@ -0,0 +1,729 @@
...
+
+    const location = router.buildLocation({
+      to: '/internal/internal-123/details',
+      params: { internalId: 'internal-123' },
+    })
+
+    expect(location.maskedLocation).toBeDefined()
+    expect(location.maskedLocation!.pathname).toBe('/public/public-123')
+    // Ensure the pathname doesn't contain 'undefined'
+    expect(location.maskedLocation!.pathname).not.toContain('undefined')
```
nlynzaad: this assertion is already covered above. if it is '/public/public-123' then we already know it doesnt contain undefined

packages/router-core/tests/mask.test.ts
```diff
@@ -0,0 +1,729 @@
...
+      params: { userId: '42' },
+    })
+
+    expect(location.maskedLocation).toBeDefined()
+    expect(location.maskedLocation!.pathname).toBe('/users/user-42')
+    expect(location.pathname).toBe('/admin/users/42')
+    expect(location.maskedLocation!.pathname).not.toContain('undefined')
+  })
+
+  test('should not have undefined params in masked location when param names differ', () => {
```
nlynzaad: This test is unnecessary. its outcome is already covered in all the above tests. does this test anything new than what was covered above?

packages/router-core/tests/mask.test.ts
```diff
@@ -0,0 +1,729 @@
...
+      to: '/posts/5/comments/10',
+      params: { postId: '5', commentId: '10' },
+    })
+
+    expect(location.maskedLocation).toBeDefined()
+    expect(location.maskedLocation!.pathname).toBe(
+      '/articles/article-5/replies/10',
+    )
+    // Verify no undefined values
+    expect(location.maskedLocation!.pathname).not.toContain('undefined')
```
nlynzaad: this assertion is already covered above. if it is '/articles/article-5/replies/10' then we already know it doesnt contain undefined

hakanshehu:
Hi @nlynzaad thanks for the feedback.

I addressed the comments in the core mask tests. I've also added some routes and a simple e2e test for react and solid. Let me know if I should use another approach or improve anything for the new tests.

PR #5767, opened 2025-11-05
fix(router-plugin): prevent double initialization of shared module-level variables

KyleAMathews:
## Summary

Fixes an issue where module-level variables (like TanStack collections) used in both loader and component were being initialized twice due to code splitting.

## Problem

When a route file defines module-level variables that are referenced by both:
- Non-split parts (e.g., loader)  
- Code-split parts (e.g., component)

The variable would be initialized twice:
1. Once in the reference file (for loader)
2. Once in the split component file

This broke singleton patterns and caused unwanted side effects, as reported in Discord where collections and queries were being created twice.

## Solution

**Reference File (`compileCodeSplitReferenceRoute`):**
- Detects module-level const/let variables used by BOTH split and non-split properties
- Automatically exports these shared variables
- Example: `const collection = {...}` → `export const collection = {...}`

**Split File (`compileCodeSplitVirtualRoute`):**
- Detects module-level variables (excluding functions) used by split nodes
- Removes their local declarations
- Imports them from the reference file instead
- Example: `const collection = {...}` → `import { collection } from "route.tsx"`

## What It Handles ✅

**Top-level const/let declarations with object/data values:**
```tsx
// Collections and queries (primary use case)
const collection = { name: 'todos', preload: async () => {} }
const query = new Query().from('users')

// Configuration objects
const config = { apiUrl: '...', timeout: 5000 }

// Data structures
const data = [1, 2, 3]
const map = new Map()
```

## What It Doesn't Handle ❌

**Function declarations/expressions** (intentionally excluded):
```tsx
// Arrow functions
const myHelper = () => 'foo'

// Function expressions
const myHelper = function() { return 'foo' }

// Function declarations
function myHelper() { return 'foo' }
```

**Other declaration types:**
```tsx
// Classes
class MyClass { }

// Non-module-level variables
if (something) {
  const local = 'foo'  // Inside a block
}
```

**Why exclude functions?** Functions are typically pure/stateless and duplication is less problematic. The fix targets the specific use case of stateful objects (collections, queries) that shouldn't be instantiated multiple times.

## Test Plan

- [x] Added test case `shared-module-variable.tsx` demonstrating the fix
- [x] All 250 router-plugin tests passing
- [x] Verified reference file exports shared variables
- [x] Verified split file imports instead of duplicating
- [x] Existing code splitting behavior preserved

## Example

**Before:**
```tsx
// Reference file
const collection = { name: 'todos', preload: async () => {} }

// Split component file  
const collection = { name: 'todos', preload: async () => {} } // Duplicate!
```

**After:**
```tsx
// Reference file
export const collection = { name: 'todos', preload: async () => {} }

// Split component file
import { collection } from "route.tsx"
```

### DIFF
--- a/e2e/react-router/basic-file-based-code-splitting/src/routeTree.gen.ts
+++ b/e2e/react-router/basic-file-based-code-splitting/src/routeTree.gen.ts
@@ -8,6 +8,8 @@
 // You should NOT make any changes in this file as it will be overwritten.
 // Additionally, you should also exclude this file from your linter and/or formatter to prevent it from being checked or modified.
 
+import type { CreateFileRoute, FileRoutesByPath } from '@tanstack/react-router'
+
 import { Route as rootRouteImport } from './routes/__root'
 import { Route as WithoutLoaderRouteImport } from './routes/without-loader'
 import { Route as ViewportTestRouteImport } from './routes/viewport-test'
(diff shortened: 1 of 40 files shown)


packages/router-plugin/src/core/router-code-splitter-plugin.ts
```diff
@@ -66,6 +66,10 @@ export const unpluginRouterCodeSplitterFactory: UnpluginFactory<
   let ROOT: string = process.cwd()
   let userConfig: Config
 
+  // Cache of shared module-level variables that were exported in reference files
+  // Key: base filename (without query params), Value: Set of exported identifier names
+  const sharedExportsCache = new Map<string, Set<string>>()
```
schiller-manuel, line 71: how does this work with HMR? 

also how does this work across multiple vite environments? i assume this will work since we have the plugin per environment?

KyleAMathews: Relaying back Claude's research:

---

  Re: sharedExportsCache and HMR/environments

  Good questions - after digging into this, I need to verify both concerns:

  1. Multiple Vite Environments

  Current behavior: The cache is shared across all Vite environments (SSR, client, etc.) because unplugin instances are shared by default in Vite 6.

  Is this correct? It should work if both SSR and client environments process route files identically. However, if there are environment-specific code paths that
   affect which variables get exported, this could be problematic.

  Potential fix if needed: Use the Map<Environment, Set<string>> pattern:
  const sharedExportsCache = new Map<string, Map<string, Set<string>>>()
  // Key: environment name, Value: Map of filename → exported identifiers

  Or use Vite's perEnvironmentPlugin helper to create separate plugin instances.

  2. HMR (Hot Module Replacement)

  Current assumption: Reference file transforms before virtual files, so cache is always fresh.

  Actual uncertainty: I cannot find documentation guaranteeing transform order during HMR. Since virtual files (route.tsx?tsr-split=component) dynamically import
   from the reference file, HMR invalidation propagates through the import graph - but the transform order isn't guaranteed.

  Potential issue: If a virtual file transforms before its reference file during HMR, it could read stale cached data.

  Potential fix: Explicitly invalidate virtual modules when reference file changes using handleHotUpdate:
  handleHotUpdate({ file, server }) {
    if (/* is reference file */) {
      // Invalidate all virtual chunks for this route
      // Then they'll re-transform with fresh cache
    }
  }

  Question: Have you observed any issues with this in practice, or is this theoretical? Happy to add explicit invalidation if needed.

schiller-manuel: so i think multiple environments (in start) dont matter since we instantiate the code splitter per environment

see https://github.com/TanStack/router/blob/854a7ef23dcd7b986af32e4efbb91861af58b9ca/packages/start-plugin-core/src/start-router-plugin/plugin.ts#L224-L238

schiller-manuel: about invalidation: it might be enough if we clear the cache of a specific id AFTER the virtual file was processed so nothing is carried over the next time this is processed

can you please test this with HMR by modifying an affected route file repeatedly and then check whether it still works?

packages/router-plugin/tests/code-splitter/snapshots/react/1-default/shared-alias-chain.tsx
```diff
@@ -0,0 +1,17 @@
+console.warn("[tanstack-router] These exports from \"shared-alias-chain.tsx\" will not be code-split and will increase your bundle size:\n- alias\nFor the best optimization, these items should either have their export statements removed, or be imported from another location that is not a route file.");
```
SeanCassiere, line 1: This export warning shouldn't be needed for exports that are essentially being marked as "safe".

schiller-manuel:
closing in favor of https://github.com/TanStack/router/pull/6659

PR #5790, opened 2025-11-07
fix: set progressiveChunkSize to infinity

### LINKED ISSUE #5383: Tamstack Start base div hidden on JS disabled

Start

### Describe the bug

At the latest version of Tanstack Start if you disable JS in browser the page is empty because the div that contains all content has param “hidden”. 

later

1. Load page with JS disabled
2. See the problem

### Expected behavior

Expected to see content

### Screenshots or Videos

_No response_

### Platform

- Bundler: [e.g. vite]

### Additional context

_No response_

andyabih:
fixes #5383

In reference to this: https://github.com/facebook/react/commit/18212ca960ee2f0acf538c2198f7ba36c3042ecd

Currently, when the shell hits ~12.5KB, any suspense boundary larger than 500 bytes triggers an empty fallback. 

This results in pages loading without the content at first, and then having them render shortly after.

This PR simply adds a way to increase that size to whatever the user deems fit if he wishes to avoid this behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Modified server-side rendering streaming behavior to ensure consistent chunk handling during content delivery.

### DIFF
--- a/packages/react-router/src/ssr/renderRouterToStream.tsx
+++ b/packages/react-router/src/ssr/renderRouterToStream.tsx
@@ -24,7 +24,7 @@ export const renderRouterToStream = async ({
     const stream = await ReactDOMServer.renderToReadableStream(children, {
       signal: request.signal,
       nonce: router.options.ssr?.nonce,
-      progressiveChunkSize: router.options.ssr?.progressiveChunkSize,
+      progressiveChunkSize: Number.POSITIVE_INFINITY,
     })
 
     if (isbot(request.headers.get('User-Agent'))) {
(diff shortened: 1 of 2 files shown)


schiller-manuel:
`progressiveChunkSize` is a react config, thus cannot be in router-core

instead of making this a router option at all, i would just set the default to infinity and then expose a new `reactOpts` (or ideally named better) that has `progressiveChunkSize`so a custom handler could configure this if needed.

cc @brhx

brhx:
agree - @andyabih please could you update your PR just to have the progressive chunk size set to Infinity for now?

andyabih:
@brhx - Done! 

Happy to implement a PR with the proper options when you guys decide on the best way to tackle this. Pretty invested in this specifically due to SEO concerns, so always happy to help.

PR #5791, opened 2025-11-07
Add direction-aware view transitions to view transitions example repo

maurer2:
Hello, I noticed that the example repo for view transitions doesn’t yet include a page demonstrating direction-aware view transitions, so I added this functionality by modifying the existing posts example.

The post page now compares the old and new `postId` values to determine the transition direction. It disables view transitions if the direction can't be determined or if the user navigates to the same page. There’s a bit of manual work involved in extracting the ID from the location. I wasn’t able to get the approach described in [main.tsx](https://github.com/TanStack/router/blob/main/examples/react/view-transitions/src/main.tsx#L22) working, since the `TSR_index` value only seems to only increase with each navigation, making it unsuitable for determining the transition direction. Perhaps there's an easier way to get the previous ID and next ID from the parsed locations.

Cheers

## Summary by CodeRabbit

* **New Features**
  * Enhanced view transitions with dynamic direction awareness—transitions now intelligently adapt based on navigation direction.
  * Added backwards transition variant with corresponding animations for enhanced visual feedback during backwards navigation.


birkskyum:
@maurer2 can you update the same in the solid example as well?

birkskyum:
Oh, just realized there wasn't a solid view-transitions example yet - it's being added here
- https://github.com/TanStack/router/pull/5815

birkskyum:
@maurer2 , can you also make a PR for the solid-router example?

maurer2:
Sure, I will give it a go. 

Cheers

PR #5805, opened 2025-11-10
fix(router): prioritize static suffix over dynamic segment in route m…

### LINKED ISSUE #5799: Route matching prioritizes dynamic segment over static suffix

Router

### Describe the bug

When using routes with patterns like `/{-$owner}/posts/new` and `/posts/$id`, the router matches `/posts/new` to the dynamic route `/posts/$id` instead of the more specific route `/{-$owner}/posts/new` which has a static suffix `/new`.

https://stackblitz.com/edit/github-yal98jzd-fv8zbmmx

1. Open the reproduction URL
2. Click on "Posts"
3. Click on "New Post"

### Expected behavior

The URL `/posts/new` should match the route pattern `/{-$owner}/posts/new` because:

* It has a static suffix (`/new`) which provides more specificity
* The optional path parameter `{-$owner}` can match an empty string

**Observed**: The URL `/posts/new` matches `/posts/$id` instead, treating `new` as the value for the `$id` parameter.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.134.15
- OS: macOS
- Browser: Chrome, Safari, Firefox
- Bundler: vite
- Bundler Version: 7.2.2

### Additional context

_No response_

LuggaPugga:
This fix corrects the route matching logic to ensure that static suffixes are matched
before dynamic segments (e.g., `/user/settings` is prioritized over `/user/:id`).
Closes #5799 

## Summary by CodeRabbit

* **Improvements**
  * Route matching now collects exact matches and prefers routes that end with static segments before falling back to fuzzy matching, improving selection predictability.
* **Tests**
  * Added parameterized tests validating static-suffix vs. dynamic-parameter route selection across multiple scenarios.

### DIFF
--- a/packages/router-core/tests/processRouteTree.test.ts
+++ b/packages/router-core/tests/processRouteTree.test.ts
@@ -443,13 +443,19 @@ describe('processRouteTree', () => {
       {
         routes: ['/{-$owner}/posts/new', '/posts/$id'],
         pathname: '/posts/new',
-        expected: { route: '/{-$owner}/posts/new', params: { owner: undefined } },
+        expected: {
+          route: '/{-$owner}/posts/new',
+          params: { owner: undefined },
+        },
         description: 'prefers static suffix over dynamic param',
       },
       {
         routes: ['/{-$owner}/posts/edit', '/posts/$id'],
         pathname: '/posts/edit',
-        expected: { route: '/{-$owner}/posts/edit', params: { owner: undefined } },
+        expected: {
+          route: '/{-$owner}/posts/edit',
+          params: { owner: undefined },
+        },
         description: 'prefers static suffix with different word',
       },
       {
@@ -461,7 +467,10 @@ describe('processRouteTree', () => {
       {
         routes: ['/users/{-$org}/settings', '/users/$id'],
         pathname: '/users/settings',
-        expected: { route: '/users/{-$org}/settings', params: { org: undefined } },
+        expected: {
+          route: '/users/{-$org}/settings',
+          params: { org: undefined },
+        },
         description: 'works with optional params before static',
       },
       {
@@ -473,7 +482,10 @@ describe('processRouteTree', () => {
       {
         routes: ['/api/{-$version}/docs', '/api/$endpoint'],
         pathname: '/api/docs',
-        expected: { route: '/api/{-$version}/docs', params: { version: undefined } },
+        expected: {
+          route: '/api/{-$version}/docs',
+          params: { version: undefined },
+        },
         description: 'prefers static suffix in nested routes',
       },
     ])(


birkskyum:
@LuggaPugga can you add a test for this?

LuggaPugga:
> @LuggaPugga can you add a test for this?

Just did and applied the coderabbit suggestion as well :)

nlynzaad:
@Sheraff I'm assuming the work you are doing with the matching and others will be intersecting with this. maybe just hold this till you have completed your refactor/rewrite?

Sheraff:
> @Sheraff I'm assuming the work you are doing with the matching and others will be intersecting with this. maybe just hold this till you have completed your refactor/rewrite?

Yes, what I'm doing fixes this issue. I don't know when I'll be finished though, it's a pretty major rewrite

birkskyum:
@Sheraff , and it's this PR that's conflicting, right?
- https://github.com/TanStack/router/pull/5722

Sheraff:
> @Sheraff , and it's this PR that's conflicting, right?
> 
> * [refactor(router-core): Process routeTree into segment tree instead of flatRoutes [WIP] #5722](https://github.com/TanStack/router/pull/5722)

Yes, that's the one.

packages/router-core/src/router.ts
```diff
@@ -2684,26 +2686,65 @@ export function getMatchedRoutes<TRouteLike extends RouteLike>({
...
+          route.path !== '/' && (matchedParams as Record<string, string>)['**']
+
+        if (isFuzzy) {
+          fuzzyMatch ??= { foundRoute: route, routeParams: matchedParams }
         } else {
-          foundRoute = route
-          routeParams = matchedParams
-          break
+          const routeSegments = parsePathname(route.fullPath, parseCache)
+          const pathSegments = parsePathname(trimmedPath, parseCache)
```
nlynzaad, line 2713: we only need to parse this once right? since this is the trimmedPath we can move this above the for loop?

packages/router-core/src/router.ts
```diff
@@ -2684,26 +2686,65 @@ export function getMatchedRoutes<TRouteLike extends RouteLike>({
...
+          const lastRouteSegment = getLastNonSlashSegment(routeSegments)
+          const lastPathSegment = getLastNonSlashSegment(pathSegments)
```
nlynzaad, line 2715: we only need to parse this once right? since this is the trimmedPath we can move this above the for loop?

packages/router-core/src/router.ts
```diff
@@ -2684,26 +2686,65 @@ export function getMatchedRoutes<TRouteLike extends RouteLike>({
...
+          const routeCaseSensitive =
+            route.options?.caseSensitive ?? caseSensitive ?? false
+          const normalizeSegmentValue = (segment?: Segment) => {
```
nlynzaad, line 2718: Bit of a nitpick but can we move this above the for loop and just add the routeCaseSensitive as a paramater. This way its only declared once and not with every iteration that has matching params

packages/router-core/src/router.ts
```diff
@@ -2684,26 +2686,65 @@ export function getMatchedRoutes<TRouteLike extends RouteLike>({
...
+    if (exactMatches.length > 0) {
+      exactMatches.sort(
+        (a, b) => Number(b.endsWithStatic) - Number(a.endsWithStatic),
+      )
+      foundRoute = exactMatches[0]!.route
+      routeParams = exactMatches[0]!.routeParams
+    } else if (fuzzyMatch) {
       foundRoute = fuzzyMatch.foundRoute
       routeParams = fuzzyMatch.routeParams
     }
```
nlynzaad, line 2750: we should still break out as early possible and if no route is ending with a matched static the original case of first route being matched can still apply.

```suggestion

          if (endsWithStatic) {
            foundRoute = route
            routeParams = matchedParams
            break
          }

          if (!foundRoute) {
            foundRoute = route
            routeParams = matchedParams
          }
        }
      }
    }

    if (!foundRoute && fuzzyMatch) {
      foundRoute = fuzzyMatch.foundRoute
      routeParams = fuzzyMatch.routeParams
    }
  }
```

nlynzaad:
@LuggaPugga thanks a lot for this. 

We are a bit concerned about the performance impact this might have and given that #5722 solves this issue, as well as a host of performance enhancements, it's probably best to put this on hold for a bit and see how #5722 progresses

LuggaPugga:
> @LuggaPugga thanks a lot for this.
> 
> We are a bit concerned about the performance impact this might have and given that #5722 solves this issue, as well as a host of performance enhancements, it's probably best to put this on hold for a bit and see how #5722 progresses

Should I go ahead and address the suggested changes, or is it better to wait for #5722 before updating this?

Sheraff:
> Should I go ahead and address the suggested changes, or is it better to wait for #5722 before updating this?

If all goes well 🤞, we're shipping #5722 this week end, and I incorporated your unit tests to make sure this case was handled. So maybe wait and see.

nlynzaad:
@LuggaPugga #5722 has been merged. I will close this PR now.

Thanks again for making this PR and the accompanying tests. 

It definitely helped to ensure the rewrite addressed this issue.

PR #5860, opened 2025-11-13
docs(react-start): serialization adapters

m9tdev:
Added an example on how to add serialization adapters for React Start. Not 100% sure if the right place is in the server function docs. LMK otherwise!

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive "Custom serialization adapters" guide with a complete example (includes a bigint adapter) for handling complex types.
  * Guide published in both React and Solid server-functions guides; the React guide includes the new section in two locations for easier discovery.
  * No changes to public APIs or declarations.

### DIFF
--- a/docs/start/framework/react/guide/server-functions.md
+++ b/docs/start/framework/react/guide/server-functions.md
@@ -270,6 +270,31 @@ export default defineConfig({
 })
 ```
 
+### Custom serialization adapters
+
+You can create custom serialization adapters to handle complex types that can't be serialized by default.
+
+Example:
+
+```ts
+// src/start.ts
+import { createStart } from "@tanstack/react-start";
+import { createSerializationAdapter } from "@tanstack/react-router";
+
+const bigIntAdapter = createSerializationAdapter({
+  key: "bigint",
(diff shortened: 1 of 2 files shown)


birkskyum:
@m9tdev can you add the same to the solid docs so they stay in sync?

schiller-manuel:
we need to document serialization adapters more generally, not just for server functions. they apply for beforeLoad and loader as well. a separate doc page should be added for this. want to do this when i find the time :)

m9tdev:
@birkskyum Should they be the same for Solid? I have no experience with Solid, but if they are the same I'll add them!

@schiller-manuel Maybe we can add this in the meantime to save others some time when looking for it.

birkskyum:
Yes it's the same for solid, except for the package name being solid-start

m9tdev:
@birkskyum done! There was nothing in there yet (except some frontmatter), but I've added it for Solid.

LadyBluenotes:
Serialization adapters need to be documented across Start’s serialization boundaries, not only server functions. Since #7796 now tracks that broader work, I’m going to close this PR in favor of it. Please feel free to open another PR updating it!

PR #5867, opened 2025-11-14
refactor(router-core): simplify decodePath implementation, improve performance

Sheraff:
Using a stateful regex w/ `.exec()` we can decode a string chunk by chunk while avoiding strings in the ignore list. This is less code and more performant.

bench: ~1.5x faster

```
 ✓  @tanstack/router-core  tests/decode.bench.ts > decodePath 1300ms
     name                    hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · old decodePath  298,375.87  0.0030  0.7446  0.0034  0.0034  0.0045  0.0046  0.0087  ±0.31%   149188
   · new decodePath  450,487.87  0.0020  0.0934  0.0022  0.0023  0.0028  0.0029  0.0032  ±0.09%   225244

 BENCH  Summary

   @tanstack/router-core  new decodePath - tests/decode.bench.ts > decodePath
    1.51x faster than old decodePath
```

## Summary by CodeRabbit

* **Refactor**
  * Improved URL/path decoding in the router for more reliable handling of unusual encodings and edge cases, reducing decoding errors and improving navigation stability.
  * Switched to a more robust, iterative decoding approach for predictable behavior across different path formats.

* **Tests**
  * Updated test expectations to standardize percent-encoding casing for consistent validation.

### DIFF
--- a/packages/router-core/src/utils.ts
+++ b/packages/router-core/src/utils.ts
@@ -505,6 +505,7 @@ function decodeSegment(segment: string): string {
 }
 
 export function decodePath(path: string, decodeIgnore?: Array<string>): string {
+  if (!path) return path
   const re = decodeIgnore
     ? new RegExp(`${decodeIgnore.join('|')}`, 'gi')
     : /%25|%5C/gi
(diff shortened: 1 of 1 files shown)


nlynzaad (approved):
This looks great, was looking for a way to just walk the string forward, but didn't think of this way. 🚀

PR #5874, opened 2025-11-15
refactor(router-core): interpolatePath can cache the parsed path template

Sheraff:
We can squeeze a little more performance out of `interpolatePath` by caching the parsed representation of the template path to avoid having to call `parseSegment` again on repeated interpolations of the same path.

Cache is keyed on the `path` and not the `params`, so it should have a pretty good hit rate.

This function is not very slow (~1-2MHz), but it is called on `buildLocation` and `matchRoutesInternal` (2 hot paths). But I'm not sure this change is worth the 1.4x we get out of it...

bench: 1.4x

```
 ✓  @tanstack/router-core  tests/interpolate.bench.ts > interpolatePath 1293ms
     name                         hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · old interpolatePath  272,652.50  0.0032  0.8732  0.0037  0.0036  0.0047  0.0050  0.0193  ±0.38%   136327
   · new interpolatePath  413,747.62  0.0022  0.0959  0.0024  0.0024  0.0027  0.0031  0.0034  ±0.13%   206874

 BENCH  Summary

   @tanstack/router-core  new interpolatePath - tests/interpolate.bench.ts > interpolatePath
    1.52x faster than old interpolatePath
```

## Summary by CodeRabbit

* **New Features**
  * Path interpolation now supports optional caching, improving performance when processing dynamic route parameters
  * Exported ParsedSegment type for advanced TypeScript use cases in custom routing implementations

* **Performance Improvements**
  * Route matching and path interpolation operations now leverage segment caching for enhanced efficiency across repeated path processing

### DIFF
--- a/packages/router-core/src/new-process-route-tree.ts
+++ b/packages/router-core/src/new-process-route-tree.ts
@@ -20,7 +20,7 @@ const OPTIONAL_PARAM_W_CURLY_BRACES_RE =
   /^([^{]*)\{-\$([a-zA-Z_$][a-zA-Z0-9_$]*)\}([^}]*)$/ // prefix{-$paramName}suffix
 const WILDCARD_W_CURLY_BRACES_RE = /^([^{]*)\{\$\}([^}]*)$/ // prefix{$}suffix
 
-type ParsedSegment = Uint16Array & {
+export type ParsedSegment = Uint16Array & {
   /** segment type (0 = pathname, 1 = param, 2 = wildcard, 3 = optional param) */
   0: SegmentKind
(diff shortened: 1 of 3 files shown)


schiller-manuel:
what's the bundle size impact here?

Sheraff:
> what's the bundle size impact here?

I haven't measured (do we have something to measure that?) but yeah that was my blocker too: the perf gains don't really seem worth it here for the increase in code.

LadyBluenotes:
Stale~

PR #5875, opened 2025-11-15
refactor(router-core): buildLocation can skip decoding the next pathname if we dont encode it in the first place

Sheraff:
## Summary by CodeRabbit

* **New Features**
  * Added optional parameter encoding control to path interpolation, allowing path parameters to be used as raw values without encoding when configured, providing greater flexibility in route handling.

* **Improvements**
  * Enhanced hash extraction logic in route location parsing.
  * Refined path parameter handling during route construction for improved consistency.


LadyBluenotes:
Stale~

PR #5897, opened 2025-11-18
refactor(router-core): add error when building route tree if trying to add children routes to an index route

Sheraff:
It seems some theoretically invalid route tree configurations used to work before https://github.com/TanStack/router/pull/5867 but don't work anymore.

This PR proposes that we add some runtime checks (when `NODE_ENV !== 'production'`) to warn users of such patterns.

Those env checks persist in the `dist` build of tanstack/router-core distributed through npm

Screenshot 2025-11-19 at 09 49 11

But get erased at build-time (tested w/ Vite rollup 7.1.12) when the production environment is set (`vite build --mode production` or simply `vite build`)

Screenshot 2025-11-19 at 09 51 15

## Summary by CodeRabbit

* **Bug Fixes**
  * Added development-time validation that surfaces clear errors for invalid route configurations—prevents index routes from having children and ensures parent/child route relationships are consistent, helping catch routing setup mistakes earlier.

### DIFF
--- a/packages/router-core/src/route.ts
+++ b/packages/router-core/src/route.ts
@@ -1744,12 +1744,10 @@ export class BaseRoute<
     THandlers
   > = (children) => {
     if (process.env.NODE_ENV !== 'production') {
-      if (!this.isRoot && this.fullPath.endsWith('/')) {
-        invariant(
-          false,
-          `Cannot add children to index route '${this.id}'. Index routes cannot have child routes.`,
-        )
-      }
+      invariant(
+        this.isRoot || !this.fullPath.endsWith('/'),
+        `Cannot add children to index route '${this.id}'. Index routes cannot have child routes.`,
+      )
     }
     return this._addFileChildren(children) as any
   }
--- a/packages/router-core/src/route.ts
+++ b/packages/router-core/src/route.ts
@@ -1675,21 +1675,6 @@ export class BaseRoute<
       )
     }
 
-    if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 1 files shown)


packages/router-core/src/route.ts
```diff
@@ -1707,6 +1707,25 @@ export class BaseRoute<
...
+          `Parent route with id '${this.parentRoute.id}' returned by getParentRoute on '${this.id}' is an index route and cannot have child routes.`,
+        )
+        invariant(
+          this.parentRoute.children && this.parentRoute.children.includes(this),
+          `Parent route with id '${this.parentRoute.id}' returned by getParentRoute has no child route with id '${this.id}'. Did you forget to call .addChildren()?`,
+        )
+      }
+      if (this.children) {
+        invariant(
+          this.isRoot || !this.fullPath.endsWith('/'),
```
schiller-manuel, line 1724: is the negation incorrect here?

Sheraff: Hmmm in `invariant` the 1st param must be `true` to *not* throw, right? If this is correct, then I think the condition is good:
- If there are children
  - the route must either be the root
  - or not be an index (i.e. not `endsWith('/')`)

---

Just checked the docs https://www.npmjs.com/package/tiny-invariant:
```ts
invariant(truthyValue, 'This should not throw!');

invariant(falsyValue, 'This will throw!');
// Error('Invariant violation: This will throw!');
```

This seems correct to me then

LadyBluenotes:
Stale~

PR #5899, opened 2025-11-19
Prevent the root path "/" from being removed

cxa:
This PR prevents the root path "/" from being emptied by `removeTrailingSlash`.

## Release Notes

* **Bug Fixes**
  * Improved handling of trailing slashes in layout-type routes to correctly exclude the root path from trailing-slash removal.

### DIFF
--- a/packages/router-generator/src/generator.ts
+++ b/packages/router-generator/src/generator.ts
@@ -1247,7 +1247,7 @@ ${acc.routeTree.map((child) => `${child.variableName}Route: typeof ${getResolved
     )
 
     if (
-      node._fsRouteType === 'layout' ||
+      (node._fsRouteType === 'layout' && node.cleanedPath !== '/') ||
       node._fsRouteType === 'pathless_layout'
     ) {
       node.cleanedPath = removeTrailingSlash(node.cleanedPath)


birkskyum:
@cxa there's a failing unit test that needs some attention.

schiller-manuel:
first of all we need to understand why this fix is needed. when would this occur?

cxa:
> first of all we need to understand why this fix is needed. when would this occur?

Without this fix, file route like `_auth.index/route.tsx` will generate a wrong `path: ''` route instead of the correct `path: '/'` one.  (Since `route.tsx` is treated as `layout`, maybe we should use a new `FsRouteType` for it?)

schiller-manuel:
> > first of all we need to understand why this fix is needed. when would this occur?
> 
> Without this fix, file route like `_auth.index/route.tsx` will generate a wrong `path: ''` route instead of the correct `path: '/'` one. (Since `route.tsx` is treated as `layout`, maybe we should use a new `FsRouteType` for it?)


nesting `route.tsx` under `index` is not a supported usage of file based routing. 
if this is only relevant for this, we won't merge this

cxa:
> nesting `route.tsx` under `index` is not a supported usage of file based routing. if this is only relevant for this, we won't merge this

Makes sense, but if this PR enables nesting `index/route.tsx` without causing any issues, I don’t see why it would be rejected.

P.S. Is there anywhere that mentions nested `index/route.tsx` isn’t supported? (used since 2023 but recent updates made it broken)

PR #5905, opened 2025-11-19
Fix: Prevent Flicker When Navigation Happens During Render

vsh1h:
This PR prevents a brief flicker that occurred when navigation happened during render, causing nested <Outlet /> to temporarily render null. Navigation is now triggered before paint (React) or on mount (Solid), so the UI transitions smoothly without flashing.

### Bug Fix
- Eliminates the temporary blank/empty state shown when navigating during render.
- Ensures nested <Outlet /> content renders immediately with no flicker.

## Summary by CodeRabbit

* **Refactor**
  * Optimized navigation timing to ensure proper execution phase synchronization across React and Solid framework examples.
  * Updated navigation effect handling to use layout-phase execution for improved consistency.

### DIFF
--- a/packages/react-router/src/useNavigate.tsx
+++ b/packages/react-router/src/useNavigate.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import { useLayoutEffect } from './utils'
 import { useRouter } from './useRouter'
 import type {
   AnyRouter,
@@ -67,10 +68,7 @@ export function Navigate<
     TMaskFrom,
     TMaskTo
   > | null>(null)
-  const useIsomorphicLayoutEffect =
-    typeof window !== 'undefined' ? React.useLayoutEffect : React.useEffect
-
-  useIsomorphicLayoutEffect(() => {
+  useLayoutEffect(() => {
     if (previousPropsRef.current !== props) {
       navigate(props)
       previousPropsRef.current = props


vsh1h:
Hi @birkskyum!
I’ve worked on this and opened a PR with a fix for the flicker caused by navigation during render. When you have a moment, I’d appreciate your review. Please let me know if there’s anything you’d like adjusted or improved — happy to iterate.
Thanks!

birkskyum:
@vsh1h , for solid it seems like the only change is in an example? so this pr will effectively only impact the react-router. Is that because the solid-router already behaves as expected, if used correctly?

vsh1h:
@birkskyum ,
Hi! Yes,  the Solid side required only a small update in the example because Solid Router already handles this correctly when used in the intended way.
Solid’s navigate() runs before paint due to its synchronous update model, so it doesn’t hit the same flicker issue that React does. The only adjustment I made was to align the example with the recommended pattern (calling navigate() inside onMount / isomorphic layout effect), ensuring consistency across both implementations.
So the functional impact of this PR is mainly on the React router, while the Solid example update simply reflects the correct usage and avoids confusion for users.
Please let me know if there’s anything you’d like adjusted or improved — happy to update.
Thanks!

packages/react-router/src/useNavigate.tsx
```diff
@@ -67,7 +67,10 @@ export function Navigate<
     TMaskFrom,
     TMaskTo
   > | null>(null)
-  React.useEffect(() => {
+  const useIsomorphicLayoutEffect =
```
SeanCassiere: We should have this as a utility exported out of the utils file, please that one instead.

vsh1h: @SeanCassiere 
Thanks for the clarification!
I've updated the PR to use the useLayoutEffect utility from the shared utils file instead of redefining it.
Let me know if you'd like any further adjustments.

SeanCassiere (approved):

PR #5910, opened 2025-11-19
feat(start-plugin-core): vite preview support

edmundhung:
This adds `vite preview` support with a new preview server plugin that loads the built server bundle and serves SSR requests. Both the `react-start` and `solid-start` basic e2e projects have been updated to re-run their test suites in preview mode.

## Summary by CodeRabbit

* **New Features**
  * Added preview-server integration to enable running built apps in preview mode.
  * Added a runtime preview mode toggle for e2e runs.

* **Tests**
  * Added preview-mode e2e test scripts and included preview validation in test orchestration.
  * Test configs updated to support preview-specific ports and execution paths.

* **Chores**
  * Added a small runtime helper to detect preview mode from the environment.


packages/start-plugin-core/src/preview-server-plugin/plugin.ts
```diff
@@ -0,0 +1,64 @@
...
+              }
+
+              const webReq = new NodeRequest({ req, res })
+              const webRes: Response = await serverBuild.fetch(webReq)
+
+              // Temporary workaround
+              // Vite preview's compression middleware doesn't support flattened array headers that srvx sets
+              // Call writeHead() before srvx to avoid corruption
+              res.setHeaders(webRes.headers)
+              res.writeHead(webRes.status, webRes.statusText)
```
edmundhung, line 53: For context:

Vite preview server uses `@polka/compression` to [handle compression](https://github.com/vitejs/vite/blob/acfe939e1f7c303c34b0b39b883cc302da767fa2/packages/vite/src/node/preview.ts#L237). This middleware overwrites the `res.writeHead` method, which [expects headers to be an object](https://github.com/lukeed/polka/blob/02cbdb529ddca0a9f3d225e2abb2931924219cc3/packages/compression/index.js#L85) and uses a `for...in` loop to iterate over them.

However, [`srvx` passes headers as a flattened array](https://github.com/h3js/srvx/blob/2cf3282e38d5020037afee9bf8256133a6ef5f76/src/adapters/_node/send.ts#L41-L63) to `writeHead()`. When the middleware's `for...in` loop runs through an array, it iterates over **array indices** (`'0'`, `'1'`, `'2'`, ...) instead of the actual header names.

This causes header corruption where a header like:
```
Content-Type: text/html; charset=utf-8
```

Becomes:
```
0: Content-Type
1: text/html; charset=utf-8
```

The workaround is to call `res.writeHead()` directly before `sendNodeResponse()`, which sets `res.headersSent = true` and causes `sendNodeResponse()` to skip header writing entirely.

I will try to submit a patch to either Vite or `@polka/compression`, but we will likely want to keep this workaround for compatibility with earlier versions of Vite.

birkskyum:
Tested this out, and it seems to work as expected

edmundhung:
Thanks for the review @birkskyum @SeanCassiere!

Issue #5913, opened 2025-11-19
Client middleware context (sendContext) is not passed to server middleware when a server function accepts FormData.


empotts:
### Which project does this relate to?

Start

### Describe the bug

Issue: Client middleware context (sendContext) is not passed to server middleware when a server function accepts FormData.
Expected Behavior
When a middleware defines both client and server handlers:
The client handler runs and sends context via sendContext (line 12-15 in middleware.ts)
The server handler receives that context in its context parameter (line 18 in middleware.ts)
The context should be available throughout the request chain
Actual Behavior
When a server function accepts FormData:
The client middleware runs and creates testString via sendContext
The server middleware runs, but context.testString is undefined
This causes an error: "BUG REPRODUCED: testString is missing from server middleware context!"
Root Cause
The bug appears to be in how TanStack Start serializes and transmits context when the request body is FormData. The sendContext data from the client middleware is not being properly included in the request when FormData is used, unlike when JSON or other data types are sent.

### Your Example Website or App
https://github.com/empotts/tanstack-middleware-formdata-bug
This stackblitz doesnt really work for either server function but download the repo, pnpm i, pnpm dev
https://stackblitz.com/~/github.com/empotts/tanstack-middleware-formdata-bug

### Steps to Reproduce the Bug or Issue

Click the button for the formdata server function and it fails. 

### Expected behavior

I expect the context to be passed from the client to the server middleware. This works for normal server functions but not with form data.

### Screenshots or Videos

<img width="592" height="591" alt="Image" src="https://github.com/user-attachments/assets/2c7b5e0a-fa75-41a4-af50-83532c2dfa4f" />

[SERVER MIDDLEWARE] Received context: {
  t: { t: 10, i: 0, p: { k: [Array], v: [Array], s: 1 }, o: 0 },
  f: 31,
  m: []
}

Server Fn Error!

Error: BUG REPRODUCED: testString is missing from server middleware context!
    at context (/Users/epotts/Downloads/tanstack-middleware-formdata-bug/src/middleware.ts:23:19)
    at applyMiddleware (/Users/epotts/Downloads/tanstack-middleware-formdata-bug/node_modules/.pnpm/@tanstack+start-client-core@1.136.17/node_modules/@tanstack/start-client-core/src/createServerFn.ts:638:10)
[...]

PR #5935, opened 2025-11-21
fix(start-server-core): deserialize tss context when using formdata

empotts:
Fixes https://github.com/TanStack/router/issues/5913 where the client `TSS_CONTEXT` is not deserialized when using formdata

## Summary by CodeRabbit

* **New Features**
  * Added a FormData context test route with a UI to invoke two server actions and view results.

* **Refactor**
  * Improved server-side context deserialization for more reliable context handling (no public API changes).

* **Tests**
  * Added end-to-end tests verifying context propagation for FormData and simple server actions.


schiller-manuel:
thanks! we should also add a testcase to the e2e server functions test

empotts:
> thanks! we should also add a testcase to the e2e server functions test

Added a test similar to my reproducer and ran it. Seems to work along with the other tests.

empotts:
Anything else I need to add?

PR #5936, opened 2025-11-21
feat(router-core): validate params while matching

Sheraff:
## summary

Use the `params.parse` option (https://tanstack.com/router/latest/docs/framework/react/api/router/RouteOptionsType#paramsparse-method) to validate path params while matching a route. This allows us to look for *another match* if the validation fails.

Pros: more flexible routing. Any `$param` can now become a regex route segment (or any validation you like).

Cons: routes with a `params.parse` method cannot share the same node as other routes of otherwise the same shape. If used frequently, this increases branching, which increases the number of branches we need to explore to find a match.

For more details, 
- refer to the doc written in this PR https://github.com/TanStack/router/pull/5936/changes#diff-d45fd81a1b75da56cf458c0e766d31c3e2cf1615ea86c3e25b20c55fb5d572c7
- or to the RFC written for this feature https://github.com/TanStack/router-private/blob/rfc-001-params-parse-while-matching/RFC-001-params-parse-matching.md

---

Some details:
- routes w/ `params.parse` have priority over routes without (all else being equal)
- `extractParams` is now "resumable" so we can call it at several point while matching without repeating work (at the expense of allocating more objects)
- we now have `index` and `pathless` nodes in the segment tree so that previously "invisible" segments can carry "route matching metadata", i.e. in this case `params.parse` and `skipRouteOnParseError`

## Summary by CodeRabbit

* **New Features**
  * Route-level parameter parsing: routes can provide custom parse handlers for path params.
  * New route option to control behavior when param parsing fails: skipRouteOnParseError (with optional priority).

* **Documentation**
  * Added guide on validating path parameters during matching with examples and guidance.
  * Expanded docs for allowed characters and configuration for path params.


Sheraff:
@coderabbitai review

docs/router/framework/react/api/router/RouteOptionsType.md
```diff
@@ -88,6 +88,27 @@ The `RouteOptions` type accepts an object with the following properties:
...
+  priority?: number
+}
+```
+
+- Optional
+- By default, when a route's `params.parse` function throws an error, the route will match and then show an error state during render. With `skipRouteOnParseError.params` enabled, the router will skip routes whose `params.parse` function throws and continue searching for alternative matching routes.
+- See [Guides > Path Params > Validating path parameters during matching](../../guide/path-params#validating-path-parameters-during-matching) for detailed usage examples.
+
+> [!IMPORTANT]
+> **Performance impact**: This option has a **non-negligible performance cost** and should not be used indiscriminately. Routes with `skipRouteOnParseError` are placed on separate branches in the route matching tree instead of sharing nodes with other dynamic routes. This reduces the tree's ability to efficiently narrow down matches and requires testing more route, even for routes that wouldn't match the path structure alone.
```
beaussan: **nitpick(non-blocking):** ‏I found the sentence hard to read as it is, here is a suggestion to keep is simple
```suggestion
> **Performance impact**: This option has a **non-negligible performance cost** and should only be enabled when needed. Routes with `skipRouteOnParseError` are placed on separate branches in the route matching tree instead of sharing nodes with other dynamic routes. This reduces the tree's ability to efficiently narrow down matches and requires testing more route, even for routes that wouldn't match the path structure alone.
```

docs/router/framework/react/guide/path-params.md
```diff
@@ -738,6 +738,303 @@ function ShopComponent() {
 
 Optional path parameters provide a powerful and flexible foundation for implementing internationalization in your TanStack Router applications. Whether you prefer prefix-based or combined approaches, you can create clean, SEO-friendly URLs while maintaining excellent developer experience and type safety.
 
+## Validating Path Parameters During Matching
+
+> [!WARNING]
+> The `skipRouteOnParseError` option is currently **experimental** and may change in future releases.
+
+> [!IMPORTANT]
+> **Performance cost**: This feature has a **non-negligible performance cost** and should not be used indiscriminately. It creates additional branches in the route matching tree, reducing matching efficiency and requiring more route evaluations. Use it only when you genuinely need type-specific routes at the same path level.
```
beaussan: **nitpick(non-blocking):** ‏I found the sentence hard to read as it is, here is a suggestion to keep is simple

```suggestion
> **Performance cost**: This feature has a **non-negligible performance cost** and should only be enabled when needed. It creates additional branches in the route matching tree, reducing matching efficiency and requiring more route evaluations. Use it only when you genuinely need type-specific routes at the same path level.
```

beaussan (approved):
Docs LGTM, only one nitpick about wording

<details>
<summary>
<i>What are the prefixes on the comment and what do they mean?</i>

</summary>


My comments are formatted according to the [conventional comment](https://conventionalcomments.org/) convention to help you identify what they are, and how much they are a hard requirement or not. Don't hesitate to reach out if you w ant to learn more

</details>

packages/router-core/src/route.ts
```diff
@@ -1188,9 +1188,42 @@ export interface UpdatableRouteOptions<
   in out TBeforeLoadFn,
 > extends UpdatableStaticRouteOption,
     UpdatableRouteOptionsExtensions {
-  // If true, this route will be matched as case-sensitive
+  /**
+   * Options to control route matching behavior with runtime code.
+   */
+  skipRouteOnParseError?: {
```
SeanCassiere, line 1198: Add a JSDoc annotation for `@experimental` mentioned this can change. Maybe even a link to the docs RouteOptionsType reference page.

If you search the codebase we should have something similar to copy off of marked with `@link`.

docs/router/framework/react/guide/path-params.md
```diff
@@ -738,6 +738,303 @@ function ShopComponent() {
...
+
+By default, TanStack Router matches routes based purely on URL structure. A route with `/$param` will match any value for that parameter, and validation via `params.parse` happens later in the route lifecycle. If validation fails, the route shows an error state.
```
SeanCassiere:
```suggestion
By default, TanStack Router doing route matching purely base on URL structure. A route with `/$param` will match any value for that parameter, with validation via `params.parse` happening later in the route lifecycle. If validation fails, the route is set into an 'error state'.
```

Issue #5930, opened 2025-11-21
Links always scroll to URL hash on hover


Mario-Eis:
### Which project does this relate to?

Router

### Describe the bug

After an update of react router a bug was introduced to our application. When there is a hash fragment in the URL to scroll to a specific headline, a scroll will be triggered as soon as any react router link on the page is hovered. I pinned it down to the last working version of react router that is 1.134.18. The next version on NPM is 1.134.20, which breaks the behaviour already. So I would suggest, the bug was intoduced in 1.134.19 or 1.134.20

### Your Example Website or App

https://stackblitz.com/edit/github-nsuiogjw?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

1. Click on "About" link
2. Click on "To hash" link
3. Scroll back up again
4. Hover the cursor over the "Home" link

### Expected behavior

Result will be: Browser scrolls to the hash position
Expected is: No scroll at all when moving the cursor over a link.

### Platform

- Router Version: 1.134.20 - 1.139.0
- OS: Windows
- Browser: Brave
- Browser Version: latest
- Bundler: vite
- Bundler Version: latest

Mario-Eis:
Related to: https://github.com/TanStack/router/issues/5931
There is some additinal information about a connection of the issue to preloading.

Shreyas-Sarkar:
@Mario-Eis @birkskyum  If this issue is approved I would like to work on this!

birkskyum:
It is confirmed, @Shreyas-Sarkar that would be great - thanks! If possible, make a e2e test case for solid and react using the route provided in the reproduction.

PR #5943, opened 2025-11-22
fix: Links always scroll to URL hash on hover (#5930)

Shreyas-Sarkar:
# Fix: Prevent hash scroll on preload-only transitions by gating handleHashScroll to true navigations

## What was the issue (#5930)

Hovering a Link with defaultPreload set to "intent" (or any programmatic preload) caused the router to trigger a hash scroll even though no navigation occurred. This resulted in the page unexpectedly jumping to an element associated with a hash fragment (for example, #id) when the user only hovered over a link.

In versions prior to v1.134.19, this did not occur. Between v1.134.18 and v1.134.20, router-core began running updateMatch inside router.startTransition. As a result, framework adapters (React and Solid) emitted onResolved during preload transitions. Both adapters unconditionally called handleHashScroll inside onResolved, so preloads triggered hash scrolling despite no URL change.

This caused the following behavior:

1. Navigate to a page containing a hash target.
2. Scroll down the page.
3. Hover any Link to trigger preload.
4. The page jumps back to the hash position even though no navigation occurred.

## The fix

In both framework Transitioner implementations (React and Solid):

1. Compute changeInfo using getLocationChangeInfo once during resolution.
2. Continue emitting onResolved using this changeInfo.
3. Only invoke handleHashScroll(router) when changeInfo.hrefChanged is true, ensuring that hash scrolling occurs only after an actual URL change.
4. No changes were made to router-core or scroll-restoration internals.

This restores the intended behavior:

- Hash scrolling occurs after real navigations.
- Preload-only transitions do not trigger any scrolling.
- Scroll restoration, SSR behavior, and navigation semantics remain unchanged.


Shreyas-Sarkar:
@birkskyum Please review this PR.

birkskyum:
@Shreyas-Sarkar there are no files changed?

Shreyas-Sarkar:
@birkskyum seems like an error from my end, I'll retry

Issue #5804, opened 2025-11-10
setState callback was called twice


zWingz:
### Which project does this relate to?

Start

### Describe the bug

I just call setState(prev => prev + 1) once, but it called twice

### Your Example Website or App

no

### Steps to Reproduce the Bug or Issue

- npx gitpick TanStack/router/tree/main/examples/react/start-basic start-basic
- modify `index.tsx` to
```tsx
import { createFileRoute } from '@tanstack/react-router'
import { useState, useCallback } from 'react'

export const Route = createFileRoute('/')({
  component: Home,
})

function Home() {
  console.log('render home');
  const [count, setCount] = useState(0)
  const click = useCallback(() => {
    setCount(prev => {
      console.log('call set COUnt');
      return prev + 1
    })
  }, [])
  return (
    <div className="p-2">
      <h3>Welcome Home!!!</h3>
      <button onClick={click} className="p-2 bg-blue-500 text-white rounded">
        Count is {count}
      </button>
    </div>
  )
}
```

click the button, It will exec the callback twice

<img width="325" height="738" alt="Image" src="https://github.com/user-attachments/assets/e7f3dfd9-b5e7-4c73-a96b-6a78fbe46fe7" />

### Expected behavior

exec the setState callback onde

### Screenshots or Videos

<img width="325" height="738" alt="Image" src="https://github.com/user-attachments/assets/2b8136ec-ea99-4318-be25-95dd812bf63c" />

### Platform

- Router / Start Version: latest
- OS: macOS
- Browser: Chrome

### Additional context
the same code in codesandbox just only exec the callback once
https://codesandbox.io/p/sandbox/keen-cray-v3v93w?file=%2Fsrc%2FApp.js%3A9%2C37

<img width="545" height="696" alt="Image" src="https://github.com/user-attachments/assets/72688933-20d7-4d3e-b99d-d8391cb846d2" />

Divyapahuja31:
@zWingz  i would like to work on this issue... can u plss assign it to me ?

schiller-manuel:
@Divyapahuja31 we don't assign issues. just create a PR that resolves this

Divyapahuja31:
@zWingz sure

schiller-manuel:
> the same code in codesandbox just only exec the callback once
[...]

PR #5948, opened 2025-11-23
fix(start-basic): remove StrictMode to prevent double setState callba…

Divyapahuja31:
(#5804)

 Title: Fix: Remove StrictMode in start-basic example to prevent double setState callback
Description:
This PR resolves issue #5804 where calling setState(prev => prev + 1) inside the start-basic example caused the functional update callback to execute twice.
This behavior occurs because React 18’s StrictMode intentionally double-invokes state updater functions in development, which can be confusing when testing simple examples.

What This PR Does
Removes React.StrictMode usage in the example’s hydration entry (client.tsx).
Adds a clear comment explaining:
Why StrictMode is not used in this example.
That React 18 double-invocations are expected behavior.
Ensures that clicking the button increments the counter only once, matching the expected example output.

Why This Change Is Needed
Developers following the start-basic example were encountering unexpected double state updates due to StrictMode.
Removing StrictMode makes the example behave consistently with expectations and avoids confusion.

Testing
Ran the start-basic example.
Confirmed that setState functional callback runs once after the change.
Verified that no other behavior is affected.

Related Issue
Fixes: #5804

## Summary by CodeRabbit

* **New Features**
  * Added a new React example project demonstrating proper client initialization with React 18 hydration capabilities.

### DIFF
--- a/examples/react/start-basic/src/client.tsx
+++ b/examples/react/start-basic/src/client.tsx
@@ -1,0 +1,13 @@
+import { hydrateRoot } from 'react-dom/client'
+import { StartClient } from '@tanstack/react-start/client'
+
+// React 18 StrictMode causes setState functional updates
+// to run twice. This example intentionally disables StrictMode
+// so the counter increments only once when clicked.
+//
+// See: https://react.dev/reference/react/StrictMode
+
+hydrateRoot(
+  document,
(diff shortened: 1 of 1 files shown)


schiller-manuel:
no need to fix anything here. this is expected

PR #5973, opened 2025-11-26
fix(start-server-core): h3 2.0.0-beta.5

birkskyum:
Reverts TanStack/router#5959

We need a replacement for event._res

## Summary by CodeRabbit

* **Chores**
  * Updated server core dependency to a newer version.
* **Bug Fixes**
  * Improved robustness in response handling by adding null-safety checks to preserve status and status text when available and avoid errors when missing.

### DIFF
--- a/packages/start-server-core/package.json
+++ b/packages/start-server-core/package.json
@@ -79,7 +79,7 @@
     "@tanstack/router-core": "workspace:*",
     "@tanstack/start-client-core": "workspace:*",
     "@tanstack/start-storage-context": "workspace:*",
-    "h3-v2": "npm:h3@2.0.0-beta.4",
+    "h3-v2": "npm:h3@2.0.0-beta.5",
     "seroval": "^1.4.0",
     "tiny-invariant": "^1.3.3"
   },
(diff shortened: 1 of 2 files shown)


pi0:
`event._res` instance is accessable by [`event[Symbol.for("h3.internal.event.res")]`](https://github.com/h3js/h3/blob/5405a1b2a23a7a4ad66a651a2743706e89c31493/src/event.ts#L14-L16) (but naturaly internals can change in any future version... `event.res` is more reliable and if any reason you can't use it, please let me know to find a solution not sure why #5975 closed)

birkskyum:
@pi0 , thank you. I'll try to take this stepwise - staying on the internal value with `h3.internal.event.res` we're now upgraded to rc.2:
- #6159  (released in [1.142.1](https://github.com/TanStack/router/releases/tag/v1.142.1))

Next step is the shift to the public event.res
- #6162 

And finally, the bump to 2.0.1-rc.6, which I notice cause failing ci due to header merging behavior changes, so there's some work needed to adapt
- #6161

PR #6000, opened 2025-12-01
feat: Add flag for disabling JSON-first search param parsing

adbjo:
By default, the router parses search params using `JSON.parse` and converts string values to native types (numbers, booleans). This causes strange behavior, for example, it adds quotes around numerical strings:
`http://my-app.com?p=123` changes to `http://my-app.com?p="123"`

It also causes rounding issues when large numbers are converted into integers, for example:

`http://my-app.com?p=1000000000000001110` becomes `http://my-app.com?p="1000000000000001200"`

Which means that resource IDs cannot be used in search params, which I guess is a common use-case.

This PR solves this by adding a `inferTypes` option to `parseSearchWith` and allows passing `null` as parser / serializer callbacks to `parseSearchWith` and `stringifySearchWith`. This let's developers opt out of JSON parsing completely as follows:
```js
const router = createRouter({
  routeTree,
  ...,
  parseSearch: parseSearchWith(null, { inferTypes: false }),
  stringifySearch: stringifySearchWith(null, null),
});
```

Also adds a shorthand for this configuration, looks like this:
```js
const router = createRouter({
  routeTree,
  ...,
  search: { raw: true },
});
```

## Summary by CodeRabbit

* **New Features**
  * New router option to treat search params as raw strings (preserves exact values).
  * Option to disable built-in serialization/parsing so values pass through unchanged.

* **Refactor**
  * Search param parsing/stringifying made more configurable with explicit raw vs. inferred-type modes; router respects the new mode and defaults accordingly.

* **Documentation**
  * Docs and guide updated to explain raw-search behavior and its trade-offs.

* **Tests**
  * Added round-trip tests for inferred-type and raw-string modes.


adbjo:
Fixes: https://github.com/TanStack/router/issues/6044

packages/router-core/tests/searchParams.test.ts
```diff
@@ -2,43 +2,44 @@ import { describe, expect, test } from 'vitest'
...
-    [{ 0: 1 }, '?0=1'],
-    [{ 'foo=bar': 1 }, '?foo%3Dbar=1'],
-    [{ '{}': 1 }, '?%7B%7D=1'],
-    [{ '': 1 }, '?=1'],
-    [{ '=': '=' }, '?%3D=%3D'],
-    [{ '=': '', '': '=' }, '?%3D=&=%3D'],
-    [{ 'foo=2&bar': 3 }, '?foo%3D2%26bar=3'],
-    [{ 'foo?': 1 }, '?foo%3F=1'],
-    [{ foo: 'bar=' }, '?foo=bar%3D'],
-    [{ foo: '2&bar=3' }, '?foo=2%26bar%3D3'],
```
Sheraff, line 37: Would it be possible to put those back the other way around? It might be a bit of a nitpick, but having these flipped
```diff
- [{}, ''],
+ ['', {}],
```
makes it harder to see what changed and what didn't. And for a PR like this, we're really trying to review this stuff character by character.

adbjo: good point, flipped them back to how it was

Sheraff:
Looking at the tests, it looks like this change prevents primitives from being parsed correctly. For example in `packages/react-router/tests/router.test.tsx:2993:42` we can see that instead of receiving `true` and `123` (as `boolean` and `number` respectively) we get strings.

```diff
  {
-   "migrated": true,
-   "productId": 123,
+   "migrated": "true",
+   "productId": "123",
  }
```

adbjo:
> Looking at the tests, it looks like this change prevents primitives from being parsed correctly. For example in `packages/react-router/tests/router.test.tsx:2993:42` we can see that instead of receiving `true` and `123` (as `boolean` and `number` respectively) we get strings.
> 
> 
> 
> ```diff
> 
>   {
> 
> -   "migrated": true,
> 
> -   "productId": 123,
> 
> +   "migrated": "true",
> 
> +   "productId": "123",
> 
>   }
> 
>

Interpreting search params as json primitives is a bug which I try to fix in this PR.

Fortunately, this library doesn't rely on JSON.parse for type safety anyway.

Sheraff:
Would this PR mean that existing URLs might become invalid? Say someone has shared a link to `/foo?bar="123"`, it would previously be parsed as `{ bar: '123' }` but would now become `{ bar: '"123"' }`?

If that's the case, then it's a hard change to ship. It might have to be done in a major release.

adbjo:
> Would this PR mean that existing URLs might become invalid? Say someone has shared a link to `/foo?bar="123"`, it would previously be parsed as `{ bar: '123' }` but would now become `{ bar: '"123"' }`?
> 
> 
> 
> If that's the case, then it's a hard change to ship. It might have to be done in a major release.

Yes, it would. But I would argue that existing URLs already are broken.

sandy2008:
Hi :) I fixed the issues with a non-breaking change way: https://github.com/TanStack/router/pull/6946

adbjo:
@coderabbitai resume

adbjo:
@coderabbitai resume

adbjo:
@Sheraff Hey, I have updated this PR so that it preserves the current behavior, but allows developers to opt out of the JSON parsing and type inference more easily. Please have a look.

adbjo:
@coderabbitai resume

LadyBluenotes:
This now preserves existing behavior, so I’m going to keep it open while we decide between this API and #6946

@Sheraff if you can spare some time can you decide whether this should stay opened or if it needs a new review?

PR #6004, opened 2025-12-01
docs: add commands each example

enBonnet:
I was trying to use a starter but it took me some tries to copy the proper URL and get the command to use it, so I came back to add the link for each example, it could make it easy for other.

I'll look like this after changes:

|Screenshot 2025-12-01 105024|
|-|
|Screenshot 2025-12-01 105045|

## Summary by CodeRabbit

* **Documentation**
  * React quick-start guide: added collapsible "Show command" sections that pair each example with a corresponding GitPick shell command for quick reference.
  * Content-only update — no changes to navigation, behavior, or public APIs were made.


LadyBluenotes:
Sorry this sat. I’m going to close it because quick-start.md has since been merged into the Getting Started guide, which includes the GitPick command pattern, and each example README now has its exact clone command. This covers the original usability issue without repeating a command block for every example.

Issue #6012, opened 2025-12-02
Missing optional param when using pathless layoutes


JohanDavidsson:
### Which project does this relate to?

Router

### Describe the bug

I have two pathless layoutes. Both have a optional language param route in them. When one of them gets matched the param is missing from useParams.

### Your Example Website or App

https://stackblitz.com/edit/github-fkabbqxu?file=src%2Froutes%2F__root.tsx

### Steps to Reproduce the Bug or Issue

1. Go to link "/sv (first pathless)"
2. See useParams returns nothing, should return {language: "sv"}

### Expected behavior

useParams should, if i'm not mistaking, return {language: "sv"}, but returns nothing.

### Platform

- Router / Start Version: 1.139.12. 
- OS: macOS
- Browser: Chrome
- Browser Version: 142
- Bundler: Vite
- Bundler Version: 7.1.2

### Additional context

Stopped working when updating from 1.131.36 to 1.139.12. Worked as expected in 1.131.36

Sheraff:
Interesting issue you got there.

First of all, I'm not really sure how this happened, but there is a "hidden" route in the stackblitz you linked, that I do not see in the files, but that does get generated by the dev server: `/{-$year}/{-$month}/{-$day}`. It can be seen by looking in the `routeTree.gen.ts`.

Trying to make a minimal reproduction of your issue in unit tests directly in the core of the repo (`packages/router-core`) I get a weird result.
```ts
describe('edge-cases', () => {
  it('#6012: optional index at root with param extraction', () => {
    const tree = {
      "id": "__root__",
      "fullPath": "/",
      "path": "/",
      "isRoot": true,
      "options": {},
      "children": [
        {
          "id": "/{-$year}/{-$month}/{-$day}",
          "fullPath": "/{-$year}/{-$month}/{-$day}",
          "path": "{-$year}/{-$month}/{-$day}",
          "isRoot": false,
          "options": {}
        },
        {
          "id": "/_pathless/{-$language}/",
          "fullPath": "/{-$language}/",
          "path": "{-$language}/",
          "isRoot": false,
          "options": {}
        }
      ]
    }
    const { processedTree } = processRouteTree(tree)
    const result = findRouteMatch(`/sv`, processedTree)
    expect(result?.route.id).toBe('/_pathless/{-$language}/')
    expect(result?.params).toEqual({ language: 'sv' })
  })
})
```

This is what we would want to happen, but what I get when running this is:
```sh
[...]

PR #6015, opened 2025-12-02
fix(router-core): matching competing optional routes uses proper fullPath for params extraction

Sheraff:
Fixes https://github.com/TanStack/router/issues/6012 (maybe)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved route precedence so index routes are selected correctly at the root, preserving proper full path resolution and avoiding incorrect not-found assignments.
  * Better parameter extraction for optional root index routes, ensuring language and other params are captured reliably.

* **Tests**
  * Added edge-case tests (including URI-decoding scenarios) to validate the above routing behaviors.


Sheraff:
Fix is confirmed

Screenshot 2025-12-02 at 23 46 12

Issue #6011, opened 2025-12-02
router.matchRoute doesn't resolve TS type with pathless (layout) routes


vladrasenko:
### Which project does this relate to?

Router

### Describe the bug

When trying to use the `router.matchRoute` function with a path that is contained within a pathless route, the resulting match result is typed to `false`. Running the function returns the correct value. It's only TS issue.

### Your Example Website or App

https://stackblitz.com/edit/github-trezcbpr?file=src%2Fmain.tsx

### Steps to Reproduce the Bug or Issue

1. Create route `routes/_pathless/route.tsx`
2. Create nested route `routes/_pathless/nested/$id.tsx`
3. Call `router.matchRoute` (use global router instance) with `"/nested/$id.tsx"`
```ts
  // match has type 'false'
  const match = router.matchRoute({
    to: '/nested/$id',
    params: {
      id: 'someId',
    },
  });

  return match && match.id; // Property 'id' does not exist on type 'never'.(2339)
```

### Expected behavior

return type is `false | ResolveParams<"/_pathless/nested/$id">` instead of `false`

### Screenshots or Videos

<img width="421" height="176" alt="Image" src="https://github.com/user-attachments/assets/92bed1a2-3b1b-4df3-823f-928c911ae84e" />

<img width="691" height="247" alt="Image" src="https://github.com/user-attachments/assets/e5031980-8c5b-4484-b181-57b1f811551e" />

<img width="375" height="177" alt="Image" src="https://github.com/user-attachments/assets/d474a2af-3c74-4091-8f76-3959455f277c" />

### Platform

- Router Version: 1.139.12
- OS: macOS
- Browser: Chrome
- Browser Version: 142.0.7444.176
- Bundler: vite
- Bundler Version: 7.1.7

### Additional context

Maybe the same reason as at https://github.com/TanStack/router/issues/1680

PR #6018, opened 2025-12-03
fix(router-core): correct TypeScript param inference for routes nested inside _pathless directories (#6011)

devel-maverick:
**Description**

This PR fixes a TypeScript inference issue where route parameters are not detected when a route file is placed inside a _pathless directory. Runtime behavior was already correct, but TypeScript always inferred router.matchRoute() as false, causing parameter access to fail at compile time.

**Problem**

When using a structure like:

`routes/_pathless/nested/$id.tsx`

the type-level path becomes:

`"/_pathless/nested/$id"`

Even though _pathless is removed from the actual URL at runtime, the type system still sees it as a real segment. This prevents $id from being extracted, producing:

```
const match = router.matchRoute({ 
  to: "/nested/$id", 
  params: { id: "123" } 
})

// match inferred as false only
match?.id // Property 'id' does not exist
```

 **Fix**

We normalize the type-level path by stripping _pathless before param extraction:

`StripPathless<TFrom>, StripPathless<TTo>`

Then apply this normalization inside ValidateParams:

```
export type ValidateParams<
  TRouter extends AnyRouter = RegisteredRouter,
  TTo extends string | undefined = undefined,
  TFrom extends string = string,
> = PathParamOptions<
  TRouter,
  StripPathless<TFrom>,
  TTo extends string ? StripPathless<TTo> : TTo
>

```

This ensures TypeScript sees the correct path:

`"/nested/$id"`

allowing parameter extraction to work as intended.

 **Result (after fix)**
```
const match = router.matchRoute({
  to: "/nested/$id",
  params: { id: "123" }
})

// ✔ match: false | { id: string }
match?.id // correctly typed

```

fixes: (#6011)

## Summary by CodeRabbit

* **Improvements**
  * Path validation in the router now features automatic normalization of special routing segments for consistent and reliable behavior. This enhancement reduces configuration edge cases, improves routing predictability, and ensures stable handling across complex scenarios with multiple parameters and nested paths, making the routing system more robust and maintainable.

### DIFF
--- a/packages/router-core/src/typePrimitives.ts
+++ b/packages/router-core/src/typePrimitives.ts
@@ -11,6 +11,11 @@ import type { AnyRouter, RegisteredRouter } from './router'
 import type { UseParamsResult } from './useParams'
 import type { UseSearchResult } from './useSearch'
 import type { Constrain, ConstrainLiteral } from './utils'
+// Remove any `_pathless` folder segments for type extraction
+export type StripPathless<T extends string> =
+  T extends `${infer A}/_pathless/${infer B}`
+    ? `${A}/${B}`
+    : T;
 
 export type ValidateFromPath<
   TRouter extends AnyRouter = RegisteredRouter,
@@ -29,11 +34,16 @@ export type ValidateSearch<
   TFrom extends string = string,
 > = SearchParamOptions<TRouter, TFrom, TTo>
 
+
 export type ValidateParams<
   TRouter extends AnyRouter = RegisteredRouter,
   TTo extends string | undefined = undefined,
   TFrom extends string = string,
-> = PathParamOptions<TRouter, TFrom, TTo>
+> = PathParamOptions<
+  TRouter,
+  StripPathless<TFrom>,
+  TTo extends string ? StripPathless<TTo> : TTo
+>
 
 /**
  * @private


schiller-manuel:
very likely we need the same fix as in https://github.com/TanStack/router/pull/1690

we also need type tests

devel-maverick:
Thanks! Yes — I can mirror the approach from #1690.

I'll update this PR to:

Apply the same normalization logic inside useMatchRoute for layout routes

Add the missing type tests to ensure StripPathless behaves correctly, including nested cases.

Also included the recursive StripPathless fix suggested by CodeRabbit so multiple _pathless segments are fully stripped.

Let me push an update shortly.

schiller-manuel:
we wouldn't need any additional typescript types it the other fix works out

devel-maverick:
Hey! Just checking in —
Should I wait for the other related fix to land first before updating this PR?
Or is this one good to proceed independently?

Happy to change the implementation either way. Let me know if there's anything blocking the merge, since I can update this PR pretty quickly.

LadyBluenotes:
This is now being addressed in #7571 using the match-route type resolution approach discussed above, rather than stripping pathless segments. I’m going to close this PR in favor of that implementation

PR #6049, opened 2025-12-09
fix: route context being undefined when navigating to a new route faster than beforeLoad resolves

### LINKED ISSUE #6040: [react-router, solid-router] Route context is missing when navigating away from a route too quickly

Router

### Describe the bug

When instantly navigating away from a route (like directly inside an effect), the layout above get's  `undefined` as route context, if the route context finishes loading after the navigation.

_The example is in solid-router, but the same issue exists with react-router._

https://github.com/jakst/tsr-missing-route-context

1. Start the dev server and open up the app
2. Notice the alert being printed about missing route context (which should always exist)

### Expected behavior

I expect the route context to always be defined

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.140.0
- OS: MacOS
- Browser: Chrome
- Browser Version: 143
- Bundler: vite
- Bundler Version: 7.2.6

### Additional context

_No response_

jakst:
This is just the unit tests. No fix is implemented. Leaving that to you guys. Interestingly, I couldn't get the bug to reproduce in the solid-js unit test, even though that's where we're having the issue. But the bug reproduces in the react-router unit test.

Fixes #6040

### DIFF
--- a/packages/react-router/tests/routeContext.test.tsx
+++ b/packages/react-router/tests/routeContext.test.tsx
@@ -9,6 +9,7 @@ import {
 import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
 import { z } from 'zod'
 
+import { useEffect } from 'react'
 import {
   Link,
   Outlet,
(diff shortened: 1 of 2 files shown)


LadyBluenotes:
This is stale so im going to close it. Feel free to reopen a new one against main

PR #5997, opened 2025-11-30
TanStack Router for Vue

birkskyum:
Since a Vue single-file component (SFC) only can export one component, and most tanstack router routes export multiple (main component, error compoentn, not found component etc.) the iteration here uses either JSX or multiple vue files like this:

## File-Based Routing Conventions

optional vue-specific route files

| Suffix/Pattern | Purpose |
|---------------|---------|
| `.component.vue` | The component rendered for the route |
| `.errorComponent.vue` | Error boundary component for the route |
| `.notFoundComponent.vue` | Not found component for the route |
| `.pendingComponent.vue` | Pending component for the route |

existing

| Suffix/Pattern | Purpose |
|---------------|---------|
| `.ts` | Route configuration (loader, validateSearch, head, etc.) |
| `.lazy.ts` | Lazy-loaded route configuration |
| `_layout` prefix | Layout routes that wrap child routes |
| `_` suffix (e.g., `posts_.$postId`) | Unnested routes (break out of parent layout) |
| `(groupName)` directory | Route groups (organizational, don't affect URL) |
| `$param` | Dynamic route parameters |

PR #6055, opened 2025-12-10
TanStack Start for Vue

birkskyum:
# Vue Start

This is a continuation of
- #5997 

This PR adds packages
- `@tanstack/vue-start`
- `@tanstack/vue-start-client`
- `@tanstack/vue-start-server`

These e2e suites are added as well
 `e2e/vue-start/basic` 
 `e2e/vue-start/query-integration` (ssr vue-query)

#### Vue SSR Deferred Script - few changes in router-core

Vue lacks React's `suppressHydrationWarning`. When TSR's `$_TSR.c()` runs during HTML parsing, it removes script elements before Vue hydrates. Vue sees the mismatch and errors. The changes to router-core are necessary to suppress these warnings in vue, to get behavior similar to react.

Specifically, these small changes to router-core in `tsrScript.ts` and `ssr-client.ts`:

1. `$_TSR_DEFER` flag - Vue sets this before TSR scripts run
2. `c()` returns early if flag is set
3. `cleanup()` - Vue calls this after hydration to do the actual cleanup

React/Solid are not impacted, because they never set `$_TSR_DEFER`, so behavior is unchanged.


KazariEX:
> Vue lacks React's `suppressHydrationWarning`.

Could https://vuejs.org/api/ssr#data-allow-mismatch satisfy the use case?

PR #6070, opened 2025-12-11
fix: accumulate all parent context

### LINKED ISSUE #6068: Nested context from async beforeLoad can be unexpected undefined. causing errors in child routes

Router

### Describe the bug

Using parent route context that is computed from an `async beforeLoad` function can cause the context to resolve in a child route's `async beforeLoad` to undefined. This causes unexpected errors because the context type does not expect undefined to be a valid value.

A version of this issue was previously reported and considered solved (https://github.com/TanStack/router/issues/4998#issuecomment-3638522564).

https://stackblitz.com/edit/github-pv1k3znv?file=src%2Froutes%2Fposts.%24postId.tsx

refresh example on url https://mklilqjgazgithub-g3em--3000--365214aa.local-credentialless.webcontainer.io/posts/1

### Expected behavior

I would expect that context is stable when going down the route chain from root -> parent -> child, even for async beforeLoad logic. 

### Screenshots or Videos

_No response_

### Platform

- Router: v1.140.5
- OS: macOs
- Browser: Chrome
- Browser Version: latest
- Bundler: vite
- Bundler Version: latest

### Additional context

_No response_

schiller-manuel:
fixes #6068

## Summary by CodeRabbit

* **Refactor**
  * Consolidated internal context handling in the routing system to improve code consistency and reduce redundancy across multiple operations.

### DIFF
--- a/packages/router-core/src/load-matches.ts
+++ b/packages/router-core/src/load-matches.ts
@@ -50,6 +50,29 @@ const resolvePreload = (inner: InnerLoadContext, matchId: string): boolean => {
   )
 }
 
+/**
+ * Builds the accumulated context from router options and all matches up to (and optionally including) the given index.
+ * Merges __routeContext and __beforeLoadContext from each match.
+ */
+const buildMatchContext = (
+  inner: InnerLoadContext,
+  index: number,
+  includeCurrentMatch: boolean = true,
+): Record<string, unknown> => {
+  const context: Record<string, unknown> = {
+    ...(inner.router.options.context ?? {}),
+  }
+  const end = includeCurrentMatch ? index : index - 1
+  for (let i = 0; i <= end; i++) {
+    const innerMatch = inner.matches[i]
+    if (!innerMatch) continue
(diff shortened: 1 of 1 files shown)


packages/router-core/src/load-matches.ts
```diff
@@ -50,6 +50,29 @@ const resolvePreload = (inner: InnerLoadContext, matchId: string): boolean => {
...
+ * Builds the accumulated context from router options and all matches up to (and optionally including) the given index.
+ * Merges __routeContext and __beforeLoadContext from each match.
+ */
+const buildMatchContext = (
+  inner: InnerLoadContext,
+  index: number,
+  includeCurrentMatch: boolean = true,
+): Record<string, unknown> => {
+  const context: Record<string, unknown> = {
+    ...(inner.router.options.context ?? {}),
```
Sheraff, line 63: undefined is spreadable, we can skip creating an object here
```suggestion
    ...(inner.router.options.context),
```

PR #6092, opened 2025-12-13
feat: page-level hydration via `hydrate` route option

tannerlinsley:
## Summary
- Add **page-level hydration** control for TanStack Start via a new route `hydrate` option and router `defaultHydrate` option.
- **Not** React 18's "Selective Hydration" (priority-based subtree hydration). This is a different granularity: a route either hydrates on initial load or it doesn't.
- When hydration is disabled for the matched route tree, skip hydration-specific client assets while keeping SSR HTML/head output intact.
- Wired across **all three router adapters** (react-router, solid-router, vue-router) with shared logic in `router-core`.

## API
```tsx
// Route-level opt out
export const Route = createFileRoute("/legal/privacy")({
  hydrate: false,
  component: PrivacyPage,
})

// Router-level default
  routeTree,
  defaultHydrate: false,
})
```

## Behavior
- `hydrate` defaults to `true` unless overridden by `defaultHydrate`.
- If any matched route explicitly sets `hydrate: false`, the page does not hydrate on initial load.
- Conflicting explicit route settings (`false` + `true` in the same match) are detected; a DEV-only warning is emitted from both SSR and client paths (dedupes per router instance).
- With `hydrate: false`, initial-load hydration-specific scripts/preloads are excluded:
  - client entry script import/assets
  - modulepreload links
  - hydration payload bootstrap
- SSR output that should remain still renders:
  - route HTML
  - head metadata
  - non-hydration scripts/styles/assets

## Navigation behavior (this release)
`hydrate: false` controls **initial page load** only. Once the app has hydrated on any route in a tab, an SPA navigation into a `hydrate: false` route renders normally — the route does not unhydrate mid-session.

A future release will add a router option to configure this — for example, performing a hard navigation or document-replace when entering a `hydrate: false` route, which will also enable removing those routes' code from the main client bundle entirely. The current PR is forward-compatible: the future option lands additively, defaults preserve today's behavior, and no API in this PR locks us out of that direction.

If you need a hard navigation on a per-link basis today, `<Link to="/legal/privacy" reloadDocument>` is the escape hatch.

## Implementation Highlights
- **router-core**: typed `HydrateOption`, `defaultHydrate` router option, `hydrate` field on route options/matches, shared `getHydrateStatus` helper, and a new `ScriptFilter` hook on `RouterOptions`.
- **@tanstack/start-client-core**: the Start-specific strip helpers (`isClientEntryScript`, `stripClientEntryImport`, `clientEntryScriptFilter`). Router-only consumers don't ship these.
- **react-router, solid-router, vue-router**: each adapter's `Scripts` and head preload-links logic now consumes the shared `getHydrateStatus` + `router.options.scriptFilter` hook, filtering client-entry scripts and gating `serverBufferedScript` / modulepreload links when hydration is disabled.
- **start-server-core**: marks generated client-entry script tags with `data-tsr-client-entry` so adapter `Scripts` renderers can robustly identify and strip them. Auto-registers `clientEntryScriptFilter` in `createStartHandler.getRouter` (server) and `hydrateStart` (client) — zero authoring change for Start users.
- Router-only SSR consumers can opt in with `createRouter({ scriptFilter: clientEntryScriptFilter })` or supply their own filter.

## Tests
- Unit coverage in `packages/react-router/tests/Scripts.test.tsx` validates the filter mechanism.
- Strip-helper unit coverage lives in `packages/start-client-core/tests/clientEntryScript.test.ts`.
- E2E coverage mirrored across all three Start fixtures:
  - `e2e/react-start/basic/{src/routes,tests}/hydrate-{false,true}.tsx`
  - `e2e/solid-start/basic/{src/routes,tests}/hydrate-{false,true}.tsx`
  - `e2e/vue-start/basic/{src/routes,tests}/hydrate-{false,true}.tsx`

## Docs
- `docs/start/framework/react/guide/page-level-hydration.md`
- `docs/start/framework/solid/guide/page-level-hydration.md` (uses standard ref/replace pattern)
- Both linked from `docs/start/config.json`
- Covers: the `hydrate` option, `defaultHydrate`, inheritance, conflict detection, `ssr` combinations, a plain-JS interactivity example using `head().scripts` + event delegation, and the current navigation behavior (with forward-looking note on the upcoming nav-mode option).

## Compatibility
- No breaking changes; this is opt-in and existing behavior remains hydrated by default.


packages/start-server-core/src/router-manifest.ts
```diff
@@ -16,22 +16,33 @@ export async function getStartManifest() {
...
+  // Add client entry (will be filtered when hydrate: false)
   rootRoute.assets.push({
     tag: 'script',
     attrs: {
       type: 'module',
       async: true,
+      'data-tsr-client-entry': 'true',
     },
-    children: script,
+    children: `import('${startManifest.clientEntry}')`,
```
SeanCassiere: I think we could use a bit more commenting here, just so we explcitly point it out for anyone looking at this in the future, that when hydrate: false, then these scripts tagged with data-tsr-client-entry='true' will outright not be loaded into the DOM.
Also, for the default use-case, we should probably use the preview packages and init a few apps to make sure that HMR really is working just fine. Mostly, since the current flow of having the HMR script be synchronously imported in before even starting to bring in the client entry was something that got settled on (by Manuel) after a decent bit of trial and error.

packages/react-router/src/Scripts.tsx
```diff
@@ -57,14 +85,62 @@ export const Scripts = () => {
           nonce,
         },
         children,
-      })),
-    }),
+      }))
+
+      // If hydrate is false, remove client entry imports but keep React Refresh for HMR
+      if (!shouldHydrate) {
```
SeanCassiere: This bit, might make more sense for it to be function that lives in `router-core`, to then be consumed by React, Solid, and Vue.

packages/react-router/src/hydrate-status.ts
```diff
@@ -0,0 +1,30 @@
+export function getHydrateStatus(
```
SeanCassiere: Also to be bounced into `router-core`.

packages/react-router/src/Scripts.tsx
```diff
@@ -57,14 +85,62 @@ export const Scripts = () => {
...
+          .filter(Boolean) as Array<RouterManagedTag>
+      }
+
+      return allScripts
+    },
     structuralSharing: true as any,
   })
 
+  // Warn about conflicting hydrate options
+  if (hydrateStatus.hasConflict) {
```
schiller-manuel: we don't have such a warning for `ssr: false` etc., we just use the "least common denominator".
do we need this warning here really?
also, if we do, we should only emit in DEV

docs/start/framework/react/guide/page-level-hydration.md
```diff
@@ -0,0 +1,475 @@
...
+// Static server-rendered page (no JavaScript)
+ssr: true, hydrate: false
+// ✅ Renders on server ✅ Data loads on server ❌ NOT interactive
+
+// Client-only page
+ssr: false, hydrate: true
+// ❌ Renders on client ❌ Data loads on client ✅ Interactive on client
+
+// This combination doesn't make sense
+ssr: false, hydrate: false
```
schiller-manuel, line 51: we probably should warn if this combination happens?

docs/start/framework/react/guide/page-level-hydration.md
```diff
@@ -0,0 +1,475 @@
...
+- Server-side rendering
+- Loader data (during SSR only)
+- Meta tags from `head()`
+- External scripts from `head()`
+- CSS and stylesheets
+- Images and static assets
+
+## Inheritance
+
+A child route inherits the `hydrate` configuration of its parent. If **any route** in the match has `hydrate: false`, the entire page will not be hydrated:
```
schiller-manuel, line 286: should this be flippable via a config?

so that you can also express

"If **any route** in the match has `hydrate: true`, the entire page **will** be hydrated"

docs/start/framework/react/guide/page-level-hydration.md
schiller-manuel: since ssr and hydrate props can conflict, should we use a single config option to express both?

schiller-manuel: or should we make it configurable whether `ssr:false` wins over `hydrate:false`?

packages/react-router/src/hydrate-status.ts
```diff
@@ -0,0 +1,41 @@
...
+      | undefined
+
+    const routeHydrate = route?.options?.hydrate
+    const hydrateOption = routeHydrate ?? defaultHydrate
+
+    if (hydrateOption === false) {
+      hasExplicitFalse = true
+    } else if (hydrateOption === true && routeHydrate !== undefined) {
+      hasExplicitTrue = true
+    }
```
Sheraff: Could we skip some iterations w/ a `for` loop and a `break` here?

```suggestion
    if (hydrateOption === false) {
      hasExplicitFalse = true
    } else if (hydrateOption === true && routeHydrate !== undefined) {
      hasExplicitTrue = true
    }
    if (hasExplicitFalse && hasExplicitTrue) break
```

docs/start/framework/react/guide/page-level-hydration.md
tannerlinsley: The combo is documented as 'avoid this' but isn't enforced at runtime by design — `ssr: false` + `hydrate: false` produces an empty page (no SSR render, no client render), which is closer to user error than a partial conflict that needs guarding.

A DEV warning here would coexist with the existing hydrate-conflict warning, which now DEV-warns reachably from both SSR and client paths (see 11b487551a). Happy to add a runtime warning specifically for the `ssr:false + hydrate:false` combo too if you'd prefer — it's a small follow-up. Resolving for now since the doc note is in place.

docs/start/framework/react/guide/page-level-hydration.md
tannerlinsley: `defaultHydrate` covers the inverse case: set `defaultHydrate: false` globally, then opt routes into hydration with `hydrate: true`.

The current rule ('any matched `hydrate: false` wins') is also the safer direction — a route author who explicitly opts out shouldn't be silently overridden by a child. The DEV conflict warning (now reachable on SSR too) surfaces any unintended mix.

If real demand emerges for the flipped semantics, a future opt-in flag can land additively without breaking existing behavior. Resolving for now.

PR #6121, opened 2025-12-17
fix: generator perf and pathless routes relative to parent 

schiller-manuel:
## Summary by CodeRabbit

* **Refactor**
  * Improved route-generation internals with precompiled pattern matching and a prefix-based lookup to speed and stabilize parent/child resolution.
  * Consolidated route file processing and route-tree generation for more consistent path, root-piece, and update handling.

* **Chores**
  * Safer and clearer path/token transformation utilities and reduced redundant computations.

* **Tests**
  * Expanded unit tests to cover new prefix-based lookup and sorting behaviors.

### DIFF
--- a/packages/router-generator/src/utils.ts
+++ b/packages/router-generator/src/utils.ts
@@ -295,28 +295,28 @@ const splitPathRegex = /[/-]/g
 const leadingDigitRegex = /^(\d)/g
 
 const toVariableSafeChar = (char: string): string => {
-    if (alphanumericRegex.test(char)) {
-      return char // Keep alphanumeric characters and underscores as is
-    }
-
-    // Replace special characters with meaningful text equivalents
-    switch (char) {
-      case '.':
-        return 'Dot'
-      case '-':
-        return 'Dash'
-      case '@':
-        return 'At'
-      case '(':
-        return '' // Removed since route groups use parentheses
-      case ')':
-        return '' // Removed since route groups use parentheses
-      case ' ':
-        return '' // Remove spaces
-      default:
-        return `Char${char.charCodeAt(0)}` // For any other characters
-    }
+  if (alphanumericRegex.test(char)) {
+    return char // Keep alphanumeric characters and underscores as is
   }
+
+  // Replace special characters with meaningful text equivalents
+  switch (char) {
+    case '.':
+      return 'Dot'
+    case '-':
+      return 'Dash'
+    case '@':
+      return 'At'
+    case '(':
+      return '' // Removed since route groups use parentheses
+    case ')':
+      return '' // Removed since route groups use parentheses
+    case ' ':
+      return '' // Remove spaces
+    default:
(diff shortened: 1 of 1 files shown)


packages/router-generator/src/utils.ts
```diff
@@ -143,13 +280,23 @@ export function determineInitialRoutePath(
...
+const trailingSplatRegex = /\$$/g
+const bracketSplatRegex = /\$\{\$\}/g
+const dollarSignRegex = /\$/g
+const splitPathRegex = /[/-]/g
+const leadingDigitRegex = /^(\d)/g
+
 export function routePathToVariable(routePath: string): string {
   const toVariableSafeChar = (char: string): string => {
-    if (/[a-zA-Z0-9_]/.test(char)) {
+    if (alphanumericRegex.test(char)) {
```
Sheraff: I haven't benchmarked it, but I keep seeing *seemingly performant* code use code point comparisons instead of regexes for this kind of stuff. Like
```suggestion
    const code = char.charCodeAt(0)
    if (code > 123 && code < 456) {
```

packages/router-generator/src/utils.ts
```diff
@@ -172,25 +319,35 @@ export function routePathToVariable(routePath: string): string {
     }
   }
```
Sheraff: Very nitpicky but if we're chasing performance, I believe moving this pure function outside of the scope of `routePathToVariable` would allow it to be JIT optimized faster

packages/router-generator/src/utils.ts
```diff
@@ -361,86 +518,54 @@ export function removeLastSegmentFromPath(routePath: string = '/'): string {
...
+
+/**
+ * Extracts non-nested segments from a route path.
+ * Used for determining parent routes in non-nested route scenarios.
+ */
+export function getNonNestedSegments(routePath: string): Array<string> {
+  nonNestedSegmentRegex.lastIndex = 0
+  const matches = [...routePath.matchAll(nonNestedSegmentRegex)]
+  const result: Array<string> = []
+  for (const match of matches) {
```
Sheraff: I'm not sure you need to copy this into an array, the result of `.matchAll` should be iterable

packages/router-generator/src/utils.ts
```diff
@@ -361,86 +518,54 @@ export function removeLastSegmentFromPath(routePath: string = '/'): string {
...
+    const beforeStr = routePath.substring(0, match.index)
+    openBracketRegex.lastIndex = 0
+    closeBracketRegex.lastIndex = 0
+    const openBrackets = (beforeStr.match(openBracketRegex) || []).length
+    const closeBrackets = (beforeStr.match(closeBracketRegex) || []).length
```
Sheraff:
```suggestion
    const openBrackets = beforeStr.match(openBracketRegex)?.length ?? 0
    const closeBrackets = beforeStr.match(closeBracketRegex)?.length ?? 0
```

PR #6172, opened 2025-12-21
fix(examples): fix paraglide middleware usage in start-i18n-paraglide example

jasperfue:
This PR fixes an issue where the `start-i18n-paraglide` example causes infinite 307 redirect loops in production environments (e.g., Netlify) or when cookie strategies are involved.
**The Issue:**
The `paraglideMiddleware` automatically de-localizes the URL (e.g., `/en/about` -> `/about`) in the `request` object passed to its callback.
However, the example also configures a `rewrite` in the TanStack Router:
```ts
rewrite: {
  input: ({ url }) => deLocalizeUrl(url),
  // ...
}
```

By passing the *already de-localized* request from the middleware to `handler.fetch(request)`, TanStack Start receives a URL that conflicts with its internal routing/rewrite logic during SSR. This leads to mismatching states where the middleware tries to enforce cookies/redirects while the router tries to handle the URL, resulting in a redirect loop.

**The Fix:**
We strictly use `paraglideMiddleware` to set the `AsyncLocalStorage` context (for translations), but we pass the **original** `req` to `handler.fetch`. This allows TanStack Router to handle the URL rewriting via its configured `rewrite` property as intended.

**Note:**
- The link to the source code in the README (`.../paraglide-js/examples/tanstack-start`) also seems to be broken/404, but fixing this inline code snippet is the most important part for users getting started.
- I noticed the link to the paraglide monorepo example is dead.

## Release Notes

* **Documentation**
  * Improved integration guides for Paraglide with Vite and TanStack Start frameworks, featuring comprehensive configuration examples and best practices.
  * Introduced URL rewriting examples showcasing enhanced localization capabilities.
  * Updated server middleware integration documentation with refined implementation patterns.

### DIFF
--- a/examples/react/start-i18n-paraglide/README.md
+++ b/examples/react/start-i18n-paraglide/README.md
@@ -1,6 +1,6 @@
 # TanStack Start example with Paraglide
 
-This example shows how to use Paraglide with TanStack Start. The source code can be found [in the Paraglide monorepo](https://github.com/opral/monorepo/tree/main/inlang/packages/paraglide/paraglide-js/examples/tanstack-start).
+This example shows how to use Paraglide with TanStack Start.
 
 - [TanStack Router Docs](https://tanstack.com/router)
 - [Paraglide Documentation](https://inlang.com/m/gerre34r/library-inlang-paraglideJs)


schiller-manuel:
i need to look closer into that, this might just be a bug in router instead the example being incorrect

jasperfue:
@schiller-manuel I thought about that too, but wasn't sure.

schiller-manuel:
thank you for providing the correct solution here.
closing this PR in favor of https://github.com/TanStack/router/pull/6211

PR #6197, opened 2025-12-23
fix(react-router): HMR not updating inline arrow function components

### LINKED ISSUE #6195: HMR updates not applied for __root.tsx changes (requires full reload)

Router

### Describe the bug

HMR updates are sent but not applied for changes to `__root.tsx`. Vite logs show the HMR update is sent (`hmr update /src/routes/__root.tsx`), but the browser does not reflect the changes. A full page reload is required.

See below

1. Create fresh app with `npx create-tsrouter-app@latest --template file-router`
2. Run `npm run dev`
3. Open http://localhost:3000 in browser
4. Edit `src/routes/__root.tsx` - add `<h1>Test</h1>` before `<Outlet />`
5. Save the file
6. Observe: Vite logs `hmr update /src/routes/__root.tsx` but page does not update
7. Manually refresh - now the change appears

### Expected behavior

Changes to `__root.tsx` should hot-reload without requiring a full page refresh, similar to other route files.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.132.0
- OS: macOS
- Browser: Chrome
- Browser Version: 143.0.0.0
- Bundler: Vite
- Bundler Version: 7.3.0

### Additional context

_No response_

longfellowone:
## Summary

Fixes #6195

Fixes HMR (Hot Module Replacement) not updating the UI when route components are defined as inline arrow functions. This regression appeared around v1.128.3 (worked in v1.128.0).

### Problem

When a route uses an inline arrow function for its `component` option (common in file-based routing), React Refresh cannot register the component for HMR updates. The router's HMR handler compensates by calling `router.invalidate()` to trigger updates, but the `MatchInner` component's `useRouterState` selector did not include the `invalid` field. This meant React didn't detect any state change and didn't re-render, so the UI stayed on the old component even though HMR was triggered.

Named function components still work because React Refresh can register them directly, bypassing this issue.

### Solution

Add `invalid: match.invalid` to the `MatchInner` component's `useRouterState` selector. This ensures React detects the state change when `invalidate()` is called and re-renders the component tree, picking up the new component reference.

### Changes
- `packages/react-router/src/Match.tsx`: Add `invalid` to the selector (1 line)
- `packages/react-router/tests/router.test.tsx`: Add regression test that:
  - Renders component v1
  - Swaps to component v2 (simulating HMR)
  - Calls `invalidate()`
  - Verifies new component v2 is rendered

## Test plan

- [x] Verified test fails without fix (shows "Version 1" after swap)
- [x] Verified test passes with fix (shows "Version 2" after swap)
- [x] All 816 existing tests pass
- [x] Manually tested HMR in `examples/react/basic-file-based` with inline arrow function

## Summary by CodeRabbit

* **New Features**
  * The match object now exposes an `invalid` property for improved routing state visibility.

* **Tests**
  * Added regression tests simulating development invalidation to confirm swapped components render correctly.
  * Added tests verifying render and loader invocation counts during invalidation with and without async loaders to ensure minimal, correct re-renders.

### DIFF
--- a/packages/react-router/tests/router.test.tsx
+++ b/packages/react-router/tests/router.test.tsx
@@ -1583,7 +1583,9 @@ describe('invalidate', () => {
     render(<RouterProvider router={router} />)
     await act(() => router.load())
 
-    expect(await screen.findByTestId('test-component')).toHaveTextContent('Version 1')
+    expect(await screen.findByTestId('test-component')).toHaveTextContent(
+      'Version 1',
+    )
     const initialCallCount = renderTracker.mock.calls.length
 
     // Simulate HMR: swap component (keep using same tracker)
@@ -1594,7 +1596,9 @@ describe('invalidate', () => {
 
     await act(() => router.invalidate())
 
-    expect(await screen.findByTestId('test-component')).toHaveTextContent('Version 2')
+    expect(await screen.findByTestId('test-component')).toHaveTextContent(
+      'Version 2',
+    )
 
     // Count renders after invalidate
     const totalCalls = renderTracker.mock.calls.length
@@ -1633,9 +1637,7 @@ describe('invalidate', () => {
         const loaderData = testRoute.useLoaderData()
         renderTracker('v1', loaderData)
         return (
-          <div data-testid="test-component">
-            Version 1 - {loaderData.data}
-          </div>
+          <div data-testid="test-component">Version 1 - {loaderData.data}</div>
         )
       },
     })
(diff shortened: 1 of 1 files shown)


schiller-manuel:
how often is the new component then rendered then after invalidation?
should only be rendered once.

longfellowone:
The component renders exactly **once** after `invalidate()`.

I've added two render count verification tests that prove this:
- `renders minimal times after invalidate without loader` - asserts exactly 1 render
- `renders minimal times after invalidate with async loader` - asserts exactly 1 render (loader also called once)

The tests track render calls across the component swap and verify only 1 render occurs after `invalidate()`. React's automatic batching handles the `invalid: false→true→false` transitions efficiently.

You can see the test output in CI or run locally:
```bash
pnpm test:unit --run -t "render count verification"
```

schiller-manuel:
please remove the console logs

schiller-manuel:
about the async loader test: component does not consume the loader data, so it shouldn't rerender because of loader finishing. please modify the test to consume the loader data and also let the loader return new data the second time. then check if new component has new loader data

PR #6157, opened 2025-12-20
fix(router-core): enhance basePath handling

nlynzaad:
This PR addresses issues raised in #5200 and #5202.

It correctly sets the href using the rewrite url path and adds additional logic in the output section of the rewrite to check for paths that is already prefixed with the base path.

We also add an e2e test on router to test for redirects without reloadDocument = true.

This PR replaces #5202 and resolves #5200

## Summary by CodeRabbit

* **New Features**
  * Added a "/redirect" route and a UI button that navigates to it; visiting "/redirect" immediately redirects to the About page.

* **Bug Fixes**
  * Programmatic redirects now respect the app basepath; location construction aligned with rewritten paths.

* **Refactor**
  * Centralized basepath handling to improve consistency.

* **Tests**
  * Added and updated end-to-end tests covering basepath, redirect behavior, and navigation assertions.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #6205, opened 2025-12-23
fix(router-core): make navigate publicHref Aware

nlynzaad:
This PR addresses issues raised in #5200 and #5202.

Following #6201 this replaces #6157 as this PR makes better use of the distinctions between href and publicHref.

We also add an e2e test on router to test for redirects with and without reloadDocument = true.

## Summary by CodeRabbit

* **New Features**
  * Added two new redirect routes for quick navigation (/redirect and /redirectReload).
  * Support for redirects that optionally trigger a full page reload.
  * Navigation now accepts an optional public-facing URL to handle rewritten or absolute navigations.

* **Tests**
  * Added end-to-end tests verifying redirect behavior and basepath handling (with and without full reload).

### DIFF
--- a/packages/router-core/src/link.ts
+++ b/packages/router-core/src/link.ts
@@ -353,6 +353,7 @@ export interface NavigateOptionProps {
    * @link [API Docs](https://tanstack.com/router/latest/docs/framework/react/api/router/NavigateOptionsType#href)
    */
   href?: string
+  /** @internal */
   publicHref?: string
 }
 


packages/router-core/src/link.ts
```diff
@@ -353,6 +353,7 @@ export interface NavigateOptionProps {
    * @link [API Docs](https://tanstack.com/router/latest/docs/framework/react/api/router/NavigateOptionsType#href)
    */
   href?: string
+  publicHref?: string
```
schiller-manuel, line 357: we should not add this here because it's just an internal prop

packages/router-core/src/router.ts
```diff
@@ -2029,20 +2029,35 @@ export class RouterCore<
    *
    * @link https://tanstack.com/router/latest/docs/framework/react/api/router/NavigateOptionsType
    */
-  navigate: NavigateFn = async ({ to, reloadDocument, href, ...rest }) => {
-    if (!reloadDocument && href) {
+  navigate: NavigateFn = async ({
```
schiller-manuel, line 2032: can we add the publicHref somehow else as a prop here so we can use it but dont show it in the public API?

nlynzaad: marked it as internal. checked the .d.ts and confirmed with preview builds its not available in the public facing types.

packages/router-core/src/link.ts
nlynzaad: marked it as internal. checked the .d.ts and confirmed with preview builds its not available in the public facing types.

Issue #6152, opened 2025-12-19
staticFunctionMiddleware does not respect basePath configuration


schiller-manuel:
### Discussed in https://github.com/TanStack/router/discussions/6146

<div type='discussions-op-text'>

<sup>Originally posted by **busy-dog** December 19, 2025</sup>
## Description of the Issue

staticFunctionMiddleware defaults to mounting __tsr/staticServerFnCache at the root path /__tsr/...
Even if base or basePath is set in SSR or Router configuration, the URLs generated by the middleware still access from the root path, leading to 404 errors when accessing resources.

## Reproduction Steps

Configure vite.config.ts, set base: '/docs/', and deploy to Ng with corresponding resource routing configured.
Open the browser and navigate to any route. The chromedevtools network shows /__tsr/staticServerFnCache/xxx.json 404.

Expected behavior
The actual request should be to /docs/__tsr/staticServerFnCache/xxx.json

## Possible Solution
Add a parameter to the getStaticCacheUrl method in the staticFunctionMiddleware.ts file of start-static-server-functions, or try reading the basePath to correct the fetch url
</div>

nktnet1:
Relevant PR: #5970.

Is there anything we can do to help push this issue forward? Thank you!

nktnet1:
If anyone's searching for a workaround in the meantime, see here:

- https://github.com/nktnet1/webview-kiosk/pull/251

PR #6215, opened 2025-12-25
fix(start-plugin): allow trailing slash in basepath when pre-rendering

### LINKED ISSUE #6023: Prerender fails when using custom base path - attempts to fetch / instead of configured base path

Start

### Describe the bug

When configuring a custom base path in Vite along with TanStack Router's basepath, the prerenderer still attempts to crawl from / instead of the configured base path, resulting in a "Failed to fetch /: Not Found" error during build.

https://github.com/busy-dog/tanstack-start-demo

1. Configure a custom base path in vite.config.ts:

```ts title=vite.config.ts
const BASEPATH = '/gitlab/34e28e1e78c44601b0974c2097b85a27/caelum-doc-dist/';

export default defineConfig({
  base: BASEPATH,
  define: {
    BASEPATH: JSON.stringify(BASEPATH),
  },
  plugins: [
    tanstackStart({
      spa: {
        enabled: true,
        prerender: {
          outputPath: 'index.html',
          enabled: true,
          crawlLinks: false,
        },
      },
      pages: [
        { path: `/docs` },
        { path: `/api/search` },
      ],
    }),
  ],
});
```

2. Configure router with basepath in router.tsx:
```ts title=router.tsx
export function getRouter() {
  return createTanStackRouter({
    routeTree,
    basepath: BASEPATH,
    // ... other options
  });
}
```
3. Run vite build

Build fails during prerender phase with:

```bash
[prerender] Crawling: /
error: Failed to fetch /: Not Found
  at <anonymous> (/node_modules/@tanstack/start-plugin-core/dist/esm/prerender.js:126:23)
```

### Expected behavior

The prerenderer should respect the configured base path and attempt to fetch from /gitlab/34e28e1e78c44601b0974c2097b85a27/caelum-doc-dist/ instead of /.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.139.14
- OS: macOS
- Browser: Chrome
- Browser Version: 141
- Bundler: vite
- Bundler Version: 7.2.6

### Additional context

_No response_

nlynzaad:
This PR resolves #6023 by ensuring trailing slashes is stripped from the base-url setting in vite before pre-rendering is executed.

Also added pre-rendering modes to react-start/custom-basepath with and without trailing slashes in the base url.

## Summary by CodeRabbit

* **New Features**
  * Added prerender build variants, including optional trailing-slash mode.
  * E2E runs now cover SSR, prerender, and prerender-with-trailing-slash.
  * Test runner now selects server mode at runtime and exposes mode/port env for tests.
  * App routes and API calls now derive the runtime base URL from the running router origin.

* **Chores**
  * E2E scripts reorganized into modular start/stop and mode-specific commands.
  * Tests include a mode-aware utility and updated redirect assertions for basepath changes.

### DIFF
--- a/packages/router-core/src/rewrite.ts
+++ b/packages/router-core/src/rewrite.ts
@@ -26,19 +26,21 @@ export function rewriteBasepath(opts: {
   basepath: string
   caseSensitive?: boolean
 }) {
-  const normalizePathCase = (path: string) => {
-    return opts.caseSensitive ? path : path.toLowerCase()
-  }
-
   const trimmedBasepath = trimPath(opts.basepath)
   const normalizedBasepath = `/${trimmedBasepath}`
   const normalizedBasepathWithSlash = `${normalizedBasepath}/`
-  const checkBasepath = normalizePathCase(normalizedBasepath)
-  const checkBasepathWithSlash = normalizePathCase(normalizedBasepathWithSlash)
+  const checkBasepath = opts.caseSensitive
+    ? normalizedBasepath
+    : normalizedBasepath.toLowerCase()
+  const checkBasepathWithSlash = opts.caseSensitive
+    ? normalizedBasepathWithSlash
+    : normalizedBasepathWithSlash.toLowerCase()
 
   return {
     input: ({ url }) => {
-      const pathname = normalizePathCase(url.pathname)
+      const pathname = opts.caseSensitive
+        ? url.pathname
+        : url.pathname.toLowerCase()
 
       // Handle exact basepath match (e.g., /my-app -> /)
       if (pathname === checkBasepath) {
@@ -50,12 +52,7 @@ export function rewriteBasepath(opts: {
       return url
     },
     output: ({ url }) => {
-      const pathname = normalizePathCase(url.pathname)
-      const base = pathname.startsWith(checkBasepathWithSlash)
-        ? ''
-        : trimmedBasepath
-
-      url.pathname = joinPaths(['/', base, url.pathname])
+      url.pathname = joinPaths(['/', trimmedBasepath, url.pathname])
       return url
     },
   } satisfies LocationRewrite


e2e/react-start/custom-basepath/package.json
```diff
@@ -6,9 +6,16 @@
...
+    "build:prerender": "MODE=prerender vite build && tsc --noEmit",
+    "build:prerender:trailing": "MODE=prerender TRAILING_SLASH=true vite build && tsc --noEmit",
     "preview": "vite preview",
     "start": "tsx express-server.ts",
-    "test:e2e": "rm -rf port*.txt; playwright test --project=chromium"
+    "test:e2e:startDummyServer": "node -e 'import(\"./tests/setup/global.setup.ts\").then(m => m.default())' &",
+    "test:e2e:stopDummyServer": "node -e 'import(\"./tests/setup/global.teardown.ts\").then(m => m.default())'",
+    "test:e2e:prerender": "rm -rf port*.txt; MODE=prerender playwright test --project=chromium",
+    "test:e2e:prerender:trailing": "rm -rf port*.txt; MODE=prerender TRAILING_SLASH=true playwright test --project=chromium",
+    "test:e2e:ssrMode": "rm -rf port*.txt; playwright test --project=chromium",
```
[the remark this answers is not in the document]
nlynzaad: ci is in a unix enviroment and this is specific for e2e tests

nlynzaad:
e2e tasks are passing locally. need to investigate why these are failing in the CI

packages/start-server-core/src/createStartHandler.ts
```diff
@@ -210,6 +211,13 @@ export function createStartHandler<TRegister = Register>(
 
     try {
       const url = new URL(request.url)
+      const basePath = joinPaths(['/', ROUTER_BASEPATH])
+
+      if (!url.pathname.startsWith(basePath)) {
```
schiller-manuel, line 216: when would the url not start with the basepath? only during prerendering?
if yes, then i dont like the idea of mixing this into each request handling :/

nlynzaad: when running in vite preview for example the request url is not prepended with the custom basepath while SERVER_FN_BASE is. This is true in both prerender and non-prerender cases as long as the basepath is used.

Note using the express server this is not a problem since it is served correctly hence the check first to see if the the update needs to be applied.

nlynzaad: From the looks of the issue description this might be similar cause to #6152

edmundhung: This seems to assume the users app route will never match the base path. Would it be risky?

nlynzaad:
closing this in favour of #6304 that is a cleaner solution. It addresses the original issue and the problems with the preview server

PR #6093, opened 2025-12-13
fix: serial head execution

schiller-manuel:
fixes #6085
fixes #4785
fixes #4889

## Summary by CodeRabbit

* **Bug Fixes**
  * Router waits for all loaders to settle, defers applying head metadata until after loaders complete, logs head errors without stopping other heads, and improves redirect and 404 handling.
* **Chores**
  * Removed the not-found "via-head" route and its UI links; removed related prerender exclusion entries.
* **Tests**
  * E2E not-found tests narrowed to loader-only thrower scenarios (removed head thrower).

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #6222, opened 2025-12-25
fix: wait for async loaders before executing head functions

### LINKED ISSUE #6221: Route head function can still execute before loader function finishes

Router

### Describe the bug

A follow-up of PR #6093 - In certain scenarios, `head` function can still run before `loaderData` is ready, resulting in incorrect head computation, see an example below.

n/a

Say we have a blog site, and accessing an article requires auth. Let's start with the user being logged out.

1. Visit `/article/123`, shows "Not found" page content, and a generic title.
2. Log in, and for the sake of argument, the user is redirected to `/dashboard`. 
3. Click browser's "Back" button, i.e. navigate back to `/article/123`.

### Expected behavior

Now we can see an inconsistent state: the document body shows the article content correctly, but we still get the generic title, instead of the article's title, as we expect to see.

The reason is, the `head` function, which is supposed to return the article title from its `loaderData`, runs before the loader function can actually finish, and thus, `loaderData` is not available at the time of execution. (On the other hand, the article content shows correctly, because `Route.useLoaderData()` works as expected.)

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.143.7
- OS: macOS
- Browser: Chrome
- Browser Version: 144.0
- Bundler: Vite
- Bundler Version: 7.3.0

### Additional context

_No response_

yanghuidong:
Fixes #6221 

The bug: When loaders run asynchronously (stale-while-revalidate), loadRouteMatch returns immediately while the loader runs in the background. Promise.allSettled(inner.matchPromises) waits for loadRouteMatch promises, not the actual loader completion, causing head() to execute with undefined loaderData.

The fix: After loadRouteMatch promises settle, explicitly wait for all loaderPromises to complete before executing head functions. This ensures loaderData is available when head() executes.

Reproduction scenario:
1. Navigate to authenticated route (e.g., /article/123)
2. Delete auth cookie, reload (shows 'not found')
3. Login, redirect to dashboard
4. Click back button to /article/123
   - Before fix: Article loads but title shows fallback (loaderData undefined)
   - After fix: Article loads with correct title (loaderData available)

Fixes the issue identified in PR #6093 follow-up investigation.

## Summary by CodeRabbit

* **New Features**
  * Staged load/head handling to avoid stale async results and ensure page metadata reflects the latest navigation
  * New example app with routing, head metadata, error/not-found UI, client auth helpers, and dev/preview modes
  * Playwright multi-mode e2e harness and local test server for SPA/SSR/prerender/preview

* **Bug Fixes**
  * Improved cleanup and guards when navigation occurs during async loaders to prevent stale state and redundant head runs

* **Chores**
  * Updated e2e start script invocation and added test-results to ignore list

* **Tests**
  * Comprehensive e2e tests validating head() behavior with async loaders (SSR-focused)


packages/router-core/src/load-matches.ts
```diff
@@ -900,6 +900,16 @@ export async function loadMatches(arg: {
     // Use allSettled to ensure all loaders complete regardless of success/failure
     const results = await Promise.allSettled(inner.matchPromises)
 
+    // Wait for async loaders to complete before executing head functions
+    // loadRouteMatch may return immediately while loaders run asynchronously in the background
+    // We need to wait for the actual loaderPromise, not just the loadRouteMatch promise
+    await Promise.all(
```
schiller-manuel: we definitely don't want to block for async loaders.

yanghuidong: I guess it depends on what we're trying to do in the head function, right? For setting the document title, blocking until the loaderData is ready at least gets us the correct title, while not blocking, under certain scenarios, can get us nothing. Also, for page title, is blocking that bad?

Or do you see a non-blocking way to ensure page title is returned reliably? Thanks.

schiller-manuel: we probably need to detect if we have async loaders and then schedule to serially re-run all head() functions

yanghuidong: Are you saying, we let the head functions run a first-pass, non-blocking (current behavior), and then we check if there're any remaining unresolved loader promises, and if yes, then we await them (`Promise.all`), and then run the head functions for the second time?

schiller-manuel: possibly yes. the second run of head() must also run detached like the async loaders itself do

yanghuidong: made an attempt, let me know if that could work?

packages/router-core/src/load-matches.ts
```diff
@@ -919,24 +940,21 @@ export async function loadMatches(arg: {
...
+    const asyncLoaderPromises = inner.matches
+      .map((match) => {
+        const currentMatch = inner.router.getMatch(match.id)
+        return currentMatch?._nonReactive.loaderPromise
+      })
+      .filter(Boolean)
+
+    if (asyncLoaderPromises.length > 0) {
+      // Schedule re-execution after all async loaders complete (non-blocking)
+      Promise.all(asyncLoaderPromises).then(() => executeAllHeadFns(inner))
```
schiller-manuel: can the detached loaders throw? if yes, should we execute the head chain? probably yes

schiller-manuel: we should probably store a reference to this promise so we can override it in case a new round of loaders has already started before the detached head() executed.

what if this happens while the detached head() runs?

schiller-manuel:
we need tests for this. unit tests and e2e test

packages/router-core/src/load-matches.ts
yanghuidong: Comparing location before executing the head seems to work? I tested with a slow fetcher, and navigate away before it can finish, and that check prevents the head function from executing. 

Now if the head already starts executing, and we navigate to a new route, then the new head will execute next, and override the old value, hopefully? Not sure. Will think about it more after sleep. Insights/advice welcome as always.

yanghuidong:
> we need tests for this. unit tests and e2e test

There's one e2e test in a local branch that I'm using, is there any guideline for where to place it, and how to name it? It's currently called `e2e/solid-start/test-head-bug`, will that work? Should I merge it into here sooner, or later? Or even in a separate PR?

packages/router-core/src/load-matches.ts
schiller-manuel: comparing location is not enough IMO as you can stay on the same location and invalidate and trigger loaders.

schiller-manuel:
please use an existing e2e project (e2e/react-start/basic for example) and add new routes in there and playwright tests

schiller-manuel:
can you please move this into e2e/react-start/basic?
this issue should not require query

yanghuidong:
> can you please move this into e2e/react-start/basic? this issue should not require query

Sure, but solid-start work as well, right? I don't know react, sorry.
I'm current trying to fix the PR / Test failures, if I'm stuck for longer, might need your help :)

yanghuidong:
Still getting PR test failures:

```
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 2 ⎯⎯⎯⎯⎯⎯⎯

 FAIL   @tanstack/solid-router  tests/link.test.tsx > Link > when navigation to . from /posts while updating search from /
Error: expect(element).toHaveTextContent()

Expected element to have text content:
  Page: 2
Received:
  Page: 1
 ❯ tests/link.test.tsx:792:25
    790|     expect(window.location.search).toBe('?page=2&filter=inactive')
    791| 
    792|     expect(updatedPage).toHaveTextContent('Page: 2')
       |                         ^
    793|     expect(updatedFilter).toHaveTextContent('Filter: inactive')
    794|   })

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/2]⎯

 FAIL   @tanstack/solid-router  tests/store-updates-during-navigation.test.tsx > Store doesn't update *too many* times during navigation > redirection in preload
AssertionError: expected 2 to be 7 // Object.is equality

- Expected
+ Received

- 7
+ 2

 ❯ tests/store-updates-during-navigation.test.tsx:159:21
    157|     // Any change that increases this number should be investigated.
    158|     // Note: Solid has different update counts than React due to diffe…
    159|     expect(updates).toBe(7)
       |                     ^
    160|   })
    161| 

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[2/2]⎯


 Test Files  2 failed | 43 passed (45)
      Tests  2 failed | 761 passed | 1 skipped (764)
```

Wondering if they were caused by introducing the re-run of head functions (`Promise.allSettled`)?

schiller-manuel:
yes this is caused by rerunning head. needs to adapt the counts for the scenarios where a detached loader runs

schiller-manuel:
looking at the e2e test it might make sense to create a new separate e2e project for this specific topic

schiller-manuel:
i have a different impl instead of loadGeneration here: https://github.com/yanghuidong/tanstack-router/pull/1

this uses a pattern we already use in router

@Sheraff what do you think?

yanghuidong:
EDIT: the current main d2b3fa184b84d09b4e10308d137206a556cda41e (upstream) passes all tests: https://cloud.nx.app/runs/5N4Ica413A

On this PR branch, currently failing 1/515 tests: https://cloud.nx.app/runs/LCQ2XHmybf

However, when I check out that passing commit locally, and run the single test, it also fails, consistently, suggesting a flaky situation? Here's a summary by Claude. @schiller-manuel do you think this could be why?

```
  Test Failure on Commit d2b3fa184b84d09b4e10308d137206a556cda41e

  Commit Details:
  Commit: d2b3fa184b84d09b4e10308d137206a556cda41e
  Author: Birk Skyum <74932975+birkskyum@users.noreply.github.com>
  Date:   Tue Dec 30 11:51:26 2025 +0100
  Message: test(vue-start, solid-start): temp remove flaky raw stream tests (#6254)

  Test Environment:
  Node.js: v22.18.0
  pnpm: 10.24.0
  OS: macOS 12.5 (Darwin 21.6.0, x86_64)

  Commands Executed:
  git checkout d2b3fa184b84d09b4e10308d137206a556cda41e
  cd packages/solid-router
  pnpm vitest run link.test.tsx -t "when navigation to . from /posts while updating search from /" --reporter=verbose

  Test Result:
  FAIL  |@tanstack/solid-router| tests/link.test.tsx > Link > when navigation to . from /posts while updating search from /

  Error: expect(element).toHaveTextContent()

  Expected element to have text content:
    Page: 2
  Received:
    Page: 1
   ❯ tests/link.test.tsx:792:25
      790|     expect(window.location.search).toBe('?page=2&filter=inactive')
      791|
      792|     expect(updatedPage).toHaveTextContent('Page: 2')
         |                         ^
      793|     expect(updatedFilter).toHaveTextContent('Filter: inactive')
      794|   })

  Test Files  1 failed (1)
        Tests  1 failed | 1 passed | 112 skipped (114)
  Duration  4.76s

  Key Observation:
  The next test in the suite PASSES (line 912):
  ✓ when navigation to . from /posts while updating search from / and using base path (38ms)

  The only difference is that the passing test has an extra await screen.findByTestId('current-page') on line 903 that gives Solid's reactivity time to propagate the router state update.

  CI vs Local:
  This test appears to pass in CI (https://cloud.nx.app/runs/5N4Ica413A) but fails consistently locally, suggesting a timing-dependent race condition that manifests differently based on environment speed.
```

---

Looking into `test:pr` failures, one of them, afaik, is not caused by changes in this PR, as the test seems to also fail on the main branch:

```
 FAIL   @tanstack/solid-router  tests/link.test.tsx > Link > when navigation to . from /posts while updating search from /
Error: expect(element).toHaveTextContent()

Expected element to have text content:
  Page: 2
Received:
  Page: 1
 ❯ tests/link.test.tsx:792:25
    790|     expect(window.location.search).toBe('?page=2&filter=inactive')
    791| 
    792|     expect(updatedPage).toHaveTextContent('Page: 2')
       |                         ^
    793|     expect(updatedFilter).toHaveTextContent('Filter: inactive')
    794|   })
```

Confirm, or refute?

yanghuidong:
> i have a different impl instead of loadGeneration here: [yanghuidong#1](https://github.com/yanghuidong/tanstack-router/pull/1)
> 
> this uses a pattern we already use in router
> 
> @Sheraff what do you think?

Much simpler, the added router state having a much clearer purpose, just self-documenting overall, thanks for the teaching, I know you previously hinted at this, I just didn't understand back then.

yanghuidong:
> However, when I check out that passing commit locally, and run the single test, it also fails, consistently, suggesting a flaky situation?

@schiller-manuel could you provide some guidance on how to proceed with passing the PR test suite? whenever you have some extra time of course.

schiller-manuel:

PR #6265, opened 2026-01-01
docs(react/isr): use `getResponseHeaders` in example

FEgor04:
I've faced an issue when trying to use cache middleware from [ISR article](https://tanstack.com/start/latest/docs/framework/react/guide/isr#%23using-middleware-for-cache-headers).
Currently it uses `result.response.headers` to access response headers.
However it turns out that `result.response` is undefined, which causes, even though type annotations say that it is not. It seems like a bug, since TypeScript allows a code that leads to a runtime error.
See commit 6c6ec1b884284774ae2796513ce723171f418b6c and screenshot below.
image

Then I tried to use `result.headers`, which is not present in `RequestServerResult`. Setting them did not work either. See commit 3b417f685cae7804be0a31db8e0da7e2a9b51b39
image

In the end I've tried using `getResponseHeaders` and it worked! See commit 52fc400c50c2b7fb1b9c7df29a4816c36850dd88
I've updated the ISR article in the docs to use `getResponseHeaders`. Not sure what to with `result.response` being undefined

## Summary by CodeRabbit

* **Documentation**
  * Updated React Start Incremental Static Regeneration guide with improved middleware header management patterns. The revised examples demonstrate a refined approach to response header modification using utility functions, providing clearer implementation guidance while preserving existing functionality.


brenelz (approved):

Hardanish-Singh (approved):

LadyBluenotes:
Thanks for tracking this down, and sorry this sat for so long. The middleware runtime has changed since this was opened: current `main` now guarantees `response: Response` on request middleware `next()` results, and the focused Start server regression mutates `result.response.headers` and verifies that the header reaches the final response. That behavior landed through #7497, building on #5517.

The existing ISR example is valid again. Switching only this block to `getResponseHeaders()` would also leave the ETag and cache-monitoring examples on the same page using `result.response.headers`, so I think we can close this as no longer necessary. The reproduction was useful in exposing the original mismatch, so thank you for that!

PR #6276, opened 2026-01-02
fix(start-core): use headers.entries() instead of Object.entries

### LINKED ISSUE #5407: `setResponseHeaders` and `setResponseStatus` don't work in global middleware

Start

### Describe the bug

When setting response headers using `setResponseHeaders` or response status using `setResponseStatus` in a global middleware server handler, the headers and status are not set.

`setResponseHeader` works as expected.

https://github.com/nickprinsloo/tss-example

Given the following `start.ts`:

```
import { createMiddleware, createStart } from "@tanstack/react-start";
import { setResponseHeaders, setResponseStatus } from "@tanstack/react-start/server";

const requestMiddleware = createMiddleware().server(({ next }) => {
  const headers = new Headers()
  headers.set("X-Hello", "World")
  headers.set("X-Foo", "Bar")
  setResponseHeaders(headers)
  setResponseStatus(404)
  return next()
})

export const startInstance = createStart(() => {
  return {
    requestMiddleware: [requestMiddleware],
  };
});
```

I'd expect the custom headers + status to be present on the response.

### Expected behavior

As a user, I expect to see the headers and status of the response to be those set in the global middleware.

### Screenshots or Videos

_No response_

### Platform

- Start: ^1.132.48
- Node: v22.19.0
- macOS Tahoe

### Additional context

_No response_

FEgor04:
There was a bug when setting headers with `setResponseHeaders` did not change them. The root cause was that `Object.entires` returned empty array instead of actual headers. The fix is to use `headers.entries()` instead.
This PR covers the issue with e2e test and fixes it.

Fixes #5407

## Summary by CodeRabbit

* **New Features**
  * Global server middleware now supports setting custom response headers via new middleware capabilities.

* **Tests**
  * Added test coverage validating response header middleware functionality and behavior.

### DIFF
--- a/packages/start-server-core/src/request-response.ts
+++ b/packages/start-server-core/src/request-response.ts
@@ -208,7 +208,7 @@ export function setResponseHeaders(
   headers: TypedHeaders<ResponseHeaderMap>,
 ): void {
   const event = getH3Event()
-  for (const [name, value] of Object.entries(headers)) {
+  for (const [name, value] of headers.entries()) {
     event.res.headers.set(name, value)
   }
 }


FEgor04:
Didn't notice another PR for that issue. Will close that.

PR #6285, opened 2026-01-03
Fixes #6270

### LINKED ISSUE #6270: server function transformation plugin should exclude `node_modules`

Start

### Describe the bug

When using `@cloudflare/vite-plugin` instead of `target: "cloudflare-module"`, the SSR build fails with:

 ` [tanstack-start-core::server-fn:ssr] 'return' outside of function at position XXXXXX`

  The `[tanstack-start-core::server-fn:ssr]` transformation is parsing third-party library code in `node_modules` that it doesn't parse when using `target: "cloudflare-module"`. The parser encounters JavaScript patterns it can't handle.

  **Suggested fix:** The server function transformation plugin should exclude `node_modules` from processing, since server functions are defined in user code, not libraries. But to still allow any tanstack packages from node_modules to be transformed:

See PR https://github.com/TanStack/router/pull/6285 

https://github.com/jasongitmail/tanstack-xlsx-repro

```
git clone https://github.com/jasongitmail/tanstack-xlsx-repro.git
cd tanstack-xlsx-repro
bun install
```
See [steps in README](https://github.com/jasongitmail/tanstack-xlsx-repro?tab=readme-ov-file#reproduction).

### Expected behavior

SSR build should succeed with `@cloudflare/vite-plugin`, matching the behavior of `target: "cloudflare-module"`.

### Screenshots or Videos

_No response_

### Platform

```json
    "@tanstack/react-query": "5.83.0",
    "@tanstack/react-router": "1.144.0",
    "@tanstack/react-router-devtools": "^1.144.0",
    "@tanstack/react-router-with-query": "1.130.17",
    "@tanstack/react-start": "1.145.3",
    "@tanstack/router-core": "1.144.0",
    "@tanstack/router-devtools-core": "^1.144.0",
    "@tanstack/router-generator": "1.145.2",
    "@tanstack/router-plugin": "1.145.2",

    "@vitejs/plugin-react": "^4.6.0",
    "@vitejs/plugin-react-oxc": "^0.2.3",
    "vite": "^7.0.4",
    "vite-env-only": "^3.0.3",
    "vite-tsconfig-paths": "^5.1.4",
```

Mac 26.1

### Additional context

_No response_

jasongitmail:
I'm running this patch in production successfully.

Fixes #6270 

## Why this works TLDR

 The compiler must skip third-party node_modules (which may contain non-standard JavaScript that Babel can't parse), but must still transform @tanstack/* packages—which contain isomorphic functions whose server-only imports need to be dead-code-eliminated to prevent node:async_hooks from leaking into client bundles.
 
 ## Why this works 
 
The TanStack Start compiler transforms isomorphic functions by replacing server-only code paths and running dead code elimination. Without excluding node_modules, it also transforms third-party packages (like xlsx-js-style) which may contain non-standard JavaScript that Babel can't parse, causing build failures.

The negative lookahead (?!@tanstack) is critical: TanStack's internal packages (@tanstack/start-client-core, etc.) contain isomorphic functions with top-level imports from @tanstack/start-storage-context. These packages must still be transformed so the .server() callbacks and their node:async_hooks imports are eliminated from client bundles.

## Summary by CodeRabbit

* **Chores**
  * Improved the compiler plugin's code transformation logic to better handle dependency exclusion patterns, enhancing build processing during compilation.

### DIFF
--- a/packages/start-plugin-core/src/start-compiler-plugin/plugin.ts
+++ b/packages/start-plugin-core/src/start-compiler-plugin/plugin.ts
@@ -232,7 +232,7 @@ export function startCompilerPlugin(
       transform: {
         filter: {
           id: {
-            exclude: [new RegExp(`${SERVER_FN_LOOKUP}$`), /node_modules/],
+            exclude: [new RegExp(`${SERVER_FN_LOOKUP}$`), /node_modules\/(?!@tanstack)/],
             include: TRANSFORM_ID_REGEX,
           },
           code: {


jasongitmail:
I can add code rabbit's suggested comment if you'd like. Edits are also open to maintainers:

```ts
// Exclude third-party node_modules but allow @tanstack packages
// because they contain isomorphic functions that require transformation
exclude: [new RegExp(`${SERVER_FN_LOOKUP}$`), /node_modules\/(?!@tanstack)/],
```

schiller-manuel:
> The compiler must skip third-party node_modules 

i don't agree with this. you could very well have server functions installed as an npm package.
first of all, we need a reproducer for the non-transformable issue.
then we can look at how we can prevent the problem.
the parser should be able to parse any valid JS, so let's check whats going on there.
aside from that, we should make it configurable which paths / packages the compiler traverses. that's an API we need to think about

brenelz:
Yeah I kinda agree with @schiller-manuel that this doesn't seem ideal. Is there a specific dependency causing you to need this patch?

jasongitmail:
Closing. I think the include filter might handle this already. I removed the patch from my code and somehow it's working today without the reported issue...I'm not sure what coincidence of variables of dependency versions caused the earlier observed issue. Apologies. Will close the issue too.

PR #6293, opened 2026-01-04
Fix(router-core): normalize pipe characters in paths to prevent redirect loops

### LINKED ISSUE #6288: App breaks when route params contain pipe character(i.e "|")

Start

### Describe the bug

I am facing a bug where pipe symbol i.e "|" breaks app when navigating to a dynamic route that contains the symbol as param, e.g a string like "ER|2024FHTd"

This seems to only be happening in tanstack start, not router.

You can navigate to the route via link without Breaking how ever, if you reload page afterwards or try to navigate directly to it, app breaks

https://tanstack-start-param-issue.vercel.app/

1. Create a dynamic route in tanstack start app e.g $reference.tsx
2. Create a link or button that navigates to /$reference with a pipe symbol in the param value
3. Navigate to route via link/button - navigation works, 
4. Reload page after navigation or navigate directly to the page by entering the url - app breaks, shows ERR_TOO_MANY_REDIRECTS

### Expected behavior

Route should not break when navigating to a dynamic route with pipe

### Screenshots or Videos

Image

### Platform

- Start Version: [e.g. 1.121.0]
- macOS
- Browser: Chrome
- Browser Version: 143.0.7499.170
- Bundler: vite
- Bundler Version: 7.1.7

### Additional context

_No response_

jong-kyung:
fixes #6288

This PR fixes an issue where using a pipe character (`|`) in route parameters caused an infinite redirect loop (`ERR_TOO_MANY_REDIRECTS`)

#### The Problem
Browsers automatically encode `|` characters to `%7C` in the URL when navigating directly. However, the router's internal normalization logic (`cleanPath`) was treating `|` as a raw character.
This caused a mismatch between the current location (encoded by browser) and the expected location (unencoded by router), triggering a redirect loop where the router tried to redirect to the unencoded path, but the browser re-encoded it upon request.

#### The Fix
I updated the `cleanPath` utility function to explicitly normalize `|` characters to `%7C`. This ensures that the router's internal path representation matches the browser's encoding behavior, preventing the unnecessary redirect loop.

### Changes

- **packages/router-core/src/path.ts**: Updated `cleanPath` to replace `|` with `%7C`.
- **e2e/react-router/basic-file-based/tests/app.spec.ts**: Added an E2E test case to verify that routes with pipe characters in parameters render correctly without redirects.

## Summary by CodeRabbit

* **New Features**
  * Support for pipe character (|) in URL paths with automatic percent-encoding
  * New dynamic route exposed in navigation for testing/using pipe-containing paths

* **Tests**
  * Added end-to-end tests validating navigation, href encoding, and page rendering when route parameters include the pipe character

### DIFF
--- a/packages/router-core/src/path.ts
+++ b/packages/router-core/src/path.ts
@@ -19,7 +19,7 @@ export function joinPaths(paths: Array<string | undefined>) {
   )
 }
 
-/** Remove repeated slashes from a path string. */
+/** Remove repeated slashes and replace | to %7C from a path string. */
 export function cleanPath(path: string) {
   // remove double slashes
   return path.replace(/\/{2,}/g, '/').replace(/\|/g, '%7C')


nlynzaad:
@jong-kyung thanks for this. as part of my review process, I like to recreate the error with the e2e tests before validating the fix. I copied the e2e test over into main and the tests passed without a problem. I also tested this manually in the browser and could not replicate the problem.

We need to ensure the e2e test fails with the expected error and the solution resolves it correctly to ensure future breakages are avoided.

I do see the error in the linked issue but unfortunately, I don't see the source. Would it be possible to link to a github repo or stackblitz link where the error re-occurs?

nlynzaad:
I see now you aren't the author of the original issue. Do you possibly have a recreation of the original problem?

jong-kyung:
Hi @nlynzaad, thanks for reviewing

Apologies for the confusion regarding the E2E test. The test passed because I added it to a CSR-only. This issue (ERR_TOO_MANY_REDIRECTS) is specific to SSR environments (TanStack Start), caused by a mismatch between browser encoding and server-side normalization during direct navigation.

You can reproduce the issue using the links below:
- GitHub Repo: https://github.com/monarcode/tanstack-start-param-issue
- StackBlitz: https://stackblitz.com/~/github.com/monarcode/tanstack-start-param-issue

Could you please confirm the issue using the reproduction links? Once confirmed, please let me know, and I will rewrite the tests to properly target the SSR environment based on your feedback.

Thanks again for your time and review, and sorry for the confusion

jong-kyung:
Hi @nlynzaad, I wanted to follow up on this PR.

Since our last discussion, I’ve updated the tests to properly target the SSR environment (e2e/react-start/...) using the reproduction case.

I've currently kept the changes as fixup commits to make it easier for you to review the updates. Once you're satisfied with the changes, I'll squash them to clean up the commit history.

Looking forward to your feedback!

nlynzaad:
@jong-kyung thank you for the effort with this. 

I've been taking a deep dive into this issue to properly understand the root causes. The root issue here is the different manners in how certain characters are encoded/decoded in browsers and server side. One such character is the `|` sign which it seems chromium browsers specifically treats in a special manner, while everybody else seems to follow the WHATWG standard more closely. There is also a difference in how this is treated within search parameters vs when included in the path. 

These differences require us to ideally normalize this at the entry point into start so we can always deal with a standardized url irrespective of the source and server in use.

It is worth noting as well that `|` is not allowed in filenames on windows filesystems so we need to take that into account when we setup the e2e tests with the best option to add this as a virtual route.

After these findings I submitted a new PR #6389 that addresses the above more holistically.

jong-kyung:
> @jong-kyung thank you for the effort with this. 
> 
> 
> 
> I've been taking a deep dive into this issue to properly understand the root causes. The root issue here is the different manners in how certain characters are encoded/decoded in browsers and server side. One such character is the `|` sign which it seems chromium browsers specifically treats in a special manner, while everybody else seems to follow the WHATWG standard more closely. There is also a difference in how this is treated within search parameters vs when included in the path. 
> 
> 
> 
> These differences require us to ideally normalize this at the entry point into start so we can always deal with a standardized url irrespective of the source and server in use.
> 
> 
> 
> It is worth noting as well that `|` is not allowed in filenames on windows filesystems so we need to take that into account when we setup the e2e tests with the best option to add this as a virtual route.
> 
> 
> 
> After these findings I submitted a new PR #6389 that addresses the above more holistically. 

@nlynzaad Thank you for looking into this. I agree that addressing the root causes regarding browser/server encoding differences and standards is the better way

PR #6295, opened 2026-01-04
fix(react-start): Update useServerFn return type to be undefined after redirect

canac:
`useServerFn` returns a promise that resolves to `undefined` when the server function redirects. This PR updates the return type to reflect that.

## Summary by CodeRabbit

* **Improvements**
  * Mutation hooks now allow empty/void results; success callbacks and state updates run only when data is returned.
  * Server-facing hooks now accept void-returning flows (e.g., redirects), so caller return types can be a value or void.

### DIFF
--- a/e2e/react-start/basic-auth/src/hooks/useMutation.ts
+++ b/e2e/react-start/basic-auth/src/hooks/useMutation.ts
@@ -1,7 +1,7 @@
 import * as React from 'react'
 
 export function useMutation<TVariables, TData, TError = Error>(opts: {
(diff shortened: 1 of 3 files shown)


canac:
@schiller-manuel I hope you don't mind me pinging you. Do you have thoughts on this change? It will unfortunately add `| undefined` to the return type of callbacks returned by `useServerFn`, which I could see being annoying for server functions that don't redirect. But on the other hand, those callbacks currently return `undefined` when server functions redirect, causing runtime errors.

PR #6302, opened 2026-01-05
Parallel Routes RFC

tannerlinsley:


docs/rfcs/parallel-route-slots/README.md
```diff
@@ -0,0 +1,1508 @@
...
+
+The parent route discovers its slots after composition and can reference them type-safely in its component.
+
+### Global Slots (Children of Root)
+
+Global slots are `@slotName` files at the routes root - they become children of `__root.tsx`:
+
+```ts
+// @modal.tsx - defined in isolation, no parent reference needed
+export const Route = createSlotRoute({
```
beaussan: **question:** ‏I looked at the examples, and they don't have the same "id in function" as other routes, is that is an omission to keep it simple or the actual final api ?

If so, I would prefer to have consistency, and have the id in function paradigme too, something like this:
```suggestion
export const Route = createSlotFileRoute('/@modal')({
```

docs/rfcs/parallel-route-slots/README.md
```diff
@@ -0,0 +1,1508 @@
...
+export const Route = createFileRoute('/dashboard')({
+  component: Dashboard,
+})
+
+function Dashboard() {
+  return (
+    <div className="grid grid-cols-3 gap-4">
+      <aside>
+        {/* Type-safe: Route knows about @activity from composition */}
+        <Route.SlotOutlet name="activity" />
```
beaussan: **question:** ‏Will this also work with the `getRouteApi` function ? I think it should, but I don't see that present in the RFC

docs/rfcs/parallel-route-slots/README.md
```diff
@@ -0,0 +1,1508 @@
...
+- Works well with TypeScript
+
+**Cons:**
+
+- Different shapes depending on where you're reading from
+- Complex to type correctly
+
+---
+
+**Current leaning:** Option A (strip prefix) or Option C (dedicated hook) seem most ergonomic, but this needs more research and community input before deciding.
```
beaussan: **thought:** ‏I would tends towards option C (dedicated hook) because it also remove the complexities around conflict handling: validate search gets the the search without the prefix, runtime too, so if you change the parallel route name, you don't have to change your validation and runtime deeply

docs/rfcs/parallel-route-slots/README.md
```diff
@@ -0,0 +1,1508 @@
...
+
+1. **Search param access within slots**: How should slot components access their namespaced search params? See detailed options in the [Search Params](#open-question-search-param-access-within-slots) section. This affects DX significantly and needs community input.
+
+2. **Preloading**: How should `<Link slots={{ modal: {...} }} preload="intent">` work? Should it preload just the slot's route, or also affect main route preloading?
+
+3. **shouldRevalidate**: Should slots participate in the main route's revalidation cycle, or have completely independent revalidation?
+
+4. **Devtools**: How should the devtools visualize parallel slots? Separate trees? Merged view?
+
+5. **Code splitting**: Should slot route trees be lazy-loadable independently of the routes that render them?
```
beaussan: And can slots support the `.lazy()` api ? This api critical for monorepo support where the route definition is not linked to the components, and we need to join components after the cli generated the route tree

beaussan (commented):
Overall, I think this is a great move! Most of the content seems great to me, I only left a couple of questions, with the perspective of the monorepo support

docs/rfcs/parallel-route-slots/README.md
```diff
@@ -0,0 +1,1508 @@
...
+})
+```
+
+---
+
+## Route Definition
+
+### Slots Are Defined in Isolation
+
+Just like child routes, **slots don't need to be declared on their parent route**. They're defined as separate `@slotName` files and automatically composed into the route tree at generation time.
```
SeanCassiere: Are multiple slots using the same name possible?

Like having the slot name of `widget` for these routes, but the slotted route could be vastly different.

- `/dashboard/overview?@widget=/` - System widgets for various bits and bobs
- `/dashboard/orders?@widget=/` - Order related widgets

I'm aware of the route-scoped part of it, but I'm wondering what'd happen when global slots come into play.

Could use clarification here.

docs/rfcs/parallel-route-slots/README.md
```diff
@@ -0,0 +1,1508 @@
...
+Slots can have their own `beforeLoad` for slot-specific auth/guards:
+
+```ts
+// @modal/users.$id.tsx
+export const Route = createSlotRoute({
+  path: '/users/$id',
+  beforeLoad: async ({ params }) => {
+    const canViewUser = await checkPermission(params.id)
+    if (!canViewUser) {
+      throw redirect({ slot: 'modal', to: '/unauthorized' })
```
SeanCassiere: I assume we defer to whichever slot route throws first right?

docs/rfcs/parallel-route-slots/README.md
SeanCassiere: A + C are the strongest for me. The dedicated hook is a bit of a toss up, since the hook would be bound using `from` to the Route. The only part where the dedicated hook really makes sense for me is instances where the same search param exists on parent and slot. Where this is an explicit API for a Slot and probably wouldn't inherit from the parent.

docs/rfcs/parallel-route-slots/README.md
```diff
@@ -0,0 +1,1508 @@
...
+
+---
+
+## Open Questions for Implementation
+
+1. **Search param access within slots**: How should slot components access their namespaced search params? See detailed options in the [Search Params](#open-question-search-param-access-within-slots) section. This affects DX significantly and needs community input.
+
+2. **Preloading**: How should `<Link slots={{ modal: {...} }} preload="intent">` work? Should it preload just the slot's route, or also affect main route preloading?
+
+3. **shouldRevalidate**: Should slots participate in the main route's revalidation cycle, or have completely independent revalidation?
```
SeanCassiere: Might be a bit of both. Even if we choose to add an option for having its own cycle, we should probably also have a way for it to be able to use the main routing as well.

I'm mostly coming at this from the mental-model-complexity side of things. That being said, when you choose this you are automatically subscribing to extra hurdles, but as usual its the "sensible defaults" we want to get right here.

docs/rfcs/parallel-route-slots/README.md
tannerlinsley: This is just an error of omission. I'm still massaging a lot of stuff in here. Don't fret.

docs/rfcs/parallel-route-slots/README.md
```diff
@@ -0,0 +1,555 @@
...
+---
+
+## Slot Routes
+
+Slot routes use `createSlotRoute` with standard route options:
+
+```ts
+// @modal.users.$id.tsx
+export const Route = createSlotRoute({
+  path: '/users/$id',
```
SeanCassiere, line 131: Would these slot arounds be able to participate in some of the more advanced route descriptions? Like:

- Route param prefixes and suffixes
- Optional route params
- Character escaping in path

docs/rfcs/parallel-route-slots/README.md
```diff
@@ -0,0 +1,555 @@
...
+```ts
+// dashboard.@adminPanel.tsx
+export const Route = createSlotRoute({
+  enabled: ({ context }) => context.user.role === 'admin',
+  loader: () => fetchAdminStats(),
+  component: AdminPanel,
+})
+```
+
+### Slot Metadata
```
SeanCassiere, line 155: Is there a story here for using route context as well? or would that be too further down the tree for that to have any effect?

SeanCassiere: Also, on the story of route context, what'd be the way of having the route be aware of its caller?

Like having a Pricing Card, that could be aware of its activator to either show more or less options.

I'm thinking search params and context being options. Just wanted to clarify where we'd stand on this.

docs/rfcs/parallel-route-slots/README.md
SeanCassiere: Off the bat, I think character escaping could be an issue, since the it'd interfere with slot route search params.

matclayton:
This looks like exactly what we need to tidy up out modal mess, particularly is preload intents are fully supported :) Looking forward to a future with it!

matclayton:
Just wondered if theres been any progress on parallel routes, as it would allow us to do really performant modals with preloading, something we've been keen to tackle for while. Really looking forward to it :)

amav96:
The community need this feature! please

ericchernuka:
This looks amazing and fills an immediate need we have. Anything blocking the progress other than time/resourcing?

RND332:
+1

lerothias:
+1

PR #6305, opened 2026-01-05
fix: prerendering with docker build

### LINKED ISSUE #6275: Prerendering does not work when building Docker container

Start

### Describe the bug

When trying to build a TanStack Start project in SPA mode with Docker Vite fails to connect to the server. The build works fine outside of Docker.

```
3.276 [nitro] ✔ You can preview this build using bun run .output/server/index.mjs
3.277 [prerender] Prerendering pages...
3.305
3.305  ╭───────────────── [Build Info] ────────────────────╮
3.305  │                                                   │
3.305  │  - Build Directory: .output                       │
3.305  │  - Date: 1/2/2026, 5:55:52 AM                     │
3.305  │  - Nitro Version: 3.0.1-20251230-165713-6e801e22  │
3.305  │  - Nitro Preset: bun                              │
3.305  │                                                   │
3.305  ╰───────────────────────────────────────────────────╯
3.305
3.305 ℹ Spawning preview server...
3.305 ℹ bun run ./server/index.mjs
3.305
3.324 ➜ Listening on: http://localhost:46317/ (all interfaces)
3.841 [prerender] Concurrency: 10
3.842 [prerender] Crawling: /
3.855 [prerender] Prerendered 0 pages:
3.859 error: Unable to connect. Is the computer able to access the url?
3.859   path: "http://localhost:3000/",
3.859  errno: 0,
3.859   code: "ConnectionRefused"
```

I've also tried forcing the Nitro port to run at 3000 which ends up with the Vite prerender server failing to start instead.

https://github.com/NormySan/tanstack-start-spa-nitro

1. Download repository
2. Build the docker image with `docker build -t tanstack-start/nitro .`

### Expected behavior

I expect it to be able to connect to the server and to the pre-rendering.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: Router 1.132.0 and Start 1.132.0
- Nitro: Nightly at latest
- Runtime: Bun
- OS: Mac OS
- Bundler: Vite
- Bundler Version: 7.1.7

### Additional context

Similar to the following issue:
https://github.com/TanStack/router/issues/5967#issuecomment-3704465609

edmundhung:
Fix #6275

This binds the vite preview server to 127.0.0.1 to avoid Docker build failure caused by resolved network URLs pointing to non-listening interfaces.

## Summary by CodeRabbit

* **Chores**
  * Updated preview server configuration to bind to the loopback address for local development.

### DIFF
--- a/packages/start-plugin-core/src/prerender.ts
+++ b/packages/start-plugin-core/src/prerender.ts
@@ -271,6 +271,7 @@ async function startPreviewServer(
     return await vite.preview({
       configFile: viteConfig.configFile,
       preview: {
+        host: '127.0.0.1',
         port: 0,
         open: false,
       },


packages/start-plugin-core/src/prerender.ts
```diff
@@ -271,6 +271,7 @@ async function startPreviewServer(
     return await vite.preview({
       configFile: viteConfig.configFile,
       preview: {
+        host: '127.0.0.1',
```
SeanCassiere, line 274: We should probably make the host configurable for the prerenderer, since there are instances where people might want to set it to `0.0.0.0` as well.

edmundhung: Users can already control the host via the preview server config as I mentioned in the original issue. I wonder if it would be simpler to rely on the existing setting and document how it affects prerendering, instead of introducing a prerender-specific option here.

schiller-manuel: yes instead of hard coding here, we should document this, especially for the common docker case.

PR #6321, opened 2026-01-07
feat: add support for Zod v4 while maintaining v3 compatibility

### LINKED ISSUE #6138: Can't use zod v4

Start

### Describe the bug

If I try to use zod 4 in project I get errors when starting vite because router-generator and start-core relies on zod v3 api', the old `z.function().returns`, its now `z.function({input: [], output:[]})`. 

https://stackblitz.com/edit/github-9eq11eyj?file=package.json

`npm install`
`npm run dev`

### Expected behavior

Server should start

### Screenshots or Videos

Image

### Platform

- Router / Start Version: latest
- OS: macos
- Browser: all
- Browser Version: latest
- Bundler: vite
- Bundler Version: vite 7 and 8 beta

### Additional context

I guess it's not possible to use zod v4 yet? otherwise get into weird issues where theres multiple versions of zod installed.

zackarychapple:
This change adds runtime detection to support both Zod v3 and v4 simultaneously, allowing users to upgrade to Zod v4 without breaking existing v3 users.

Changes:
- Added compatibility helpers that detect Zod version at runtime
- Updated z.function() calls to use new v4 API when available
- Falls back to v3 API for existing installations
- Updated zod dependency to "^3.25.0 || ^4.0.0" in all packages

Technical Details:
- Detects Zod v4 by checking for '_zod' property on schemas
- Uses z.function({ input: [], output: ... }) for v4
- Uses z.function().args(...).returns(...) for v3
- Minimum v3.25.0 required for compatibility layer support

Affected packages:
- @tanstack/router-generator
- @tanstack/router-plugin
- @tanstack/start-plugin-core

Fixes #6138
Related to #4322, #4092

I verified this using a pnpm patch on zephyr cloud's main app.  @swalker326

## Summary by CodeRabbit

* **Chores**
  * Updated zod dependencies across router packages to support both v3.25+ and v4.0+, enabling broader compatibility with current and future releases while maintaining full backward compatibility with existing functionality.

### DIFF
--- a/packages/router-generator/package.json
+++ b/packages/router-generator/package.json
@@ -72,7 +72,7 @@
     "recast": "^0.23.11",
     "source-map": "^0.7.4",
     "tsx": "^4.19.2",
-    "zod": "^3.24.2"
+    "zod": "^3.25.0 || ^4.0.0"
   },
   "devDependencies": {
     "@tanstack/react-router": "workspace:*"
--- a/packages/router-generator/src/config.ts
+++ b/packages/router-generator/src/config.ts
@@ -4,6 +4,29 @@ import { z } from 'zod'
 import { virtualRootRouteSchema } from './filesystem/virtual/config'
 import type { GeneratorPlugin } from './plugin/types'
 
+// Helper to create a function schema compatible with both Zod v3 and v4
+function createFunctionSchema() {
+  // Try Zod v4 syntax first
+  if (typeof (z as any).function === 'function') {
+    try {
+      // Check if this is Zod v4 by testing for the new API
+      const testSchema = z.string()
+      if ('_zod' in testSchema) {
+        // Zod v4: use new function API
+        return (z as any).function({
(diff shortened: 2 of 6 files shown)


schiller-manuel:
why is this needed? a package manager with isolated installs should just work fine with multiple versions of zod

zackarychapple:
> why is this needed? a package manager with isolated installs should just work fine with multiple versions of zod

Good question, we did temporarily get around this by pinning tanstack router to use zod v3, which "worked" but is not ideal. Having to maintain package specific overrides is a pain in the behind.  

Also since zod 4 has a ton of performance improvements and has been out for over half a year we wanted to move our ecosystem to it. 

But since we know that tanstack ecosystem does not typically do hard cut overs / breaking changes the goal of this PR was to maintain compatibility with v3 of zod, while allowing those of us who want to update to be able to do so.

PR #6343, opened 2026-01-10
feat: collect styles in dev

### LINKED ISSUE #3023: How to use CSS Modules with TanStack Start?

Start

### Describe the bug

I'm struggling to make CSS Modules work properly. Currently, the styles are being loaded only on the client, causing a flash of unstyled content (FOUC) during initial render.
I tried in this repo https://github.com/tanstack/router/tree/main/examples/react/start-basic

https://github.com/nikolayemrikh/start-css-modules-bug

1. Clone repo https://github.com/nikolayemrikh/start-css-modules-bug
2. Add CSS Modules support by creating a .module.css file and importing it into a React component.
3. Run the server and open the app in a browser.
4. Observe that the styles are applied only after the client-side hydration, causing a flash of unstyled content (FOUC).

### Expected behavior

I would like to see a working example or documentation on how to properly configure TanStack Router with CSS Modules and SSR.
The desired behavior is to have styles included in the server-rendered HTML, so there is no flash of unstyled content (FOUC).

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 91.1]

### Additional context

_No response_

schiller-manuel:
fixes #3023

## Summary by CodeRabbit

* **New Features**
  * Dev-mode route-scoped CSS collection and serving for improved SSR styling across React, Solid, and Vue.
  * New example projects demonstrating CSS Modules integration for React, Solid, and Vue with Vite configs.

* **Tests**
  * End-to-end test suites added to validate global CSS and CSS Modules behavior, hydration, and client-side navigation in all three framework samples.

### DIFF
--- a/packages/react-router/tests/Scripts.test.tsx
+++ b/packages/react-router/tests/Scripts.test.tsx
@@ -217,7 +217,7 @@ describe('ssr HeadContent', () => {
       <RouterProvider router={router} />,
     )
     expect(html).toEqual(
-      `<title>Index</title><meta name="image" content="image.jpg"/><meta property="og:description" content="Root description"/><meta name="description" content="Index"/><meta name="last-modified" content="2021-10-10"/><meta property="og:image" content="index-image.jpg"/>`,
+      `<title>Index</title><meta name="image" content="image.jpg"/><meta property="og:description" content="Root description"/><meta name="description" content="Index"/><meta name="last-modified" content="2021-10-10"/><meta property="og:image" content="index-image.jpg"/><link rel="stylesheet" href="/@tanstack-start/styles.css?routes=__root__%2C%2F" data-tanstack-start-dev-styles="true"/>`,
     )
   })
 })


packages/react-router/src/HeadContent.tsx
```diff
@@ -203,6 +203,39 @@ export const useTags = () => {
...
+
+  React.useEffect(() => {
+    // After hydration, remove the SSR-rendered dev styles link
+    document
+      .querySelectorAll('[data-tanstack-start-dev-styles]')
+      .forEach((el) => el.remove())
+  }, [])
+
+  // Build the same href on both server and client for hydration match
+  const href = `/@tanstack-start/styles.css?routes=${encodeURIComponent(routeIds.join(','))}`
```
dgreif, line 227: This doesn't appear to use the vite `base` path if one is set, leading to a 404 when trying to fetch the styles

packages/start-plugin-core/src/dev-server-plugin/plugin.ts
dgreif, line 70: I noticed this return gets hit if I have the `nitro` plugin running as well, causing the new middleware to not get registered, leading to a 404 for `/@tanstack-start/styles.css`

PR #6349, opened 2026-01-10
fix: prevent crash in buildLocation when params stringification fails (#6310)

### LINKED ISSUE #6310: Params parse/stringify behavior doesn't match types when narrowing params

Router

### Describe the bug

Hello 👋

If I use `params.parse` to narrow valid param value (my exact use case is language in URL and narrowing valid values only to "en" and "pl", my use case also involves parsing them into "en-US" and "pl-PL" but I narrowed the issue down) these are narrowed both in `stringify` function and everywhere where I can provide a value for it (e.g. `Link` or `redirect` or `navigate`).

Image

However, the reality is that `stringify` function (where types says that it's only called with one of narrowed values) is also called with invalid values when they're visited.  You can see, even TypeScript complaining that it never happens.

Image

Is this expected behavior? I tried to look into the code, but I couldn't fully grasp why `stringify` is called if params were never parsed in the first place. Is `notFound` allowed to be called like this in `params.parse`? 

While debugging I also spotted one more thing - `stringify` behaves weirdly if error is thrown inside. Blank page is rendered and no log to be found anywhere. I stumbled upon it cause I used Effect schema, which threw error in `stringily` when I tried to decode "en-US" or "pl-PL" into "en"/"pl", because it was actually getting different value sometimes. 

https://github.com/pawelblaszczyk5/tanstack-router-not-found-params-problem

1. Create `params.parse` function in route with path param that narrows string value into set of allowed values, otherwise throws `notFound` (or `redirect` - it also works in that case)
2. Create `params.stringify` where types say it'll only ever be called with one of narrowed values
3. Observe it being called also with values that weren't parsed correctly

### Expected behavior

Types match the runtime behavior. Probably `stringify` is never called with value that wasn't properly parsed?

### Screenshots or Videos

Provided them above in description

### Platform

- Router / Start Version: 1.145.7
- OS: macOS
- Browser: Reproducible everywhere
- Browser Version: Same
- Bundler: Vite
- Bundler Version: 7.2.7

### Additional context

_No response_

Ju-MINJAE:
### Description
Fixes #6310

I investigated the issue using the reproduction repository provided in the issue.
Currently, `buildLocation` attempts to execute `stringify` even when `params.parse` fails or throws an error. In SSR environments, this causes the server to throw a 500 error (crash) instead of rendering the `NotFound` boundary.

### Proposed Solution
I wrapped the `stringify` execution in a `try-catch` block within `buildLocation`.
If stringification fails, it now silently ignores the error. This prevents the app from crashing and allows the router to render the `NotFound` component correctly.

**Before:**
before

**After:**
after

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved router stability by implementing graceful error handling for parameter serialization. Navigation will no longer fail unexpectedly due to serialization errors.

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -1701,7 +1701,11 @@ export class RouterCore<
           const fn =
             route.options.params?.stringify ?? route.options.stringifyParams
           if (fn) {
-            Object.assign(nextParams, fn(nextParams))
+            try {
+              Object.assign(nextParams, fn(nextParams))
+            } catch (err) {
+              //
+            }
           }
         }
       }


pawelblaszczyk5:
I'm not fully sure would I call this solving the issue I reported. I see that it partially fixes the problem, but underlaying issue is still here. If framework tells me a function will be called with given set of arguments, I expect it to be called with these arguments. Not to be called sometimes with different arguments but errors that comes from it (or any other flaw in my logic) to be caught. Obviously, it's up to maintainers, but my intuition tingles here

Ju-MINJAE:
@pawelblaszczyk5  Thanks for the insight! I completely agree that fixing the underlying issue (validation logic) would be the ideal solution.
This PR was primarily intended as a safety net to prevent the immediate crash in production. However, if you have a specific idea or could point me to where the validation flow should be corrected upstream, I'd be more than happy to update this PR to address the root cause!

packages/router-core/src/router.ts
```diff
@@ -1701,7 +1701,11 @@ export class RouterCore<
           const fn =
             route.options.params?.stringify ?? route.options.stringifyParams
           if (fn) {
-            Object.assign(nextParams, fn(nextParams))
+            try {
```
brenelz, line 1704: im not sure if just ignoring the exception is a good fix

Ju-MINJAE: Should I add a console.warn inside a process.env.NODE_ENV !== 'production' check as CodeRabbit suggested?

```ts
catch (err) {
-  //
+  if (process.env.NODE_ENV !== 'production') {
+    console.warn(
+      `Failed to stringify params for route ${route.id}:`,
+      err,
+    )
+  }
 }
 ```

PR #6373, opened 2026-01-12
feat: add Route.redirect and getRouteApi.redirect

schiller-manuel:
## Release Notes

* **New Features**
  * Added Route.redirect() method: a type-safe redirect function bound to the route, with auto-filled from parameter for cleaner relative redirects (e.g., "./overview").
  * Added getRouteApi().redirect() for programmatic redirects outside route definitions with automatic context awareness.

* **Documentation**
  * Expanded redirect documentation with examples covering standalone usage, route-bound patterns, and RouteApi approaches for various redirect scenarios.

### DIFF
--- a/docs/router/framework/react/api/router/redirectFunction.md
+++ b/docs/router/framework/react/api/router/redirectFunction.md
@@ -19,6 +19,8 @@ The `redirect` function accepts a single argument, the `options` to determine th
 
 ## Examples
 
+### Using the standalone `redirect` function
+
 ```tsx
 import { redirect } from '@tanstack/react-router'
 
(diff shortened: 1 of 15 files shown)


docs/router/framework/react/api/router/redirectFunction.md
```diff
@@ -51,3 +53,59 @@ const route = createRoute({
...
+```tsx
+import { getRouteApi } from '@tanstack/react-router'
+
+const routeApi = getRouteApi('/dashboard/settings')
+
+// In a beforeLoad or loader callback
+function checkAuth() {
+  if (!user) {
+    // Type-safe redirect with automatic 'from' parameter
+    throw routeApi.redirect({
```
kettanaito, line 98: Wouldn't this produce a confusing navigation history? Via `getRouteApi`, you can produce a redirect where `from` points to a different route where you've actually issued the redirect. This example is fine, but the API itself has this footgun.

schiller-manuel: this API just allows you to use relative notation for building redirect targets. has nothing to do with navigation history

kettanaito: My bad, misread the intention then. Excited for this one to land!

PR #6387, opened 2026-01-15
fix(router): ensure useParams returns parsed params when strict is false

### LINKED ISSUE #6385: useParams({ strict: false }) returns unparsed params

Router

### Describe the bug

"version": "1.147.3",

when using useParams with { strict: false }, we are getting unparsed values back, but they are still typed as if they were parsed. 

This is caused by useMatch match.params no longer being parsed (compared to the version we upgraded from) match. _strictParams is parsed, but doesn't contain params from the child routes.

This is the patch we applied to fix the issue for us, it seems a bit blunt though but I hope it helps the team quickly narrow down the issue.

[@tanstack+react-router+1.147.3.patch](https://github.com/user-attachments/files/24624734/%40tanstack%2Breact-router%2B1.147.3.patch)

https://stackblitz.com/edit/github-3khftpeu

from the repro url, click the "version" link

https://stackblitz.com/edit/github-3khftpeu

Image

have a child route - createFileRoute with:

  params: {
    parse: (params) => ({
      ...params,
      "version": parseInt(params["version"]),
    }),
    stringify: (params) => ({
      ...params,
      "version": `${params["version"]}`,
    }),
  },

then in a parent route, do:

const { version } = useParams({ strict: false });
console.log(version); // in the ide, shows up as `number | undefined` but in the console, shows up as a string.

### Expected behavior

I expect the runtime values of useParams to be equal to their type.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.147.3
- OS: macOS
- Browser: Chrome
- Browser Version: 143.0.7499.193
- Bundler: vite
- Bundler Version: 7.1.4

### Additional context

_No response_

jong-kyung:
fixes #6385

This PR fixes issue #6385, where `useParams({ strict: false })` in a parent route would return unparsed parameter values from a child route.

### The Fix

The fix the issue by ensuring that parsed parameters from child routes are correctly propagated to their parents when `strict: false` is used.

1.  **Correct Parameter Propagation**: The router's internal matching logic was updated. When a route match is being processed, it now correctly looks up the previous match's state, which includes its parsed parameters, and merges them. This ensures the parsed values are available throughout the route hierarchy.

2.  **Performance Optimization**: As part of the fix, the mechanism for retrieving previous route matches was optimized. The implementation was changed from using `Array.prototype.find()` to a `Map`-based lookup. This improves performance from O(n) to O(1), which is especially beneficial in applications with many routes.

### Changes

-   **`packages/router-core/src/router.ts`**: Updated the core routing logic to correctly handle parameter propagation and implemented the `Map`-based optimization for match retrieval.
-   **`packages/react-router/tests/useParams.test.tsx`**: Added a new unit test to specifically verify that `useParams({ strict: false })` receives parsed values.
-   **`e2e/react-router/basic-file-based/**`**: Added new routes and an e2e test to confirm the fix in a real-world file-based routing scenario.

## Summary by CodeRabbit

* **New Features**
  * Added a "Parsed params with strict false" page and two routes: /params-ps/strict-false and /params-ps/strict-false/$version; parent reads a numeric version param non-strictly and offers links to versioned child pages.

* **Tests**
  * Added end-to-end and unit tests verifying non-strict param parsing and numeric values across parent→child navigation.

* **Chores**
  * Routing performance optimization for faster internal lookups.


jong-kyung:
oops, clicked the wrong button and accidentally closed this! 😅 Reopening now.

packages/router-core/src/router.ts
```diff
@@ -1533,8 +1534,13 @@ export class RouterCore<
       const route = this.looseRoutesById[match.routeId]!
       const existingMatch = this.getMatch(match.id)
 
-      // only execute `context` if we are not calling from router.buildLocation
+      // Update the match's params
+      const previousMatch = previousMatchesByRouteId.get(match.routeId)
+      match.params = previousMatch
+        ? replaceEqualDeep(previousMatch.params, routeParams)
+        : routeParams
```
Sheraff, line 1610: Maybe we shouldn't set the `match.params` in the 1st iteration then? Because it looks like we're now doing the work twice now.

jong-kyung: @Sheraff I tried removing the second update, but the regression (#6385) reappeared.

Here is why the second update is necessary:
The router processes matches from **Parent to Child**. When the Parent match is created in the first loop, the Child hasn't been processed yet, so `routeParams` doesn't strictly contain the Child's parsed params at that moment.

Since `strict: false` allows Parents to access Child params, we must update the Parent's `match.params` **after** the entire loop finishes (when `routeParams` is fully populated).

I've optimized it by piggybacking on the existing second loop (used for Context) instead of creating a new one, so the performance impact is minimal.

Please let me know if I misunderstood anything. Thank you!

jong-kyung:
@nlynzaad 
can you please review this?

jong-kyung:
@nlynzaad Thank you for the review and merge !

PR #6392, opened 2026-01-16
fix(start-client-core): allow middleware to catch server function handler errors

### LINKED ISSUE #6381: Middleware does not catch error thrown by serverFn

Start

### Describe the bug

Hi , 

First of all, Thanks for the project!.

I have an issue with middleware catching errors thrown by server fn. I have a server function that uses a middleware. That middleware try catch the next() to hide possible leak of information thrown by unhandled errors.
But for my surprise, the thrown error is not caught. 

Note: I also tried WITH and WITHOUT await (...return await next()....) .Some examples in docs, show return next without awaiting.

```
	component: App,
	loader: async () => getName(),
});

const middleware = createMiddleware().server(async ({ next }) => {
	try {
		return await next();
	} catch (error) {
		console.error("Middleware caught an error:", error);
		return new Response("An error occurred while processing your request.", {
			status: 500,
		});
	}
});

const getName = createServerFn()
	.middleware([middleware])
	.handler(async () => {
		throw new Error("This message should be caught by middleware");
		// return "TanStack Start";
	});

function App() {
	const name = Route.useLoaderData();
	console.log("Name is ", name);
return <div>hi</div>
}
```

https://github.com/esoteloferry/tanstackstart-bug-catch-middleware

1. Git clone minimal reproduction repro https://github.com/esoteloferry/tanstackstart-bug-catch-middleware
2. npm i
3. npm run dev
4. check localhost:3000

### Expected behavior

I would suspect error message would be "An error occurred while processing your request." , which is the error message in the catch of middleware.

### Screenshots or Videos

Image

### Platform

- Router / Start Version:v1.149.3
- OS: macOS
- Browser:Chrome
- Browser Version: 143.0.7499.193
- Bundler: vite
- Bundler Version: 7.3.1

### Additional context

This used to work, I think in v1.143 . Not sure.

jjcantu:
## Summary & Purpose
Fixes #6381 - Middleware can now properly catch errors thrown by server function handlers.

## Changes
- `packages/start-client-core/src/createServerFn.ts`: Modified `executeMiddleware` to throw errors from `result.error` instead of silently returning them in context, allowing parent middleware to catch them with try/catch
- `e2e/react-start/server-functions/src/routes/middleware/catch-handler-error.tsx`: Added test route demonstrating middleware error catching
- `e2e/react-start/server-functions/tests/server-functions.spec.ts`: Added e2e test verifying middleware can catch handler errors

## Summary by CodeRabbit

* **New Features**
  * Middleware can intercept and transform errors from server functions; a new UI route lets users trigger and view the transformed error.

* **Bug Fixes**
  * Improved error propagation so downstream errors surface correctly through the middleware stack.

* **Tests**
  * Added end-to-end test validating middleware error interception and the UI behavior.

### DIFF
--- a/e2e/react-start/server-functions/src/routes/middleware/catch-handler-error.tsx
+++ b/e2e/react-start/server-functions/src/routes/middleware/catch-handler-error.tsx
@@ -1,0 +1,67 @@
+import { createFileRoute } from '@tanstack/react-router'
+import { createMiddleware, createServerFn } from '@tanstack/react-start'
+import { useState } from 'react'
+
+const errorCatchingMiddleware = createMiddleware({ type: 'function' }).server(
+  async ({ next }) => {
+    try {
+      return await next()
+    } catch (error) {
+      const originalMessage =
+        error instanceof Error ? error.message : 'Unknown error'
+      throw new Error(`Middleware caught and transformed: ${originalMessage}`)
+    }
+  },
+)
+
+const $serverFnThatThrows = createServerFn({ method: 'GET' })
+  .middleware([errorCatchingMiddleware])
+  .handler(() => {
+    throw new Error('This error should be caught by middleware')
+  })
+
+export const Route = createFileRoute('/middleware/catch-handler-error')({
+  component: RouteComponent,
+})
+
+function RouteComponent() {
+  const [error, setError] = useState<string | null>(null)
+  const [loading, setLoading] = useState(false)
+
+  const handleClick = async () => {
+    setLoading(true)
+    setError(null)
+    try {
+      await $serverFnThatThrows()
+    } catch (e) {
+      setError(e instanceof Error ? e.message : String(e))
+    } finally {
+      setLoading(false)
+    }
+  }
+
+  return (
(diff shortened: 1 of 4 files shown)


e2e/react-start/server-functions/src/routes/middleware/catch-handler-error.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { useState } from 'react'
+
+const errorCatchingMiddleware = createMiddleware({ type: 'function' }).server(
+  async ({ next }) => {
+    try {
+      return await next()
+    } catch (error) {
+      // Return a custom result instead of the original error
+      // The types don't expose `result` but runtime handles it
+      return next({
```
schiller-manuel: throw a different error here and observe it in the tests.
returning a different value from middleware is currently not supported from the types, so we should not do in the tests

jjcantu: You're right, @schiller-manuel. It is now throwing an error.

PR #6398, opened 2026-01-16
refactor: don't reparse upon navigation

schiller-manuel:
## Summary by CodeRabbit

* **New Features**
  * Navigation operations are now asynchronous and return a result indicating success or blocked.
  * Navigation state now includes cached route-match snapshots to speed back/forward navigation.

* **Bug Fixes**
  * Prevents unnecessary route reloads for internal navigations via a new skip-load guard.

* **Tests**
  * Test suites updated to await asynchronous navigation and DOM updates.

### DIFF
--- a/packages/router-core/src/index.ts
+++ b/packages/router-core/src/index.ts
@@ -458,5 +458,4 @@ export {
   executeRewriteInput,
   executeRewriteOutput,
 } from './rewrite'
-export { INTERNAL_NAV_MARKER } from './router'
 export type { LocationRewrite, LocationRewriteFunction } from './router'


packages/router-core/src/router.ts
```diff
@@ -1264,44 +1292,68 @@ export class RouterCore<
...
-    const matchedRoutesResult = this.getMatchedRoutes(next.pathname)
-    const { foundRoute, routeParams, parsedParams } = matchedRoutesResult
-    let { matchedRoutes } = matchedRoutesResult
-    let isGlobalNotFound = false
+    // Fast-path: use snapshot hint if valid (all route IDs must exist)
+    const snapshot = opts?.snapshot
+    const snapshotValid =
+      snapshot &&
+      snapshot.routeIds.length > 0 &&
+      snapshot.routeIds.every((id) => this.routesById[id])
```
Sheraff, line 1283: Why is this check necessary? (mainly `snapshot.routeIds.every((id) => this.routesById[id])`)

Can we not trust what is passed here?

schiller-manuel: theoretically if you reload you could get a new bundle served. this check tries to detect if the route tree changed. however it's not sufficient probably as the routes might still exist but would match otherwise.

packages/router-core/src/router.ts
```diff
@@ -1783,17 +1877,21 @@ export class RouterCore<
       // If a rewrite function is provided, use it to rewrite the URL
       const rewrittenUrl = executeRewriteOutput(this.rewrite, url)
 
+      // Use encoded URL path for href (consistent with parseLocation)
+      const encodedHref = url.href.replace(url.origin, '')
```
nlynzaad, line 1877: can probably just build this using url.pathname + url.search + url.hash here

packages/router-core/src/router.ts
nlynzaad: this checks if all snapshot ids exist in the routes, is this sufficient in the case where there are new routes that hasn't been loaded in the snapshot yet? specifically, is this.routeIds.every((id) => snapshot.routesById[id]) not necessary?

arekrgw:
I think this exact PR broke my apps navigation. Before @tanstack/react-router@1.153.2 it worked and from this version it does not.

Details:

First when navigating to the article page I do it like that: 
image

Then on my article page I have back button which uses `back` param as an url. Back param usually look like this: `/stories?page=1&query=&limit=20`
image

When I click on it, router shows 404 page, when I reload page then it shows correct page without any URL changes.

Should it be done differently from now on or something is off?

s3rbug:
my route params is always undefined after this PR changes

```ts
import { createRootRouteWithContext, createRoute } from '@tanstack/react-router';

const rootRoute = createRootRouteWithContext()();

const layoutRoute = createRoute({
    id: 'layout',
    component: AppPageLayout,
});

const lobbyRoute = createRoute({
    getParentRoute: () => layoutRoute,
    path: '/',
}).lazy(() => import('./lobby/LobbyPage').then((m) => m.Route));

const itemsIndexRoute = createRoute({
    getParentRoute: () => layoutRoute,
    path: 'items',
}).lazy(() => import('./item/ItemsPage').then((m) => m.Route));

const itemRoute = createRoute({
    path: 'items/view/$itemId',
    loader: (props) => {
        // props.params.itemId is undefined after @tanstack/react-router@1.153.2
    },
}).lazy(() => import('./item/ItemPage').then((m) => m.Route));

export const routeTree = rootRoute.addChildren([
  itemRoute,
  layoutRoute.addChildren([itemsIndexRoute]),
  lobbyRoute
]);
```

iffa:
Seeing a similar issue, our app with lots of search param usage now causes a "not found" every time search params change. Once you reload the page, it works.

arekrgw:
> Seeing a similar issue, our app with lots of search param usage now causes a "not found" every time search params change. Once you reload the page, it works.

~~I've pulled newest changes today and it works now for me~~

Ignore me, it does not work, went back to stable version

iffa:
> > Seeing a similar issue, our app with lots of search param usage now causes a "not found" every time search params change. Once you reload the page, it works.
> 
> ~I've pulled newest changes today and it works now for me~
> 
> Ignore me, it does not work, went back to stable version

Yep, version `1.154.13` still has the same issue, `1.153.1` is the latest one that works.

schiller-manuel:
we reverted this and will revisit this perf optimization at a later point

PR #6409, opened 2026-01-18
fix(react-router): walk parent route chain for notFoundComponent lookup

### LINKED ISSUE #6351: main.tsx defaultNotFound component is being rendered instead a root pathless route notFoundComponent

Router

### Describe the bug

root pathless **notFound** is not rendering instead of ```main.tsx``` **defaultNotFoundComponent** is being triggered.

https://github.com/rahulvils/tanstack-router-not-found

1. Clone the repository
2. Run ```pnpm i```
3. Run ```pnpm dev```
4. Open ```http://localhost:3000``` in your browser
5. Click on the **Skill Agent** button in the home page, You will be redirected to skill-agent page
6. Now manually go the the route to trigger not-found component ```http://localhost:3000/agents/skill-agen```
7. The not-found component in the ```_authenticated/agents/route.tsx``` will be triggered (Which is expected)
8. Now go the location in the codebase ```_authenticated/agents/route.tsx``` and comment the ```notFoundComponent```
9. Now navigate to the same link ```http://localhost:3000/agents/skill-age``` the main.tsx **defaultNotFoundComponent** is being triggered instead of ```_authenticated.tsx``` **notFound** component

### Expected behavior

For root pathless route (say ```routes/_authenticated.tsx```) the **notFound** component is not being triggered and ```main.tsx``` **defaultNotFoundComponent** is being triggered. I didn't change the default behaviour of notFoundMode (which is fuzzy)

### Screenshots or Videos

_No response_

### Platform

- Router Version: 1.147.0
- OS: Linux (Ubuntu 24.04)
- Browser: Chrome
- Browser Version: 143.0.7499.192 (Official Build) (64-bit)
- Bundler: Vite
- Bundler Version: 7.1.1

### Additional context

_No response_

davidclimbing:
## What Changed

Refactored `new-process-route-tree.ts` to ensure pathless routes (e.g., `_authenticated`) are correctly processed during fuzzy matching, allowing their `notFoundComponent` to trigger.

**Files Changed:**
- `packages/router-core/src/new-process-route-tree.ts`
  - Removed `skipOnParamError` guard for pathless node creation.
  - Added explicit `node.route` assignment to pathless nodes.

## Why

**Fixes #6351 **

Pathless layout routes were being skipped during fuzzy matching when a child route failed to match. This prevented their `notFoundComponent` from rendering, causing the error to bubble up to the root incorrectly.

This change ensures pathless nodes are properly flagged in the segment tree so they are considered valid candidates for handling `notFound` errors.

## How Tested

**New Tests:**
- Added `packages/router-core/tests/pathless_fuzzy_matching.test.ts` to verify:
  - Exact matching works for children.
  - Fuzzy matching correctly finds the pathless layout for missing paths.

**Results:**
- ✅ All new tests passed.
- ✅ All 846 regression tests in `router-core` passed.

**Closes #6351 **

## Summary by CodeRabbit

* **Bug Fixes**
  * Not-found handling now searches up the route hierarchy to render the nearest ancestor's not-found component when a child lacks one, improving fallback behavior for nested routes.

* **Tests**
  * Added test confirming pathless layout routes inherit and use a parent's not-found component instead of the global default.


Sheraff:
Thank your for your PR, the issue is clear and based on the video (https://github.com/TanStack/router/issues/6351#issuecomment-3750555670) it does seem like a real bug to me.

However:

1. in the test you are proposing in this PR, I would disagree that `_layout` should be fuzzy matched

In the video example we have these routes
```
__root__
  _authenticated
    agents.route
      skill-agent.index
```
So when the path to match is `/agents/foo`, then `skill-agent.index` is not matched, but `agents.route` should be (so the `notFoundComponent` in `_authenticated` should be used).

But in your case
```
__root__
  _layout
    child
```
matching `/missing` does not match `child` at all, so I don't see why it should include the `_layout` in its matched routes.

2. following this explanation, I'm not entirely sure the fix you are proposing is correct. It might be. But skipping some nodes in the tree is good for performance, so ideally I'd like to understand why `getNodeMatch` does not fuzzy match the `agents.route` in the video case. I haven't tried, but it feels like it should pass this condition

https://github.com/TanStack/router/blob/17045eb011fa0826d9192dd0bcc11b3f81eebb5a/packages/router-core/src/new-process-route-tree.ts#L1021-L1029

schiller-manuel:
this PR breaks a lot of tests. check the CI output

davidclimbing:
> 2. following this explanation, I'm not entirely sure the fix you are proposing is correct. It might be. But skipping some nodes in the tree is good for performance, so ideally I'd like to understand why `getNodeMatch` does not fuzzy match the `agents.route` in the video case. I haven't tried, but it feels like it should pass this condition


Thank you @Sheraff for your patient feedback!

### What was wrong with my previous approach

**I realize now that my previous test case was incorrect, and I am sorry for any confusion this caused.** As @Sheraff pointed out, the test was flawed because the `/missing` path was structured in a way that it didn't actually fall under the `_layout` (pathless route) hierarchy. This led me to mistakenly investigate `new-process-route-tree.ts`, which was not the source of the problem. Your feedback helped me realize that the issue was specifically about how the rendering layer identifies the correct `notFoundComponent` within a nested, pathless route structure.

### Root cause analysis (Issue #6351)

The problem exists in the `renderRouteNotFound()` function. When a "Not Found" state occurs (e.g., navigating to `/agents/non-existent`):

1. The router correctly performs a fuzzy-match to the `agents` route.
2. The `agents` route itself does not have a `notFoundComponent`.
3. **The Bug:** Instead of traversing up to the immediate parent (the pathless `_authenticated` route), the code was skipping the parent hierarchy and falling back directly to the `defaultNotFoundComponent` defined in the root.
4. **Expected:** Even for pathless routes, the router should check the parent chain recursively to find the nearest available `notFoundComponent`.

### The fix

I have updated `renderRouteNotFound()` to correctly walk up the parent route chain. This ensures that pathless layout routes (like `_authenticated`) are properly checked for their `notFoundComponent` before reaching the global default.

```typescript
// Recursively walk up the parent route chain to find the nearest notFoundComponent
let routeWithNotFound: AnyRoute | undefined = route
while (routeWithNotFound && !routeWithNotFound.options.notFoundComponent) {
  routeWithNotFound = routeWithNotFound.parentRoute
}

```

### Test case

I added a robust test case in `packages/react-router/tests/not-found.test.tsx` that mirrors the structure reported in issue #6351:

* **Structure:** `__root__` → `_authenticated` (pathless) → `agents` → `skill-agent`
* **Scenario:** Navigate to `/agents/non-existent`.
* **Verification:** Confirms that `_authenticated`'s `notFoundComponent` is rendered, ensuring pathless parents are no longer ignored.

### Files changed

* `packages/react-router/src/renderRouteNotFound.tsx` - The fix
* `packages/react-router/tests/not-found.test.tsx` - Regression test for issue #6351

rahulvils:
@Sheraff can you please verify this :)

rahulvils:
@schiller-manuel can you please verify this PR

PR #6006, opened 2025-12-01
fix(start): allow request middleware to return Response

JBR-0100:
Fixes #5846. This change allows `RequestMiddleware` to return a `Response` object directly, as expected by the issue report. Updated type tests to reflect this change.

## Summary by CodeRabbit

* **New Features**
  * Middleware server functions can now return a Response object directly (sync or Promise) in addition to previous result types, allowing more flexible response handling.

* **Tests**
  * Added tests covering direct Response returns (sync and async) to validate the widened return behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #6413, opened 2026-01-18
feat: server function middleware short circuit

schiller-manuel:
runtime only, types TBD


uncvrd:
This would be real awesome to help mock 3rd party API requests for testing purposes, would the current workaround be to `return Response.json();` in the middleware until something like this is implemented? I got the idea from this PR https://github.com/TanStack/router/pull/6006/changes

PR #6416, opened 2026-01-18
fix: encoded url hash fragment matching

nlynzaad:
This PR addresses some issues when matching url hashes.

As part of #6389 it was noticed that when using encoded characters in hashes these would not be matched correctly. This is addressed by making sure the encoded hash fragment is decoded before matching for Link active properties and others. 

It was also noticed that in the case of solid hash matching for Link active properties was not working when doing SSR. This is due to how hashes is handled between server and client and compounded with how Solid's reactivity works. When a request is sent to the server during SSR the hash is removed. When received on client the hash is then processed but a rerender does not occur and hence not applying the active props. This is addressed by ensuring that the hash check is only done on client and not accounted for on server. 

Tests have been added for the above

## Summary by CodeRabbit

* **New Features**
  * Added a new route to exercise URL hash handling (non-ASCII and special chars) and hash-aware navigation links with active styling.

* **Bug Fixes**
  * Decode hash fragments when parsing locations.
  * Avoid evaluating hash-based active state during server-side rendering.

* **Tests**
  * Added end-to-end tests covering direct and router navigation with encoded URL hashes.

* **Chores**
  * Excluded the new hash route from prerendering.

### DIFF
--- a/docs/start/framework/react/guide/middleware.md
+++ b/docs/start/framework/react/guide/middleware.md
@@ -641,12 +641,13 @@ await myServerFn({
 
 When custom fetch implementations are provided at multiple levels, the following precedence applies (highest to lowest priority):
 
-| Priority    | Source             | Description                                     |
-| ----------- | ------------------ | ----------------------------------------------- |
-| 1 (highest) | Call site          | `serverFn({ fetch: customFetch })`              |
-| 2           | Later middleware   | Last middleware in chain that provides `fetch`  |
-| 3           | Earlier middleware | First middleware in chain that provides `fetch` |
-| 4 (lowest)  | Default            | Global `fetch` function                         |
+| Priority    | Source             | Description                                          |
+| ----------- | ------------------ | ---------------------------------------------------- |
+| 1 (highest) | Call site          | `serverFn({ fetch: customFetch })`                   |
+| 2           | Later middleware   | Last middleware in chain that provides `fetch`       |
+| 3           | Earlier middleware | First middleware in chain that provides `fetch`      |
+| 4           | createStart        | `createStart({ serverFns: { fetch: customFetch } })` |
+| 5 (lowest)  | Default            | Global `fetch` function                              |
 
 **Key principle:** The call site always wins. This allows you to override middleware behavior for specific calls when needed.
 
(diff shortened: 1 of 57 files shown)


e2e/react-start/basic-rsc/src/routeTree.gen.ts
```diff
@@ -249,6 +249,7 @@ import type { getRouter } from './router.tsx'
 import type { createStart } from '@tanstack/react-start'
 declare module '@tanstack/react-start' {
   interface Register {
+    ssr: true
```
schiller-manuel: dont touch this file please. this example will soon be gone

nlynzaad: been excluding it everytime it came up as a change, but must have slipped through on this one. will remove it from the PR.

nlynzaad:
@birkskyum when using hash fragments in url's these aren't available server side (the browser does not send this to the server), so during SSR the hash value is blank, but the value is provided once we are client side.

However Solid does not treat this update in a reactive manner, i.e while the hash is updated, the location accessor does not notify subscribers to this change. This creates problems for cases when using the hash value i.e. Link active props, conditional rendering, etc.

I've added a hacky workaround in the attached Solid e2e for this and made a change to the Link component to exclude the hash when checking if the hash should be used in the active props when server side, which solves the problem for Link active props.

Ideally the location should update all the subscribers once the hash value is received, my knowledge of Solid is just too broken to comprehend where or what needs to change to do this the right way.

Do you have any recommendations on what needs to change in solid-router for this or is this the best way to deal with this update?

birkskyum:
@nlynzaad  I think this workaround to Link looks reasonable for solid v1.

packages/solid-router/src/link.tsx
```diff
@@ -230,7 +230,8 @@ export function useLinkProps<
         }
       }
 
-      if (local.activeOptions?.includeHash) {
+      // url hash is not available on server, so do not evaluate this here when on server
+      if (local.activeOptions?.includeHash && !router.isServer) {
```
theoludwig, line 234: This fix should be backported to the React package as well.

As there is the same hydratation issue in React, even after upgrading to TanStack Start v1.153.1.

nlynzaad: Thanks, could you create a reproducer in a new issue. i will look at it a bit later. We have tests for this specifically so would need to see what it is we missed.

PR #6421, opened 2026-01-19
fix(react-start): Apply Link active properties correctly on SSR when including hash  

nlynzaad:
This PR applies the same fix that was done in #6389 for Solid to React. As mentioned by @theoludwig in the comments this affects React as well and was missed by the tests.

This PR also updates and aligns the tests across React, Solid and Vue. 

## Summary by CodeRabbit

* **New Features**
  * Added a toggle to show/hide the URL hash on the special-chars page and added a second hash-enabled link.

* **Bug Fixes**
  * Made hash-based active-state matching more consistent across navigation and hydration.

* **Tests**
  * Updated end-to-end tests to cover both hash links, the toggle interaction, and verification of hash content after navigation.


theoludwig:
Thank you for working on this!

Edit: I tried to test the PR using pkg.pr.new with:
```yaml
"@tanstack/react-router": "https://pkg.pr.new/TanStack/router/@tanstack/react-router@6421"
"@tanstack/react-start": "https://pkg.pr.new/TanStack/router/@tanstack/react-start@6421"
```

However, it doesn't seem to resolve the hydratation issue. And now all my links that point to the same page are considered "active", even with `includeHash: true`, and `exact: true`.

nlynzaad:
> Thank you for working on this!
> 
> Edit: I tried to test the PR using pkg.pr.new with:
> 
> ```yaml
> "@tanstack/react-router": "https://pkg.pr.new/TanStack/router/@tanstack/react-router@6421"
> "@tanstack/react-start": "https://pkg.pr.new/TanStack/router/@tanstack/react-start@6421"
> ```
> 
> However, it doesn't seem to resolve the hydratation issue. And now all my links that point to the same page are considered "active", even with `includeHash: true`, and `exact: true`.

picked this up during testing and hence the draft status, digging through it to find a more holistic solution

nlynzaad:
@theoludwig could you test the latest builds and advise if the problem is resolved for you. When including hashes for active options you can remove exact:true as that overrides the hash checks.

theoludwig:
> @theoludwig could you test the latest builds and advise if the problem is resolved for you. When including hashes for active options you can remove exact:true as that overrides the hash checks.

Oh indeed, the issue is now resolved! I needed to do `rm -rf "$(pnpm store path)"` otherwise pnpm wouldn't pick the change from pkg.pr.new.

I end up doing something like this for checking if it needs `exact: true` or `includeHash: true`:
```tsx
const SidebarItemCreate = createLink(SidebarItemRaw)
export const SidebarItem: LinkComponent<React.FC<SidebarItemRawProps>> = (props) => {
  return (
    <SidebarItemCreate
      activeProps={{
        isActive: true,
      }}
      activeOptions={{
        exact: props.hash == null,
        includeHash: props.hash != null,
      }}
      {...props}
    />
  )
}
```

Thank you again! Looking forward to getting this released.

PR #6426, opened 2026-01-20
fix(router-core): fix handle AbortError logic when rapid navigation

### LINKED ISSUE #6388: Aborted loader request renders errorComponent and on rapid param navigation with undefined error

Router

### Describe the bug

When rapidly navigating between the same param route with different params, an aborted loader request causes the route’s errorComponent to render because useLoaderData() is undefined.
Even though a new loader request is sent and resolves successfully, the route never re-renders and remains stuck in the error state.

This happens during fast navigation and appears to be related to how aborted loader requests are handled.

https://stackblitz.com/edit/vitejs-vite-cwhys51i?file=src%2Fmain.tsx

	1.	Start on /
	2.	Click Param 1
	3.	Quickly click Param 2
	4.	Quickly click Param 1 again
	5.	Observe error state

### Expected behavior

Aborted loader requests should not trigger errorComponent.
While a new loader request is in flight, the route should remain pending, and once the latest loader resolves successfully, the route should render normally.

### Screenshots or Videos

https://github.com/user-attachments/assets/5e98c3c0-2c38-4a41-ba6f-4c5c3260156c

### Platform

- Router Version: 1.150.0
- OS: macOS
- Browser: Chrome
- Browser Version: 143.0.7499.170 (Official Build) (arm64)
- Bundler: vite
- Bundler Version: 7.3.1

### Additional context

This is easy to reproduce with:
	•	Slow backend responses
	•	Fast param-based navigation

It looks like an aborted request puts the route into an error state, even though a newer loader request is sent in the background and succeeds.

leesb971204:
fixes #6388 

  Summary

  - Fixed a bug where rapid navigation between parameterized routes caused errorComponent to render incorrectly

  Problem

  When rapidly navigating between parameterized routes (e.g., /something/1 → /something/2 → /something/1), the errorComponent was incorrectly triggered with a TypeError.

  Root cause: The existing AbortError handler set status: 'success' without setting loaderData. When the component tried to access loaderData, it was undefined, causing a TypeError that was caught by the error boundary.

  Solution

  Distinguish between two types of AbortError by checking abortController.signal.aborted:

  | AbortError Type                                 | signal.aborted | Handling                                             |
  |-------------------------------------------------|----------------|------------------------------------------------------|
  | Intentional (loader throws AbortError directly) | false          | Original behavior (set status to 'success')          |
  | Navigation abort (cancelMatches triggered)      | true           | Don't update match state, only resolve loaderPromise |

  Additionally, after awaiting a previous loaderPromise, check if the match status is still 'pending'. If so, the previous loader was aborted and a fresh loader needs to run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed rapid navigation between parameterized routes so aborted loads no longer trigger error UI and final route renders reliably.

* **Tests**
  * Added an end-to-end test reproducer validating rapid param-route navigation does not show error components.
  * Updated existing navigation store-update tests to reflect corrected update counts during navigation.

### DIFF
--- a/packages/react-router/tests/loaders.test.tsx
+++ b/packages/react-router/tests/loaders.test.tsx
@@ -914,23 +914,29 @@ test('reproducer for #6388 - rapid navigation between parameterized routes shoul
   render(<RouterProvider router={router} />)
   await act(() => router.latestLoadPromise)
 
+  const pendingComponent = screen.findByTestId('pending-component')
   expect(await screen.findByTestId('home-page')).toBeInTheDocument()
 
   const param1Link = await screen.findByTestId('link-to-param-1')
   fireEvent.click(param1Link)
+  expect(await pendingComponent).toBeInTheDocument()
 
   const param2Link = await screen.findByTestId('link-to-param-2')
   fireEvent.click(param2Link)
+  expect(await pendingComponent).toBeInTheDocument()
 
   fireEvent.click(param1Link)
+  expect(await pendingComponent).toBeInTheDocument()
 
   await act(() => router.latestLoadPromise)
 
   expect(onAbortMock).toHaveBeenCalled()
   expect(errorComponentRenderCount).not.toHaveBeenCalled()
   expect(screen.queryByTestId('error-component')).not.toBeInTheDocument()
+  expect(await pendingComponent).not.toBeInTheDocument()
 
   const paramPage = await screen.findByTestId('param-page')
   expect(paramPage).toBeInTheDocument()
+  expect(paramPage).toHaveTextContent('Param Component 1 Done')
   expect(loaderCompleteMock).toHaveBeenCalled()
 })


leesb971204:
@schiller-manuel 
can you please review this?

leesb971204:
@nlynzaad 
can you please review this? I think manuel is too busy now.

nlynzaad:
@leesb971204 thanks for this.

I had a look and my only real concern was the change in the control flow on loadRouteMatch.

I've merged with main and made some changes (most just isolating the running of the loader) to address the above.

packages/solid-router/tests/link.test.tsx
```diff
@@ -775,7 +775,7 @@ describe('Link', () => {
       '/posts?page=2&filter=inactive',
     )
 
-    await fireEvent.click(updateSearchLink)
+    fireEvent.click(updateSearchLink)
```
nlynzaad, line 778: This isn't a promise so no need to await it

packages/solid-router/tests/link.test.tsx
```diff
@@ -786,8 +786,11 @@ describe('Link', () => {
     const updatedFilter = await screen.findByTestId('current-filter')
 
     // Verify search was updated
-    expect(window.location.pathname).toBe('/posts')
-    expect(window.location.search).toBe('?page=2&filter=inactive')
+    // Wait for navigation to complete and search params to update
```
nlynzaad, line 789: We should wait for navigation to complete

packages/vue-router/tests/link.test.tsx
nlynzaad, line 1: These are promises and we should wait for these to complete before making assertions

packages/router-core/src/load-matches.ts
```diff
@@ -773,12 +780,17 @@ const loadRouteMatch = async (
...
       if (error) {
-        handleRedirectAndNotFound(inner, match, error)
+        handleRedirectAndNotFound(inner, matchAfterWait, error)
       }
-    } else {
+
+      if (matchAfterWait.status !== 'pending') {
+        return matchAfterWait
+      }
+    }
```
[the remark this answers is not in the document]
nlynzaad: latest commit should have resolved these

leesb971204:
> @leesb971204 thanks for this.
> 
> I had a look and my only real concern was the change in the control flow on loadRouteMatch.
> 
> I've merged with main and made some changes (most just isolating the running of the loader) to address the above.

Awesome! Thnks for this!

PR #6443, opened 2026-01-21
docs(router): URL encoding note on <Link/> and how to opt out

bodadotsh:
Took me weeks to find out about the config option `pathParamsAllowedCharacters`, otherwise I was trying some hack to stop the encoding in the URL path params.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that special characters in URL parameters are URL-encoded by default (e.g., "@" → "%40").
  * Added examples demonstrating how encoded paths appear at runtime.
  * Documented how to opt out or customize parameter encoding via router configuration, including a usage example for the configuration option.

### DIFF
--- a/docs/router/framework/react/api/router/linkComponent.md
+++ b/docs/router/framework/react/api/router/linkComponent.md
@@ -39,11 +39,11 @@ function Component() {
 By default, param values with characters such as `@` will be encoded in the URL:
 
 ```tsx
-// url path will be `/%40foo` 
-<Link to="/$username" params={{ username: "@foo" }} />
+// url path will be `/%40foo`
+<Link to="/$username" params={{ username: '@foo' }} />
 ```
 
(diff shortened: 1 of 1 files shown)


docs/router/framework/react/api/router/linkComponent.md
```diff
@@ -35,3 +35,21 @@ function Component() {
...
 ```
+
+By default, param values with characters such as `@` will be encoded in the URL:
+
+```tsx
+// url path will be `/%40foo` 
+<Link to="/$username" params={{ username: "@foo" }} />
+```
+
+To opt-out, update the `pathParamsAllowedCharacters` config on the router
```
nlynzaad: It's already mentioned in the [api](https://tanstack.com/router/latest/docs/framework/react/api/router/RouterOptionsType#pathparamsallowedcharacters-property) section but it adds value to mention it here. 

Could we change `pathParamsAllowedCharacters` to be a link to the mentioned api section?

PR #6444, opened 2026-01-21
fix(router-core): only rerender Link on search change if search prop is a function

Sheraff:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Link component's search parameter tracking so changes are only tracked when configured, improving update behavior for static vs. dynamic search options.
  * Reduces unnecessary updates and ensures navigation reflects the intended search behavior, leading to more predictable link updates and fewer unexpected re-renders.

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -93,8 +93,7 @@ export function useLinkProps<
 
   // subscribe to search params to re-build location if it changes
   const currentSearch = useRouterState({
-    select: (s) =>
-      options.search ? s.location.search : null,
+    select: (s) => (options.search ? s.location.search : null),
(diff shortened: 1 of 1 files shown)


packages/react-router/src/link.tsx
```diff
@@ -93,7 +93,8 @@ export function useLinkProps<
 
   // subscribe to search params to re-build location if it changes
   const currentSearch = useRouterState({
-    select: (s) => s.location.search,
+    select: (s) =>
+      typeof options.search === 'function' ? s.location.search : null,
```
Sheraff: Maybe this then?
```suggestion
      options.search ? s.location.search : null,
```

PR #6447, opened 2026-01-21
refactor(router-core): skip URL constructor in buildLocation

Sheraff:
`URL` objects are expensive to create and to update. Ideally we should minimize their use in the hot path. This PR proposes 2 changes:
- in `buildLocation` skip creating a `URL` if there is no `rewrite`
- during router initialization, skip creating a `rewrite` if there is no user-provided `basepath` nor any `options.rewrite`

---

Profiling: The purple blocks are `URL > parse` and `set pathname > update`. They are quite expensive to run, and completely gone after this PR (in the specific simple case we're testing, i.e. no `rewrite`, no `basepath`)

Before
Screenshot 2026-01-24 at 10 20 14

After
Screenshot 2026-01-24 at 10 20 26

## Summary by CodeRabbit

* **Refactor**
  * Public navigation surface simplified: location objects now expose href and publicHref and an external flag; internal URL objects removed.
  * Link behavior unified: components use publicHref for external/display URLs and router-created hrefs for in-app navigation.

* **Tests**
  * Test assertions updated to reference href/publicHref and external semantics instead of internal URL objects.

### DIFF
--- a/packages/router-core/src/utils.ts
+++ b/packages/router-core/src/utils.ts
@@ -623,9 +623,9 @@ export function decodePath(path: string, decodeIgnore?: Array<string>): string {
  */
 export function encodeNonAscii(path: string): string {
   // eslint-disable-next-line no-control-regex
-	if (!/[^\u0000-\u007F]/.test(path)) return path
-	// eslint-disable-next-line no-control-regex
-	return path.replace(/[^\u0000-\u007F]/gu, encodeURIComponent)
+  if (!/[^\u0000-\u007F]/.test(path)) return path
+  // eslint-disable-next-line no-control-regex
+  return path.replace(/[^\u0000-\u007F]/gu, encodeURIComponent)
 }
 
 /**


packages/router-core/src/router.ts
```diff
@@ -1883,20 +1883,48 @@ export class RouterCore<
...
+        href,
+        // Lazy URL getter - only constructs URL when accessed
+        // This eliminates URL construction from hot link rendering path
+        get url(): URL {
+          if (!_cachedUrl) {
+            const url = new URL(fullPath, origin)
+            _cachedUrl = rewrite ? executeRewriteOutput(rewrite, url) : url
+          }
+          return _cachedUrl
+        },
```
Sheraff: Do we really need to keep this getter?

packages/react-router/src/link.tsx
```diff
@@ -128,20 +128,27 @@ export function useLinkProps<
...
+      ? next.maskedLocation.publicHref
+      : next.publicHref
+
+    // Check if publicHref is a full URL (different origin from rewrite)
+    const isFullUrl =
+      publicHref.startsWith('http://') || publicHref.startsWith('https://')
+    if (isFullUrl) {
+      // Full URL means rewrite changed the origin - treat as external-like
+      return { href: publicHref, external: false }
+    }
```
Sheraff: Do we really need to "extract" this information? Can't we just make it a part of the Location object?

dsonet:
Less objects created means less memory usage.

Sheraff:
> Less objects created means less memory usage.

i agree, but if the changes aren't reflected in measurable metrics, then something must be wrong

PR #6448, opened 2026-01-21
refactor(router-core): getEnumerableOwnKeys fast path for 'no symbol key' common case, 6-11% faster replaceEqualDeep

Sheraff:
This entire PR is basically just this one line: most objects going through `replaceEqualDeep` won't have symbols, so return early if that's the case:
```ts
if (symbols.length === 0) return names
```

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal key-enumeration to reduce memory allocations for plain objects and improve handling of symbol keys.

* **Tests**
  * Added comprehensive tests for key-enumeration covering symbol keys, non-enumerable properties, mixed scenarios, frozen/sealed objects, and typical router-state patterns.
  * Updated two test expectations to reflect a corrected update count during navigation.


packages/react-router/tests/store-updates-during-navigation.test.tsx
```diff
@@ -241,7 +241,7 @@ describe("Store doesn't update *too many* times during navigation", () => {
     // This number should be as small as possible to minimize the amount of work
     // that needs to be done during a navigation.
     // Any change that increases this number should be investigated.
-    expect(updates).toBe(7)
+    expect(updates).toBe(8)
```
Sheraff: This entire PR is basically just an early return `if (symbols.length === 0) return names` in `replaceEqualDeep`. No test failed in `utils.test`, and all the added tests were added before refactoring, just in case.

And yet... the number of store updates changed... Why?

PR #6464, opened 2026-01-23
refactor(router-core): buildLocation performance by simplifying 'from' location parsing

Sheraff:
The `buildLocation` function needs some information about the `from` location.

Before this PR, to get this information we would build a full match branch through `matchRoutes`, but with a `_buildLocation: true` argument that skipped *some* things.

This PR proposes that we implement a fully separate function that can skip much more work. We also allow for the case where `from` is the current location to short-circuit search validation and params validation.

Before
Screenshot 2026-01-23 at 14 40 34

After
Screenshot 2026-01-23 at 15 04 23

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed the optional build-location flag from route-matching configuration (public interface change).

* **Performance Improvements**
  * Added a lightweight route-matching path to speed up path computation and validation.

* **Bug Fixes**
  * Consolidated parameter parsing and unified error propagation for more consistent handling during route resolution.

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -1570,7 +1570,7 @@ export class RouterCore<
     //   destRoutes: matchedRoutes,
     //   _includeValidateSearch: true,
     // })
-    
+
     // Accumulate search validation through route chain
     const accumulatedSearch = { ...location.search }
     for (const route of matchedRoutes) {
@@ -1599,7 +1599,12 @@ export class RouterCore<
       const strictParams: Record<string, unknown> = { ...routeParams }
       for (const route of matchedRoutes) {
         try {
-          extractStrictParams(route, routeParams, parsedParams ?? {}, strictParams)
+          extractStrictParams(
+            route,
+            routeParams,
+            parsedParams ?? {},
+            strictParams,
+          )
         } catch {
           // Ignore errors, we're not actually routing
         }


packages/router-core/src/router.ts
```diff
@@ -1650,6 +1649,84 @@ export class RouterCore<
...
+      try {
+        const validatedSearch =
+          validateSearch(route.options.validateSearch, {
+            ...accumulatedSearch,
+          }) ?? {}
+        accumulatedSearch = { ...accumulatedSearch, ...validatedSearch }
+      } catch {
+        // Ignore validation errors - they'll be caught later during full matching
+      }
+    }
```
Sheraff: This is very similar to search accumulation in the regular `matchRoutesInternal`. Could we somehow share the code?

packages/router-core/src/router.ts
```diff
@@ -1650,6 +1649,84 @@ export class RouterCore<
...
+        } else if (route.options.skipRouteOnParseError) {
+          // Use pre-parsed params from route matching for skipRouteOnParseError routes
+          for (const key in accumulatedParams) {
+            if (key in parsedParams) {
+              accumulatedParams[key] = parsedParams[key]
+            }
+          }
+        }
+      }
+      params = accumulatedParams
```
Sheraff: This is very similar to search accumulation in the regular `matchRoutesInternal`. Could we somehow share the code?

packages/router-core/src/router.ts
```diff
@@ -1650,6 +1633,73 @@ export class RouterCore<
...
+    // for (const route of matchedRoutes) {
+    //   try {
+    //     Object.assign(
+    //       accumulatedSearch,
+    //       validateSearch(route.options.validateSearch, accumulatedSearch),
+    //     )
+    //   } catch {
+    //     // Ignore errors, we're not actually routing
+    //   }
+    // }
```
Sheraff: @schiller-manuel any idea if we should just run the `validateSearch` or if we need to run the full `applySearchMiddleware` (with `_includeValidateSearch`) when building the `from` data for `buildLocation`?

I do know that we need to run one of them, but idk which one.

PR #6475, opened 2026-01-23
feat: cache routeTree on the server in-between requests

schiller-manuel:
## Summary by CodeRabbit

* **New Features**
  * Server-side route-tree caching to speed up route resolution and reuse processed results.
  * Structured route processing that exposes optimized lookup maps for faster path and ID resolution.

* **Chores**
  * Adjusted router initialization flow to apply cached or newly built route data consistently.

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -38,7 +38,7 @@ import {
   executeRewriteOutput,
   rewriteBasepath,
 } from './rewrite'
-import type { LRUCache } from './lru-cache';
+import type { LRUCache } from './lru-cache'
 import type {
   ProcessRouteTreeResult,
   ProcessedTree,


packages/router-core/src/router.ts
```diff
@@ -1120,9 +1155,17 @@ export class RouterCore<
       },
```
Sheraff, line 1157: is it OK that this won't be called again when reusing the cached process tree?

It seems to be doing a bunch of important things
```ts
this.parentRoute = this.options.getParentRoute?.()
this._path = path as TPath
this._id = id as TId
this._fullPath = fullPath as TFullPath
this._to = trimPathRight(fullPath) as TrimPathRight<TFullPath>
```

schiller-manuel: this sets path, id, to, fullPath of a route. those are static

Sheraff: oh right, and they are on the `routeTree` itself, which *is* what survives across requests anyway, regardless of this PR.

Ok so it is fine then

CBell045:
I'm using oRPC and after upgrading from 1.155.0 to 1.156.0 my server HMR doesn't seem to be working; I have to quit and restart the dev server for any of my route changes to take effect. Do you think it could be due to this PR?

PR #6479, opened 2026-01-24
fix(react-router): remove async/defer attributes during hydration to prevent warnings

### LINKED ISSUE #6438: Assets: async script triggers hydration mismatch

Router

### Describe the bug

I believe there is a regression with this https://github.com/TanStack/router/pull/5980/files or maybe the one before that https://github.com/TanStack/router/pull/5895/files#diff-5319ec26839a90fd840b11ecc7f23fa4b8ff5c311338056cd032204f8448ee45R146-R154.
By specifying "async: true", it will trigger the hydration warning.

Image

https://stackblitz.com/edit/github-52nzeepx

I tried to reproduce with vite/react (no tanstack router) and simple renderToPipeableStream & hydrateRoot. It gives the same result
```tsx
function AppFailedToSuppressHydrationWarning() {
  if (typeof window === 'undefined') {
    return <script src="https://accounts.google.com/gsi/client" async></script>
  }

  return <script dangerouslySetInnerHTML={{ __html: "" }} suppressHydrationWarning={true} async></script>
}

function AppSucceededToSuppressHydrationWarning() {
  if (typeof window === 'undefined') {
    return <script src="https://accounts.google.com/gsi/client"></script>
  }

  return <script dangerouslySetInnerHTML={{ __html: "" }} suppressHydrationWarning={true}></script>
}

function AppHoisted() {
  return <div>
    This should belong to the body

    {/* this will go to the head with renderToPipeableStream */}
    <script src="https://accounts.google.com/gsi/client" async={true}></script>
  </div>
}

export default AppFailedToSuppressHydrationWarning
 ```

The suppressHydrationWarning doesn't seem to work with async script.

### Expected behavior

No hydration mismatch warning

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.141.1
- OS: Ubuntu 24.04.3 LTS
- Browser: Chrome
- Browser Version: Version 143.0.7499.192 (Official Build) (64-bit)
- Bundler: vite
- Bundler Version: 7.2.7

### Additional context

Already reported here:
https://discord.com/channels/719702312431386674/1457054816831541289

### PR DESCRIPTION
**Pull request by Kyujenius**
### Description

Fixes #6438

When using `async` or `defer` attributes on script tags via `useHead` or `Scripts`, React throws a hydration warning even if `suppressHydrationWarning` is used. This is because of how React handles hydration mismatches for specific attributes on script tags during streaming/suspense boundaries.

This PR solves the root cause by stripping `async`, `defer`, and `src` attributes from the script tag **on the client-side only** (during hydration).

- **Server**: Renders full `<script src="..." async />` to ensure the browser loads it.
- **Client**: Renders a placeholder `<script />` (via `Asset.tsx`) that matches the server's hydration expectation for a controlled script tag, effectively preventing the mismatch warning.

### Solution

The `Asset` component now explicitly destructures and removes these attributes when `!router.isServer`:

```tsx
if (!router.isServer) {
  const { src: _src, async: _async, defer: _defer, ...rest } = attrs || {}
  return (
    <script
      suppressHydrationWarning
      dangerouslySetInnerHTML={{ __html: '' }} // Explicitly empty
      {...rest}
    ></script>
  )
}
```

image

### Verification

New tests added to cover these scenarios:
1.  **E2E (`script-duplication.spec.ts`)**: Verifies that no hydration warnings are logged to the console when loading a page with `async` scripts.
2.  **Unit (`Scripts.test.tsx`)**: Verifies that the server output correctly includes `async`/`defer` attributes (using regex to be resilient to attribute ordering).

## Summary by CodeRabbit

* **New Features**
  * Added an `/async-scripts` route demonstrating asynchronous and deferred script loading.
  * Included navigation link to the new async-scripts page.

* **Bug Fixes**
  * Fixed script attribute handling to properly support `async` and `defer` attributes without hydration warnings.

* **Tests**
  * Added test suite for async/defer script loading and hydration validation.


hiendaovinh:
My question is, why do we have to strip them and render a placeholder <script /> in the first place?

Issue #6455, opened 2026-01-22
Hydration error on direct navigation to non-root routes in SPA mode


bbertold:
### Which project does this relate to?

Start

### Describe the bug

When deploying a TanStack Start application with `spa` mode enabled to Cloudflare, directly navigating to a non-root route (e.g., `/example`) results in a React hydration error. The same route works correctly when accessed via client-side navigation (e.g., clicking a `<Link>` from `/`).

On direct navigation to a non-root route, React throws a hydration mismatch error and the page visibly flashes:
```
Hydration failed because the server rendered HTML didn't match the client. 
As a result this tree will be regenerated on the client.
```

Error reference: https://react.dev/errors/418?args[]=HTML&args[]=

This issue appears to be directly related to Cloudflare's SPA handling configuration:
```json
"not_found_handling": "single-page-application"
```

### Your Example Website or App

https://github.com/bbertold/tanstack-start-spa

### Steps to Reproduce the Bug or Issue

1. Clone the reproduction repository
2. Build the application: `pnpm build`
3. Preview the build: `pnpm preview`
4. Open browser and navigate directly to `/example` in the URL bar
5. Press Enter
6. Observe the hydration error in console and visible flash of content

### Expected behavior

The page should load without hydration errors.

### Platform

- Router / Start Version: 1.154.8
- OS: macOS
- Browser: Chrome (144.0.7559.59)
- Bundler: Vite (7.1.7)
- Wrangler version: 4.60.0

pablor21:
I'm facing the same error, with caddy as a server.  Only happens, as @bbertold  says, wen navigating directly to a child route, not on client side navigation.

kyu-rong:
Hi @bbertold and @pablor21!

I've been looking into this issue and recently submitted a PR (#6481) that I hope might help resolve this.

From my analysis, it seems the root cause is that in SPA mode, servers (like Cloudflare or Caddy) serve the root index.html (the shell) for all direct navigations to child routes. React then tries to hydrate this root HTML as if it were the specific child route's content, which causes the mismatch error and the visible flash.

In the PR, I've proposed a small fix that:

Injects a marker (window.__TSS_SPA_SHELL__) into the SPA shell during build. Note: I adopted the window.__TSS_* naming convention already used in the codebase (e.g., window.__TSS_START_OPTIONS__) to ensure consistency with the existing architecture.
Allows the client to detect this marker and safely skip hydration if the path doesn't match, performing a fresh client-side render instead.
[...]

PR #6481, opened 2026-01-24
fix(start): handle Cloudflare SPA fallback hydration mismatch

Kyujenius:
## Problem
When deploying a TanStack Start application to Cloudflare Pages with SPA mode enabled, directly navigating to non-root routes (e.g., `/example`) causes React hydration errors.
**Root Cause:**
- Cloudflare's SPA fallback serves `index.html` (app shell) for all routes
- Client expects to hydrate the deep route (e.g., `/example`)
- Server HTML contains the root shell content
- Mismatch causes hydration error and visible content flash
**Reproduction:**
See Issue #6455 and <https://github.com/bbertold/tanstack-start-spa>
## Solution
Implemented a two-part fix:
### 1. Build-time Marker Injection (`start-plugin-core`)
- Inject `window.__TSS_SPA_SHELL__ = true` into the SPA shell during prerendering
- Only applied to the shell file (identified by `spa.prerender.outputPath`)
- Robust injection with fallback: tries `</head>` first, then `<body>` if not found
### 2. Client-side Detection (`start-client-core`)
- Check for `__TSS_SPA_SHELL__` marker before hydration
- If marker exists AND current path ≠ basepath, skip server hydration
- Let RouterProvider mount the router freshly (client render)

image

This fix is server-agnostic and will resolve similar hydration issues on servers like Caddy, Nginx, or Vercel when configured for SPA fallback 😄 👍 

### Unit Tests (11 new tests)
- Normal hydration scenarios (no marker, root path)
- SPA shell fallback detection (deep links, trailing slashes)
- Custom basepath handling (3 scenarios)
- Existing matches handling (2 scenarios)
- Router configuration verification (2 tests)
All tests passing ✅

## Summary by CodeRabbit

* **New Features**
  * SPA shell support: client now conditionally skips or performs hydration based on a runtime SPA-shell marker and the current URL path (respecting basepath).
  * Prerender now injects a SPA-shell marker into HTML to enable the above behavior.

* **Tests**
  * Added comprehensive tests covering hydration fallback scenarios, basepath handling, existing router matches, and adapter merging.

### DIFF
--- a/packages/start-client-core/src/client/hydrateStart.ts
+++ b/packages/start-client-core/src/client/hydrateStart.ts
@@ -1,4 +1,7 @@
 import { hydrate } from '@tanstack/router-core/ssr/client'
+import {
+  trimPathRight,
+} from '@tanstack/router-core'
 
 import { ServerFunctionSerializationAdapter } from './ServerFunctionSerializationAdapter'
 import type { AnyStartInstanceOptions } from '../createStart'
@@ -44,7 +47,8 @@ export async function hydrateStart(): Promise<AnyRouter> {
   if (!router.state.matches.length) {
     if (
       window.__TSS_SPA_SHELL__ &&
-      window.location.pathname !== (router.options.basepath || '/')
+      trimPathRight(window.location.pathname) !==
+        trimPathRight(router.options.basepath || '/')
     ) {
       // If we are loading the SPA shell (index.html) and the path is not the root,
       // it means we are in a fallback scenario (e.g. Cloudflare SPA fallback).
--- a/packages/start-client-core/src/client/hydrateStart.ts
+++ b/packages/start-client-core/src/client/hydrateStart.ts
@@ -8,6 +8,12 @@ import { getRouter } from '#tanstack-router-entry'
 // eslint-disable-next-line import/no-duplicates,import/order
 import { startInstance } from '#tanstack-start-entry'
 
+declare global {
+  interface Window {
+    __TSS_SPA_SHELL__?: boolean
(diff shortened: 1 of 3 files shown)


schiller-manuel:
why do we need this? what exactly is mismatching here? we should already handle this correctly in router IMO

bbertold:
I debugged this with a minimal repro I made on the latest version v1.157.16: https://github.com/bbertold/tanstack-start-spa. The problem is that with SPA mode in Start and Cloudflare, the same HTML is served for all routes. When loading any route, the HTML contains:
```
matches: [
  { i: "__root__/", ... },
  { i: "//", ... }
],
```
This causes `router.state.matches.length` to be truthy all the time, so the existing check passes and hydration proceeds even when the current route is not `/`. This leaves the router attempting to hydrate the non-root route with root route content. This is the mismatch, you can see the visual flash of the root content before correction (especially with dev tools CPU slowdown).
The issue is that the current check only verifies matches exist, not whether they match the current URL. This also explains why client-side navigation is unaffected.

bbertold:
Hey @schiller-manuel, just wanted to follow up on this. To be sure, I've updated the repro today to the latest versions of TanStack Start and Wrangler, and I'm still experiencing the SPA hydration issue.  I completely understand if there’s a lot going on and I’m not trying to rush anything. This is currently blocking our deployment though, so I mainly wanted to check whether the general approach in this PR seems off to you, or if it’s possible that I’m misconfiguring something on my end. Thanks a lot for taking the time to look into this!

PR #6482, opened 2026-01-24
refactor: react link ssr perf

schiller-manuel:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed hydration mismatches by returning stable, no-handler anchor props during server rendering and applying transition/navigation props only after hydration.

* **Improvements**
  * Centralized server-aware URL handling and safer external-link resolution (blocks dangerous protocols).
  * More consistent active-link detection between server and client (search/hash handling).
  * Consolidated class/style merging and streamlined preload/intersection behavior.

* **Note**
  * No public API signature changes.

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -261,6 +261,10 @@ export function useLinkProps<
       }
 
       // Hash is not available on the server
+      if (activeOptions?.includeHash) {
+        return false
+      }
+
       return true
     })()
 
@@ -504,8 +508,8 @@ export function useLinkProps<
         }
       }
 
-      if (isHydrated && activeOptions?.includeHash) {
-        return s.location.hash === next.hash
+      if (activeOptions?.includeHash) {
+        return isHydrated && s.location.hash === next.hash
       }
       return true
     },


packages/react-router/src/link.tsx
```diff
@@ -93,6 +94,279 @@ export function useLinkProps<
...
+
+    const next = router.buildLocation({ ...options, from: options.from } as any)
+
+    const hrefOption = (() => {
+      if (disabled) {
+        return undefined
+      }
+      let href = next.maskedLocation
+        ? next.maskedLocation.url.href
+        : next.url.href
```
Sheraff: Can we use something other than `.url` for this? I'm removing it in https://github.com/TanStack/router/pull/6447

schiller-manuel: possibly? public href?

packages/react-router/src/link.tsx
Sheraff, line 1: should this (can this) be applied to solid and vue too?

schiller-manuel: I think so yes, needs @birkskyum probably

Issue #5031, opened 2025-08-28
File-based routing generates URLs with spaces when folder names contain spaces


yhoungdev:
### Which project does this relate to?

Router

### Describe the bug

When using file-based routing in TanStack Router, if a folder name contains a space, the generated route includes %20 in the URL. While technically correct, it leads to awkward and less user-friendly URLs. It might be better to disallow spaces in folder names or automatically normalize them (e.g., replace with - or _) instead of producing %20 in the path.

Yeah, i know by default the browser will encode the space " " to %20 as seen in the first image 
<img width="576" height="82" alt="Image" src="https://github.com/user-attachments/assets/34e5c9f3-6561-42cf-8ec2-8b9f7cbba39d" />

Here is the codebase. 
<img width="709" height="254" alt="Image" src="https://github.com/user-attachments/assets/12fc272f-63c3-4832-8da2-c2f77bb89788" />

Question / Suggestion:
Would this be considered a bug or a feature request?
It might be useful if TanStack Router could either:

Warn or disallow spaces in folder names, or

Provide an option (e.g., via vite.config.ts) to normalize spaces automatically (e.g., convert to - or _), instead of producing %20 in the path.

### Your Example Website or App

it is closed source 

### Steps to Reproduce
1. Under the folder `routes/` with a space in its name (e.g. ` folder/`).
2. Add an `index.tsx` inside it.
3. Start the dev server and open the route.

**Result:** URL becomes `/%20folder`.  
**Expected:** Should warn or normalize instead of `%20`.

### Expected Behavior
Instead of producing `%20` in the URL, TanStack Router should either:  
- Warn or disallow spaces in folder names, or  
- Normalize spaces automatically (e.g., replace with `-` or `_`).  

### Platform

- Router / Start Version: ```v1.129.7```
- OS: Mac 
- Browser: Chrome and Dia
- Bundler:  Vite
- Bundler Version: ^6.0.5

schiller-manuel:
we could think of some hook that gets called by the generator, but TBH i don't think this is a common issue

theobiabo:
Yea, its not a common iss, but can actually be an oversight.

nlynzaad:
@yhoungdev encoding spaces as %20 for URLs is in line with the [URL standards](https://url.spec.whatwg.org/). Not encoding the space would result in an invalid URL.

[...]

PR #6520, opened 2026-01-26
fix(router-core): use encodeURIComponent for splat route params

### LINKED ISSUE #6519: SSR redirect loop for splat routes with URL-encoded characters (%20, parentheses)
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

When using splat routes (/$) with URLs containing encoded special characters (e.g., %20 for spaces, %28/%29 for parentheses), SSR causes an infinite 307 redirect loop.

Root cause: The encodeParam function in path.ts uses encodeURI() for splat params, which doesn't encode spaces or parentheses. When comparing the original URL (/path/file%20name.pdf) with the rebuilt URL (/path/file name.pdf), they don't match, triggering a redirect with unencoded characters in the Location header.

Reproduction:
1. Create a splat route ($.tsx)
2. Navigate to a URL with spaces: /owner/repo/blob/main/my%20file.pdf
3. Refresh the page
4. Observe infinite 307 redirect loop

Fix: Replace encodeURI(value) with value.split('/').map(segment => encodeURIComponent(segment)).join('/') to properly encode each path segment.

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@tanstack/router-core@1.157.3` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@tanstack/router-core/dist/esm/path.js b/node_modules/@tanstack/router-core/dist/esm/path.js
index b178989..af2c810 100644
--- a/node_modules/@tanstack/router-core/dist/esm/path.js
+++ b/node_modules/@tanstack/router-core/dist/esm/path.js
@@ -120,7 +120,9 @@ function encodeParam(key, params, decoder) {
   const value = params[key];
   if (typeof value !== "string") return value;
   if (key === "_splat") {
-    return encodeURI(value);
+    // Use encodeURIComponent for each segment to properly encode spaces and special characters
+    // but preserve forward slashes as path separators
+    return value.split('/').map(segment => encodeURIComponent(segment)).join('/');
   } else {
     return encodePathParam(value, decoder);
   }
diff --git a/node_modules/@tanstack/router-core/dist/esm/utils.js b/node_modules/@tanstack/router-core/dist/esm/utils.js
index d966d48..1b1469c 100644
--- a/node_modules/@tanstack/router-core/dist/esm/utils.js
+++ b/node_modules/@tanstack/router-core/dist/esm/utils.js
@@ -219,8 +219,24 @@ function decodePath(path, decodeIgnore) {
   return result;
 }
 function encodeNonAscii(path) {
-  if (!/[^\u0000-\u007F]/.test(path)) return path;
-  return path.replace(/[^\u0000-\u007F]/gu, encodeURIComponent);
+  // Encode both non-ASCII characters and special characters that should be encoded in URLs
+  // Split by / to preserve path separators, encode each segment
+  return path.split('/').map(segment => {
+    // Encode special characters: spaces, parentheses, brackets, etc.
+    if (/[\s()[\]{}#]/.test(segment)) {
+      return segment.split('').map(char => {
+        if (/[\s()[\]{}#]/.test(char)) {
+          return encodeURIComponent(char);
+        }
+        return char;
+      }).join('');
+    }
+    // Also encode non-ASCII
+    if (/[^\u0000-\u007F]/.test(segment)) {
+      return segment.replace(/[^\u0000-\u007F]/gu, encodeURIComponent);
+    }
+    return segment;
+  }).join('/');
 }
 function buildDevStylesUrl(basepath, routeIds) {
   const trimmedBasepath = basepath.replace(/^\/+|\/+$/g, "");

```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

thejasonxie:
Summary
Fixes SSR redirect loop for splat routes with URL-encoded characters (spaces, plus signs, brackets, etc.)
 Problem
When a splat route contains URL-encoded characters like `%20` (space), `%2B` (plus), or `%5B` (bracket):
1. User navigates to `/docs/file%20name.pdf`
2. TanStack Router decodes the `_splat` param using `decodeURIComponent` → `file name.pdf`
3. When rebuilding the URL for SSR, the router used `encodeURI()` which doesn't encode spaces
4. This produces `/docs/file name.pdf` instead of `/docs/file%20name.pdf`
5. URL mismatch triggers a 307 redirect loop
 Solution
Changed `encodeParam()` in `path.ts` to use `encodeURIComponent` for each path segment instead of `encodeURI` on the entire splat value:
// Before
return encodeURI(value)
// After  
return value.split('/').map((segment) => encodeURIComponent(segment)).join('/')
This properly encodes spaces (%20), plus signs (%2B), brackets (%5B/%5D), and other special characters that encodeURI leaves unencoded, while still preserving forward slashes as path separators.
Changes
- packages/router-core/src/path.ts: Updated encodeParam() function
- packages/router-core/tests/path.test.ts: Added tests for splat params with special characters
- packages/router-core/tests/utils.test.ts: Added tests for encodeNonAscii() (for coverage)
Testing
All 1030 existing tests pass, plus 18 new tests for special character encoding.

Resolves #6519

## Summary by CodeRabbit

* **Bug Fixes**
  * Splat path segments now encode per segment with a fast-path for already-safe strings, fixing encoding for spaces, plus signs, "@", hashes, brackets, etc.

* **New Features**
  * Added a public utility to percent‑encode whitespace and non‑ASCII characters in path-like URLs; exported utility name updated.

* **Tests**
  * Added and updated tests covering splat edge cases, whitespace/non‑ASCII/emoji encoding, and stricter "@" percent‑encoding expectations.


nlynzaad:
@thejasonxie thanks for this. 

I agree with the change to treat the various path segments for splat params similar to how path params is treated. encodePathParam uses encodeURIComponent underlying so made a change to rather use this and standardise across both path params and splat params.

There were a few tests that failed, these were expected due to the change from encodeURI to encodeURIComponent and I updated those accordingly.

nlynzaad:
just a quick note.

this does not resolve [#5031](https://github.com/TanStack/router/issues/5031). 

The problem there was in fact that a space in a file-based route name results in a %20 i.e. ` test.tsx` becomes `/%20test` and `test me.tsx` becomes `/test%20me`. This handling is correct and your problem highlights exactly why this should be happening, i.e. spaces in paths should be encoded as %20 in order to follow URL standards.

packages/router-core/src/path.ts
```diff
@@ -248,7 +248,12 @@ function encodeParam(
 
   if (key === '_splat') {
     // the splat/catch-all routes shouldn't have the '/' encoded out
-    return encodeURI(value)
+    // Use encodeURIComponent for each segment to properly encode spaces,
+    // plus signs, and other special characters that encodeURI leaves unencoded
+    return value
```
nlynzaad, line 255: @Sheraff just checking with you if there are any performance implications or improvements you can think of here.

It is a fairly hot path so just want to ensure we aren't undoing some of the hard work you have been putting in.

packages/router-core/src/path.ts
```diff
@@ -248,7 +248,12 @@ function encodeParam(
...
+      .split('/')
+      .map((segment) => encodePathParam(segment, decoder))
+      .join('/')
```
Sheraff, line 258: Because this is more expensive to run than before, and this tends to be in the hot path of most things in Router, could we add an early return check?

Maybe something like this? (untested!)
```ts
if (/^[a-zA-Z0-9\-._~!/]*$/.test(value)) return value
```

Sheraff:
The issue (https://github.com/TanStack/router/issues/6519) and the added tests seem to imply that there is an issue with `encodeNonAscii` but I do not see it modified in this PR. Was it not actually part of the problem?

packages/router-core/src/path.ts
thejasonxie: I've added the early return check

thejasonxie:
> The issue (#6519) and the added tests seem to imply that there is an issue with `encodeNonAscii` but I do not see it modified in this PR. Was it not actually part of the problem?

Yes, encodeNonAscii is part of the problem. Here's the flow that causes the SSR redirect loop:
1. Request URL: /path/file%20name.pdf
2. parseLocation decodes pathname → /path/file name.pdf (actual space)
3. SSR calls buildLocation({ to: latestLocation.pathname })
4. Since pathname doesn't contain $, interpolatePath returns it unchanged
5. encodeNonAscii(fullPath) is called, but spaces are ASCII (0x20), so they weren't encoded
6. nextLocation.publicHref = /path/file name.pdf
7. Compare: /path/file%20name.pdf !== /path/file name.pdf → redirect triggered → loop

I've added a second commit that updates encodeNonAscii to also encode whitespace characters. The fix:
- Encodes whitespace (spaces → %20, tabs → %09, etc.)
- Encodes non-ASCII/unicode characters (including emojis)
- Preserves other ASCII special characters (parentheses, brackets, etc.) as they're valid in URL paths

Both fixes are needed:
- encodeParam fix handles the $ splat route param case
- encodeNonAscii fix handles the direct pathname case in buildLocation

thejasonxie:
Issue happens from accessing /file-name%20(3).pdf? in the url then refreshing which crashes

Updated patch to fix my issue:

```patch
diff --git a/node_modules/@tanstack/router-core/dist/esm/path.js b/node_modules/@tanstack/router-core/dist/esm/path.js
index 114353f..212f57d 100644
--- a/node_modules/@tanstack/router-core/dist/esm/path.js
+++ b/node_modules/@tanstack/router-core/dist/esm/path.js
@@ -121,7 +121,11 @@ function encodeParam(key, params, decoder) {
   const value = params[key];
   if (typeof value !== "string") return value;
   if (key === "_splat") {
-    return encodeURI(value);
+    // Early return if value only contains URL-safe characters (performance optimization)
+    if (/^[a-zA-Z0-9\-._~!/]*$/.test(value)) return value;
+    // Use encodeURIComponent for each segment to properly encode spaces and special characters
+    // but preserve forward slashes as path separators
+    return value.split('/').map(segment => encodeURIComponent(segment)).join('/');
   } else {
     return encodePathParam(value, decoder);
   }
diff --git a/node_modules/@tanstack/router-core/dist/esm/utils.js b/node_modules/@tanstack/router-core/dist/esm/utils.js
index ebada3c..b5f5ca5 100644
--- a/node_modules/@tanstack/router-core/dist/esm/utils.js
+++ b/node_modules/@tanstack/router-core/dist/esm/utils.js
@@ -226,8 +226,10 @@ function decodePath(path, decodeIgnore) {
   return result;
 }
 function encodeNonAscii(path) {
-  if (!/[^\u0000-\u007F]/.test(path)) return path;
-  return path.replace(/[^\u0000-\u007F]/gu, encodeURIComponent);
+  // Encode non-ASCII characters and whitespace that should be encoded in URLs
+  // eslint-disable-next-line no-control-
+  // biome-ignore lint/suspicious/noControlCharactersInRegex: intentional ASCII range check
+  if (!/\s|[^\u0000-\u007F]/.test(path)) return path;
+  return path.replace(/\s|[^\u0000-\u007F]/gu, encodeURIComponent);
 }
 function buildDevStylesUrl(basepath, routeIds) {
   const trimmedBasepath = basepath.replace(/^\/+|\/+$/g, "");

```

packages/router-core/src/utils.ts
```diff
@@ -636,10 +636,27 @@ export function decodePath(path: string, decodeIgnore?: Array<string>): string {
...
+      // Encode whitespace and non-ASCII characters
+      // Use spread operator to properly handle multi-byte unicode characters (emojis)
+      return [...segment]
+        .map((char) => {
+          if (needsEncoding.test(char)) {
+            return encodeURIComponent(char)
+          }
+          return char
+        })
+        .join('')
```
Sheraff: Why do we need char-by-char encoding? Can't we just `return encodeURIComponent(segment)`?

thejasonxie: `encodeURIComponent` encodes all characters except: A-Z a-z 0-9 - _ . ! ~ * ' ( )

This means it would encode characters like @, $, &, +, ,, ;, =, : which are valid in URL paths and were not being encoded before.

The function is named encodeNonAscii and its purpose is specifically to encode:
- Whitespace (spaces → %20, tabs → %09)
- Non-ASCII/Unicode (emojis, accented characters, etc.)

Using `encodeURIComponent(segment)` would be a breaking change - paths like `/user@domain/file` would become `/user%40domain/file`.

However, the char-by-char approach can be simplified. Instead of:
```ts
return [...segment]
  .map((char) => needsEncoding.test(char) ? encodeURIComponent(char) : char)
  .join('')
```

We can use a regex replace:
```ts
return segment.replace(/\s|[^\u0000-\u007F]/gu, encodeURIComponent)
```

thejasonxie: switched to regex

Sheraff: Yeah I think that `encodeNonAscii` is poorly named, it's goal is to "encode like `new URL` would, but without requiring all the parsing that `new URL` does".

thejasonxie: Renamed it to `encodePathLikeUrl` and updated jsdoc

PR #6527, opened 2026-01-27
support trusted types enforcement in react-start

hybrist:
Currently `@tanstack/react-start` fails during route transitions when trusted types are enabled. This makes it impossible to use it with strict CSPs that enable all the "XSS-prevention bells and whistles".

### DIFF
--- a/e2e/react-start/csp/src/routes/__root.tsx
+++ b/e2e/react-start/csp/src/routes/__root.tsx
@@ -1,9 +1,29 @@
+/// <reference types="vite/client" />
+
 import {
   createRootRoute,
   HeadContent,
   Outlet,
   Scripts,
 } from '@tanstack/react-router'
+
+export const requiresTrustedTypes: boolean =
+  import.meta.env.VITE_CSP === 'strict'
+
+declare const trustedTypes: any
+
+function createPolicy<T>(name: string, policy: T): T {
+  if (!requiresTrustedTypes) return policy
+
+  if (typeof trustedTypes !== 'undefined') {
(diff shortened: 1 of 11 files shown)


e2e/react-start/csp/src/routes/__root.tsx
```diff
@@ -1,10 +1,30 @@
...
   createRootRoute,
   HeadContent,
   Outlet,
   Scripts,
 } from '@tanstack/react-router'
 
+export const requiresTrustedTypes: boolean =
+  import.meta.env.VITE_CSP === 'strict'
+
+declare const trustedTypes: any
```
hybrist, line 13: These types should make it into TS proper soon™ given that there's now two engines (Safari + Chromium) that ship the feature.

hybrist:
@schiller-manuel Updated with an integration into the existing CSP test. The checked in test results show the general issue (client-side script creation after hydration).

packages/react-router/src/utils.ts
```diff
@@ -123,3 +123,22 @@ export function useForwardedRef<T>(ref?: React.ForwardedRef<T>) {
...
+
+interface TanStackPolicy {
+  createHTML: (s: string) => string
+  createScript: (s: string) => string
+  createScriptURL: (s: string) => string
+}
+
+declare const trustedTypes: any
+
+const tanStackPolicy: TanStackPolicy = {
```
hybrist, line 135: NOTE: This policy should really be locked down, at least to some extent. E.g. `createHTML` shouldn't just bless _any_ string when its purpose is to bless static strings like `''`.

PR #6542, opened 2026-01-28
feat(router-core, react-router, solid-router, vue-router): isDangerousProtocol uses customizable allowlist

Sheraff:
- there can be an infinity of protocols (custom protocols are a thing)
- `blob:` and `data:` are widely used protocols (so we should have a way to allow them), but they can also be considered dangerous if displaying untrusted links (so we should have a way to block them)
- an allowlist is usually a better security practice than a blocklist
- causing security issues is bad (block everything)
- breaking links on the web is bad (allow everything)

List of CVEs about unsafe URL protocols:
- https://www.cve.org/CVERecord?id=CVE-2025-57665 (same as us, mentions javascript:, data:, file:)
- https://www.cve.org/CVERecord?id=CVE-2025-31477 (tauri, mentions file:, smb:, or nfs:)
- https://www.cve.org/CVERecord?id=CVE-2025-56200 (improper parsing, i think we're good here)
- https://www.cve.org/CVERecord?id=CVE-2024-37359 (server side, might not be relevant to us)
- https://www.cve.org/CVERecord?id=CVE-2024-34343 (nuxt, improper parsing, we're good here)
- https://www.cve.org/CVERecord?id=CVE-2024-28175 (mentions javascript:)

## Summary by CodeRabbit

* **New Features**
  * Added a configurable protocolAllowlist option and exposed DEFAULT_PROTOCOL_ALLOWLIST for React, Solid, and Vue packages.

* **Behavior Changes**
  * Link, navigation, and redirect protocol checks now honor the router allowlist; redirects with disallowed protocols are rejected and warnings updated.

* **Documentation**
  * Docs updated with protocolAllowlist usage, defaults, and examples.

* **Tests**
  * Expanded tests for allowed/blocked protocols, custom allowlists, and integration with router behavior.


packages/router-core/src/redirect.ts
```diff
@@ -124,17 +123,6 @@ export function redirect<
...
-  if (
-    !opts._builtLocation &&
-    typeof opts.href === 'string' &&
-    isDangerousProtocol(opts.href)
-  ) {
-    throw new Error(
-      `Redirect blocked: unsafe protocol in href "${opts.href}". Only ${SAFE_URL_PROTOCOLS.join(', ')} protocols are allowed.`,
-    )
-  }
-
```
Sheraff, line 137: We don't need to check here, since a `redirect()` is always handled by something else that *will* check, right?

hiendv:
I believe I have the same problem as yours. `x-safari-https` stops working for us after the update.
But from my POV, the whitelist works better than the blacklist. The current default whitelist is reasonable and works with most cases. We just need a way to config additional allowed protocols.

docs/router/framework/react/api/router/RouterOptionsType.md
```diff
@@ -144,6 +144,41 @@ The `RouterOptions` type accepts an object with the following properties and met
 - When `true`, disables the global catch boundary that normally wraps all route matches. This allows unhandled errors to bubble up to top-level error handlers in the browser.
 - Useful for testing tools, error reporting services, and debugging scenarios.
 
+### `protocolBlocklist` property
```
hiendv: I think we better stick with a sensible default with whitelist
e.g. `['http:', 'https:', 'mailto:', 'tel:'`

And allow users to specify `protocolWhitelist` or `protocolAllowList`

docs/router/framework/react/api/router/RouterOptionsType.md
```diff
@@ -144,6 +144,41 @@ The `RouterOptions` type accepts an object with the following properties and met
...
+
+- Type: `Array<string>`
+- Optional
+- Defaults to `DEFAULT_PROTOCOL_BLOCKLIST` which includes:
```
hiendv: Then we wouldn't need `DEFAULT_PROTOCOL_BLOCKLIST`

packages/react-router/src/link.tsx
hiendv, line 1: Can you add a test case for `<Link />`?

Something like this

```tsx
describe('weird links', () => {
  test('should work', async () => {
    const rootRoute = createRootRoute()
    const inputs = [
      'x-safari-https://example.com',
      'googlechromes://example.com',
      'intent://example.com#Intent;scheme=https;end',
    ]
    const indexRoute = createRoute({
      path: '/',
      component: () => {
        return inputs.map((input) => <Link to={input} key={input} />)
      },
    })

    const router = createRouter({
      routeTree: rootRoute.addChildren([indexRoute]),
      history,
    })

    render(<RouterProvider router={router} />)

    const links = await screen.findAllByRole('link')
    const hrefs = links.map((el) => el.getAttribute('href'))
    expect(hrefs).toBe(inputs)
  })
})
```

packages/router-core/src/router.ts
```diff
@@ -2241,9 +2256,9 @@ export class RouterCore<
       // otherwise use href directly (which may already include basepath)
       const reloadHref = !hrefIsUrl && publicHref ? publicHref : href
 
-      // Block dangerous protocols like javascript:, data:, vbscript:
+      // Block dangerous protocols like javascript:, blob:, data:
```
hiendv, line 2266: Why do we have to change this?

packages/router-core/src/utils.ts
```diff
@@ -536,14 +536,23 @@ function decodeSegment(segment: string): string {
...
-export const SAFE_URL_PROTOCOLS = ['http:', 'https:', 'mailto:', 'tel:']
+export const DEFAULT_PROTOCOL_ALLOWLIST = [
+  // Standard web navigation
+  'http:',
+  'https:',
+
+  // Common browser-safe actions
+  'mailto:',
+  'tel:',
+  'sms:',
```
Sheraff:
```suggestion
```

docs/router/framework/react/api/router/RouterOptionsType.md
```diff
@@ -144,6 +144,37 @@ The `RouterOptions` type accepts an object with the following properties and met
 - When `true`, disables the global catch boundary that normally wraps all route matches. This allows unhandled errors to bubble up to top-level error handlers in the browser.
 - Useful for testing tools, error reporting services, and debugging scenarios.
 
+### `protocolAllowlist` property
+
+- Type: `Array<string>`
+- Optional
+- Defaults to `DEFAULT_PROTOCOL_ALLOWLIST` which includes:
+  - Web navigation: `http:`, `https:`
+  - Common browser-safe actions: `mailto:`, `tel:`, `sms:`
```
Sheraff:
```suggestion
  - Common browser-safe actions: `mailto:`, `tel:`
```

packages/router-core/src/router.ts
```diff
@@ -470,6 +471,15 @@ export interface RouterOptions<
    */
   disableGlobalCatchBoundary?: boolean
 
+  /**
+   * An array of URL protocols to allow in links, redirects, and navigation.
+   * Absolute URLs with protocols not in this list will be rejected.
+   *
+   * @default DEFAULT_PROTOCOL_ALLOWLIST (http:, https:, mailto:, tel:, sms:)
```
Sheraff:
```suggestion
   * @default DEFAULT_PROTOCOL_ALLOWLIST (http:, https:, mailto:, tel:)
```

Sheraff (commented):
remove `:sms` protocol from defaults

PR #6561, opened 2026-01-31
fix(react-router): prevent stale closures in useMatchRoute with React Compiler

### LINKED ISSUE #4499: useMatchRoute not updating when using React Compiler

Router

### Describe the bug

When using the function returned by `useMatchRoute` in combination with React Compiler, it looks the result is only evaluated on the initial render.

When using the repro stackblitz, you can remove the react compiler config, which will make the repro case act as I would expect it to.

https://stackblitz.com/edit/github-jbydd83d?file=src%2Fmain.tsx

Using the reproduction scenario:

1. Click on the "To About"-link
2. The "Matched route name" doesn't change
3. Refresh
4. The "Matched route name" updates to the correct label

### Expected behavior

The "Matched route name" label changes on each navigation change.

### Screenshots or Videos

_No response_

### Platform

- OS: Linux
- Browser: Chromium
- Version: 137

### Additional context

_No response_

kamalbennani:
## Problem

React Compiler's aggressive memoization causes `useMatchRoute` to return stale results after navigation (issue #4499). The hook uses `useCallback` with only the stable `router` reference as a dependency, but the callback reads mutable internal router state (`location`, `resolvedLocation`, `status`).

When React Compiler optimizes the code, it memoizes the callback with captured stale state, preventing it from seeing navigation changes.

### ⚠️ Important: When the Bug Manifests

The bug **specifically manifests when `matchRoute` results are consumed in `useEffect` dependencies**:

```typescript
function Component() {
  const matchRoute = useMatchRoute()
  const isActive = !!matchRoute({ to: '/home' })

  useEffect(() => {
    // Do something when route becomes active
    if (isActive) doSomething()
  }, [isActive])  // ← React Compiler creates stale closure here
}
```

**Without the useEffect pattern**, React Compiler doesn't trigger the problematic memoization, which is why some tests may pass without the fix. However, consuming `matchRoute` results in `useEffect` is a common real-world pattern, making this fix essential.

## Solution

This PR fixes the issue by:

1. **Capturing router state explicitly** - Store the result of `useRouterState` in a variable
2. **Adding state to dependencies** - Include `routerState` in the `useCallback` dependency array
3. **Ensuring callback recreation** - The callback now recreates when navigation occurs, providing fresh values to components using the useEffect pattern

## Test Added

Added a comprehensive e2e test project (`e2e/react-router/react-compiler/`) that:
- Enables React Compiler via `babel-plugin-react-compiler`
- Demonstrates the useEffect pattern (lines 30-33 in main.tsx)
- Tests navigation between routes
- Verifies match status updates correctly

The e2e test includes console logging to capture the behavior when the bug would manifest.

## Related Issue

Fixes #4499

## Breaking Changes

None - This is a bug fix that maintains the same public API.

---

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a stale-closure issue in route state handling to ensure consistent route matching during navigation.

* **New Features**
  * Added a minimal demo app to exercise route matching and link active-state behavior.

* **Tests**
  * Added end-to-end tests and test setup to validate navigation, match state updates, and active-link highlighting with the React Compiler.

* **Chores**
  * Added project config files and ignore patterns to support the new e2e test environment.


schiller-manuel:
we should add a new, small e2e test project that has react-compiler enabled and that demonstrates this issue. please do this as a separate commit

kamalbennani:
Hey @schiller-manuel, thanks for the suggestion, I've added a comprehensive e2e test project in a separate commit:

E2E Test Project: e2e/react-router/react-compiler/
  - Enables React Compiler via babel-plugin-react-compiler: ^1.0.0
  - Demonstrates the exact pattern that triggers the bug (see src/main.tsx lines 30-33)
  - Tests navigation between routes and verifies match status updates correctly

Key Discovery: The bug specifically manifests when matchRoute results are consumed in useEffect dependencies:

```tsx
  const isHome = !!matchRoute({ to: '/home' })
  useEffect(() => {
    console.log('isHome', isHome)
  }, [isHome])  // ← React Compiler's aggressive memoization creates stale closure here
```

Without this useEffect pattern, React Compiler doesn't trigger the problematic memoization. This is why the test checks DOM state directly rather than through useEffect - but the demo app (main.tsx) includes the useEffect pattern to demonstrate the real-world scenario.

The e2e tests pass with the fix, confirming the solution works correctly with React Compiler enabled.

schiller-manuel:
is this a react compiler bug that we are working around here?

kamalbennani:
No, this is not a React Compiler bug - this is us fixing improper hook dependency management in our code.

The Problem:

Our useMatchRoute hook was using useCallback with only the router reference as a dependency:

```typescript
  return React.useCallback((opts) => {
    return router.matchRoute(opts) // reads router.latestLocation, router.state
  }, [router]) // ❌ router reference is stable, but its internal state is mutable
```

The router reference never changes, but the callback reads mutable internal state (router.latestLocation, router.state.resolvedLocation, etc.). This violates React's rules - we're depending on values that aren't in the dependency array.

Why It Worked Before:

When navigation occurs, useRouterState triggers a re-render of the component. If the callback is called during render, it works fine because it reads the router's current internal state:

```typescript
  // This works - callback called during render
  return <div>{matchRoute({ to: '/home' }) ? 'Active' : 'Inactive'}</div>
```

However, if the callback result is used in a useEffect, that effect won't re-run after navigation because the callback reference hasn't changed, even though the component re-rendered:

```typescript
  // This breaks - useEffect doesn't see the callback reference change
  useEffect(() => {
    const isHome = matchRoute({ to: '/home' })
    if (isHome) {
       doSomething()
    }
  }, [matchRoute]) // ❌ Effect won't re-run after navigation
```

Why React Compiler Exposes It:

React Compiler enforces proper memoization semantics. When it sees useCallback([router]) with a stable reference, it assumes the callback never needs to update. This exposes our improper dependency management.

The Fix:

We now explicitly subscribe to the router state and include it in dependencies:

```typescript
  const routerState = useRouterState({
    select: (s) => [s.location.href, s.resolvedLocation?.href, s.status]
  })

  return React.useCallback((opts) => {
    return router.matchRoute(opts)
  }, [router, routerState]) // ✅ Explicit dependencies on values that actually change
```

This is the correct pattern. As the React team moves toward requiring proper memoization (removing the burden from developers to manually optimize), our libraries need to follow best practices. This fix ensures our code works correctly with React Compiler and future React versions.

kamalbennani:
Hey @schiller-manuel, if you have any further comments, I'd love to address them.

JeremyMoeglich:
This would be great to have, it's the only place I have react compiler disabled via "use no memo" right now.

kamalbennani:
Hey @schiller-manuel @tannerlinsley, I know you guys have a lot to deal with, but I was wondering if there is anything I could do to get this PR approved?

yslpn:
Hey @kamalbennani, thanks for working on this! I ran into the same issue (ref [#4499](https://github.com/TanStack/router/issues/4499)) and spent some time digging into the root cause across all three frameworks (React, Vue, Solid). I'd like to propose an alternative approach that fixes the problem more cleanly.

The fix proposed in this PR adds `router.state` as a dependency — but the function body never actually reads `router.state`. It reads live stores inside `router.matchRoute()` at call time. The linter flags this (`react-hooks/exhaustive-deps`), so the PR adds an `// eslint-disable-next-line` comment to suppress it. We're passing a value we don't genuinely use just to make the dependency array non-empty — which is a form of deception.

**The same problem exists in Vue and Solid**

Looking at the Vue and Solid implementations, there is a similar pattern: values are read from reactive stores solely for their side-effect of registering a dependency, then immediately discarded. The actual matching still happens against live stores inside `router.matchRoute()`.

**Vue** — `routerState.value` is read and thrown away, while `router.matchRoute()` reads live stores independently:
```ts
const routerState = useStore(router.stores.matchRouteDeps, (value) => value)

const matchRoute = Vue.computed(() => {
  // Access routerState to establish dependency
  routerState.value  // ← read and discarded, result unused
  return router.matchRoute(rest as any, { pending, caseSensitive, fuzzy, includeSearch })
})
```

**Solid** — same pattern, `matchRouteDeps.get()` is called just for the signal tracking side-effect:
```ts
return Solid.createMemo(() => {
  router.stores.matchRouteDeps.get()  // ← read and discarded, result unused
  return router.matchRoute(rest as any, { pending, caseSensitive, fuzzy, includeSearch })
})
```

**A cleaner approach: `matchRouteWithSnapshot`**

Instead of patching the dependency arrays, the idea is to extract the matching logic into a standalone pure function in `router-core`. For example `matchRouteWithSnapshot`

Each framework adapter would subscribe to a dedicated `matchRouteDeps` store and pass the result **directly as data** into this function:

**React** — `snapshot` is a genuine `useCallback` dependency:
```ts
const snapshot = useStore(router.stores.matchRouteDeps, (d) => d)

return useCallback((opts) => {
  return matchRouteWithSnapshot(router, rest, matchOpts, snapshot)
}, [router, snapshot])  // snapshot is real data, not a phantom dep
```

**Vue** — `routerState.value` is read inside `computed()`, creating a genuine reactive dependency:
```ts
const routerState = useStore(router.stores.matchRouteDeps, (v) => v)

return (opts) => computed(() =>
  matchRouteWithSnapshot(router, rest, matchOpts, routerState.value)
)
```

**Solid** — `matchRouteDeps.get()` is read inside `createMemo()`, tracking it as a signal:
```ts
return (opts) => createMemo(() =>
  matchRouteWithSnapshot(router, rest, matchOpts, router.stores.matchRouteDeps.get())
)
```

**Benefits**

- No phantom dependencies, no `eslint-disable` comments
- `matchRoute` function reference genuinely updates when and only when router navigation state changes
- Vue/Solid no longer discard read values — they pass them as real inputs to the matching function
- `router.matchRoute()` becomes a thin wrapper calling the same function
- Can be fully typed: `matchRouteWithSnapshot` carries the same generic signature as `MatchRouteFn`

Happy to open a separate PR with this approach if it looks good to you / maintainers.

cc @schiller-manuel

yslpn:
> we should add a new, small e2e test project that has react-compiler enabled and that demonstrates this issue. please do this as a separate commit

And to me, adding an extra `e2e` with a compiler seems redundant. As if you could just add one unit test that covers this case. And all the other cases could also be covered by simply manually adding `useMemo`/`useCallback` in the right places.

Lyra-Inspectly:
This appears to not have moved since April, is there any status on this?

ryami333:
@schiller-manuel where are we at with this? The author has offered to address any concerns but none have been raised. Seems there was no objection to the approach – so can this not be merged?

LadyBluenotes:
I'm sorry it's taken a while to get to this, but, full disclosure, this PR might have a heavy rebase (or have to be redone) soon. This predates the current `matchRouteDeps` store architecture and conflicts with #7805 (which is a pretty substantial rewrite of similar logic to address related bugs), which changes the same `Matches.tsx` path and pending-match behavior.

So, I did bring this up to @Sheraff about possibly adding to his PR, but if you want to continue with this one it would need to be adapted on top of #7805 and verified with a valid compiler-enabled fixture.

For coverage, I think this needs:

- one focused unit test asserting when the returned callback identity remains stable and when location, resolved location, status, or pending state must invalidate it;
- one compiler-enabled browser test for the actual Home → About → Home result;
- the fixture to satisfy React's effect dependency rules.

The separate compiler E2E project makes sense because this behaviour depends on compiler output, but the CSS-only test currently exercises the same transition as the primary test and can be consolidated.

I can understand it might be a lot to do now, but if you decide to go ahead with fixing the above, please let us know :)

PR #6605, opened 2026-02-06
chore(vite.config): enable coverage collection by default

sukvvon:
## Summary
- Add `@vitest/coverage-v8` as a devDependency
- Add `coverage: { enabled: true }` to all 32 packages that have a `test` block in `vite.config.ts`

## Why v8 over istanbul?
- Considered using istanbul for consistency with TanStack Query and Form, but chose v8 for technical reasons
- Istanbul instruments source code by inserting counter code (e.g., `cov_2egp37ota4().s[31]++`), which breaks snapshot tests like those in `router-plugin`
- v8 measures coverage at the engine level without modifying source code, so snapshot tests are unaffected
- v8 is also the default provider in Vitest, requiring no explicit `provider` configuration

## Summary by CodeRabbit

* **Tests**
  * Enabled test coverage collection across the entire monorepo to generate comprehensive code quality metrics and provide improved visibility into test effectiveness and coverage across all packages.

* **Chores**
  * Added test coverage infrastructure tooling as a dependency to support automated coverage analysis, detailed reporting, and metrics tracking during test execution.


schiller-manuel:
why?

sukvvon:
@schiller-manuel Having coverage enabled by default makes it easier to identify untested areas and helps maintain test quality as the codebase evolves. This also aligns with TanStack Query and Form, where `coverage: { enabled: true }` is already configured in every package's `vite.config.ts`.

PR #6612, opened 2026-02-07
Docs tweaks for pre-rendering 

arackaf:
## Summary by CodeRabbit

* **Documentation**
  * Updated static prerendering configuration guidance. Prerendering is now disabled by default.

### DIFF
--- a/docs/start/framework/react/guide/static-prerendering.md
+++ b/docs/start/framework/react/guide/static-prerendering.md
@@ -19,10 +19,10 @@ export default defineConfig({
   plugins: [
     tanstackStart({
       prerender: {
-        // Enable prerendering
-        enabled: true,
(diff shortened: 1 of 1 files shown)


docs/start/framework/react/guide/static-prerendering.md
```diff
@@ -19,10 +19,10 @@ export default defineConfig({
   plugins: [
     tanstackStart({
       prerender: {
-        // Enable prerendering
-        enabled: true,
+        // Switch to true to enable prerendering
+        enabled: false,
```
arackaf, line 23: Do we need the enabled flag at all? Can we not enable prerendering by virtue of the prerender section existing at all?

Like, what would ever be the purpose of having 

```ts
prerender: {
  enabled: false
}
```

Issue #6604, opened 2026-02-06
Error when parsing data streamed from server function


ceski23:
### Which project does this relate to?

Start

### Describe the bug

When streaming data from server function to client, first messages aren't properly decoded and there is error regarding parsing JSON in console

```
Invalid JSON line: {"t":32,"i":5,"f":{"t":0,"s":0}}{"t":32,"i":5,"f":{"t":0,"s":1}} SyntaxError: Unexpected non-whitespace character after JSON at position 32 (line 1 column 33)
    at JSON.parse (<anonymous>)
    at serverFnFetcher.js?v=31b5285e:226:30
overrideMethod	@	installHook.js:1
await in serverFnFetcher		
(anonymous)	@	index.tsx:27
(anonymous)	@	index.tsx:30
```

```tsx
const streamNumbers = createServerFn({ method: 'GET' }).handler(async function*() {
  const signal = getRequest().signal

  // await setTimeout(50, undefined, { signal })

  for (let i=0; !signal.aborted; i++) {
    yield i
    await setTimeout(1000, undefined, { signal })
  }
})

function App() {
  const [messages, setMessages] = useState<Array<number>>([]);

  useEffect(() => {
    const abortController = new AbortController();

    (async () => {
      for await (const value of await streamNumbers({ signal: abortController.signal })) {
        setMessages((prev) => [...prev, value])
      }
    })()

    return () => abortController.abort()
  }, [])

  return <pre>{JSON.stringify(messages, null, 2)}</pre>
}
```

Seems like some race-condition because if we add small delay before starting sending data (that 50ms setTimeout in code above) then everything works and all data is properly decoded on client.

### Your Example Website or App

https://github.com/ceski23/tanstack-start-streaming-repro

### Steps to Reproduce the Bug or Issue

1. Start app
2. Wait for data to stream from server
3. Received data will start from 2 instead of 0

### Expected behavior

All data should be received on client and there shouldn't be parsing errors in console

### Platform

- Router / Start Version: 1.158.3
- OS: macOS/Windows
- Browser: Chrome
- Browser Version: 144
- Bundler: Vite
- Bundler Version: 7.1.7

schester44:
Yeah I am seeing this as well.

```SyntaxError: Unexpected non-whitespace character after JSON at position 448```


```ts
export const globalSearchAction = createServerFn({ method: "GET" })
  .middleware([actionTracingMiddleware, actionAuth])
  .inputValidator(schema)
[...]

PR #6613, opened 2026-02-07
fix(start-client-core): fix NDJSON stream buffer boundary bug in processServerFnResponse

schester44:
Disclaimer: This code was written by Claude. I can clean it up for merge or feel free to just treat it as a POC of a bug in the system.  See Issue #6604

## Summary

This PR fixes a bug in `processServerFnResponse` that causes JSON parse errors when using async generator server functions that stream NDJSON responses.

## The Bug

When processing NDJSON streaming responses, the buffer handling code had a bug where remaining lines after extracting the first object would lose their trailing newline separator.

The issue occurred in this code path:

```js
if (buffer.endsWith('\n')) {
  const lines = buffer.split('\n').filter(Boolean)
  // ...
  buffer = lines.slice(1).join('\n')  // BUG: loses trailing newline!
}
```

When the first chunk contained multiple complete JSON lines (e.g., `{"id":1}\n{"id":2}\n`), after extracting the first line, the remaining buffer would be `{"id":2}` instead of `{"id":2}\n`. 

When the next chunk arrived, it would be concatenated directly, resulting in invalid JSON like `{"id":2}{"id":3}\n`, causing:

```
SyntaxError: Unexpected non-whitespace character after JSON at position 459
```

## The Fix

The fix simplifies the first-object parsing logic to consistently use `indexOf('\n')` and `slice()`, which correctly preserves the buffer contents after each extracted line. This removes the buggy `buffer.endsWith('\n')` branch entirely.

Also included a test using the buggy implementation to demonstrate that the original code fails as described.

## Reproduction

This bug manifests when using async generator server functions like:

```typescript
export const searchAction = createServerFn({ method: 'GET' })
  .handler(async function* ({ data }): AsyncGenerator<SearchResultBatch> {
    // Yield results as searches complete
    for (const type of typesToSearch) {
      const results = await searchFn(type, data.query)
      if (results.length > 0) {
        yield { type, results }
      }
    }
  })
```

When multiple yields happen quickly and get batched into the same network chunk, the client-side parser fails.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced robustness of server function response streaming by improving data parsing logic. Now correctly handles edge cases where JSON data spans multiple network chunks.

* **Tests**
  * Added comprehensive test coverage for server function response parsing, including multi-chunk scenarios, error handling, and boundary conditions.

### DIFF
--- a/packages/start-client-core/src/client-rpc/serverFnFetcher.ts
+++ b/packages/start-client-core/src/client-rpc/serverFnFetcher.ts
@@ -297,24 +297,14 @@ async function processServerFnResponse({
       throw new Error('Stream ended before first object')
     }
 
-    // common case: buffer ends with newline
-    if (buffer.endsWith('\n')) {
-      const lines = buffer.split('\n').filter(Boolean)
-      const firstLine = lines[0]
-      if (!firstLine) throw new Error('No JSON line in the first chunk')
-      firstObject = JSON.parse(firstLine)
-      firstRead = true
-      buffer = lines.slice(1).join('\n')
-    } else {
-      // fallback: wait for a newline to parse first object safely
-      const newlineIndex = buffer.indexOf('\n')
-      if (newlineIndex >= 0) {
-        const line = buffer.slice(0, newlineIndex).trim()
-        buffer = buffer.slice(newlineIndex + 1)
-        if (line.length > 0) {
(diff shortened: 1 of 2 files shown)


schiller-manuel:
NDJSON will be gone soon

PR #6615, opened 2026-02-08
perf(solid-start): 51% improvement

huseeiin:
improve performance by 51%

## Summary by CodeRabbit

* **Refactor**
  * Switched server-side rendering response implementation to a more efficient response mechanism; end-user behavior is unchanged (status codes, headers, and rendered output remain the same).
* **Chores**
  * Added an underlying dependency to support the new response mechanism; no visible API or behavior changes.

### DIFF
--- a/packages/solid-router/src/ssr/renderRouterToStream.tsx
+++ b/packages/solid-router/src/ssr/renderRouterToStream.tsx
@@ -2,10 +2,10 @@ import * as Solid from 'solid-js/web'
 import { isbot } from 'isbot'
 import { transformReadableStreamWithRouter } from '@tanstack/router-core/ssr/server'
 import { makeSsrSerovalPlugin } from '@tanstack/router-core'
+import { FastResponse } from 'srvx'
 import type { JSXElement } from 'solid-js'
 import type { ReadableStream } from 'node:stream/web'
 import type { AnyRouter } from '@tanstack/router-core'
-import { FastResponse } from 'srvx'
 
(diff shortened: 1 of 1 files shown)


packages/solid-router/src/ssr/renderRouterToStream.tsx
birkskyum, line 57: Can we just use hte FastResponse here instead of making sideeffects on the  globalThis?

huseeiin: > Can we just use hte FastResponse here instead of making sideeffects on the globalThis?

i understand, but the reason i'm patching `globalThis.Response` so user can benefit even when they return `Response` in their server routes, etc.

birkskyum: The package has `"sideEffects": true` like CodeRabbit point to [here](https://github.com/TanStack/router/pull/6615#discussion_r2779271806). Do you think that will be a problem? We're still on rollup.

huseeiin: > The package has `"sideEffects": true` like CodeRabbit point to [here](https://github.com/TanStack/router/pull/6615#discussion_r2779271806). Do you think that will be a problem? We're still on rollup.

looking at the tanstack start output (`dist/server/server.js`) everything is externalized so srvx will be imported as `import srvx from 'srvx'` anyway

PR #6616, opened 2026-02-08
perf(start): use FastResponse from srvx

birkskyum:
Similar to this change in solid-router, to keep the frameworks in sync
- #6615 

Except, direct use of FastResponse instead of global patch

# Flamegraph bench

## Req/Sec (10 runs each)

| Run | Baseline | FastResponse |
|-----|----------|--------------|
| 1   | 7,057    | 7,543        |
| 2   | 6,955    | 7,405        |
| 3   | 6,843    | 7,739        |
| 4   | 7,028    | 7,856        |
| 5   | 7,789    | 8,260        |
| 6   | 8,236    | 7,832        |
| 7   | 7,785    | 7,716        |
| 8   | 7,874    | 8,193        |
| 9   | 7,398    | 8,213        |
| 10  | 7,469    | 7,679        |
| **Mean** | **7,443** | **7,844** |
| **Stdev** | ~456 | ~283 |

**FastResponse: +5.4% higher req/sec**

## Avg Latency (ms)

| Run | Baseline | FastResponse |
|-----|----------|--------------|
| 1   | 1.25     | 1.08         |
| 2   | 1.20     | 1.44         |
| 3   | 1.22     | 1.01         |
| 4   | 1.16     | 0.95         |
| 5   | 0.99     | 0.76         |
| 6   | 0.82     | 0.90         |
| 7   | 1.04     | 1.08         |
| 8   | 0.98     | 0.81         |
| 9   | 1.28     | 0.81         |
| 10  | 1.08     | 1.06         |
| **Mean** | **1.10** | **0.99** |

**FastResponse: -10% lower avg latency**

## p99 Latency (ms)

| Run | Baseline | FastResponse |
|-----|----------|--------------|
| 1   | 7        | 5            |
| 2   | 4        | 9            |
| 3   | 4        | 3            |
| 4   | 4        | 4            |
| 5   | 4        | 3            |
| 6   | 3        | 3            |
| 7   | 6        | 4            |
| 8   | 5        | 3            |
| 9   | 11       | 3            |
| 10  | 3        | 7            |
| **Mean** | **5.1** | **4.4** |

**FastResponse: -14% lower p99 latency**

## Summary

| Metric | Baseline | FastResponse | Delta |
|--------|----------|--------------|-------|
| Req/Sec (mean) | 7,443 | 7,844 | **+5.4%** |
| Avg Latency (mean) | 1.10ms | 0.99ms | **-10.0%** |
| p99 Latency (mean) | 5.1ms | 4.4ms | **-13.7%** |
| Stdev (req/sec) | ~456 | ~283 | **lower variance** |

FastResponse won 9 out of 10 runs on req/sec. The improvement is consistent and statistically meaningful.

## Summary by CodeRabbit

* **Chores**
  * Added a new runtime dependency (srvx) to React Router and Vue Router packages.
  * Switched server-side rendering response implementation across router packages to use the new runtime response type, affecting how SSR responses are constructed.

### DIFF
--- a/packages/react-router/src/ssr/renderRouterToStream.tsx
+++ b/packages/react-router/src/ssr/renderRouterToStream.tsx
@@ -10,8 +10,6 @@ import type { AnyRouter } from '@tanstack/router-core'
 import type { ReadableStream } from 'node:stream/web'
 import type { ReactNode } from 'react'
 
-globalThis.Response = FastResponse
-
 export const renderRouterToStream = async ({
   request,
   router,
@@ -38,7 +36,7 @@ export const renderRouterToStream = async ({
       router,
       stream as unknown as ReadableStream,
     )
-    return new Response(responseStream as any, {
+    return new FastResponse(responseStream as any, {
       status: router.state.statusCode,
       headers: responseHeaders,
     })
@@ -81,7 +79,7 @@ export const renderRouterToStream = async ({
       router,
       reactAppPassthrough,
     )
-    return new Response(responseStream as any, {
+    return new FastResponse(responseStream as any, {
       status: router.state.statusCode,
       headers: responseHeaders,
     })
--- a/packages/solid-router/src/ssr/renderRouterToStream.tsx
+++ b/packages/solid-router/src/ssr/renderRouterToStream.tsx
@@ -7,8 +7,6 @@ import type { JSXElement } from 'solid-js'
 import type { ReadableStream } from 'node:stream/web'
 import type { AnyRouter } from '@tanstack/router-core'
 
-globalThis.Response = FastResponse
-
 export const renderRouterToStream = async ({
   request,
   router,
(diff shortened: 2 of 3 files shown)


Sheraff:
I think this optimization might be "gaming" the benchmark because it appears to be a node+srvx only optimization.
- implementation adds a `_toNodeResponse()` method which appears to be the main optimization https://github.com/h3js/srvx/blob/dbeeb0177ba020447fe95845b934416d20d3c292/src/adapters/_node/response.ts
- and then `srvx` knows to call this method https://github.com/h3js/srvx/blob/dbeeb0177ba020447fe95845b934416d20d3c292/src/adapters/_node/send.ts#L16

Maybe rather than putting this directly in the source code, we could advise users to augment the global object w/ FastResponse?

Like in the docs somewhere "if you're on Node and are using Nitro/h3/srvx, you can do `globalThis.Response = FastResponse` for a free 5% speed up"

Or maybe if someone has a magic idea to "detect" node+srvx and then do it automagically?

birkskyum:
Okay let's do the docs note for now

PR #6632, opened 2026-02-10
Restore scroll when navigating back from a resetScroll=false navigation without having scrolled

### LINKED ISSUE #6595: Scroll position should be remembered for intermediate pages when navigating with resetScroll=false

Router

### Describe the bug

Starting on page A, if I scroll somewhere and then navigate to from pages A -> B -> C with resetScroll=false and without scrolling the page on B, the scroll position is maintained as expected. However, when navigating back to B with the back arrow, it resets the scroll position back to the top. Then going back to A remembers the original scroll position on A. 

This is presumably because no scroll event happened on page B, so it didn't store any scroll position information for the entry for B at that point. I suspect when using resetScroll=false it should copy over A's scroll position entry to B when the navigation happens.

This also appears to happen when using a nested scrollable element with scrollToTopSelectors. 

https://github.com/TomVHPresto/scroll-restoration-mcve

1. Start on page /a
2. Scroll down some amount
3. Click any link to go to /b
4. Do not scroll on /b and click a link to go to /c
5. Click the back arrow in the browser - /b scrolls to the top of the page 

### Expected behavior

I would expect scroll position to be remembered for /b when going back from /c.

### Screenshots or Videos

_No response_

### Platform

- Router Version: 1.158.0
- OS: Windows
- Browser: Chrome
- Browser Version: 144.0.7559.110
- Bundler: vite
- Bundler Version: 7.3.1

### Additional context

_No response_

TomVHPresto:
Fixes https://github.com/TanStack/router/issues/6595

In the case of scroll to top selectors, this PR updates the scroll restoration code to use the string selector itself as the selector stored in onScroll, rather than generating a CSS selector on the fly. This then is copied across when navigating with resetScroll=false in order to remember the scroll positions without having a scroll event firing. If the CSS selector were created dynamically it'd risk becoming stale and referring to the wrong element between navigations.

This does mean the problem will still happen for scroll to top selectors initialized using a function call, which this PR doesn't change - I think this is a reasonable trade-off.

This PR also updates the tests. I've updated the basic scroll restoration test to use the default scroll restoration tests - its compatible with all of the old tests, and seems more reasonable to check against the default behaviour for the basic tests.

## Summary by CodeRabbit

* **New Features**
  * Replaced the top navigation with a persistent fixed sidebar and adjusted main content layout.
  * Improved scroll restoration so scroll positions for the page and key page elements are preserved across navigation.

* **Tests**
  * Added an end-to-end regression test verifying scroll positions are maintained during navigation and back navigation.

### DIFF
--- a/e2e/react-router/basic-scroll-restoration/src/main.tsx
+++ b/e2e/react-router/basic-scroll-restoration/src/main.tsx
@@ -288,7 +288,6 @@ const router = createRouter({
   routeTree,
   defaultPreload: 'intent',
   scrollRestoration: true,
-  getScrollRestorationKey: (location) => location.pathname,
   scrollToTopSelectors: ['#sidebar'],
 })
 
--- a/e2e/react-router/basic-scroll-restoration/src/main.tsx
+++ b/e2e/react-router/basic-scroll-restoration/src/main.tsx
@@ -20,21 +20,42 @@ const rootRoute = createRootRoute({
 function RootComponent() {
   return (
     <>
-      <div className="p-2 flex gap-2 sticky top-0 border-b bg-gray-100 dark:bg-gray-900">
-        <Link to="/" className="[&.active]:font-bold">
-          Home
-        </Link>{' '}
-        <Link to="/about" className="[&.active]:font-bold">
-          About
-        </Link>
-        <Link to="/about" resetScroll={false}>
-          About (No Reset)
-        </Link>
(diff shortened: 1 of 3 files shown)


schiller-manuel:
closing in favor of https://github.com/TanStack/router/pull/7447

schiller-manuel:
closing in favor of https://github.com/TanStack/router/pull/7447

PR #6670, opened 2026-02-15
WIP remove pending from store (2nd attempt)

Sheraff:
## Summary by CodeRabbit

* **Breaking Changes**
  * Removed the public pendingMatches property; use router.state.status === 'pending' and route-matching against the latest location to detect pending matches.

* **Documentation**
  * Updated API docs with migration notes on inferring pending state and retrieving pending matches.

* **Tests**
  * Added pending-transition tests and updated existing test expectations for navigation/store-update behavior.

* **Devtools/UX**
  * Devtools and tooling now surface pending matches based on status + latest-location matching.

### DIFF
--- a/packages/router-plugin/tests/add-hmr/snapshots/solid/arrow-function@true.tsx
+++ b/packages/router-plugin/tests/add-hmr/snapshots/solid/arrow-function@true.tsx
@@ -25,7 +25,8 @@ if (import.meta.hot) {
         router.resolvePathCache.clear();
         walkReplaceSegmentTree(newRoute, router.processedTree.segmentTree);
         const filter = m => m.routeId === oldRoute.id;
-        if (router.state.matches.find(filter) || router.state.pendingMatches?.find(filter)) {
+        const hasPendingRouteMatch = router.state.status === 'pending' && router.matchRoutes(router.latestLocation).some(filter);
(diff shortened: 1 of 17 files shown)


packages/router-plugin/tests/add-hmr/snapshots/solid/arrow-function@true.tsx
```diff
@@ -25,7 +25,8 @@ if (import.meta.hot) {
         router.resolvePathCache.clear();
         walkReplaceSegmentTree(newRoute, router.processedTree.segmentTree);
         const filter = m => m.routeId === oldRoute.id;
-        if (router.state.matches.find(filter) || router.state.pendingMatches?.find(filter)) {
+        const hasPendingRouteMatch = router.state.status === "pending" && router.matchRoutes(router.latestLocation).some(filter);
```
SeanCassiere, line 28: Perhaps, defer having to ever having to run this IF condition which if successful checks for pending matches and keep it to only do so when the left-hand of the OR fails?

```suggestion
        const checkHasPendingRouteMatch = () => router.state.status === "pending" && router.matchRoutes(router.latestLocation).some(filter);
        if (router.state.matches(filter) || checkHasPendingRouteMatch()) {
```

PR #6674, opened 2026-02-15
refactor(router-devtools-core): restore visibility into internalized cached and pending matches

Sheraff:
after having internalized `pendingMatches` and `cachedMatches` (https://github.com/TanStack/router/pull/6673), they are not accessible outside of `router-core` anymore.

This PR restores the devtools visibility into those matches by exposing them as a separate store, that will not have the overhead of being consumed throughout the app.

In a follow-up PR, we can further reduce the overhead of reactivity by using `createServerStore` instead of actual tanstack/store when not running on the client, or not running in dev mode.

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -912,6 +912,14 @@ type RouterStateStore<TState> = {
   setState: (updater: (prev: TState) => TState) => void
 }
 
+type DevtoolsMatchesState = {
+  pendingMatches?: Array<AnyRouteMatch>
+  cachedMatches: Array<AnyRouteMatch>
+}
+
+const filterRedirectedMatches = (matches: Array<AnyRouteMatch>) =>
+  matches.filter((match) => match.status !== 'redirected')
+
 function createServerStore<TState>(
   initialState: TState,
 ): RouterStateStore<TState> {
@@ -946,6 +954,10 @@ export class RouterCore<
 
   // Must build in constructor
   __store!: Store<RouterState<TRouteTree>>
+  __storeDevtoolsMatches = new Store<DevtoolsMatchesState>({
+    pendingMatches: [],
+    cachedMatches: [],
+  })
   options!: PickAsRequired<
     RouterOptions<
       TRouteTree,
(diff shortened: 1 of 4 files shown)


SeanCassiere (commented):
If we're storing Reactive matches in a separate Store, I wonder if it makes sense to take Form's approach of exposing that store and using `useStore` for subscription to be documented.

```ts
const routerInternalPendingMatches = useStore(router.internalReactiveStore, s => s.pendingMatches);
```

This has the added benefit of providing a migration path for the dozens of users who were accessing the pending matches. We can say that, based on your use case, you should probably be just checking the `status === 'pending'`, but if you absolutely need the pending and cached matches then you'll need to do this "extra legwork" to access it instead.

Sheraff:
> If we're storing Reactive matches in a separate Store, I wonder if it makes sense to take Form's approach of exposing that store and using `useStore` for subscription to be documented.
> 
> ```ts
> const routerInternalPendingMatches = useStore(router.internalReactiveStore, s => s.pendingMatches);
> ```
> 
> This has the added benefit of providing a migration path for the dozens of users who were accessing the pending matches. We can say that, based on your use case, you should probably be just checking the `status === 'pending'`, but if you absolutely need the pending and cached matches then you'll need to do this "extra legwork" to access it instead.

That's a good point. I do like keeping stuff "private" / "internal" for more freedom for us, but if some people do actually use those things, we could mention it in the docs

Sheraff:
> If we're storing Reactive matches in a separate Store, I wonder if it makes sense to take Form's approach of exposing that store and using `useStore` for subscription to be documented.
> 
> ```ts
> const routerInternalPendingMatches = useStore(router.internalReactiveStore, s => s.pendingMatches);
> ```
> 
> This has the added benefit of providing a migration path for the dozens of users who were accessing the pending matches. We can say that, based on your use case, you should probably be just checking the `status === 'pending'`, but if you absolutely need the pending and cached matches then you'll need to do this "extra legwork" to access it instead.

This would prevent an optimisation I was planning though: to not actually use a reactive store outside of `NODE_ENV === 'development'`. But then maybe w/ alien-signals (and in a case of zero subscribers) this wouldn't actually make a difference? I was planning on using `createServerStore` for prod, which is basically a pass-through store: same API, no reactivity.

PR #6677, opened 2026-02-16
fix(start): force close preview server connections after prerendering

mattkinnersley:
## Summary

The prerender step hangs indefinitely after successfully completing when the application imports modules that create persistent handles (websocket connections, flush intervals, timers, etc.).

During prerendering, `startPreviewServer` loads the full server bundle via Nitro's preview plugin. If the application uses libraries that create persistent connections at module scope (e.g. PostHog analytics, Zero sync engine, OpenAuth client), those handles keep the Node.js event loop alive. Calling `previewServer.close()` only shuts down the HTTP server — it does not clean up resources from dynamically imported application modules.

This causes the build process to hang after printing "Prerendered N pages", requiring a manual `^C` to terminate.

## Fix

Force-close all active connections before closing the preview server, since it is an ephemeral server used only for prerendering during the build step. Graceful shutdown of application-level resources is not necessary here.

## Test plan

- Build a TanStack Start app with `spa: { enabled: true }` that imports a
  library creating persistent handles (e.g. a websocket client)
- Run `vite build`
- Verify the build completes and exits cleanly after prerendering

### DIFF
--- a/packages/start-plugin-core/src/prerender.ts
+++ b/packages/start-plugin-core/src/prerender.ts
@@ -109,6 +109,7 @@ export async function prerender({
     logger.error(error)
   } finally {
     await previewServer.close()
+    process.exit(0)
   }
 
   function extractLinks(html: string): Array<string> {


schiller-manuel:
not a fan of this approach. there might be other things running after prerendering. users should not create those persistent connections during prerendering

mattrobrob:
> not a fan of this approach. there might be other things running after prerendering. users should not create those persistent connections during prerendering

@schiller-manuel how would someone do that if their application is built around a websocket connection? Is there a flag to say when an app is prerendering?

This wasn't an issue before either so something must have changed recently

PR #6704, opened 2026-02-19
refactor: signal based reactivity

Sheraff:
## Benchmarks

Custom benchmarks (separate from the codspeed ones), made to exaggerate the impact of this PR:
- `total_cycle_ms` is a Router test, runs on `playwright`, measures "how long to navigate 100 times" on a rerender-heavy page
- `median_req_per_sec` is a Start test, runs on `autocannon`, using the previously established "minimal", "100 links", "many nested routes" apps

Results:
- Nav total_cycle_ms (median of trials), lower is better:

|   | Benchmark | Base | Head | Change |
| --- | --- | --- | --- | --- |
⚡ | React | 853.90 | 265.40 | 3.2x
⚡ | Solid | 823.55 | 342.10 | 2.4x
⚡ | Vue | 1395.15 | 214.25 | 6.5x

- SSR median_req_per_sec (median of rounds), higher is better:

|   | Benchmark | Base | Head | Change |
| --- | --- | --- | --- | --- |
✅ | React root_fixed | 5022.20 | 5033.15 | +0.22%
✅ | React page100_fixed | 1170.50 | 1170.60 | +0.01%
✅ | React page_random | 1185.66 | 1186.00 | +0.03%
✅ | Solid root_fixed | 1177.84 | 1178.60 | -0.06%
✅ | Solid page100_fixed | 139.07 | 139.04 | +0.02%
✅ | Solid page_random | 139.97 | 139.57 | +0.29%
✅ | Vue root_fixed | 1387.64 | 1389.57 | +0.14%
✅ | Vue page100_fixed | 328.24 | 328.64 | +0.12%
✅ | Vue page_random | 284.44 | 284.80 | +0.13%

## Summary by CodeRabbit

* **New Features**
  * Enhanced server/client reactivity and SSR paths across React, Solid, and Vue; more granular per-route reactivity exposed to UIs and devtools.

* **Bug Fixes**
  * Stabilized tests by replacing flaky range assertions with exact expected values; E2E tests adjusted for deterministic renders.

* **Documentation**
  * Removed documented pendingMatches field and updated hook wording to reference pending vs active matches.

* **Refactor**
  * Router state moved to store-backed match APIs; public RouterState no longer exposes pendingMatches/cachedMatches.


packages/react-router/src/link.tsx
```diff
@@ -401,7 +399,8 @@ export function useLinkProps<
     // eslint-disable-next-line react-hooks/exhaustive-deps
     [
       router,
-      currentLocationState,
+      currentLeafMatchId,
+      currentLocation.hash,
       from,
```
Sheraff: This is not correct, the `main` implementation of react/link re-builds the location on this
```ts
  const currentLocationState = useRouterState({
    select: (s) => {
      const leaf = s.matches[s.matches.length - 1]
      return {
        search: leaf?.search,
        hash: s.location.hash,
        path: leaf?.pathname, // path + params
      }
    },
  })
```
i.e. resolved full search, resolved full path+params, live hash

Whereas this implementation has
- resolved full path+params using the `lastMatchId`
- live hash using `location.hash`

But it's missing the search

packages/router-core/src/stores.ts
```diff
@@ -0,0 +1,346 @@
...
+  // two-level byRouteId → matchStore pattern.
+  //
+  // Cache size is bounded by the route tree (routeIds are static strings
+  // defined at app init). Unwatched computed stores are inert in
+  // alien-signals — they purge dependency links and don't participate
+  // in the reactive graph until re-subscribed.
+  const matchStoreByRouteIdCache = new Map<
+    string,
+    RouterReadableStore<AnyRouteMatch | undefined>
+  >()
```
Sheraff: Use LRUCache here? We can set size to 32 because this is the limit set by the bitmask in route matching (and also a somewhat reasonable "maximum number of active routes at the same time") Or even set the limit to 100 if we really want to be safe. The point is LRUCache auto-evicts, and we only need the active matches in there anyway.

birkskyum:
I'm OK with the changes to the solid/vue packages made here.

PR #6710, opened 2026-02-20
chore: mark eslint peer dependency as optional

dotnize:
see https://github.com/TanStack/router/discussions/6706

This is to consider other linters like Oxlint that can consume ESLint plugins directly, so ESLint doesn't have to be installed for these projects.

Would it be worth mentioning Oxlint as an additional supported linter in the docs?

UPDATE: see https://github.com/TanStack/router/discussions/6706#discussioncomment-15865975

## Summary by CodeRabbit

* **Chores**
  * Updated package configuration to mark ESLint as an optional peer dependency, allowing greater flexibility for users who may not require it.

### DIFF
--- a/packages/eslint-plugin-router/package.json
+++ b/packages/eslint-plugin-router/package.json
@@ -61,5 +61,10 @@
   },
   "peerDependencies": {
     "eslint": "^8.57.0 || ^9.0.0 || ^10.0.0"
+  },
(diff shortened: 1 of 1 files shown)


dotnize:
https://github.com/TanStack/router/discussions/6706#discussioncomment-15865975

PR #6761, opened 2026-02-25
Fix(head-content-utils): deduplicate specific link tags (e.g. canonical) to allow child overrides

### LINKED ISSUE #6719: `head`: child route `links` cannot override parent route links (e.g. `rel="canonical"`)

Start

## Describe the bug

When defining `links` in the `head` option of both a parent route and a child route, the links from all matched routes are simply concatenated without any deduplication logic. This means a child route cannot override a parent route's link — for example, `<link rel="canonical">` defined in the root route cannot be overridden by a child route's canonical link. Both end up in the rendered HTML, which is invalid and harmful for SEO.

This is inconsistent with how `meta` tags are handled. Meta tags are correctly deduplicated by `name` or `property` attribute, with child routes taking priority over parent routes. Links have no such mechanism.

## Your Example Website or App

Minimal reproduction:

```tsx
// __root.tsx
  head: () => ({
    links: [
      { rel: 'canonical', href: 'https://example.com/' },
    ],
  }),
  // ...
})

// blog/index.tsx
export const Route = createFileRoute('/blog/')({
  head: () => ({
    links: [
      { rel: 'canonical', href: 'https://example.com/blog' },
    ],
  }),
  // ...
})
```

When visiting `/blog`, the rendered HTML contains **both** canonical links:

```html
<link rel="canonical" href="https://example.com/" />
<link rel="canonical" href="https://example.com/blog" />
```

Expected: only `https://example.com/blog` should be present.

## Steps to Reproduce

1. Define a `<link rel="canonical">` in the root route's `head.links`
2. Define a different `<link rel="canonical">` in a child route's `head.links`
3. Navigate to the child route
4. Inspect the rendered `<head>` — both canonical links are present

## Expected behavior

Child route's `links` with the same `rel` attribute (at least for unique-by-nature rels like `canonical`) should override the parent route's, similar to how `meta` tags are deduplicated by `name`/`property`.

## Source Code Analysis

In `headContentUtils.js`:

- **Meta tags** (lines 17–49): iterated from last match (deepest child) to first (root), deduplicated by `name`/`property` — child wins.
- **Links** (lines 75–83): simply `state.matches.map(match => match.links).flat(1)` — no dedup by `rel` or any other attribute.

The final `uniqBy` (line 140) deduplicates by `JSON.stringify`, but since two canonical links have different `href` values, they are not considered duplicates.

## How often does this bug happen?

Every time

## Platform

- @tanstack/react-router: v1.132.0
- OS: macOS
- Browser: All

/

/

### Expected behavior

/

### Screenshots or Videos

_No response_

### Platform

- @tanstack/react-router: v1.161.3
- OS: macOS
- Browser: All

### Additional context

_No response_

jong-kyung:
Closes #6719 

## Description

This PR fixes an issue where `<link>` tags defined in child routes (specifically `rel="canonical"`) could not override parent route links, resulting in duplicated output in the `<head>` which is harmful for SEO.

## Changes

To solve this safely without introducing destructive side-effects (such as accidentally removing multiple `stylesheet` or `preload` links):

1. **Selective Deduplication**: Introduced a `relsToDedupe` Set. Currently, it only targets `canonical`.
2. **Reverse Iteration**: Replicated the robust reverse-iteration logic already used for `meta` tags (traversing from the deepest child to the root). This ensures the child route's link naturally takes precedence.
3. **Performance**: Used an O(1) lookup record (`linksByRel`) for deduplication to ensure no performance degradation during the render cycle.

## Summary by CodeRabbit

* **New Features**
  * Added canonical pages at /canonical and /canonical/deep across example apps; routes are registered and exposed in the public route surfaces.
  * Head generation now deduplicates and deterministically orders canonical link tags so only one rel="canonical" appears.

* **Tests**
  * Added e2e tests verifying a single canonical link and correct href for /canonical/deep across frameworks.

* **Refactor**
  * Reworked link/head assembly to support canonical deduplication and stable ordering.


jong-kyung:
While testing the fix in the vue-start environment, I noticed that the entire head content is being duplicated during SSR/Hydration (everything from <title> to <link> appears twice in the DOM).

This duplication is causing the E2E tests to fail, as they detect two canonical links instead of one. I'm currently investigating @tanstack/vue-router to see if this is an issue with the HeadContent component or the hydration process in the Vue adapter.

image

jong-kyung:
> While testing the fix in the vue-start environment, I noticed that the entire head content is being duplicated during SSR/Hydration (everything from <title> to appears twice in the DOM).
> 
> This duplication is causing the E2E tests to fail, as they detect two canonical links instead of one. I'm currently investigating @tanstack/vue-router to see if this is an issue with the HeadContent component or the hydration process in the Vue adapter.
> 
> image

I created an issue #6764 related to this.

jong-kyung:
I'm leaving this PR **open for review**, but please note that the E2E tests for `vue-start` and `solid-start` are currently failing due to framework-specific behavior in the test environments.

Here is what I found after digging deeper:
1. **`vue-start`**: The test fails because it detects **2** canonical links. I verified via `pnpm dev` that this is a real issue—the entire `<head>` block is being duplicated during SSR/Hydration. I've opened #6764 to track this framework bug.
2. **`solid-start`**: The test fails because it detects **0** canonical links. However, when I run `pnpm dev` locally, it correctly outputs exactly **1** canonical link. This seems like a hydration timing or setup issue specific to the Playwright test environment, rather than a flaw in the logic.

@schiller-manuel, could you please review my core deduplication logic? If the logic looks good, how should we handle the E2E tests? Any insights on why the `solid-start` tests might be failing to catch the injected tags would be greatly appreciated!

jong-kyung:
@nlynzaad, since Manuel seems quite busy with other priorities right now, I was wondering if you might have some cycles to take a quick look at this. 

As noted above, the core deduplication logic is fully baked and ready for review. We are just blocked by the framework-specific hydration bugs in the E2E environment. I'd love to get your thoughts on the logic itself and hear how you'd like to proceed with the failing tests!

PR #6763, opened 2026-02-25
fix: stabilize server function manifest build output

### LINKED ISSUE #6762: Unstable build output due to unsorted manifest map keys in TanStack build artifact

Start

### Describe the bug

Hi,

I noticed an issue with the build output generated by TanStack, specifically in:

```
__tanstack-start-server-fn-resolver-[hash].mjs
```

Inside this artifact, the `manifest` variable is a map/object whose keys are not sorted deterministically.

Because of this, even when the actual function code has **not changed**, the iteration order of the `manifest` keys can differ between builds, which causes:

* The generated file content to change
* The related `[hash]` value to change randomly
* Builds to become non-deterministic / unstable

This makes it difficult to rely on caching, reproducible builds, and CI optimizations.

none

1. create some server functions 
2. vite build

### Expected behavior

The `manifest` keys should be sorted in a stable order (e.g. lexicographically) before being emitted into the build artifact, so that:

* Identical inputs always produce identical outputs
* The hash only changes when actual code changes

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.163.1
- OS: Windows
- Browser: Edge
- Browser Version: 145.0.3800.70
- Bundler: vite
- Bundler Version: 8.0.0-beta.15

### Additional context

_No response_

motopods:
`generateManifestModule` iterated `serverFnsById` via `Object.entries` without sorting, producing manifest map keys in insertion order — which varies across builds and causes non-deterministic build artifacts.

## Changes

- **`packages/start-plugin-core/src/start-compiler-plugin/plugin.ts`**: Sort manifest entries by key (alphabetically) before code generation.

```ts
const manifestEntries = Object.entries(serverFnsById)
  .sort(([a], [b]) => (a < b ? -1 : a > b ? 1 : 0)) // ensure stable key order
  .map(([id, fn]) => { ... })
```

Fix  https://github.com/TanStack/router/issues/6762

## Summary by CodeRabbit

* **Chores**
  * Emitted manifest entries are now produced in a deterministic, sorted order by function ID, improving consistency of build outputs and downstream tooling that rely on manifest ordering.


schiller-manuel:
please add a comment into the code why we are sorting here

motopods:
@schiller-manuel  Hi, the comment t has been added into the code. Please check the PR.

PR #6769, opened 2026-02-26
fix(router-core): use routeId for lifecycle hooks to fix onStay when loaderDeps change

### LINKED ISSUE #6765: onLeave/onEnter fire instead of onStay when loaderDeps change from search param updates

Router

### Describe the bug

The `onLeave`, `onEnter`, and `onStay` route lifecycle hooks use `match.id` to determine whether a route was entered, stayed, or left. However, `match.id` includes the `loaderDepsHash`, which means any change to `loaderDeps` (e.g. from search param updates) causes the match ID to change. This results in:

- `onLeave` firing on the **old** match (old deps hash)
- `onEnter` firing on the **new** match (new deps hash)  
- `onStay` **never** firing

This is inconsistent with the `loader`'s `cause` parameter, which correctly uses `routeId` (not `match.id`) to determine `"stay"` vs `"enter"`:

```typescript
// router.js — loader cause uses routeId ✅
const previousMatch = previousMatchesByRouteId.get(route.id);
const cause = previousMatch ? "stay" : "enter";
```

```typescript
// router.js — lifecycle hooks use match.id ❌
exitingMatches = previousMatches.filter(
  (match) => !newMatches.some((d) => d.id === match.id)
);
enteringMatches = newMatches.filter(
  (match) => !previousMatches.some((d) => d.id === match.id)
);
stayingMatches = newMatches.filter(
  (match) => previousMatches.some((d) => d.id === match.id)
);
```

The match ID is computed as `route.id + interpolatedPath + loaderDepsHash`, so when `loaderDeps` changes, the lifecycle hooks treat the same route as "left and re-entered" rather than "stayed".

https://stackblitz.com/edit/github-1bd7mwws?file=src%2Froutes%2Findex.tsx

1. Create a route with `loaderDeps` that depends on search params:
```tsx
const DashboardRoute = createRoute({
  path: '/dashboard',
  validateSearch: DashboardSearchSchema,
  loaderDeps: ({ search }) => search,
  loader: async ({ context, deps }) => {
    await context.queryClient.ensureQueryData(getDataQueryOptions(deps));
  },
  onLeave: () => console.log('onLeave'),
  onStay: () => console.log('onStay'),
  onEnter: () => console.log('onEnter'),
  component: DashboardComponent,
});
```
2. Navigate to `/dashboard`  
3. Observe: `onEnter` logs (correct)
4. Change a search param via `navigate({ search: (s) => ({ ...s, page: 2 }) })`
5. Observe: `onLeave` and `onEnter` log instead of `onStay`

### Expected behavior

When navigating within the same route and only search params change:
- `onStay` should fire (the route is still matched, just with different deps)
- `onLeave` should **not** fire (the user has not left the route)
- `onEnter` should **not** fire (the user was already on this route)

This would be consistent with the `loader`'s `cause` parameter which correctly reports `"stay"` in this scenario.

### Screenshots or Videos

_No response_

### Platform

- Router Version: 1.163.2
- OS: macOS
- Browser: Chrome
- Bundler: Vite

### Additional context

The practical impact is that `onLeave` cannot be used for cleanup logic that should only run when truly navigating away (e.g. invalidating TanStack Query caches). The workaround is to use a `useEffect` cleanup in the route component instead, but this defeats the purpose of having route-level lifecycle hooks.

The fix would likely involve using `routeId` instead of `match.id` when categorizing matches into `exitingMatches`, `enteringMatches`, and `stayingMatches` — the same approach already used for the loader's `cause` parameter.

sleitor:
## Summary

Fixes #6765

When `loaderDeps` change (e.g. from search param updates), the `match.id` changes because it includes the `loaderDepsHash`. The lifecycle hooks (`onLeave`/`onEnter`/`onStay`) were computing `exitingMatches`, `enteringMatches`, and `stayingMatches` using `match.id`, which caused a loaderDeps change on the **same route** to be treated as "leave + re-enter" instead of "stay".

## Root Cause

`match.id` is composed as `routeId + interpolatedPath + loaderDepsHash`. When `loaderDeps` change (e.g. due to a search param update), the hash changes and `match.id` changes, making the old and new matches appear as different entries.

The loader's `cause` parameter already handles this correctly by using `routeId`:
```ts
// router.ts — loader cause uses routeId ✅
const previousMatch = previousMatchesByRouteId.get(route.id)
const cause = previousMatch ? 'stay' : 'enter'
```

But the lifecycle hooks were using `match.id`:
```ts
// router.ts — lifecycle hooks used match.id ❌
exitingMatches = previousMatches.filter(
  (match) => !newMatches.some((d) => d.id === match.id),
)
```

## Fix

Use `match.routeId` (consistent with the loader's `cause` param logic) when computing `exitingMatches`, `enteringMatches`, and `stayingMatches`.

## Test

Added a test case in `callbacks.test.ts` that verifies `onStay` fires (not `onLeave`/`onEnter`) when navigating within the same route and only loaderDeps-affecting search params change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Separated lifecycle identity from caching identity so enter/leave/stay callbacks behave correctly when only route params or loader-driven dependencies change; ensures onStay fires instead of unnecessary onLeave/onEnter.

* **Tests**
  * Added and expanded tests and helpers covering loader-driven dependency changes, same-route caching behavior, stale/gc timing, and navigation event payloads/counts.

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -2407,11 +2407,9 @@ export class RouterCore<
                   // Commit the pending matches. If a previous match was
                   // removed, place it in the cachedMatches
                   //
-                  // Cache identity uses match.id (routeId + params + loaderDeps) so
+                  // exitingMatches uses match.id (routeId + params + loaderDeps) so
                   // navigating /foo?page=1 → /foo?page=2 correctly caches the page=1 entry.
                   let exitingMatches: Array<AnyRouteMatch> = []
-                  let enteringMatches: Array<AnyRouteMatch> = []
-                  let stayingMatches: Array<AnyRouteMatch> = []
 
                   // Lifecycle-hook identity uses routeId only so that navigating between
                   // different params/deps of the same route fires onStay (not onLeave+onEnter).
@@ -2424,16 +2422,8 @@ export class RouterCore<
                       const previousMatches = s.matches
                       const newMatches = s.pendingMatches || s.matches
 
-                      // Cache-level identity: route id + params + loaderDeps
                       exitingMatches = previousMatches.filter(
                         (match) => !newMatches.some((d) => d.id === match.id),
-                      )
-                      enteringMatches = newMatches.filter(
-                        (match) =>
-                          !previousMatches.some((d) => d.id === match.id),
-                      )
-                      stayingMatches = newMatches.filter((match) =>
-                        previousMatches.some((d) => d.id === match.id),
                       )
 
                       // Lifecycle-hook identity: routeId only (route presence in tree)


Sheraff:
@schiller-manuel do you agree that when params change (e.g. nav from `/a/123` to `/a/456`) we want to trigger the `onStay` callback for the `/a/$id` route? The PR makes sense to me but this is a slight semantic change. The docs aren't exactly super specific about this: they describe the stay/enter/leave concepts as something like "a route is matched after not being matched in the previous location". 

To me this change makes sense, just asking for a second opinion.

Sheraff:
@sleitor don't worry about the "Bundle Size" job failing, this is unrelated to your PR, i'm working on it.

sleitor:
Thanks for the heads up @Sheraff, appreciated!

packages/router-core/src/router.ts
```diff
@@ -2416,14 +2416,19 @@ export class RouterCore<
...
-                          !previousMatches.some((d) => d.id === match.id),
+                          !previousMatches.some(
+                            (d) => d.routeId === match.routeId,
+                          ),
                       )
                       stayingMatches = newMatches.filter((match) =>
-                        previousMatches.some((d) => d.id === match.id),
+                        previousMatches.some(
+                          (d) => d.routeId === match.routeId,
+                        ),
```
Sheraff, line 2443: This change might mess up the caching behavior: a cache entry *should* be route id + params + deps (what we have currently). But I may agree that as far as lifecycle events are concerned, it may not exactly correlate with the cache changes.

I had codex whip up a quick test showing the expected cache behavior https://github.com/TanStack/router/pull/6772. This test works on main and should fail w/ your PR.

Now we need to decide the semantics of those lifecycle events. Are they tied to the cache entries' lifecycles? Or are they more tied to the route itself (i.e. routeId)?

sleitor: That's a great distinction to make. I think cache entries and lifecycle events serve different purposes and shouldn't necessarily share the same identity:

- **Cache entries** are about data identity — `route + params + deps` makes sense here, because different params or deps mean different data
- **Lifecycle hooks** are about route presence — `onEnter`/`onStay`/`onLeave` describe whether a *route* is appearing, persisting, or disappearing in the matched tree

From a user perspective, if I navigate from `/posts/123` to `/posts/456`, I'd expect `onStay` for the `/posts/$id` route (the route is still there, just showing different data), not `onLeave`+`onEnter`. The cache correctly gets a new entry for `456`, but the route itself didn't leave.

That said, I'm happy to defer to whatever semantics you and @schiller-manuel decide are correct here. If the conclusion is that lifecycle events should follow cache identity, I can revert and we can look for a different fix for #6765.

schiller-manuel: the semantics as you described them are correct. so we need to fix the caching regression that @Sheraff  pointed out and then we can merge this

sleitor:
Fixed in commit c8be9de — the caching regression is resolved by separating the two concerns:

- **Cache identity** (//) still uses `match.id` (routeId + params + loaderDeps), so navigating `/foo?page=1` → `/foo?page=2` correctly moves the page=1 entry into `cachedMatches`
- **Lifecycle-hook identity** uses `routeId` only via new `hookExitingMatches`/`hookEnteringMatches`/`hookStayingMatches` variables, so `onStay` fires for same-route param changes

Sheraff's test from #6772 should now pass with this change — the previous match gets cached and reused within staleTime as expected.

sleitor:
Addressed both remaining review requests in commit eb67444 — test coverage added to `packages/router-core/tests/callbacks.test.ts`:

---

**@Sheraff — caching regression tests (from #6772)**

Added the two tests from your test PR to verify that switching lifecycle hooks to `routeId` does NOT affect cache behavior:

- `keeps previous loaderDeps variant cached and reuses it within staleTime` — navigates `/foo?page=1` → `/foo?page=2` → `/foo?page=1` and confirms the loader is only called twice (page=1 entry is found in `cachedMatches` and reused)
- `keeps previous params variant cached and reuses it within staleTime` — same for path params (`/posts/$postId`)

Both tests confirm that `match.id` (routeId + params + loaderDeps) is still the cache identity, while only the lifecycle-hook arrays (`hookExiting/Entering/StayingMatches`) use `routeId`.

---

**@SeanCassiere — `onBeforeNavigate` event tests**

Added three tests for the `onBeforeNavigate` subscription event:

1. Fires when navigating to a new route
2. Fires on every navigation including same-route `loaderDeps` changes (important for Sentry tracing — each param change should still be tracked)
3. Includes `toLocation` and `pathChanged` in the event payload

Re: `onResolved` — this event is emitted from framework-specific `Transitioner` components (React/Vue/Solid), not from `router-core` directly. Existing tests for `onResolved` live in `packages/react-router/tests/router.test.tsx`, `packages/vue-router/tests/router.test.tsx`, and `packages/solid-router/tests/router.test.tsx` — those are unaffected by this PR since `router.ts` does not touch that code path.

---

Full suite: **1017 tests pass, 0 type errors** (29 test files).

packages/router-core/src/router.ts
sleitor: Good suggestion — done in the latest commit. Merged `setupWithLoaderDeps` and `setupWithLoaderDepsForCaching` into a single `createFooRouter` helper that accepts all options (`onEnter/onLeave/onStay`, `loader`, `staleTime`). All 9 tests still pass.

sleitor:
Addressed CodeRabbit's nitpick: consolidated `setupWithLoaderDeps` and `setupWithLoaderDepsForCaching` into a single `createFooRouter(opts)` helper (commit `40e7176`). Less duplication, easier to extend.

packages/router-core/src/router.ts
```diff
@@ -2426,6 +2436,23 @@ export class RouterCore<
                         previousMatches.some((d) => d.id === match.id),
                       )
```
Sheraff: We don't use those anymore, do we?

sleitor: You're right —  and  are no longer used after the split. Only  is still needed (for populating ). Removed both in the latest commit.

PR #6778, opened 2026-02-27
fix(router-generator): skip generated route tree in watchChange to prevent infinite loop

### LINKED ISSUE #6775: Router generator infinite loop: safeFileWrite mtime race with Vite watcher
## Description

The route generator enters an infinite loop on startup, flooding the console with:

```
File /path/to/src/routeTree.gen.ts was modified by another process during processing.
```

This causes Vite to continuously full-page reload the browser, making the app unusable in development.

## Reproduction

- **OS:** macOS (Darwin 25.4.0, APFS filesystem)
- **Versions:** `@tanstack/router-plugin@1.163.2`, `@tanstack/router-generator@1.163.2`, `vite@7.3.1`
- **Route count:** ~30 routes (layout routes + leaf routes)

Minimal `vite.config.ts`:

```ts
import react from '@vitejs/plugin-react'
import { TanStackRouterVite } from '@tanstack/router-plugin/vite'

export default defineConfig({
  plugins: [TanStackRouterVite(), react()],
})
```

Running `vite` (dev server) immediately produces the infinite loop. A clean `rm routeTree.gen.ts` + restart does not help.

## Root Cause

Traced through the source — the feedback loop is:

1. Generator writes `routeTree.gen.ts` via `safeFileWrite` (mtime strategy)
2. Vite's file watcher detects the write → fires `watchChange` on the plugin
3. `isFileRelevantForRouteTreeGeneration()` returns `true` for `generatedRouteTreePath`
4. `generator.run()` is called with an `update` event for the generated file
5. `safeFileWrite` compares `beforeStat.mtimeMs` against `opts.strategy.expectedMtimeMs` — they differ (because Vite's watcher or macOS APFS updated metadata between stat calls)
6. Throws `rerun({ msg: "File ... was modified by another process during processing." })`
7. The `rerun` error is caught and re-queued → goto 4

The design assumes "external modifications will eventually cease" but Vite's watcher IS the external modifier, so the loop never terminates.

**Key code locations:**
- `router-generator/dist/esm/generator.js` → `safeFileWrite()` (line ~844)
- `router-generator-plugin.js` → `watchChange()` blindly calls `generate()` for all files

## Workaround

Adding the generated file to Vite's watcher ignore list breaks the feedback loop:

```ts
export default defineConfig({
  server: {
    watch: {
      ignored: ['**/routeTree.gen.ts'],
    },
  },
})
```

This works because the plugin still detects actual `.tsx` route file changes through `watchChange` — it only needs to stop receiving events for its own output file.

## Suggested Fix

Either:
- **Option A:** Filter out `generatedRouteTreePath` in the `watchChange` handler before calling `generate()`
- **Option B:** Add a max-retry limit to the `rerun` mechanism in `safeFileWrite`
- **Option C:** Have the plugin configure Vite's watcher to ignore the generated file automatically

Option A seems cleanest — the generator should never re-trigger from changes to its own output when running inside Vite.

## Platform Note

This may be more likely on macOS/APFS due to nanosecond timestamp precision and fast filesystem events, making the mtime race window larger. May not reproduce on ext4/Linux.

sleitor:
## Problem

On macOS (APFS) with ~30+ routes, the route generator enters an infinite loop on dev server startup, flooding the console with:

```
File /path/to/src/routeTree.gen.ts was modified by another process during processing.
```

This causes Vite to continuously full-page reload the browser.

## Root Cause

The feedback loop in `watchChange`:

1. Generator writes `routeTree.gen.ts` via `safeFileWrite`
2. Vite's file watcher detects the write → fires `watchChange`
3. `isFileRelevantForRouteTreeGeneration()` returns **`true`** for `generatedRouteTreePath`
4. `generator.run()` is called for the generated file
5. `safeFileWrite` detects mtime mismatch → throws `rerun`
6. Rerun is caught and re-queued → goto 4

The bug is that `isFileRelevantForRouteTreeGeneration` returned `true` for the generator's own output file.

## Fix

Return `false` for `generatedRouteTreePath` in `isFileRelevantForRouteTreeGeneration`. The generator should never re-run because its own output changed — only changes to actual route source files should trigger regeneration.

This is **Option A** from the issue analysis — the cleanest approach.

## Test

The user's documented workaround was:
```ts
server: { watch: { ignored: ['**/routeTree.gen.ts'] } }
```

This fix makes that workaround unnecessary by filtering at the generator level.

Fixes #6775

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue that could cause the route generation process to enter an infinite loop when monitoring files for changes. Route generation now works more efficiently without unnecessary regeneration cycles.

### DIFF
--- a/packages/router-generator/src/generator.ts
+++ b/packages/router-generator/src/generator.ts
@@ -1624,9 +1624,12 @@ ${acc.routeTree.map((child) => `${child.variableName}Route: typeof ${getResolved
 
   // only process files that are relevant for the route tree generation
   private isFileRelevantForRouteTreeGeneration(filePath: string): boolean {
-    // the generated route tree file
+    // Skip the generated route tree file itself — changes to it are caused by
+    // the generator writing its own output and must not re-trigger generation,
+    // otherwise the Vite watcher creates an infinite feedback loop:
+    // write → watchChange → isFileRelevant=true → run() → write → …
     if (filePath === this.generatedRouteTreePath) {
-      return true
+      return false
     }
 
     // files inside the routes folder


schiller-manuel:
this is not the proper fix. as i wrote in the issue, we first need a reproducer and need to find out why the mtime check fails

sleitor:
Thanks for the feedback, @schiller-manuel! You're right — I jumped ahead to a fix without first establishing a proper reproducer or confirming the root cause of the mtime check failure.

I'll look into the issue more carefully: understand why the mtime-based deduplication fails in practice (is it a sub-millisecond write race, or something else?), and put together a minimal reproducer. Once the root cause is clear, the fix can be much more targeted.

Closing this PR for now — will reopen or submit a new one once I have a proper diagnosis.

PR #6809, opened 2026-03-03
docs(start): add hostname-based multi-tenancy tutorial

harshG775:
## Summary

Adds a new React tutorial: "Hostname-Based Multi-Tenancy" for TanStack Start.

## What This Covers

- Hostname-based tenant resolution during SSR
- Server-only tenant lookup using createServerOnlyFn
- Root-level loader hydration
- Dynamic metadata handling
- Production considerations

## Why

Multi-tenancy is a common SaaS requirement. This tutorial provides
a clean SSR-first approach aligned with TanStack Start architecture.

Tested locally against latest Start version.

## Summary by CodeRabbit

* **Documentation**
  * Added a new React tutorial on hostname-based multi-tenancy covering architecture, server-side tenant detection, root-level data hydration, dynamic metadata, component usage, production considerations (caching, validation, asset handling, security), and complete examples.
  * Placed the tutorial into the React Tutorials listing for improved discovery and navigation.

### DIFF
--- a/docs/start/framework/react/tutorial/multi-tenant-setup.md
+++ b/docs/start/framework/react/tutorial/multi-tenant-setup.md
@@ -3,66 +3,178 @@ title: 'Hostname-Based Multi-Tenancy'
 description: 'Learn how to structure a hostname-based multi-tenant app using TanStack Start with React.'
 ---
 
-> This tutorial assumes TanStack Start v1.132+.
+> This tutorial assumes @tanstack/react-router: v1.132+.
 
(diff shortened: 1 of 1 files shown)


schiller-manuel:
wouldnt recommend using root loader data for this. we should use router context instead. this will soon be much easier when we finalize the start context bridge concept. for now we'll keep this open

harshG775:
Thanks for the suggestion!

I've updated the tutorial to resolve the tenant inside `router.tsx` and pass it through the router context instead of using root loader data.

This should align better with the recommended architecture.

PR #6811, opened 2026-03-03
fix: handle beforeLoad throwing notFound correctly

### LINKED ISSUE #6779: Hydration error when throwing notFound() in beforeLoad of a child route

Start

### Describe the bug

Throwing `notFound()` from beforeLoad within a child route leads to a hydration error.

```
Uncaught Error: Hydration failed because the server rendered HTML didn't match the client. As a result this tree will be regenerated on the client. This can happen if a SSR-ed Client Component used:

- A server/client branch `if (typeof window !== 'undefined')`.
- Variable input such as `Date.now()` or `Math.random()` which changes each time it's called.
- Date formatting in a user's locale which doesn't match the server.
- External changing data without sending a snapshot of it along with the HTML.
- Invalid HTML tag nesting.

It can also happen if the client has a browser extension installed which messes with the HTML before React loaded.

https://react.dev/link/hydration-mismatch

  ...
    <RouterProvider router={{...}}>
      <RouterContextProvider router={{...}}>
        <Matches>
          <SafeFragment fallback={null}>
            <Transitioner>
            <MatchesInner>
              <CatchBoundary getResetKey={function getResetKey} errorComponent={function ErrorComponent} ...>
                <CatchBoundaryImpl getResetKey={function getResetKey} onCatch={function}>
                  <MatchImpl matchId="__root__/">
                    <RootDocument>
                      <html>
                        <head>
                          <HeadContent>
                            <Asset>
                            <Asset>
                            <Asset>
                            <Asset>
                            <Asset>
                            <Asset>
                            <Asset>
                            <Asset>
                            <Asset>
                            <Asset>
                            <Asset>
                            <Asset>
                            <Asset tag="link" attrs={{rel:"style...", ...}} nonce={undefined}>
+                             <link
+                               rel="stylesheet"
+                               href="/src/styles/app.css"
+                               nonce={undefined}
+                               suppressHydrationWarning={true}
+                             >
-                             <script src="https://codesandbox.io/p/chrome-devtool/protocol/index.js">
                            ...
                        ...

    at throwOnHydrationMismatch (react-dom_client.js?v=07a9c313:3918:13)
    at beginWork (react-dom_client.js?v=07a9c313:8692:196)
    at runWithFiberInDEV (react-dom_client.js?v=07a9c313:995:72)
    at performUnitOfWork (react-dom_client.js?v=07a9c313:12559:98)
    at workLoopConcurrentByScheduler (react-dom_client.js?v=07a9c313:12555:11)
    at renderRootConcurrent (react-dom_client.js?v=07a9c313:12537:71)
    at performWorkOnRoot (react-dom_client.js?v=07a9c313:11764:152)
    at performWorkOnRootViaSchedulerTask (react-dom_client.js?v=07a9c313:13503:9)
    at MessagePort.performWorkUntilDeadline (react-dom_client.js?v=07a9c313:34:50)
```

https://codesandbox.io/p/devbox/friendly-haze-v4nyyq

1. Visit the a posts page (https://v4nyyq-3000.csb.app/posts/abc)
2. See an initial lack of styles applied and check the console

### Expected behavior

I would expect to see the styles applied on first paint 

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.163.2

### Additional context

in a child route - 
throwing error from loader ✅ 
throwing notFound from loader ✅ 
throwing error from beforeLoad ✅ 
throwing notFound from beforeLoad ❌

schiller-manuel:
fixes #6779

## Summary by CodeRabbit

* **New Features**
  * Enhanced not-found behavior: beforeLoad can trigger targeted notFound boundaries (including root) while preserving parent loader data; per-route notFoundComponent support; new demo routes and navigation to exercise deep and parent-boundary cases; exported rootRouteId for targeted notFound.

* **Documentation**
  * Updated guide and migration example to explain boundary resolution and show notFoundComponent usage instead of prior API.

* **Tests**
  * Expanded E2E and unit tests across frameworks for beforeLoad not-found, routeId targeting, deep/parent-boundary scenarios, and hydration global-not-found checks.

### DIFF
--- a/packages/router-core/src/load-matches.ts
+++ b/packages/router-core/src/load-matches.ts
@@ -962,8 +962,8 @@ export async function loadMatches(arg: {
       }
       if (isNotFound(reason)) {
         firstNotFound ??= reason
-      } else if (firstUnhandledRejection === undefined) {
-        firstUnhandledRejection = reason
+      } else {
+        firstUnhandledRejection ??= reason
       }
     }
 
--- a/packages/router-core/src/ssr/ssr-client.ts
+++ b/packages/router-core/src/ssr/ssr-client.ts
@@ -28,7 +28,14 @@ function hydrateMatch(
   match.ssr = deyhydratedMatch.ssr
   match.updatedAt = deyhydratedMatch.u
   match.error = deyhydratedMatch.e
-  match.globalNotFound = deyhydratedMatch.g
+  // Only hydrate global-not-found when a defined value is present in the
+  // dehydrated payload. If omitted, preserve the value computed from the
+  // current client location (important for SPA fallback HTML served at unknown
+  // URLs, where dehydrated matches may come from `/` but client matching marks
+  // root as globalNotFound).
+  if (deyhydratedMatch.g !== undefined) {
+    match.globalNotFound = deyhydratedMatch.g
+  }
 }
 
 export async function hydrate(router: AnyRouter): Promise<any> {
--- a/packages/router-core/tests/hydrate.test.ts
+++ b/packages/router-core/tests/hydrate.test.ts
@@ -305,7 +305,7 @@ describe('hydrate', () => {
     expect((mockMatches[0] as AnyRouteMatch).globalNotFound).toBeUndefined()
   })
 
-  it('should clear stale globalNotFound when dehydrated flag is omitted', async () => {
+  it('should preserve existing globalNotFound when dehydrated flag is omitted', async () => {
     const mockMatches = [
       {
         id: '/',
@@ -346,7 +346,7 @@ describe('hydrate', () => {
 
     await hydrate(mockRouter)
 
-    expect((mockMatches[0] as AnyRouteMatch).globalNotFound).toBeUndefined()
+    expect((mockMatches[0] as AnyRouteMatch).globalNotFound).toBe(true)
   })
 
   it('should decode dehydrated match ids before hydration lookup and SPA-mode checks', async () => {


packages/router-core/src/load-matches.ts
```diff
@@ -906,9 +890,8 @@ export async function loadMatches(arg: {
   updateMatch: UpdateMatchFn
   sync?: boolean
 }): Promise<Array<MakeRouteMatch>> {
-  const inner: InnerLoadContext = Object.assign(arg, {
-    matchPromises: [],
-  })
+  const inner: InnerLoadContext = Object.assign(arg)
```
Sheraff: This line doesn't do anything anymore, it's just 
```suggestion
  const inner: InnerLoadContext = arg
```

packages/router-core/src/load-matches.ts
```diff
@@ -919,77 +902,159 @@ export async function loadMatches(arg: {
...
+  } catch {
+    const settled = await Promise.allSettled(matchPromises)
+
+    for (const result of settled) {
+      if (result.status !== 'rejected') continue
+
+      const reason = result.reason
+      if (!firstRedirect && isRedirect(reason)) {
+        firstRedirect = reason
+      }
```
Sheraff: We could break right here, right? if we have a redirect and no notFound (yet) no need to continue?
```suggestion
      if (!firstRedirect && isRedirect(reason)) {
        throw reason
      }
```

packages/router-core/src/load-matches.ts
```diff
@@ -919,77 +902,159 @@ export async function loadMatches(arg: {
...
 
-    const readyPromise = triggerOnReady(inner)
-    if (isPromise(readyPromise)) await readyPromise
-  } catch (err) {
-    if (isNotFound(err) && !inner.preload) {
-      const readyPromise = triggerOnReady(inner)
-      if (isPromise(readyPromise)) await readyPromise
-      throw err
+  if (firstRedirect) {
+    throw firstRedirect
```
Sheraff: Should the redirection be honored if it was found "under" a `notFound`? I feel like whichever first error we see (notFound or redirect) is the one that should be used

tannerlinsley: I think first one in wins, especially here.

packages/router-core/src/load-matches.ts
tannerlinsley: I think so yes.

packages/router-core/src/load-matches.ts
```diff
@@ -74,64 +73,45 @@ const buildMatchContext = (
...
-    // Ensure we have a notFoundComponent
-    invariant(
-      routeCursor.options.notFoundComponent,
-      'No notFoundComponent found. Please set a notFoundComponent on your route or provide a defaultNotFoundComponent to the router.',
-    )
-  }
-
-  // Find the match for this route
-  const matchForRoute = inner.matches.find((m) => m.routeId === routeCursor.id)
+  const requestedRouteId = err.routeId ? String(err.routeId) : undefined
```
Sheraff: Why `String(err.routeId)` here? what else could it be if not already a string?

packages/router-core/src/load-matches.ts
schiller-manuel: this is the scenario where a loader throws a `redirect` after a child beforeLoad threw a `notFound`

i think we cannot honor the `notFound` since there is no loader data for this route and so we cannot render its route component (the types would not make the loader data optional if you throw a `redirect`)

PR #6825, opened 2026-03-05
fix(solid-router): fix pendingComponent stuck on loading with ssr: 'data-only'

### LINKED ISSUE #6824: [solid-router] pendingComponent stuck on loading screen with ssr: 'data-only'
solid-router

### Describe the bug

When using `ssr: 'data-only'` (or `ssr: false`) together with a `pendingComponent` on a route, the page gets stuck on the loading/pending screen and never transitions to the actual route component after hydration.

_Minimal reproduction steps below_

### Steps to Reproduce

1. Create a route with `ssr: 'data-only'` and a `pendingComponent`:

```tsx
export const Route = createFileRoute('/demo')({
  ssr: 'data-only',
  loader: () => fetchData(),
  pendingComponent: () => <div>Loading...</div>,
  component: DemoRoute,
})
```

2. Run with SSR enabled (TanStack Start)
3. Navigate to the route — the `pendingComponent` renders on the server, but after hydration on the client, it stays visible and the actual `component` never appears

### Expected Behavior

After hydration, the `pendingComponent` should be replaced by the actual route `component`.

### Actual Behavior

The `pendingComponent` (loading spinner) remains permanently visible and obscures the actual route component. DOM inspection reveals both elements are present in the DOM, but the server-rendered pending component is not being removed.

### Root Cause Analysis

In `packages/solid-router/src/Match.tsx`, the no-SSR branch uses:

```tsx
<Solid.Show
  when={!(isServer ?? router.isServer)}
  fallback={<Dynamic component={resolvePendingComponent()} />}
>
  <MatchInner matchId={props.matchId} />
</Solid.Show>
```

During SSR: `isServer` is `true`, so `when` is `false` → `fallback` (pendingComponent) is rendered.
During client hydration: `isServer` is `false`, so `when` becomes `true` → should show `MatchInner`.

However, SolidJS's `<Show>` component has a hydration issue where it fails to properly remove the server-rendered fallback DOM when transitioning from `fallback` → `children` during hydration. The fallback DOM stays in the document and the pending component remains visible.

### Proposed Fix

Replace the `<Show>` fallback approach with a `MatchNoSsr` helper component that:
- Renders the `pendingComponent` inside a wrapper `<div style="display:contents">`
- After hydration, hides the wrapper via `style.display = 'none'` using a ref + `createEffect`
- Renders `MatchInner` through a `<Show when={hydrated()}>` guard (using the existing `useHydrated` from `ClientOnly.tsx`)

This avoids the SolidJS `<Show>` hydration cleanup issue by keeping the pending DOM in place and hiding it via CSS instead of relying on DOM removal.

I've already prepared a PR with the fix and tests: will link shortly.

### Environment

- @tanstack/solid-router: latest (main branch)
- solid-js: 1.9.x
- TanStack Start with Vinxi/Nitro SSR

ljho01:
## Summary

Fixes #6824

When using `ssr: 'data-only'` or `ssr: false` together with a `pendingComponent`, the page gets stuck on the loading screen and never transitions to the actual route component after hydration.

### Root Cause

The previous no-SSR branch in `Match.tsx` uses:

```tsx
<Solid.Show
  when={!(isServer ?? router.isServer)}
  fallback={<Dynamic component={resolvePendingComponent()} />}
>
  <MatchInner matchId={props.matchId} />
</Solid.Show>
```

This creates a **hydration mismatch**: `when` is `false` on the server but immediately `true` on the client. SolidJS's `<Show>` fails to properly remove the server-rendered fallback DOM during this hydration transition, leaving the pending component permanently visible and obscuring the actual route content.

### Solution

Introduces a `MatchNoSsr` helper that uses `useHydrated()` (from the existing `ClientOnly.tsx`):

```tsx
const MatchNoSsr = (props) => {
  const hydrated = useHydrated()
  return (
    <Solid.Show
      when={hydrated()}
      fallback={<Dynamic component={props.pendingComponent()} />}
    >
      <MatchInner matchId={props.matchId} />
    </Solid.Show>
  )
}
```

Both server and client start with `hydrated() = false` (no mismatch), so hydration succeeds cleanly. After `onMount`, `hydrated()` flips to `true` and `<Show>` performs a **normal runtime transition** from fallback → children, which works correctly.

### Changes

- `packages/solid-router/src/Match.tsx` — Added `MatchNoSsr` component using `useHydrated()`; replaced `Solid.Switch`/`Solid.Show` block
- `packages/solid-router/tests/MatchNoSsr.test.tsx` — Client-side tests
- `packages/solid-router/tests/server/MatchNoSsr.test.tsx` — SSR tests

## Test Plan

- [x] Client test: `ssr: 'data-only'` + `pendingComponent` → actual content renders
- [x] Client test: `ssr: false` + `pendingComponent` → actual content renders
- [x] Client test: `defaultPendingComponent` with `ssr: 'data-only'` works
- [x] Client test: baseline (no ssr option) still works
- [x] Server test: `pendingComponent` renders during SSR for `data-only` routes
- [x] Server test: `pendingComponent` renders during SSR for `ssr: false` routes
- [x] Server test: `defaultPendingComponent` renders during SSR
- [x] Server test: route component is NOT rendered during SSR
- [x] Manual test: verified fix in a real TanStack Start app

## Summary by CodeRabbit

* **New Features**
  * Improved hydration-safe handling for routes opting out of SSR, ensuring correct fallback UI until client hydration.
  * Added pending (loading) indicators to posts and postId routes for selective SSR scenarios.

* **Tests**
  * Added extensive unit, server, and e2e tests validating data-only, client-only, and default pending-component behaviors across SSR and hydration.

### DIFF
--- a/e2e/solid-start/selective-ssr/tests/app.spec.ts
+++ b/e2e/solid-start/selective-ssr/tests/app.spec.ts
@@ -39,35 +39,3 @@ test.describe('selective ssr', () => {
     })
   }
 })
-
-test.describe('pendingComponent with ssr data-only/false', () => {
-  test('pendingComponent does not stay visible with ssr data-only', async ({
-    page,
-  }) => {
-    // testcase 7: posts and postId both have ssr: 'data-only' with pendingComponent
-    await page.goto('/')
-    await page.getByTestId('testcase-7-link').click()
-
-    // wait for hydration and route component to render
-    await expect(page.getByTestId('postId-heading')).toContainText('postId')
-
-    // pendingComponent must not be visible after hydration
-    await expect(page.getByTestId('posts-pending')).not.toBeVisible()
-    await expect(page.getByTestId('postId-pending')).not.toBeVisible()
-  })
-
-  test('pendingComponent does not stay visible with ssr false', async ({
-    page,
-  }) => {
-    // testcase 8: posts and postId both have ssr: false with pendingComponent
-    await page.goto('/')
-    await page.getByTestId('testcase-8-link').click()
-
-    // wait for hydration and route component to render
-    await expect(page.getByTestId('postId-heading')).toContainText('postId')
-
-    // pendingComponent must not be visible after hydration
-    await expect(page.getByTestId('posts-pending')).not.toBeVisible()
-    await expect(page.getByTestId('postId-pending')).not.toBeVisible()
-  })
-})


schiller-manuel:
we need a failing e2e test first for this.
please adapt or enhance`e2e/solid-start/selective-ssr/`

ljho01:
Closing this PR — the issue turned out to be a misconfiguration on my side, not a bug in solid-router.

I was placing `<HeadContent />` inside the `<head>` tag instead of in `<body>`, which broke hydration for `ssr: 'data-only'` routes. The Solid examples correctly show `HeadContent` in `<body>` since it uses portals internally.

Sorry for the false alarm and the unnecessary review work! 🙏

ljho01:
**Update: found additional root cause**

The `HeadContent` placement wasn't the only issue. Having **any non-route components inside the root SSR `<Suspense>` boundary** also triggers the same hydration mismatch.

Components like `<Toaster />` and `<OverlayHost>` (which run side effects at render time or during `onMount`) must be placed **outside** the `<Suspense>` that wraps `<Outlet />` in the shell component. The server generates HTML chunks per Suspense boundary, and extra components inside that boundary cause DOM node misalignment during client hydration.

See the updated comment on #6824 for the full breakdown and correct structure.

PR #6828, opened 2026-03-05
fix vite preview streaming

huseeiin:
now streaming works in `vite preview`

## Summary by CodeRabbit

* **Chores**
  * Server now sets Content-Encoding to identity only for HTML responses (no change to other response types).
  * Added a changeset entry for a patch release.
  * Added a preview script and a new e2e preview test script; added a default preview target in CI config.

* **Tests**
  * Added an end-to-end test validating incremental HTML streaming in preview mode and updated test runner configuration to support preview vs. start modes.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -55,6 +55,11 @@
       "dependsOn": ["^build"],
       "inputs": ["default", "^production"]
     },
+    "test:e2e:preview": {
+      "cache": true,
+      "dependsOn": ["^build"],
+      "inputs": ["default", "^production"]
+    },
     "test:types": {
       "cache": true,
       "dependsOn": ["^build"],


packages/start-plugin-core/src/preview-server-plugin/plugin.ts
```diff
@@ -20,6 +20,7 @@ export function previewServerPlugin(): Plugin {
           let serverBuild: any = null
 
           server.middlewares.use(async (req, res, next) => {
+            res.setHeader('content-encoding', 'identity')
```
schiller-manuel: is this only required for solid?

huseeiin: > is this only required for solid?

i don't think so. its for all streaming frameworks

PR #6832, opened 2026-03-06
benchmarks: bundle client-nav runtimes

Sheraff:
## Summary
- move client-nav benchmark mounting for react, solid, and vue into each built app bundle instead of leaving framework render APIs in the harness
- bundle the framework runtimes in the benchmark build so the harness runs the built production client code more closely, which fixes Solid's incorrect runtime resolution
- keep the benchmark setup focused on creating the container, subscribing to router renders, and driving navigation only

## Testing
- Not run in this sandbox (`benchmarks/client-nav` dependencies are not installed, so `pnpm run test:types` fails with `tsc: not found`)

## Summary by CodeRabbit

* **Refactor**
  * Unified test app mounting across frameworks with an explicit mount/unmount lifecycle and improved cleanup; returned mounts include an unmount handler.

* **Chores**
  * Adjusted benchmark build configs so certain framework libraries are no longer treated as externals, affecting bundling and production build behavior.

### DIFF
--- a/benchmarks/client-nav/vue/vite.config.ts
+++ b/benchmarks/client-nav/vue/vite.config.ts
@@ -1,11 +1,16 @@
 import { defineConfig } from 'vitest/config'
+import vue from '@vitejs/plugin-vue'
 import vueJsx from '@vitejs/plugin-vue-jsx'
 import codspeedPlugin from '@codspeed/vitest-plugin'
 
 export default defineConfig({
+  define: {
+    'process.env.NODE_ENV': JSON.stringify('production'),
+  },
   plugins: [
     !!(process.env.VITEST && process.env.WITH_INSTRUMENTATION) &&
       codspeedPlugin(),
+    vue(),
     vueJsx(),
   ],
   build: {


benchmarks/client-nav/vue/app.tsx
```diff
@@ -99,7 +99,16 @@ export function createTestRouter() {
     routeTree: root.addChildren([route]),
   })
 
-  const component = <RouterProvider router={router} />
+  const app = Vue.createApp({
+    render: () => <RouterProvider router={router} />,
+  })
```
Sheraff, line 105:
```suggestion
  const component = <RouterProvider router={router} />
  const app = Vue.createApp({
    render: () => component,
  })
```

benchmarks/client-nav/react/app.tsx
```diff
@@ -84,7 +85,13 @@ export function createTestRouter() {
     routeTree: root.addChildren([route]),
```
Sheraff:
```suggestion
const rootRoute = createRootRoute({
  component: Root,
})

const route = createRoute({
  getParentRoute: () => root,
  path: '/$id',
  component: () => <div />,
})

export function mountTestApp(container: Element) {
  const router = createRouter({
    history: createMemoryHistory({
      initialEntries: ['/0'],
    }),
    scrollRestoration: true,
    routeTree: rootRoute.addChildren([route]),
```

PR #6838, opened 2026-03-06
devtools: update to solid 2

birkskyum:
Updates router-devtools-core from solid v1 to v2.

Solid-Router etc. is still on solid v1 - this is possible because router-devtools-core is bundled.

Blocked by:
- https://github.com/TanStack/devtools/pull/371

### DIFF
--- a/packages/router-devtools-core/src/BaseTanStackRouterDevtoolsPanel.tsx
+++ b/packages/router-devtools-core/src/BaseTanStackRouterDevtoolsPanel.tsx
@@ -6,11 +6,10 @@ import {
   Match,
   Show,
   Switch,
+  createEffect,
   createMemo,
   createSignal,
-  createTrackedEffect,
   onCleanup,
-  untrack,
 } from 'solid-js'
 import { useDevtoolsOnClose } from './context'
 import { useStyles } from './useStyles'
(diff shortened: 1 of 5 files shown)


packages/router-devtools-core/src/BaseTanStackRouterDevtoolsPanel.tsx
```diff
@@ -6,9 +6,9 @@ import {
   Match,
   Show,
   Switch,
-  createEffect,
   createMemo,
   createSignal,
+  createTrackedEffect,
```
brenelz: I think we want to be careful using `createTrackedEffect`. `createEffect` split function is better

packages/router-devtools-core/src/FloatingTanStackRouterDevtools.tsx
```diff
@@ -126,16 +129,16 @@ export function FloatingTanStackRouterDevtools({
 
   const isButtonClosed = isOpen() ?? false
 
-  createEffect(() => {
+  createTrackedEffect(() => {
     setIsResolvedOpen(isOpen() ?? false)
   })
 
-  createEffect(() => {
+  createTrackedEffect(() => {
```
brenelz: maybe use createEffect here?

packages/router-devtools-core/package.json
```diff
@@ -67,9 +67,10 @@
     "tiny-invariant": "^1.3.3"
   },
   "devDependencies": {
-    "solid-js": "^1.9.10",
+    "@solidjs/web": "2.0.0-beta.1",
+    "solid-js": "2.0.0-beta.1",
     "vite": "*",
-    "vite-plugin-solid": "^2.11.10"
+    "vite-plugin-solid": "3.0.0-next.0"
```
brenelz: there is a [3.0.0-next.2](https://github.com/solidjs/vite-plugin-solid/commit/2fba63821d10593feb0d8b0b46cbb5f36d46e431) now

birkskyum:
> Do we want to change solid-router-devtools too or no?

solid-router-devtools should stay on the same solid version as solid-router and solid-start, so not in this pr

PR #6839, opened 2026-03-06
Add extraEnvironments option to TanStack Start config

afn:
## Overview

This adds the ability to add additional [vite environments](https://vite.dev/guide/api-environment) besides the default `ssr` and `client` environments to a TanStack Start build.

## Use Case

This can be used to call TanStack Server Functions from client code in an external project --- for example, an Electron app which implements an alternate front-end to your TanStack Start app.

## Usage

Pass `extraEnvironments` to the `tanstackStart()` vite plugin, containing an array of `name` (referencing a vite environment name) and `type` (either `'client'` or `'server'`).

**Example:** The following vite.config.js will generate an additional output artifact, `dist/external-api.js`, from `src/external-api.ts`, which in turn exports several Server Functions. The resulting JS file will contain the client stub for the server function.

```typescript
// vite.config.js
import { defineConfig, normalizePath } from 'vite'
import tsConfigPaths from 'vite-tsconfig-paths'
import viteReact from '@vitejs/plugin-react'

export default defineConfig({
  server: {
    port: 3000,
  },
  environments: {
    externalAPI: {
      consumer: 'client',
      build: {
        emptyOutDir: false,
        lib: {
          entry: normalizePath('src/external-api.ts'),
          name: 'external-api',
          fileName: 'external-api',
          formats: ['es'],
        },
      },
      define: {
        'process.env.TSS_SERVER_FN_BASE': JSON.stringify('http://localhost:3000/_serverFn/'),
        'window.__TSS_START_OPTIONS__': JSON.stringify({}),
      },
    },
  },
  plugins: [
    tsConfigPaths(),
    tanstackStart({
      extraEnvironments: [
        { name: 'externalAPI', type: 'client' },
      ],
    }),
    viteReact(),
  ],
})
```

## Summary by CodeRabbit

* **New Features**
  * Extra environments support: Users can now configure additional custom environments by specifying a name and type (client or server). These extra environments will be automatically built and included in the final application build alongside standard environments.

### DIFF
--- a/packages/start-plugin-core/src/plugin.ts
+++ b/packages/start-plugin-core/src/plugin.ts
@@ -89,6 +89,7 @@ export function TanStackStartVitePluginCore(
     return bundle
   }
 
+  const extraEnvironmentNames: Array<string> = []
   const environments: Array<{ name: string; type: 'client' | 'server' }> = [
     { name: VITE_ENVIRONMENT_NAMES.client, type: 'client' },
     { name: VITE_ENVIRONMENT_NAMES.server, type: 'server' },
@@ -102,6 +103,11 @@ export function TanStackStartVitePluginCore(
       type: 'server',
     })
   }
+  if (Array.isArray(startPluginOpts?.extraEnvironments)) {
+    environments.push(...startPluginOpts.extraEnvironments)
+    extraEnvironmentNames.push(...startPluginOpts.extraEnvironments.map(e => e.name))
+  }
+
   return [
     {
       name: 'tanstack-start-core:config',
@@ -322,6 +328,13 @@ export function TanStackStartVitePluginCore(
             async buildApp(builder) {
               const client = builder.environments[VITE_ENVIRONMENT_NAMES.client]
               const server = builder.environments[VITE_ENVIRONMENT_NAMES.server]
+              const extraEnvironments = extraEnvironmentNames.map(environmentName => {
+                const environment = builder.environments[environmentName];
+                if (!environment) {
+                  throw new Error(`Environment ${environmentName} not found`);
+                }
+                return environment
+              })
 
               if (!client) {
                 throw new Error('Client environment not found')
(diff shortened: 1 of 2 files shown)


schiller-manuel:
not something we'll support. possibly you can do this directly using a custom vite plugin that uses the `buildApp` hook

afn:
I understand the rationale, and in fact the approach I initially took was to create a plugin with a `config` hook to define the new environment and a `buildApp` hook to build it. ([See gist here.](https://gist.github.com/afn/bdaaebfb668d1b0d832b36c81946ac60)) This still required a minor change to `@tanstack/start-plugin-core` to export `startCompilerPlugin`, but that seemed preferable to changing its behavior.

However, there are two issues with this approach. First, since there are now two instances of the compiler, the server function ids are generated separately for the externalAPI environment build. This isn't such a big deal with the default deterministic `generateFunctionId`. But second, because the "main" compiler doesn't visit the externalAPI environment, any server functions that are _only_ called from `external-api.ts` and not from the main client code are optimized away as server-only-referenced.

Taking a step back, my goal here is to provide a way for developers to build a full-stack app with two separate front ends --- a web app and an Electron app --- and to take advantage of the ergonomics of TanStack server functions to minimize friction and boilerplate. Maybe I'm going about this the wrong way, and maybe there's already a solution out there? What I had in mind was a monorepo that looks something like:

```
+ my-app/
|-+ packages/
  |-- electron/              # Electron app using electron-vite (depends on my-app/shared and my-app/web)
  |-- shared/                # Shared UI components, etc
  |-+ web/                   # TanStack Start app with web app and server-side (depends on my-app/shared)
    |-- src/external-api.ts  # This is exported and can be consumed my-app/electron
```

and providing a vite plugin that can be added to the `web` package to allow server functions to be consumed by the `electron` package.

If this is something that's already been explored/solved and I haven't found it, I'd appreciate some pointers! And if not, I'd be open to alternative/less-intrusive ways of hooking into the TanStack compiler besides the approach in this PR.

Thanks for replying so quickly!

PR #6867, opened 2026-03-08
fix: don't re-execute stale loaders on unrelated location changes

schiller-manuel:
## Summary by CodeRabbit

* **New Features**
  * Added ability to forcefully reload stale matches via explicit router.load() calls.

* **Bug Fixes**
  * Stale-match revalidation now occurs when routes are re-entered, loader dependencies change, or router.load() is called; background reload behavior made more consistent.

* **Documentation**
  * Clarified Router.load() semantics and updated staleTime guidance.

* **Tests**
  * Updated unit and end-to-end tests and expectations (including separating loader vs. context checks) to reflect the revised reload behavior.

### DIFF
--- a/packages/router-core/src/load-matches.ts
+++ b/packages/router-core/src/load-matches.ts
@@ -852,7 +852,7 @@ const loadRouteMatch = async (
     syncMatchContext(inner, matchId, index)
 
     if (isServer ?? inner.router.isServer) {
-      return match
+      return inner.router.getMatch(matchId)!
     }
   } else {
     const prevMatch = inner.router.getMatch(matchId)! // This is where all of the stale-while-revalidate magic happens


packages/router-core/src/load-matches.ts
```diff
@@ -817,8 +842,15 @@ const loadRouteMatch = async (
   const route = inner.router.looseRoutesById[routeId]!
 
   if (shouldSkipLoader(inner, matchId)) {
+    const match = inner.router.getMatch(matchId)
+    if (!match) {
+      return inner.matches[index]!
+    }
+
+    syncMatchContext(inner, matchId, index)
```
Sheraff, line 852: Can we avoid this 2nd `syncMatchContext` call if we went through the branch that contains the 1st one? Something like
```ts
let hasSynced = false
if (...) {
} else {
  syncMatchContext(inner, matchId, index)
  hasSynced = true
}
...
if (!hasSynced)
  syncMatchContext(inner, matchId, index)
```

packages/router-core/src/load-matches.ts
```diff
@@ -817,8 +842,15 @@ const loadRouteMatch = async (
...
+
     if (isServer ?? inner.router.isServer) {
-      return inner.router.getMatch(matchId)!
+      return match
```
Sheraff: Are you sure we can just `return match` here? With the current implementation, every time we call `updateMatch()` the `match` becomes a new object, so any previously held `match` will not contain up-to-date data.

packages/router-core/src/load-matches.ts
schiller-manuel: No change needed here.
- In packages/router-core/src/load-matches.ts, those two syncMatchContext(...) calls are in mutually exclusive branches of the same loadRouteMatch invocation:
  - one is inside handleLoader(...) when we decide not to run the loader
  - the other is in the top-level if (shouldSkipLoader(...)) branch
- A single call cannot hit both, because handleLoader(...) only runs in the else branch.

packages/router-core/src/load-matches.ts
schiller-manuel: reverted this, good catch

PR #6875, opened 2026-03-09
fix: ensure lazy error component is preloaded

### LINKED ISSUE #6873: TSR ignores error component definitions in lazy routes

Router

### Describe the bug

Similar to #1311, I am having trouble making TSR recognize error components that are defined in lazy routes. They are properly displayed when defined in regular routes. I have tested this with file-based routing (see stackblitz), but it also happens if the route is later updated with `lazy()`. The setup for the latter is a bit more involved, and I do not have a MWE ready.

Dependencies in `package.json`:

```
"dependencies": {
  "@tanstack/react-router": "1.166.2",
  "@tanstack/router-plugin": "1.166.2",
  "react": "19.2.4",
  "react-dom": "19.2.4"
},
"devDependencies": {
  "@types/react": "19.2.10",
  "@types/react-dom": "19.2.3",
  "@vitejs/plugin-react-swc": "4.2.2",
  "vite": "7.3.1"
}
```

https://stackblitz.com/edit/tanstack-router-on4kqf5s

1. Open the linked stackblitz.
2. Note that file `src/routes/index.lazy.tsx` has an error component defined.
3. However, the index page of the sample project still displays the default error component.

### Expected behavior

As a developer, I expect the error components defined on lazy routes to be recognized by TSR, but that's not currently happening.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.166.2 (could be earlier)
- OS: Linux
- Browser: Firefox / Chrome
- Browser Version: 148.0 (debian) / 145.0 (ungoogled-chromium)
- Bundler: vite
- Bundler Version: 7.3.1

### Additional context

_No response_

schiller-manuel:
fixes #6873

## Summary by CodeRabbit

* **New Features**
  * Added a new lazy route at /lazy-error that demonstrates lazy-loaded route behavior with a custom error UI when the route loader fails.

* **Tests**
  * Added end-to-end and unit tests confirming the lazy route renders its specific error component (and suppresses generic error UI) when initialization throws.

### DIFF
--- a/packages/router-core/src/load-matches.ts
+++ b/packages/router-core/src/load-matches.ts
@@ -1163,11 +1163,9 @@ function preloadRouteComponents(
     .map((type) => (route.options[type] as any)?.preload?.())
     .filter(Boolean)
 
-  if (preloads.length) {
-    return Promise.all(preloads).then(() => {})
-  }
+  if (preloads.length === 0) return undefined
 
-  return undefined
+  return Promise.all(preloads) as any as Promise<void>
 }
 
 export function loadRouteChunk(


packages/router-core/src/load-matches.ts
```diff
@@ -1145,7 +1149,27 @@ export async function loadMatches(arg: {
...
+  | 'errorComponent'
+  | 'pendingComponent'
+  | 'notFoundComponent'
+
+async function preloadRouteComponents(
+  route: AnyRoute,
+  componentTypesToLoad: Array<RouteComponentType>,
+) {
+  await Promise.all(
+    componentTypesToLoad.map((type) =>
```
Sheraff: Nitpick: i'm bundle-size conscious now
```suggestion
function preloadRouteComponents(
  route: AnyRoute,
  componentTypesToLoad: Array<RouteComponentType>,
) {
  return Promise.all(
    componentTypesToLoad.map((type) =>
```

packages/router-core/src/load-matches.ts
```diff
@@ -1160,30 +1184,22 @@ export async function loadRouteChunk(route: AnyRoute) {
...
+      ? undefined
+      : componentTypesToLoad === componentTypes
+        ? (route._componentsPromise ??= preloadRouteComponents(
+            route,
+            componentTypes,
+          ).then(() => {
+            route._componentsLoaded = true
+            route._componentsPromise = undefined // gc promise, we won't need it anymore
+          }))
+        : preloadRouteComponents(route, componentTypesToLoad)
```
Sheraff, line 1212: This seems like a compact way of loading just what we need so it's nice. But I'm just pointing this out in case it was not on purpose: `errorComponent` and `notFoundComponent` are never "marked as loaded" and so they will always require a promise to be displayed (until we actually end up loading all lazy components later on).

This is probably fine because error and notFound are exceptions and not the regular case, and the browser won't actually load them a 2nd time. It's just 1 extra microtask.

Sheraff (approved):

packages/router-core/src/load-matches.ts
schiller-manuel: yes let's not optimize for the exception case

PR #6926, opened 2026-03-15
build: update to vite-config 0.5.x (rolldown)

birkskyum:
Patched similarly to:
- https://github.com/TanStack/config/pull/354

This upgrades the workspace build environment from rollup to rolldown.

## Summary by CodeRabbit

* **Chores**
  * Updated multiple @tanstack packages and developer tooling, including Vite (7.x → 8.x) and @tanstack/vite-config (0.4 → 0.5); applied patch bumps and adjusted build/module emission targets.
* **Refactor**
  * Unified the Solid Router Block API into a single exported form while preserving existing behavior and compatibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
       "vite": "$vite",
       "@benchmarks/bundle-size>vite": "^7.3.1",
       "@benchmarks/client-nav>vite": "^7.3.1",
+      "@benchmarks/ssr>vite": "^7.3.1",
       "@types/node": "$@types/node",
       "@playwright/test": "$@playwright/test",
       "@tanstack/react-query": "$@tanstack/react-query",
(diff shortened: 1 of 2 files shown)


Sheraff:
Should we worry that this has a negative impact on both bundle size and client-side navigation? I haven't had the chance to make this same migration at work, so I don't know if this is expected or not. But i'm worried lol

birkskyum:
@Sheraff  I think we absolutely should worry about that.

Will set this PR to draft until we understand it better.

birkskyum:
Re-running locally, i can't verify a regression, it appear to be within noise. How do we re-run the CodSpeed ?

The increase size in bundle is because of how rolldown treat CJS interop ( and react still being CJS-only, which comes at a cost )

- https://github.com/facebook/react/issues/11503

PR #6934, opened 2026-03-16
refactor(solid-router): untrack dynamic store creation

Sheraff:


packages/solid-router/src/Match.tsx
```diff
@@ -25,7 +25,8 @@ export const Match = (props: { matchId: string }) => {
   const match = Solid.createMemo(() => {
     const id = props.matchId
     if (!id) return undefined
-    return router.stores.activeMatchStoresById.get(id)?.state
+    return Solid.untrack(() => router.stores.activeMatchStoresById.get(id))
+      ?.state
```
Sheraff:
```suggestion
    return router.stores.activeMatchStoresById.get(id)?.state
```

PR #6938, opened 2026-03-16
feat: integrate solid v2 query

brenelz:


examples/solid/start-basic-solid-query/src/routes/deferred.tsx
```diff
@@ -41,13 +41,14 @@ function Deferred() {
 
 function DeferredQuery() {
   const deferredQuery = useQuery(() => deferredQueryOptions())
+  const data = () => deferredQuery.data
```
birkskyum: why is this necessary now?

brenelz: Hmm i don't think its required... but not 100% sure

PR #6941, opened 2026-03-16
fix(router-cli): pass process.argv.slice(2) to yargs for proper argument parsing

ThePott:
# Problem
- `tsr generate` and `tsr watch` commands exit silently without executing

# Cause
- calls `yargs()` without arguments
- docs of yargs says to pass argument to either of `yargs()` or `parse()`
- without argument of `yargs`, the parsed result is `{ _: [] }`

# Fix
- pass `process.argv.slice(2)` to yargs

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI argument handling corrected so commands process only user-provided arguments, ensuring reliable execution and help output.

* **New Features**
  * CLI now requires at least one command and enforces strict command parsing to surface invalid usage promptly.

### DIFF
--- a/packages/router-cli/src/index.ts
+++ b/packages/router-cli/src/index.ts
@@ -9,6 +9,8 @@ export function main() {
   yargs(process.argv.slice(2))
     .scriptName('tsr')
     .usage('$0 <cmd> [args]')
+    .demandCommand(1)
+    .strictCommands()
     .command('generate', 'Generate the routes for a project', async () => {
       const config = getConfig()
       await generate(config, process.cwd())


jmalmo:
Hit this exact issue in CI yesterday. Our `tsr generate` was exiting 0 but producing no output, which silently broke our typecheck step (routeTree.gen.ts is gitignored so CI has to generate it fresh). Took a while to figure out because the exit code was 0 and there was no error output at all.

We pinned to `@tanstack/router-cli@1.164.0` as a workaround, which confirmed the regression started in the 1.166.9 range.

The `process.argv.slice(2)` fix looks correct. One suggestion: it might be worth adding `demandCommand(1)` and `strictCommands()` to the yargs chain as well. That way if someone runs `tsr` with no subcommand or a typo, they get a clear error instead of a silent exit. Would have saved us a couple hours of debugging.

```diff
 yargs(process.argv.slice(2))
   .scriptName('tsr')
   .usage('$0 <cmd> [args]')
+  .demandCommand(1)
+  .strictCommands()
   .command('generate', ...)
   .command('watch', ...)
   .help().argv
```

packages/router-cli/src/index.ts
jmalmo, line 25:
```suggestion
    .demandCommand(1)
    .strictCommands()
    .help().argv
```

ThePott: Thank you for the kind guidance! I'll update yargs chain and commit.

ThePott:
Hello, I just noticed that a PR fixing the same bug has been merged (https://github.com/TanStack/router/pull/6981 ).

Since this was my first time attempting a contribution, I think my PR lacked some requirements. I'd like to know what I could have done better so my PR could have been accepted.

Issue #6044, opened 2025-12-08
Numeric search parameters wrongly parsed and wrapped with quotes


isnifer:
### Which project does this relate to?

Router

### Describe the bug

The problem is that Router interprets **Strings** that look like a **Number** as Numbers at some point.
This behavior introduces 2 major issues:

1. Numeric strings wrapped with **double quotes** — they shouldn't.
2. If the numbers (that I expect to read as strings in my schema) are too large — they are modified.

So when I will pass `&numericString=723421968459640832` to URL it will be transformed twice:

1. `723421968459640832` became `723421968459640800` (since it's too big for Number)
2. And it will be wrapped with `""` and became `&numericString="723421968459640800"`

And the code is here — https://github.com/isnifer/tanstack-issue-537/blob/master/src/routes/issue.tsx

And yes I see your response here https://github.com/TanStack/router/discussions/989#discussioncomment-8125891 to transform big numbers to add double quotes, but c'mon, it should not be double quotes for numeric strings at all.

**UPD:** I've added a second page with the same Zod schema where I need to parse `z.coerce.string()` value from `params` – and it `just works`. No double quotes, no transformation `URL => number => string`. And added example with EXPECTED issue of casting from string to number (`z.coerce.number()`) to show the difference.

### Your Example Website or App

https://tsr-issue-537.netlify.app

### Steps to Reproduce the Bug or Issue

1. Go to site
2. Click to link on top
3. Watch to search parameters in URL
4. Read steps on that page

### Expected behavior

As a user, I expected — numeric parameters even expected as strings would NOT be wrapped with double quotes and would NOT `parsed` after page transition. Since HUGE numbers will change their values to unpredictable like `723421968459640832` became `723421968459640800` in the provided example.

### Screenshots or Videos

I recorded an example and made it slow to 0.25x, you can see how the Router transforms the URL.
You can find out on the 00:08 how it makes these 2 changes.

https://github.com/user-attachments/assets/55d990ae-a286-4ee4-aa57-34f90668d843

### Platform

- Router / Start Version: 1.132.0
- OS: macOS
- Browser: Chrome
- Browser Version: 144.0.7559.4 beta (arm64)
- Bundler: vite
- Bundler Version: ^7.1.7

adbjo:
Here is a quick-fix if you need:
```ts
import { createRouter, parseSearchWith, stringifySearchWith } from '@tanstack/react-router';

function parseSearch(value: string) {
[...]

PR #6946, opened 2026-03-17
feat(): Add opt-in raw search input for validateSearch

### LINKED ISSUE #537: Numeric strings in search params issue
### Describe the bug

Consider the following scenario:

- You have a route where you accept a string as a search param. (validate it with zod)
- Pass a numeric string as a search param through the route. (e.g. `navigate({search: {test: '3'})`)
- You'll see that an error is thrown, this is because the router will receive the query param as a number, and the zod validator will consider it illegal as it is not a string

https://stackblitz.com/edit/github-fbut9a-bu53ju?file=src/IndexComponent.tsx

- Open the stackblitz
- Click on the `Set test param to 3 // should fail` button.
- An error will be thrown

### Expected behavior

I would hope the search param could be considered a string. I don't know if the casting could be done based on the validation?

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Firefox
- - Version: 110

### Additional context

_No response_

sandy2008:
## Summary

Fixes [#537](https://github.com/TanStack/router/issues/537) and [#6044](https://github.com/TanStack/router/issues/6044).

Today, `validateSearch` receives search params after the router’s default parsing/coercion step. That means numeric-looking values like `?folder=34324324235325352523` can be converted before validation, which breaks string schemas and can lose precision for large integer-like values.

This PR adds a non-breaking, opt-in way for validators to receive raw URL search values instead: `validateSearchWithRawInput(...)`.

## What changed

- Added `validateSearchWithRawInput(...)` in `router-core`
- Re-exported it from React, Solid, and Vue router packages
- Preserved existing default search parsing behavior for all current callers
- Updated route matching so wrapped validators receive raw decoded string values from the URL
- Added tests covering:
  - existing numeric parsing behavior still works
  - wrapped validators preserve numeric-looking strings from the URL
- Updated docs to clarify the default `validateSearch` input and document the new opt-in helper

## Why this approach

Changing the default parser behavior would be breaking because existing apps rely on values like `?page=1` or `?enabled=true` being coerced before validation.

This approach fixes the reported bug without changing current behavior for existing validators. Users only opt into raw string input when they need it.

## Example

```ts
validateSearch: validateSearchWithRawInput(
  z.object({
    folder: z.string(),
  }),
)
```

## Testing

Ran:

- `CI=1 NX_DAEMON=false pnpm nx run-many --target=test:unit --projects=@tanstack/router-core,@tanstack/react-router,@tanstack/zod-adapter --outputStyle=stream --skipRemoteCache`
- `CI=1 NX_DAEMON=false pnpm nx run-many --target=test:types --projects=@tanstack/router-core,@tanstack/react-router,@tanstack/solid-router,@tanstack/vue-router,@tanstack/zod-adapter --outputStyle=stream --skipRemoteCache`
- `CI=1 NX_DAEMON=false pnpm nx run-many --target=test:eslint --projects=@tanstack/router-core,@tanstack/react-router,@tanstack/solid-router,@tanstack/vue-router,@tanstack/zod-adapter --outputStyle=stream --skipRemoteCache`

## Summary by CodeRabbit

* **New Features**
  * Added validateSearchWithRawInput to opt into validating raw URL search values.
  * Search parsing now accepts a pluggable parser for custom value conversion.

* **Documentation**
  * Renamed and clarified the search validation parameter name and semantics.
  * Added guide examples demonstrating raw-URL validation usage.

* **Tests**
  * Added tests covering raw-URL search validation and preservation of raw string values.


devinrhode2:
I think we should have a global/default flag we can set too. Tanstack Router v2 can make that the default. In v1 it will of course have to be opt-in.

devinrhode2:
BONUS: for a separate PR:
  Regarding detecting and correcting bad urls.
  There could be utilities for detecting it and correcting it, but these are just utilities - users need to actually invoke them on their own. Some may have extra long numbers, some may not, who knows what kind of url behavior is going on out there... - better to not attempt to fix this for everyone and let them handle it themselves on a case by case basis. Then again, I can say for my app, I'm pretty sure I wouldn't have to think about numbers being too large for Javascript to handle.

sandy2008:
> I think we should have a global/default flag we can set too. Tanstack Router v2 can make that the default. In v1 it will of course have to be opt-in.

@devinrhode2 

Great suggestion! I've added a `defaultRawSearchInput` option to `RouterOptions`. When set to `true`, **all** route validators will receive raw URL search strings by default — no need to wrap each one with `validateSearchWithRawInput`.

```ts
const router = createRouter({
  routeTree,
  defaultRawSearchInput: true, // all validators now get raw strings
})
```

Individual routes can still use `validateSearchWithRawInput()` for fine-grained control, and the two approaches compose without conflict.

This is fully opt-in for v1. Agreed that v2 could flip the default to `true`.

sandy2008:
> BONUS: for a separate PR: Regarding detecting and correcting bad urls. There could be utilities for detecting it and correcting it, but these are just utilities - users need to actually invoke them on their own. Some may have extra long numbers, some may not, who knows what kind of url behavior is going on out there... - better to not attempt to fix this for everyone and let them handle it themselves on a case by case basis. Then again, I can say for my app, I'm pretty sure I wouldn't have to think about numbers being too large for Javascript to handle.

@devinrhode2 I will make the PR once this get merged!

devinrhode2:
Tangent: if a search param is numeric, but too large for Javascript to handle, what does router do?

IMO it should crash/throw.

It would be neat if there was built in support for BigInt.

devinrhode2:
Looking back at this, the names could be just "validateSearchV2" or something

devinrhode2:
Is it possible to consolidate this PR with other similar ones like #6044?

PR #6955, opened 2026-03-17
feat(vite-plugin-core): Spport both Vite 7 (`rollupOptions`) and Vite 8 (`rolldownOptions`)

birkskyum:
By runtime checking if rollup or rolldown is used, the plugin can now read/write to rollupOptions or rolldownOptions, thus support both vite 7 and 8

## Summary by CodeRabbit

* **New Features**
  * Added automatic Vite version detection to support both Vite 7 and Vite 8, ensuring seamless compatibility with your installed version without requiring manual configuration.

### DIFF
--- a/packages/start-plugin-core/src/utils.ts
+++ b/packages/start-plugin-core/src/utils.ts
@@ -7,9 +7,9 @@ import * as vite from 'vite'
 export const isRolldown = 'rolldownVersion' in vite
 
 /** Returns `'rolldownOptions'` when using Rolldown, `'rollupOptions'` otherwise. */
-export const bundlerOptionsKey = (
-  isRolldown ? 'rolldownOptions' : 'rollupOptions'
-)
+export const bundlerOptionsKey = isRolldown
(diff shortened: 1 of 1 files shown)


Sheraff:
I think this PR might have broken the SSR benchmarks. They are marked as skipped here
> ⏩ 3 skipped benchmarks

Looking at the commit history, this PR's merge commit seems to be the 1st with skipped benchmarks.

They seem to fail during the build
```
> vite build --config ./solid/vite.config.ts
vite v7.3.1 building client environment for production...
✓ 0 modules transformed.
✗ Build failed in 91ms
error during build:
Could not resolve entry module "solid/index.html".
```

Maybe an issue with the rolldown detection? GPT tracked it down to this https://github.com/TanStack/router/pull/6984. ~~The fix is ugly, but we can see if it does restore the benchmarks.~~ Well the fix doesn't work, but it still seems to be coming from this PR...

BTW: if it fails on the benchmarks, there is a chance it fails in real apps too, right?

PR #6956, opened 2026-03-17
Docs: Update search parameters guide for zod v4

### LINKED ISSUE #4322: Zod 4 support for `@tanstack/zod-adapter` (specifically `fallback`)

robballou:
Zod v4 does not require use of the adapter and can use the schemas directly. See https://github.com/TanStack/router/issues/4322

closes #4322

## Summary by CodeRabbit

* **Documentation**
  * Clarified Zod v3 vs Zod v4 guidance across search-params docs, specifying when to use the adapter versus direct Zod schemas.
  * Updated all examples and guides to prefer direct-schema validateSearch for Zod v4 (removed adapter wrappers in examples), adjusted default/.catch vs fallback recommendations, and refreshed step-by-step, inheritance, Link, and catch/fallback guidance to match the new patterns.

### DIFF
--- a/packages/router-core/skills/router-core/search-params/SKILL.md
+++ b/packages/router-core/skills/router-core/search-params/SKILL.md
@@ -38,11 +38,11 @@ import { createFileRoute } from '@tanstack/react-router'
 import { z } from 'zod'
 
 const productSearchSchema = z.object({
-  page: fallback(z.number(), 1).default(1),
-  filter: fallback(z.string(), '').default(''),
-  sort: fallback(z.enum(['newest', 'oldest', 'price']), 'newest').default(
+  page: z.number().default(1).catch(1),
+  filter: z.string().default(''),
+  sort: z.enum(['newest', 'oldest', 'price']).default(
     'newest',
-  ),
+  ).catch('newest'),
 })
 
 export const Route = createFileRoute('/products')({


robballou:
Additional question, there are 12 other files which mention the `zodAdapter`: should those also be updated?

birkskyum:
> Additional question, there are 12 other files which mention the zodAdapter: should those also be updated?

@robballou  Yes, that would be good to get everything modernized.

robballou:
I have updated more of the documentation while hopefully retaining the value. I am trying to give both v3 and v4 examples where it seems appropriate. In a few places, I replaced the instructions with Zod v4 only as it seems new projects would use v4.

Notes:

- Some of these changes are within LLM skills it seems like, but I have not validated that those work. 
- I left the zodAdapter use for Start's `inputValidation` because I am unsure if how that works. Since the `@tanstack/zod-adapter` only works with Zod v3, I assume something needs to be updated in the documentation.
- I did try some variations in of using the schema directly in a local project to see that they worked as expected, but not all examples were tested by me 😆

birkskyum:
@robballou thanks for helping out getting this updated!

PR #6976, opened 2026-03-19
docs: update the scratch guide to be compatible with Vite 8^

zougari47:
Vite 8 is out. and by following the **build from scratch** guide and runing the command `npm i vite` it install v 8^.

Vite 8 introduce this option `resolve.tsconfigPaths` and so no need for the plugin `vite-tsconfig-paths` anymore.

I skipped two files intentionlly containing same snippet `migrate-from-next-js.md` and `tailwindd-integration.md` because I don't know if it compatible with it yet. 

Let me know if I should also update those files.

## Summary by CodeRabbit

* **Documentation**
  * Updated React and Solid framework build-from-scratch guides with revised Vite build configuration instructions.
  * Removed vite-tsconfig-paths plugin dependency from installation steps.
  * Documentation now demonstrates path alias resolution using Vite's native configuration option in place of external plugin.
  * Changes reflected across both React and Solid framework setup guides.

### DIFF
--- a/docs/start/framework/react/build-from-scratch.md
+++ b/docs/start/framework/react/build-from-scratch.md
@@ -80,7 +80,7 @@ Alternatively, you can also use `@vitejs/plugin-react-swc`.
 and some TypeScript:
 
 ```shell
-npm i -D typescript @types/react @types/react-dom @types/node vite-tsconfig-paths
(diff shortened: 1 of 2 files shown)


birkskyum (approved):
thanks! that should be all the modification that's needed.

PR #6994, opened 2026-03-20
fix(start-plugin-core): use duck-type check for RunnableDevEnvironment to avoid dual-package hazard

### LINKED ISSUE #6982: start-plugin-core: instanceof RunnableDevEnvironment fails with vite-plus alias (dual-package hazard)
## Problem

`@tanstack/start-plugin-core`'s dev server plugin uses `isRunnableDevEnvironment(serverEnv)` (which calls `instanceof`) to guard SSR middleware registration at `src/dev-server-plugin/plugin.ts` line ~147.

When `vite` is aliased to another package via pnpm overrides (e.g. `vite: npm:@voidzero-dev/vite-plus-core@latest`), two different `RunnableDevEnvironment` class references exist in memory — `instanceof` returns `false`, SSR middleware is never registered, and all routes return 404.

This is a [dual-package hazard](https://nodejs.org/api/packages.html#dual-package-hazard).

## Reproduction

1. Create a TanStack Start project
2. Add pnpm override: `"vite": "npm:@voidzero-dev/vite-plus-core@latest"`
3. Run `pnpm dev`
4. All routes return 404 "Cannot GET /"

## Suggested Fix

Replace `instanceof` with duck-typing in `plugin.ts`:

```diff
- if (!isRunnableDevEnvironment(serverEnv) || "dispatchFetch" in serverEnv) {
+ if (!("runner" in serverEnv) || "dispatchFetch" in serverEnv) {
```

The `runner` property (a public getter) is unique to `RunnableDevEnvironment` — `FetchableDevEnvironment` and `DevEnvironment` don't have it. The existing `"dispatchFetch" in serverEnv` guard correctly handles Nitro.

## Versions

- `@tanstack/start-plugin-core`: 1.166.8
- `vite-plus` / `@voidzero-dev/vite-plus-core`: 0.1.12 (Vite 8.0.0)

## Workaround

`pnpm patch @tanstack/start-plugin-core` with the diff above.

sleitor:
## Summary

Fixes #6982

When  is aliased to another package via pnpm overrides (e.g. `vite: npm:@voidzero-dev/vite-plus-core@latest`), two different `RunnableDevEnvironment` class references exist in memory. The `isRunnableDevEnvironment()` helper uses `instanceof` which returns `false`, causing SSR middleware to never be registered and all routes returning 404.

## Changes

Replace `isRunnableDevEnvironment(serverEnv)` (instanceof-based) with a duck-type check `'runner' in serverEnv` in `packages/start-plugin-core/src/dev-server-plugin/plugin.ts`.

The `runner` property (a public getter) is unique to `RunnableDevEnvironment` — `FetchableDevEnvironment` and `DevEnvironment` don't have it. The existing `'dispatchFetch' in serverEnv` guard is preserved to correctly handle Nitro's `FetchableDevEnvironment`.

`RunnableDevEnvironment` is still imported as a **type** only (for the TypeScript cast on `serverEnv.runner.import()`), so no runtime instanceof check remains.

## Testing

- ESLint passes
- Type errors in test:types are pre-existing (missing sibling package builds in standalone test run)

## Bug Fixes
* Fixed a compatibility issue when multiple versions or alternative distributions of the build tool were loaded in the same project, ensuring the plugin works reliably in complex dependency environments.

### DIFF
--- a/.changeset/fix-6982-duck-type-runnable-dev-env.md
+++ b/.changeset/fix-6982-duck-type-runnable-dev-env.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/start-plugin-core": patch
+---
+
+fix(start-plugin-core): replace isRunnableDevEnvironment instanceof check with duck-typing to avoid dual-package hazard when vite is aliased (e.g. vite-plus pnpm override)
--- a/.changeset/fix-6982-duck-type-runnable-dev-env.md
+++ b/.changeset/fix-6982-duck-type-runnable-dev-env.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/start-plugin-core": patch
+'@tanstack/start-plugin-core': patch
 ---
 
 fix(start-plugin-core): replace isRunnableDevEnvironment instanceof check with duck-typing to avoid dual-package hazard when vite is aliased (e.g. vite-plus pnpm override)
--- a/packages/start-plugin-core/src/dev-server-plugin/plugin.ts
+++ b/packages/start-plugin-core/src/dev-server-plugin/plugin.ts
@@ -8,7 +8,12 @@ import {
   collectDevStyles,
   normalizeCssModuleCacheKey,
 } from './dev-styles'
-import type { Connect, DevEnvironment, PluginOption, RunnableDevEnvironment } from 'vite'
+import type {
+  Connect,
+  DevEnvironment,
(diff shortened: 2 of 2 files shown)


schiller-manuel:
this is not necessary unless proven otherwise.

ultra2207:
bruh....

PR #7001, opened 2026-03-21
chore: sync main branch with store refactor

brenelz:


packages/solid-router/src/routerStores.ts
```diff
@@ -0,0 +1,124 @@
...
+      // onSettled, createTrackedEffect), flush() is not reentrant and
+      // will throw, so we catch and ignore that case since Solid will
+      // flush automatically when the effect completes.
+      try {
+        Solid.flush()
+      } catch {
+        // flush() throws when called from inside onSettled or
+        // createTrackedEffect — Solid will flush automatically there.
+      }
+    },
```
Sheraff, line 116: Should we track depth to avoid flushing too much?
```ts
let depth = 0
batch: (fn) => {
  depth++
  fn()
  depth--
  if (depth === 0) {
    try {
      Solid.flush()
    } catch {}
  }
}
```

Also, just so I understand: memos keep carrying the old value until we flush?

Like if we do this:
```ts
const [foo, setFoo] = createSignal(1)
const double = createMemo(() => foo() * 2)
setFoo(2)
console.log(double())
```

This logs `2` and not `4`?

If this is the case, then we might have other issues in router-core. I've tried using `batch()` every time we are setting **multiple** stores in the same tick, but I've never used it for a **single** setter. There might be places in router-core where we try to read a derived value just after setting it synchronously.

packages/solid-router/src/Transitioner.tsx
```diff
@@ -92,49 +110,71 @@ export function Transitioner() {
...
   })
 
-  Solid.createEffect(
-    () => [previousIsLoading(), isLoading()] as const,
-    ([previousIsLoading, isLoading]) => {
-      if (previousIsLoading.previous && !isLoading) {
+  const locationState = Solid.createMemo(() => router.stores.location.state)
+  const resolvedLocationState = Solid.createMemo(
+    () => router.stores.resolvedLocation.state,
+  )
```
Sheraff: We probably don't need to wrap those in `createMemo`, they are signals already. 

Reading
```ts
router.stores.resolvedLocation.state
```

Is like reading the signal in
```ts
const [sig, setSig] = createSignal()
sig()
``` 

Same thing for `hasPendingMatches` and `isLoading` above

Sheraff:
Should we worry about the bundle size increase (router +3.93 KiB, start +9.53 KiB)? Or is it expected moving to solid v2?

(BTW, bundle-size deltas are always compared against `main`, whereas performance benchmarks are compared against the base branch of the PR, in this case `solid-router-v2-pre`, so it's a little confusing!)

packages/solid-router/src/Transitioner.tsx
brenelz: ok will make this update

packages/solid-router/src/routerStores.ts
brenelz: yeah that is my understanding

brenelz:
Im a little concerned about bundle size. I expect a little bump from solid 2 but this is a decent amount

packages/solid-router/src/routerStores.ts
brenelz: Ok added the depth check now

PR #7003, opened 2026-03-21
fix(router-core): avoid preload cleanup crashes after invalidation

Sheraff:
## Summary
- harden `loadRouteMatch` so preload cleanup does not crash when an in-flight match has already been evicted from cache
- add regression coverage for both direct cache GC and `router.invalidate()` clearing an in-flight preload
- keep the current behavior where expired preloads may be removed, while preventing the `_nonReactive` console error reported after the recent preload changes

## Testing
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/router-core:test:unit --outputStyle=stream --skipRemoteCache -- tests/load.test.ts`
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/router-core:test:types --outputStyle=stream --skipRemoteCache`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved route loader cleanup so background preloads clean up reliably and tolerate missing or evicted matches without errors.
  * Enhanced behavior during cache eviction/invalidation to avoid spurious errors when preloads are still in flight.

* **Tests**
  * Added concurrency tests covering cache eviction and invalidation interacting with in‑flight preloads.

* **Chores**
  * Added a changeset to publish a patch release.

### DIFF
--- a/.changeset/moody-kings-travel.md
+++ b/.changeset/moody-kings-travel.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
+evicting cache during preload does not error, return early
--- a/packages/router-core/src/load-matches.ts
+++ b/packages/router-core/src/load-matches.ts
@@ -834,18 +834,19 @@ const loadRouteMatch = async (
       shouldReloadInBackground
     ) {
       loaderIsRunningAsync = true
+      const matchForCleanup = prevMatch
       ;(async () => {
         try {
           await runLoader(inner, matchPromises, matchId, index, route)
-          const match = inner.router.getMatch(matchId)!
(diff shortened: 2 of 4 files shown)


Sheraff:
fixed in https://github.com/TanStack/router/pull/7805

sommeeeer:
Thanks @Sheraff ! When will it be in a release?

Sheraff:
> Thanks @Sheraff ! When will it be in a release?

very soon, a couple hours maybe

sommeeeer:
> > Thanks @Sheraff ! When will it be in a release?
> 
> very soon, a couple hours maybe

Alright perfect. Thanks :)

PR #7006, opened 2026-03-21
fix(router-core): avoid preload cleanup crashes after invalidation 2

Sheraff:
Alternative solution to https://github.com/TanStack/router/pull/7003

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved preload cache eviction errors by improving early return logic
  * Enhanced robustness when cache operations occur during active preloads
  * Improved match availability checks throughout loader execution

* **Tests**
  * Added tests verifying preload stability during cache operations like clearExpiredCache and invalidate

### DIFF
--- a/.changeset/moody-kings-travel.md
+++ b/.changeset/moody-kings-travel.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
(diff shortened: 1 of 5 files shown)


Sheraff:
fixed in https://github.com/TanStack/router/pull/7805

PR #7025, opened 2026-03-23
docs: update migrate from next docs to working example

banjo:
Current example is a bit out of date with the new Vite 8 release, and it is not fully working because it is missing the `nitro` dependency.

- Added `nitro` as a dependency and included it in the Vite configuration. This is required to run the application after build. Without it it will just do a silent 0 exit without any output.
- Removed the `vite-tsconfig-paths` dependency in favor of path alias support introduced in Vite v8.

## Summary by CodeRabbit

* **Documentation**
  * Updated React migration guide with a streamlined Vite setup and clearer example configuration.
  * Replaced an external path-alias plugin with Vite’s built-in tsconfig path resolution and simplified related Tailwind dev dependency instructions.
  * Added guidance and an example for integrating a Vite-compatible deployment plugin into the project setup.

### DIFF
--- a/docs/start/framework/react/migrate-from-next-js.md
+++ b/docs/start/framework/react/migrate-from-next-js.md
@@ -89,15 +89,15 @@ import { defineConfig } from 'vite'
 import { tanstackStart } from '@tanstack/react-start/plugin/vite'
 import viteReact from '@vitejs/plugin-react'
 import tailwindcss from '@tailwindcss/vite'
-import { nitro } from "nitro/vite";
+import { nitro } from 'nitro/vite'
 
 export default defineConfig({
   server: {
     port: 3000,
   },
   resolve: {
-      // Enables Vite to resolve imports using path aliases.
-      tsconfigPaths: true,
+    // Enables Vite to resolve imports using path aliases.
+    tsconfigPaths: true,
   },
   plugins: [
     tailwindcss(),


docs/start/framework/react/migrate-from-next-js.md
```diff
@@ -52,16 +52,16 @@ rm postcss.config.* next.config.*
 
 ### 2. Install Required Dependencies
 
-TanStack Start leverages [Vite](https://vite.dev) and TanStack Router:
+TanStack Start leverages [Vite](https://vite.dev), [nitro](https://nitro.build/) and TanStack Router:
```
birkskyum: nitro is one of multiple options for deployment vite plugins, so nitro can't be assumed. Alternatives count e.g. [cloudflare plugin](https://www.npmjs.com/package/@cloudflare/vite-plugin) and [netlify plugin](https://www.npmjs.com/package/@netlify/vite-plugin)

banjo: Yeah that is true.

Even though you could take any provider the nitro provider is used for a normal node sever, which is the most simple use case in my opinion. It is also used in the basic example of Tanstack Start in the docs.

The guide also already includes a start script that expects  nitro to be installed, so if you follow the guide as it is in its current form you'll fail on the start script.

Maybe it is a good start for people migrating over? I was a bit confused myself when migrating and had to look at a the basic example to find what was missing. 

One maybe could mention more about hosting providers in the same section.

birkskyum: I think you're right that it's a good starting point.

Basically, the only small modification I'd like to see is a wording that emphasize nitro is an option, like: "start leverages tanstack router, vite, and vite deployment plugins e.g. nitro"

banjo: Yeah that is much better, thanks for the feedback. 

I updated it like your suggestion.

PR #7042, opened 2026-03-25
fix: scroll restorating without throttling

### LINKED ISSUE #7040: Scroll restoration `onScroll` throttle saves scroll positions under wrong route key during SPA navigation

Router

### Describe the bug

We ran into a tricky scroll bug in our TanStack Start app. When users navigate between pages via SPA navigation, the destination page sometimes doesn't scroll to top. Instead it gets stuck partway down, even though `resetScroll: true` is set.

After a lot of debugging, we traced it to a race condition in the `onScroll` handler inside `setupScrollRestoration` (`scroll-restoration.ts`).

The handler reads `router.state.location` inside the throttled callback, which runs up to 100ms after the actual scroll event fired. During that 100ms window, `router.state.location` may have already changed to the new route if the user navigated. So the old page's scroll position gets saved under the new page's cache key in sessionStorage.

When `restoreScroll()` runs on the new page, it finds a cached entry with a non-zero `scrollY` (belonging to the old page) and restores to that position instead of scrolling to top.

Here's the relevant code:

```typescript
// scroll-restoration.ts, setupScrollRestoration()
const onScroll = (event: Event) => {
  // ...
  const restoreKey = getKey(router.stores.location.state) // ← read at EXECUTION time, not event time
  scrollRestorationCache.set((state) => {
    state[restoreKey].window = { scrollY: window.scrollY }
    // ...
  })
}
```
The trailing throttle captures args from the first call but `restoreKey` is not part of those args. It's read from `router.state.location` when the throttle finally executes.

## Suggested Fix

We applied two changes to `setupScrollRestoration()` in `scroll-restoration.ts` (well, the compiled `scroll-restoration.js`). Here is the full diff of our working patch:

### Fix 1: Capture `restoreKey` at scroll event time, not at throttle execution time

We split the original `onScroll` handler into two parts. The raw scroll event listener resolves the `elementSelector` and captures the `restoreKey` immediately. Then it passes both as arguments to a throttled `saveScrollPosition` function. Since the throttle preserves args from the first call in each 100ms window, the key always reflects the route that was active when the scroll actually happened.

```diff
-  const onScroll = (event) => {
-    if (ignoreScroll || !router.isScrollRestoring) {
-      return;
-    }
-    let elementSelector = "";
-    if (event.target === document || event.target === window) {
-      elementSelector = "window";
-    } else {
-      const attrId = event.target.getAttribute(
-        "data-scroll-restoration-id"
-      );
-      if (attrId) {
-        elementSelector = `[data-scroll-restoration-id="${attrId}"]`;
-      } else {
-        elementSelector = getCssSelector(event.target);
-      }
-    }
-    const restoreKey = getKey(router.state.location);
+  const saveScrollPosition = (restoreKey, elementSelector) => {
     scrollRestorationCache.set((state) => {
       const keyEntry = state[restoreKey] ||= {};
       const elementEntry = keyEntry[elementSelector] ||= {};
@@ (unchanged scroll position save logic)
       return state;
     });
   };
+  const throttledSave = throttle(saveScrollPosition, 100);
   if (typeof document !== "undefined") {
-    document.addEventListener("scroll", throttle(onScroll, 100), true);
+    document.addEventListener("scroll", (event) => {
+      if (ignoreScroll || !router.isScrollRestoring) {
+        return;
+      }
+      let elementSelector = "";
+      if (event.target === document || event.target === window) {
+        elementSelector = "window";
+      } else {
+        const attrId = event.target.getAttribute(
+          "data-scroll-restoration-id"
+        );
+        if (attrId) {
+          elementSelector = `[data-scroll-restoration-id="${attrId}"]`;
+        } else {
+          elementSelector = getCssSelector(event.target);
+        }
+      }
+      // Capture the restore key at scroll event time, not when the throttled
+      // handler executes. The throttle preserves args from the first call in
+      // each window, preventing stale keys from post-navigation state changes.
+      const restoreKey = getKey(router.state.location);
+      throttledSave(restoreKey, elementSelector);
+    }, true);
   }
```

### Fix 2: Clear stale cache entries before `restoreScroll()` in `onRendered`

This handles the second vector: browser-generated scroll events during DOM transitions that fire after `router.state.location` already changed. We clear any cached entries for the new route's key right before `restoreScroll()` reads from sessionStorage. We use the `scrollRestorationCache.set()` API which handles both the in-memory state and sessionStorage atomically.

```diff
   router.subscribe("onRendered", (event) => {
     const cacheKey = getKey(event.toLocation);
     // ...existing resetNextScroll and scrollRestoration checks...

+    // Clear stale scroll entries for the new key before restoring.
+    // During DOM transitions the browser fires scroll events after
+    // router.state.location already changed, polluting the new key's cache.
+    if (router.resetNextScroll && scrollRestorationCache) {
+      scrollRestorationCache.set((state) => {
+        delete state[cacheKey];
+        return state;
+      });
+    }
     restoreScroll({
       storageKey,
       key: cacheKey,
```

This way `restoreScroll` finds no cached entries for a new forward navigation and falls through to `scrollTo(0, 0)` as expected.

I can share in private on Discord (#Bluebeel) if needed. The website is not yet live.

1. Create a TanStack Router app with `scrollRestoration: true`
2. Have a home page with enough content to scroll to 6000px or more (multiple sections, sliders, etc.)
3. Have a destination page that is much shorter (3000-4000px)
4. Scroll down on the home page using the mouse wheel (this generates continuous scroll events)
5. **Immediately** click a `<Link>` or call `router.navigate({ resetScroll: true })`
6. The destination page is **not** scrolled to top. It's stuck at `newDocHeight - viewportHeight`

**Why it's intermittent:** The bug only shows up when the user's last scroll event falls within the 100ms throttle window before navigation. If the user scrolls, pauses for more than 100ms, then clicks, the throttle has already fired and saved under the correct key. But fast scroll-then-click hits the race condition.

**Proof from sessionStorage (`tsr-scroll-restoration-v1_3`):**

We dumped the full sessionStorage key from a real browsing session where we navigated to several games and could reproduce the bug. Here's what the polluted entries look like. These keys belong to game pages, but they contain scroll positions from the home page:

```json
"zwayg": { "window": { "scrollX": 0, "scrollY": 2975 }, ... }
"gfcvo": { "window": { "scrollX": 0, "scrollY": 2710 }, ... }
"g6rvv": { "window": { "scrollX": 0, "scrollY": 2253 }, ... }
"vr4bd": { "window": { "scrollX": 0, "scrollY": 2710 }, ... }
```

For comparison, the home page keys have the correct (high) scroll positions:

```json
"tv3zv": { "window": { "scrollX": 0, "scrollY": 6634 }, ... }
"gbcme": { "window": { "scrollX": 0, "scrollY": 6611 }, ... }
"mfqzi": { "window": { "scrollX": 0, "scrollY": 6660 }, ... }
"7b5b6": { "window": { "scrollX": 0, "scrollY": 6735 }, ... }
```

The game page keys shouldn't have a `window.scrollY` of 2710 or 2975. Those are clamped home page scroll positions that were saved under the wrong key because of the throttle race.

We also confirmed this with a Playwright test. Here's the clearest example:

```
TSR key:        v0u6x (game page)
Pre-click:      scrollY = 8961 (on home page)
Post-navigation: scrollY = 2691 (stuck on game page)
Cache for v0u6x: { "window": { "scrollX": 0, "scrollY": 8961 } }
```

`scrollY: 8961` is the home page position saved under the game page key `v0u6x`. `restoreScroll()` finds it and restores to 8961. The browser clamps it to `3771 - 1080 = 2691`. The user sees the footer instead of the game.

After applying our fix, the same test shows clean data:

```
TSR key:        2heq9 (game page)
Pre-click:      scrollY = 6083 (on home page)
Post-navigation: scrollY = 0
Cache for 2heq9: { "window": { "scrollX": 0, "scrollY": 0 } }
```

No more pollution. The game page key has `scrollY: 0` and the page scrolls to top every time.

### Expected behavior

When navigating with `resetScroll: true`, the destination page should always scroll to top (`scrollY: 0`), no matter how recently the user was scrolling on the previous page.

### Screenshots or Videos

_No response_

### Platform

- Router Version: 1.166.7
- OS: macOS (also reproduced on Windows/Linux via Playwright)
- Browser: Chrome 134, also reproducible in Firefox and Safari
- Bundler: Vite 8

### Additional context

_No response_

schiller-manuel:
fixes #7040

## Summary by CodeRabbit

* **Bug Fixes**
  * Fix scroll restoration so navigation after fast scrolling reliably restores the correct position (removes throttling regression).

* **New Features**
  * Added source and target test pages to reproduce fast-scroll navigation scenarios.

* **Tests**
  * Added an end-to-end reproduction test asserting scroll-restoration behavior across rapid-scroll navigations.

* **Chores**
  * Added changeset entries to prepare patch releases for router packages.

### DIFF
--- a/packages/router-core/src/scroll-restoration.ts
+++ b/packages/router-core/src/scroll-restoration.ts
@@ -1,6 +1,5 @@
 import { isServer } from '@tanstack/router-core/isServer'
-import minifiedScrollRestorationScript from './scroll-restoration-inline?script-string'
-import { escapeHtml, functionalUpdate, isPlainObject } from './utils'
+import { functionalUpdate, isPlainObject } from './utils'
 import type { AnyRouter } from './router'
 import type { ParsedLocation } from './location'
 import type { NonNullableUpdater } from './utils'
@@ -20,13 +19,6 @@ type ScrollRestorationCache = {
 export type ScrollRestorationOptions = {
   getKey?: (location: ParsedLocation) => string
   scrollBehavior?: ScrollToOptions['behavior']
-}
-
-type InlineScrollRestorationScriptOptions = {
-  storageKey: string
-  key?: string
-  behavior?: ScrollToOptions['behavior']
-  shouldScrollRestoration?: boolean
 }
 
 function getSafeSessionStorage() {
@@ -85,56 +77,6 @@ function createScrollRestorationCache(): ScrollRestorationCache | null {
 }
 
 export const scrollRestorationCache = createScrollRestorationCache()
-
-const defaultInlineScrollRestorationScript = `(${minifiedScrollRestorationScript})(${escapeHtml(
-  JSON.stringify({
-    storageKey,
-    shouldScrollRestoration: true,
-  } satisfies InlineScrollRestorationScriptOptions),
-)})`
-
-export function getScrollRestorationScript(
-  options: InlineScrollRestorationScriptOptions,
-) {
-  if (
-    options.storageKey === storageKey &&
-    options.shouldScrollRestoration === true &&
-    options.key === undefined &&
-    options.behavior === undefined
-  ) {
-    return defaultInlineScrollRestorationScript
-  }
-
-  return `(${minifiedScrollRestorationScript})(${escapeHtml(JSON.stringify(options))})`
-}
-
-export function getScrollRestorationScriptForRouter(router: AnyRouter) {
-  if (
-    typeof router.options.scrollRestoration === 'function' &&
-    !router.options.scrollRestoration({ location: router.latestLocation })
-  ) {
-    return null
-  }
-
-  const getKey = router.options.getScrollRestorationKey
-  if (!getKey) {
-    return defaultInlineScrollRestorationScript
-  }
-
-  const location = router.latestLocation
-  const userKey = getKey(location)
-  const defaultKey = defaultGetScrollRestorationKey(location)
-
-  if (userKey === defaultKey) {
-    return defaultInlineScrollRestorationScript
-  }
-
-  return getScrollRestorationScript({
-    storageKey,
-    shouldScrollRestoration: true,
-    key: userKey,
-  })
-}
 
 /**
  * The default `getKey` function for `useScrollRestoration`.
--- a/packages/react-router/src/scroll-restoration.tsx
+++ b/packages/react-router/src/scroll-restoration.tsx
@@ -1,4 +1,4 @@
-import { getScrollRestorationScriptForRouter } from '@tanstack/router-core'
+import { getScrollRestorationScriptForRouter } from '@tanstack/router-core/scroll-restoration-script'
 import { useRouter } from './useRouter'
 import { ScriptOnce } from './ScriptOnce'
 
--- a/packages/router-core/package.json
+++ b/packages/router-core/package.json
@@ -148,6 +148,26 @@
         "default": "./dist/cjs/isServer/client.cjs"
       }
     },
+    "./scroll-restoration-script": {
+      "browser": {
+        "import": {
+          "types": "./dist/esm/scroll-restoration-script/client.d.ts",
+          "default": "./dist/esm/scroll-restoration-script/client.js"
+        },
+        "require": {
+          "types": "./dist/cjs/scroll-restoration-script/client.d.cts",
+          "default": "./dist/cjs/scroll-restoration-script/client.cjs"
+        }
+      },
+      "import": {
+        "types": "./dist/esm/scroll-restoration-script/server.d.ts",
+        "default": "./dist/esm/scroll-restoration-script/server.js"
+      },
+      "require": {
+        "types": "./dist/cjs/scroll-restoration-script/server.d.cts",
+        "default": "./dist/cjs/scroll-restoration-script/server.cjs"
+      }
+    },
     "./package.json": "./package.json"
   },
   "sideEffects": false,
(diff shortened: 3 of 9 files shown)


packages/router-core/src/scroll-restoration.ts
```diff
@@ -237,114 +365,79 @@ export function setupScrollRestoration(router: AnyRouter, force?: boolean) {
...
+      {} as ScrollRestorationEntry)
+    nextElementEntry.scrollX = scrollX
+    nextElementEntry.scrollY = scrollY
   }
 
-  // Throttle the scroll event to avoid excessive updates
-  if (typeof document !== 'undefined') {
-    document.addEventListener('scroll', throttle(onScroll, 100), true)
-  }
+  document.addEventListener('scroll', onScroll, true)
```
Sheraff, line 226: I'm not sure we need to run on the `capture` phase here.
```suggestion
  document.addEventListener('scroll', onScroll)
```

packages/router-core/src/scroll-restoration.ts
```diff
@@ -57,47 +76,86 @@ function createScrollRestorationCache(): ScrollRestorationCache | null {
...
-    : {}
+  let state: ScrollRestorationByKey = readScrollRestorationState()
+
+  const persist = () => {
+    try {
+      safeSessionStorage.setItem(storageKey, JSON.stringify(state))
+    } catch {
+      console.warn(
+        '[ts-router] Could not persist scroll restoration state to sessionStorage.',
+      )
```
Sheraff: Could we make this dev only to reduce bundle size?

schiller-manuel: done

PR #7051, opened 2026-03-26
fix: preserve suspense promises for pending route matches

Sheraff:
## Summary
- keep a route's suspense promise alive while the match still renders as `pending`
- add a regular navigation regression test where aborting a pending route must keep rendering the pending UI
- add an `invalidate({ forcePending: true })` regression test that keeps rendering the suspense fallback instead of throwing `undefined`

## Testing
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/react-router:test:unit --outputStyle=stream --skipRemoteCache`
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/react-router:test:types --outputStyle=stream --skipRemoteCache`
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/react-router:test:eslint --outputStyle=stream --skipRemoteCache`

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented spurious error rendering during route loads and navigations; pending UI is now preserved when in-flight loaders are aborted or when reloads are forced into a pending state.

* **Tests**
  * Added regression tests covering aborted-loader navigation and forced-pending reloads to ensure pending UI remains visible and error components do not render incorrectly.

### DIFF
--- a/packages/router-core/src/load-matches.ts
+++ b/packages/router-core/src/load-matches.ts
@@ -839,7 +839,9 @@ const loadRouteMatch = async (
           await runLoader(inner, matchPromises, matchId, index, route)
           const match = inner.router.getMatch(matchId)!
           match._nonReactive.loaderPromise?.resolve()
-          match._nonReactive.loadPromise?.resolve()
+          if (match.status !== 'pending') {
+            match._nonReactive.loadPromise?.resolve()
+          }
           match._nonReactive.loaderPromise = undefined
         } catch (err) {
           if (isRedirect(err)) {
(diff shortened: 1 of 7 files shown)


packages/router-core/src/load-matches.ts
```diff
@@ -839,9 +839,10 @@ const loadRouteMatch = async (
           await runLoader(inner, matchPromises, matchId, index, route)
           const match = inner.router.getMatch(matchId)!
           match._nonReactive.loaderPromise?.resolve()
-          match._nonReactive.loadPromise?.resolve()
+          if (match.status !== 'pending') {
```
schiller-manuel: why? is another resolve() problematic?

PR #7053, opened 2026-03-26
optimize bundle-size & perf

schiller-manuel:
## Summary by CodeRabbit

* **Refactor**
  * Improved navigation lifecycle dispatch to better distinguish leaving, staying, and entering routes; streamlined error-state detection and route-matching logic.

* **New Features**
  * Added a public check to detect error states among active routes.

* **Tests**
  * Added a test ensuring same-route navigations with changed loader params are treated as a "stay" (no leave/enter).

### DIFF
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -2475,8 +2475,10 @@ export class RouterCore<
 
                   // Lifecycle-hook identity uses routeId only so that navigating between
                   // different params/deps of the same route fires onStay (not onLeave+onEnter).
-                  const nextPendingRouteIds = pendingRouteIds as Set<string> | null
-                  const nextActiveRouteIds = activeRouteIds as Set<string> | null
+                  const nextPendingRouteIds =
+                    pendingRouteIds as Set<string> | null
+                  const nextActiveRouteIds =
+                    activeRouteIds as Set<string> | null
 
                   if (nextPendingRouteIds && nextActiveRouteIds) {
                     for (const match of currentMatches) {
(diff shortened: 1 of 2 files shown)


packages/router-core/src/router.ts
```diff
@@ -2418,54 +2418,42 @@ export class RouterCore<
...
+                          match.status !== 'error' &&
+                          match.status !== 'notFound' &&
+                          match.status !== 'redirected'
+                        ) {
+                          exitingMatches.push(match)
+                        }
+                      }
+                    } else {
+                      exitingMatches = null
                     }
```
Sheraff, line 2455: `else` not necessary, `exitingMatches` is already `null`
```suggestion
                    }
```

packages/router-core/src/router.ts
```diff
@@ -2474,31 +2462,47 @@ export class RouterCore<
...
-                  ] as const) {
-                    if (!matches) continue
-                    for (const match of matches as Array<AnyRouteMatch>) {
-                      this.looseRoutesById[match.routeId]!.options[hook]?.(
+                  // Lifecycle-hook identity uses routeId only so that navigating between
+                  // different params/deps of the same route fires onStay (not onLeave+onEnter).
+                  const nextPendingRouteIds =
+                    pendingRouteIds as Set<string> | null
+                  const nextActiveRouteIds =
+                    activeRouteIds as Set<string> | null
```
Sheraff, line 2481: This doesn't seem necessary, `as` casts to the same type those variables already have

Issue #6947, opened 2026-03-17
Start: SSR dehydrated script output contains unnecessary whitespace/newlines


canmi21:
### Which project does this relate to?

Start

### Describe the bug

The inline <script> blocks generated by TanStack Start for SSR dehydration contain unnecessary blank lines and whitespace. Since this output is machine-generated serialization (Seroval streams, router state, etc.) and not user content, it could be minified without any risk of breaking functionality.

### Your Example Website or App

https://github.com/canmi21/taki

### Steps to Reproduce the Bug or Issue

View the HTML source of any SSR-rendered TanStack Start page. The dehydrated <script> tags contain patterns like:
 ```js
<script>($R=>{ ... })($R["tsr"]);    
  document.currentScript.remove()</script>
```

<img width="2082" height="1136" alt="Image" src="https://github.com/user-attachments/assets/1ae1a171-f3b3-4a54-8fda-d3586bf3eb47" />

There are blank lines between statements and unnecessary newlines within the Seroval stream constructor output.

### Expected behavior

Dehydrated script output should be compact, e.g：
 
```js
 <script>($R=>{ ... })($R["tsr"]);document.currentScript.remove()</script>
```

### Screenshots or Videos

<img width="2082" height="1136" alt="Image" src="https://github.com/user-attachments/assets/7a589b86-c72a-4a3d-8d04-bf6774b52a13" />

### Platform

- Router / Start Version: 1.166.11
- OS: any
- Browser: any
- Browser Version: any
- Bundler: vite
- Bundler Version: 8.0.0

### Additional context

Why this matters

Slightly increases HTML payload size on every SSR response before compression Makes View Source unnecessarily noisy
This is framework-generated serialization, not user-authored content there is no "<pre.>" or whitespace-sensitive context to worry about, so it is safe to strip

schiller-manuel:
yeah could potentially be minified. not a high priority though

canmi21:
I submitted PR, waiting for the review

PR #7081, opened 2026-03-31
fix(router): core trim whitespace from SSR dehydrated script output

canmi21:
Fix: https://github.com/TanStack/router/issues/6947

## Summary by CodeRabbit

* **Chores**
  * Improved script handling consistency in server-side rendering by standardizing whitespace formatting in script injection, ensuring cleaner and more consistent script output.

### DIFF
--- a/packages/router-core/src/ssr/ssr-server.ts
+++ b/packages/router-core/src/ssr/ssr-server.ts
@@ -74,13 +74,13 @@ class ScriptBuffer {
 
   constructor(router: AnyRouter) {
     this.router = router
-    // Copy INITIAL_SCRIPTS to avoid mutating the shared array
-    this._queue = INITIAL_SCRIPTS.slice()
+    // Copy INITIAL_SCRIPTS and trim to avoid unnecessary whitespace in HTML output
+    this._queue = INITIAL_SCRIPTS.map((s) => s.trim())
   }
 
   enqueue(script: string) {
     if (this._cleanedUp) return
-    this._queue.push(script)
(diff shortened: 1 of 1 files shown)


packages/router-core/src/ssr/ssr-server.ts
```diff
@@ -74,13 +74,13 @@ class ScriptBuffer {
 
   constructor(router: AnyRouter) {
     this.router = router
-    // Copy INITIAL_SCRIPTS to avoid mutating the shared array
-    this._queue = INITIAL_SCRIPTS.slice()
+    // Copy INITIAL_SCRIPTS and trim to avoid unnecessary whitespace in HTML output
+    this._queue = INITIAL_SCRIPTS.map((s) => s.trim())
```
schiller-manuel, line 78: this will not remove whitespace in between. we need proper minification. you can get this today if you minify your server build

canmi21: Good point about `.trim()` not handling internal whitespace.

That said, the official docs recommend `minify: false` for the SSR build (for easier debugging), so relying on server-side minification to fix this feels like a workaround rather than a proper solution.

Since these scripts are machine-generated serialization injected into client-facing HTML (not server-executed code), would it make sense to do a lightweight regex-based cleanup directly in `ScriptBuffer` instead? Something like:

```ts
script.replace(/\n\s*\n/g, '\n').trim()
```

This keeps server builds debuggable while still producing clean HTML output. Happy to update the PR if you think this direction works.

PR #7083, opened 2026-03-31
fix(react-router): use onShellError to destroy stream on unrecoverable SSR errors

### LINKED ISSUE #7078: `onError` in `renderRouterToStream` destroys stream even for recoverable errors, breaking Error Boundary SSR rendering

Start

### Describe the bug

In `renderRouterToStream.tsx`, the `onError` callback added in #5996 destroys the `PassThrough` stream on **any** error, including errors that are caught by React Error Boundaries (recoverable errors).

React's `renderToPipeableStream` calls `onError` for **all** errors — both recoverable (caught by Error Boundary) and unrecoverable. When the error is recoverable:

1. Error Boundary catches the error and renders `errorComponent` / `defaultErrorComponent`
2. `onShellReady` fires normally — the shell includes the error fallback UI
3. `onError` also fires → **`stream.destroy()` kills the already-piping stream**
4. Response body becomes broken → upstream proxy returns **502 Bad Gateway**

This means `defaultErrorComponent` and route-level `errorComponent` never render during SSR — users always see a 502 instead of the intended error page.

N/A (reproducible in any TanStack Start app with `defaultErrorComponent` set)

1. Set `defaultErrorComponent` in `createRouter()`:
```ts
const router = createRouter({
    routeTree,
    defaultErrorComponent: () => <MyErrorPage />,
})
```
2. Throw an error in any route component:
```tsx
function RouteComponent() {
    throw new Error('test error')
    return <div>Hello</div>
}
```
3. Access the page via SSR → 502 Bad Gateway instead of MyErrorPage

### Expected behavior

The Error Boundary should catch the error and render defaultErrorComponent during SSR, returning a valid HTML response (with status 500).

### Screenshots or Videos

_No response_

### Platform

@tanstack/react-router: 1.167.16 (any version after #5996 merge on 2025-11-29)
@tanstack/react-start: 1.167.16

#### Root Cause
In [packages/react-router/src/ssr/renderRouterToStream.tsx](https://github.com/TanStack/router/blob/main/packages/react-router/src/ssr/renderRouterToStream.tsx):

```ts
onError: (error, info) => {
    console.error('Error in renderToPipeableStream:', error, info)
    // Destroy the passthrough stream on error
    if (!reactAppPassthrough.destroyed) {
        reactAppPassthrough.destroy(
            error instanceof Error ? error : new Error(String(error)),
        )
    }
},
```
This was added in #5996 to fix memory leaks, but onError fires for recoverable errors too. Additionally, there is no onShellError handler, so truly unrecoverable errors also have no proper fallback.

#### Proposed Fix
Follow [React's recommended pattern](https://react.dev/reference/react-dom/server/renderToPipeableStream#setting-the-status-code):

```ts
onShellReady() {
    pipeable.pipe(reactAppPassthrough)
},
onShellError(error) {
    // Handle truly unrecoverable errors with a fallback response
    reactAppPassthrough.end(FALLBACK_HTML)
},
onError(error, info) {
    // Log only — do NOT destroy the stream.
    console.error('Error in renderToPipeableStream:', error, info)
},
```
- `onError`: logging/flag-setting only, no `stream.destroy()`. 
  This aligns with [React's recommended pattern](https://react.dev/reference/react-dom/server/renderToPipeableStream#setting-the-status-code) 
  where `onError` is used solely for logging — none of React's official 
  examples destroy the stream in `onError`.
- `onShellError` (new): handle unrecoverable shell errors with fallback HTML, 
  as documented in React's API reference.

mixelburg:
Fixes #7078

React's `renderToPipeableStream` calls `onError` for **all** errors — both recoverable (caught by Error Boundaries) and unrecoverable. The previous implementation destroyed the `PassThrough` stream in `onError`, which meant:

1. A route throws an error
2. React Error Boundary catches it → renders `defaultErrorComponent`
3. `onShellReady` fires normally with the error UI in the shell
4. `onError` also fires → `stream.destroy()` kills the already-piping stream
5. Response body is broken → upstream proxy returns 502

**Fix:** Follow [React's recommended pattern](https://react.dev/reference/react-dom/server/renderToPipeableStream#setting-the-status-code):
- `onError`: logging only, no stream destruction
- `onShellError` (new): handle unrecoverable shell render failures with `stream.destroy()`

`onShellError` fires only when React cannot produce any HTML at all — the true unrecoverable case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling in server-side rendering to better distinguish between shell-level errors and runtime errors, ensuring more accurate error logging and recovery.

### DIFF
--- a/packages/react-router/src/ssr/renderRouterToStream.tsx
+++ b/packages/react-router/src/ssr/renderRouterToStream.tsx
@@ -59,14 +59,16 @@ export const renderRouterToStream = async ({
                 pipeable.pipe(reactAppPassthrough)
               },
             }),
-        onError: (error, info) => {
-          console.error('Error in renderToPipeableStream:', error, info)
-          // Destroy the passthrough stream on error
+        onShellError(error) {
+          console.error('Error in renderToPipeableStream:', error)
           if (!reactAppPassthrough.destroyed) {
             reactAppPassthrough.destroy(
               error instanceof Error ? error : new Error(String(error)),
             )
           }
+        },
+        onError: (error, info) => {
+          console.error('Error in renderToPipeableStream:', error, info)
         },
       })
     } catch (e) {


mixelburg:
Closing due to inactivity. Over 3 months with no review.

PR #7086, opened 2026-04-01
[codex] Add Solid Start pending hydration regression test

ljho01:
## What changed

This adds a Solid Start regression test for `ssr: 'data-only'` with a slow loader and a `pendingComponent` that returns a DOM element.

It introduces:
- a dedicated `/mre-data-only` route under `e2e/solid-start/selective-ssr`
- a dev-only Playwright config so the hydration path is exercised against the Vite dev server
- a focused Playwright spec that asserts the route hydrates through pending and reaches `OK — loader finished` without `template is not a function` or `Hydration Mismatch`

## Why

This is intended to lock in the Solid hydration fix around `pendingComponent` on `data-only` routes.

Current `main` passes this test.
If `packages/solid-router/src/Match.tsx` is changed from:

```ts
(isServer ?? router.isServer) || resolvedNoSsr
```

to:

```ts
(isServer ?? router.isServer) && resolvedNoSsr
```

the same test fails in dev mode and the route never reaches the loaded UI.

## Impact

This gives us a targeted regression test for the exact hydration behavior that previously broke in dev mode for Solid Start.

## Validation

Ran:

```bash
cd e2e/solid-start/selective-ssr
rm -rf port*.txt && pnpm exec playwright test --config=playwright.dev.config.ts tests/pending-component-hydration.spec.ts --project=chromium
```

Also manually verified:
- current `main` / fixed `||` condition: passes
- local temporary revert to `&&`: fails
- restored `||`: passes again

### DIFF
--- a/e2e/solid-start/selective-ssr/package.json
+++ b/e2e/solid-start/selective-ssr/package.json
@@ -9,7 +9,8 @@
     "build": "vite build && tsc --noEmit",
     "preview": "vite preview",
     "start": "pnpx srvx --prod -s ../client dist/server/server.js",
-    "test:e2e": "rm -rf port*.txt; playwright test --project=chromium"
+    "test:e2e": "rm -rf port*.txt; playwright test --project=chromium",
+    "test:e2e:dev": "rm -rf port*.txt; playwright test --config=playwright.dev.config.ts --project=chromium"
   },
   "dependencies": {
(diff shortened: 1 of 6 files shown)


brenelz:
Shouldn't the tests be failing if you add a regression test?

brenelz:
Integrated what I could into this pr

https://github.com/TanStack/router/pull/7266

Issue #7032, opened 2026-03-25
Adding @module-federation/vite breaks vanilla @tanstack start project


jsg2021:
### Which project does this relate to?

Start

### Describe the bug

After generating a new project, adding @module-federation/vite breaks the build.

Similar to #4516 

### Your Example Website or App

https://github.com/jsg2021/tanstack-start-module-federation-issue

### Steps to Reproduce the Bug or Issue

`run build`

### Expected behavior

Build succeeds with normal output plus federated assets

This check probably needs to be more sophisticated: https://github.com/TanStack/router/blob/b1c0f4602df7d3950df199e6471774f0520a359b/packages/start-plugin-core/src/start-manifest-plugin/manifestBuilder.ts#L195-L200

We can produce multiple "entry" points... Is there a way to flag which one we want in this context?

### Platform

- Router / Start Version: 1.167.6
- OS: macOS/Linux
- Browser: irrelevant, in terminal.
- Bundler: Vite
- Bundler Version: 7.3.1

jsg2021:
using ` hostInitInjectLocation: 'entry', ` gets past multiple entry issues.

jsg2021:
Not sure where the best place to fix this? my instinct @module-federation/vite?

From claude:

The server-fn:client plugin transforms any module whose source contains server-fn markers (createServerFn, .handler, .validator, …). To classify each call expression, StartCompiler crawls the import graph — resolveImportKind → resolveIdCached(source) → getModuleInfo — following import bindings module-to-module ([compiler.js:869-893](vscode-webview://17i0jvls1hgs88m5r1gsa1sdr4joit1fspnq8q2rgobuj7dbmosf/node_modules/@tanstack/start-plugin-core/dist/esm/start-compiler/compiler.js#L869)).

[...]

PR #7089, opened 2026-04-01
fix(start): skip plugin-injected entries in manifest chunk scanner

rr-jimmy-multani:
Hey everyone! I'm working on a POC for TanStack Start + Module Federation and hit this issue. This is my first PR here so apologies if I've missed anything from the contributing guide.

## The problem

When using `@module-federation/vite` as a host, `vite build` fails with:

```
Error: multiple entries detected: assets/hostInit-xxx.js assets/main-xxx.js
```

`@module-federation/vite` emits additional entry chunks into the client bundle — `hostInit` (via `__mf__virtual/...`) and `remoteEntry` (via `virtual:mf-...`). The `scanClientChunks` function in `manifestBuilder.ts` throws as soon as it sees more than one `isEntry` chunk, which makes it incompatible with any bundler plugin that emits its own entry points.

This is being tracked in #7032.

## What I changed

`scanClientChunks` now uses `facadeModuleId` to detect plugin-injected entries and skip them. `facadeModuleId` preserves the original entry module ID before Rolldown resolves it to a real file path, making it a reliable discriminator:

- Chunks with `__mf__virtual` in their `facadeModuleId` → MF `hostInit` entry, skip
- Chunks with a `virtual:mf-` prefix in their `facadeModuleId` → MF `remoteEntry`, skip
- Everything else → treated as the real app entry, original behaviour preserved (including throwing on genuine duplicates)

## How I tested it

- Added 3 unit tests to `manifestBuilder.test.ts` covering both skip patterns and the existing throw-on-duplicate behaviour
- Verified `tanstack-start-example-basic` still builds cleanly
- Validated end-to-end in the POC repo: https://github.com/rr-jimmy-multani/rr-tanstack-poc

## Release Notes

* **Bug Fixes**
  * Prevented false "multiple entries" build errors by ignoring framework/plugin-injected entry chunks so only actual app entries are considered.

* **Tests**
  * Added regression tests to ensure injected entries are skipped and genuine multiple-entry errors still surface.

### DIFF
--- a/packages/start-plugin-core/src/start-manifest-plugin/manifestBuilder.ts
+++ b/packages/start-plugin-core/src/start-manifest-plugin/manifestBuilder.ts
@@ -194,12 +194,19 @@ export function scanClientChunks(
     chunksByFileName.set(bundleEntry.fileName, bundleEntry)
 
     if (bundleEntry.isEntry) {
-      if (entryChunk) {
-        throw new Error(
-          `multiple entries detected: ${entryChunk.fileName} ${bundleEntry.fileName}`,
-        )
+      // Skip injected entries from bundler plugins (e.g. @module-federation/vite
+      // emits hostInit and remoteEntry entries). These are not the app entry.
+      const facadeId = bundleEntry.facadeModuleId ?? ''
+      const isPluginInjectedEntry =
+        facadeId.includes('__mf__virtual') || facadeId.startsWith('virtual:mf-')
+      if (!isPluginInjectedEntry) {
+        if (entryChunk) {
+          throw new Error(
+            `multiple entries detected: ${entryChunk.fileName} ${bundleEntry.fileName}`,
+          )
+        }
+        entryChunk = bundleEntry
       }
-      entryChunk = bundleEntry
     }
 
     const routeFilePaths = getRouteFilePathsFromModuleIds(bundleEntry.moduleIds)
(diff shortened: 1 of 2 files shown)


packages/start-plugin-core/src/start-manifest-plugin/manifestBuilder.ts
```diff
@@ -194,12 +194,19 @@ export function scanClientChunks(
...
     if (bundleEntry.isEntry) {
-      if (entryChunk) {
-        throw new Error(
-          `multiple entries detected: ${entryChunk.fileName} ${bundleEntry.fileName}`,
-        )
+      // Skip injected entries from bundler plugins (e.g. @module-federation/vite
+      // emits hostInit and remoteEntry entries). These are not the app entry.
+      const facadeId = bundleEntry.facadeModuleId ?? ''
+      const isPluginInjectedEntry =
+        facadeId.includes('__mf__virtual') || facadeId.startsWith('virtual:mf-')
```
schiller-manuel, line 201: we will not hardcode "arbitrary" strings here. if at all, i could imagine a generic, configurable filter logic (via a callback for example). however we are restructuring plugins right now, so this has to wait

rr-jimmy-multani: No worries! Thank you for taking the time to review this! Is there a thread or PR I can follow while we wait?

rr-jimmy-multani: Hey @schiller-manuel! Would TanStack Start ever expose a programmatic API that bundler plugins could use to register entry filters themselves? Or is the expectation that this kind of integration lives entirely in user config?

PR #7107, opened 2026-04-05
fix(react-router): preserve inline head scripts in React tree after hydration

### LINKED ISSUE #7104: The content/children script tags of the head (`scripts[].children`) are lost from the HTML after hydration

Router

### Describe the bug

```
export const Route = createFileRoute('/test')({
  component: RouteComponent,
  head: () => ({
    scripts: [
      {
        async: true,
        src: `https://www.googletagmanager.com/gtag/js?id=ID`,
      },
      {
        // this is missing on hydrated HTML
        children: `window.dataLayer = window.dataLayer || [];
function gtag(){dataLayer.push(arguments);}
gtag('js', new Date());
gtag('config', 'ID');`,
      },
    ],
  }),
});
```

Children code in scripts (`<script>window.dataLayer...</script>`) disappears after hydration on first page visit, reappears if navigating first to another page.

https://stackblitz.com/edit/github-zpekad3w-lxs8xvjn?file=src%2Froutes%2Ftest%2Froute.tsx

1. Build the attached StackBlitz site (`npm run build`)
2. `node .output/server/index.mjs`
3. navigate to the test page
4. Refresh the page and check the HTML element in the console `<script>window.dataLayer...</script>` is missing.

### Expected behavior

`<script>window.dataLayer...</script>` should not disappear after hydration

### Screenshots or Videos

https://github.com/user-attachments/assets/badce2ce-7cb9-4ab0-a749-844ecb48fa8c

### Platform

- Router / Start Version: 1.168.10 / 1.167.16
- OS: Windows (WSL 2)
- Browser: Chrome
- Browser Version: 146.0.7680.178
- Bundler: vite
- Bundler Version: 8.0.3

### Additional context

_No response_

mixelburg:
Fixes #7104

## Problem

When a route's `head()` function returns `scripts[]` entries with `children` (inline script content) but no `src`, those script tags are rendered correctly on the server and during initial hydration, but **disappear from the DOM after hydration completes**.

Root cause in `Asset.tsx` — the `Script` component:

1. During SSR: renders `<script dangerouslySetInnerHTML={{ __html: children }} />` ✓
2. Pre-hydration (`!hydrated`): renders the same element to match server HTML ✓
3. After hydration: falls through to `return null` — **React unmounts the script from the DOM** ✗
4. `useEffect` fires after React commits the removal → `querySelectorAll('script:not([src])')` no longer finds the script → creates a new one and appends it → **script re-executes** (e.g. `gtag` config double-fires)

Both `solid-router` and `vue-router` already keep inline scripts rendered after hydration — only `react-router` had this gap.

## Fix

After the `!hydrated` block, add an early-return that keeps inline `children` scripts in the React tree:

```tsx
if (!attrs?.src && typeof children === 'string') {
  return <script {...attrs} dangerouslySetInnerHTML={{ __html: children }} suppressHydrationWarning />
}
```

With this change:
- React never unmounts the SSR-rendered script element
- The `useEffect` still runs, but the `existingScript` check finds the element and returns early — so the script is **not re-executed**
- Behavior is now consistent with `solid-router` and `vue-router`

Scripts with `src` are unaffected (still managed imperatively via `useEffect`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where inline scripts were being removed after page hydration, ensuring they remain active and functional throughout the page lifecycle.

### DIFF
--- a/packages/react-router/src/Asset.tsx
+++ b/packages/react-router/src/Asset.tsx
@@ -51,6 +51,11 @@ function Script({
 }) {
   const router = useRouter()
   const hydrated = useHydrated()
+  // Track whether this component instance went through the !hydrated (pre-hydration) phase.
+  // true  → component was SSR-rendered; the inline script was already executed by the browser.
+  // false → component was mounted fresh on the client (client-side navigation); useEffect must
+  //         inject an executable script because dangerouslySetInnerHTML never executes scripts.
+  const wentThroughNonHydratedPhase = React.useRef(!hydrated)
   const dataScript =
     typeof attrs?.type === 'string' &&
     attrs.type !== '' &&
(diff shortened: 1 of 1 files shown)


schiller-manuel:
please add a regression test to the e2e test projects (we already have some tests for the scripts)
please also make sure the other framework adapters (Vue, solid) are working as well

mixelburg:
Closing due to inactivity. Over 2 months with no formal review.

schiller-manuel:
you didnt react to my comments back then, why would we review it in this state?

PR #7116, opened 2026-04-08
fix: preserve unmasked route on SSR hard reload

### LINKED ISSUE #7115: Start: route masks with unmaskOnReload lose the real route on hard reload

Start

### Describe the bug

When a route mask uses `unmaskOnReload: true`, client navigation works, but a hard reload in TanStack Start SSR keeps the browser on the masked URL.

In the repro, the real route is still present in `window.history.state.__tempLocation`, but the server renders `/`. Hydration then lands on `/` instead of `/modal`, so the modal route disappears.

https://github.com/tmm/tanstack-mask-reload-repro

1. Open `/`
2. Click `Open masked modal`
3. Confirm the modal opens while the URL still shows `/`
4. Hard reload the page

### Expected behavior

After reload, the router should restore `/modal` before hydration and keep the modal rendered.

### Screenshots or Videos

No response.

### Platform

- Router / Start Version: `@tanstack/react-router` `1.168.10`, `@tanstack/react-start` `1.167.16`
- OS: macOS
- Browser: Chrome
- Browser Version: current stable
- Bundler: Vite
- Bundler Version: `7.3.2`

### Additional context

Client navigation behaves correctly. This seems specific to SSR reloads with `unmaskOnReload: true`.

tmm:
SSR hard reloads on masked routes can lose the real route because the server only sees the masked URL, so hydration can drop modal or dialog UI that depends on the unmasked route.

This injects a small pre-hydration script for masks with `unmaskOnReload` that restores `history.state.__tempLocation` before hydration (while preserving the SSR nonce).

Closes #7115.

## Summary by CodeRabbit

* **New Features**
  * Server-side rendering now preserves masked route navigation state during hard page reloads, maintaining user context and preventing loss of state across browser refreshes

* **Tests**
  * Added comprehensive test coverage for route state preservation behavior during page reload scenarios

### DIFF
--- a/.changeset/calmly-fox-smiled.md
+++ b/.changeset/calmly-fox-smiled.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-router': patch
+---
+
+Injected `unmaskOnReload` pre-hydration script during SSR so hard reloads preserved the unmasked route.
--- a/.changeset/swiftly-otter-jumped.md
+++ b/.changeset/swiftly-otter-jumped.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
+Added helpers for building `unmaskOnReload` reload scripts and route mask match patterns. This enabled SSR hard reload recovery for masked routes.
--- a/packages/react-router/src/headContentUtils.tsx
+++ b/packages/react-router/src/headContentUtils.tsx
@@ -431,7 +431,7 @@ function buildUnmaskOnReloadHeadScript(
 
   return {
     tag: 'script',
-    attrs: nonce ? { nonce } : undefined,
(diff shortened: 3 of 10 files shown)


schiller-manuel:
please open an issue first before creating a PR as this probably needs a different solution

tmm:
Thanks for the reply! I created #7115 before opening this PR.

The README.md says "We welcome issues and pull requests!" so figured it was fine to create a PR. Would be good to remove that or update the CONTRIBUTING.md if you don't want pull requests.

schiller-manuel:
sorry didn't see the issue. we accept pull requests for sure but this is probably not the correct solution. I assume there is a less invasive one.

tmm:
Got it. Happy to close this then.

PR #7136, opened 2026-04-10
fix(react-router): clear stale route errors on navigation

schiller-manuel:
fixes #7121

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an error boundary state issue where the error component from a failed route could briefly appear when navigating to a subsequent route.

* **Tests**
  * Added comprehensive test coverage to verify error states no longer leak across route navigations, including errors thrown from loaders and component rendering.

### DIFF
--- a/.changeset/ten-cougars-jump.md
+++ b/.changeset/ten-cougars-jump.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-router': patch
+---
+
+Fix a stale route error boundary state issue that could briefly render the next route's `errorComponent` after navigating away from a failed route.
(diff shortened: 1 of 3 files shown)


schiller-manuel:
@copilot 



In `@packages/react-router/src/CatchBoundary.tsx` around lines 41 - 48,
getDerivedStateFromProps is using unsafe any types and the state's resetKey is
mis-typed as string; change the method signature to use proper types (props: {
getResetKey: () => string | number } and state: { resetKey?: string | number;
error: Error | null } or the existing CatchBoundaryState type updated to make
resetKey optional and accept number|string), and update the state type
declaration so resetKey is optional (resetKey?: string | number) to match the
async initialization pattern where initial state may omit resetKey; ensure
getDerivedStateFromProps returns { resetKey } or { resetKey, error: null } with
the correctly typed resetKey.

PR #7140, opened 2026-04-10
chore: bump to h3 v2 rc.20

birkskyum:
closing https://github.com/h3js/h3/pull/1362
closing https://github.com/h3js/h3/issues/1361

Bumps `h3-v2` to `2.0.1-rc.20` and `cookie-es` to v3.

This needs a small `@tanstack/start-server-core` update because both bumps change behavior that Start currently relies on:

- `requestHandler` now translates the `URIError` thrown by `new H3Event(request)` on malformed percent-encoded request paths (e.g. `/%80`, `/%FF`, `/%E0%A4`) into a 400 Bad Request response. Previously these inputs caused an uncaught error that surfaced as a 500.

- `getCookies()` return type in h3 narrows to `Record<string, string | undefined>`, so filtering undefined allow start to keep the `Record<string, string>`

## Summary by CodeRabbit

* **Chores**
  * Bumped core dependencies (cookie and h3) and added a changeset to trigger a patch release.

* **Bug Fixes**
  * Cookie reads now omit undefined entries for more reliable cookie handling.
  * Requests with malformed URLs are handled gracefully and routed to a safe fallback.

* **Tests**
  * Updated end-to-end test utilities and fixtures for more consistent imports and exports.


e2e/e2e-utils/src/users.ts
```diff
@@ -24,7 +24,7 @@ export type User = {
   }
 }
 
-export const users: Array<User> = [
```
schiller-manuel, line 27: why was this changed?

e2e/e2e-utils/src/posts.ts
```diff
@@ -7,7 +7,7 @@ export type Post = {
   body: string
 }
 
-export const posts: Array<Post> = [
```
schiller-manuel, line 10: why was this changed?

e2e/e2e-utils/src/users.ts
birkskyum: It was because the  CI failed in this run (the commit before i "fix dummy erver imports") because of the exports

https://github.com/TanStack/router/actions/runs/24258504702/job/70836128095

e2e/e2e-utils/src/posts.ts
birkskyum: same as above

PR #7162, opened 2026-04-12
fix(start-rsc): avoid Promise.withResolvers browser dependency

tannerlinsley:
## Summary
- replace `Promise.withResolvers` in `ReplayableStream` with a local deferred helper so the RSC runtime does not depend on a newer Promise API
- avoid client bootstrap and hard refresh failures on RSC pages in browsers where `Promise.withResolvers` is unavailable
- add patch changesets for `@tanstack/react-start-rsc` and `@tanstack/react-start`

## Validation
- `pnpm exec eslint src/ReplayableStream.ts`
- `pnpm exec tsc --noEmit -p tsconfig.json`
- `pnpm exec vitest run tests/ReplayableStream.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved React Server Components replay stream compatibility by removing dependency on unsupported JavaScript APIs, preventing client bootstrap and hard refresh failures in browsers lacking newer standard support.

### DIFF
--- a/.changeset/sharp-peaches-jump.md
+++ b/.changeset/sharp-peaches-jump.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/react-start-rsc': patch
+'@tanstack/react-start': patch
+---
+
+Avoid requiring `Promise.withResolvers` in the RSC replay stream runtime so client bootstrap and hard refreshes do not fail in browsers that do not implement it.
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Closing this because the failure came from running an old Node version, not a framework bug.

PR #7169, opened 2026-04-13
fix(router-generator): normalize virtual physical subtree paths

### LINKED ISSUE #3219: Suspected bug when using virtual routes w/ nested __virtual.ts

Router

### Describe the bug

I am attempting to do a bit of a specific setup with my routes and I am getting some unexpected behavior.
The high level approach is to have a top level virtual route configuration like this
```ts
const routingRoot = path.resolve(__dirname, "./src/routes/");
const virtualRoutes = rootRoute("__root.tsx", [
  index('index.tsx'),
  physical("/s1", "../sub_section/routes/"),
  physical("/s2", "../another_section/routes/"),
]);
```

Those sub sections should be able to do a `__virtual.ts` file if they desire and federate their routes to sub directories as well
```ts
// Folder Hierarchy
// src/sub_section/routes/__virtual.ts
// src/sub_section/foo/routes/index.ts
// example src/sub_section/routes/__virtual.ts 
import {
  defineVirtualSubtreeConfig,
  index,
  layout,
  physical,
} from "@tanstack/virtual-file-routes";

export default defineVirtualSubtreeConfig([
  layout("./_layout.tsx", [
    physical("/foo", "../foo/routes/"),
  ]),
]);

```

The generator wants to generate these routes `'/s1/.//_layout/foo/' instead of what I expected `'/s1//_layout/foo/'
Not sure if this is expected but it results in the route getting not found in the browser.

https://codesandbox.io/p/devbox/rjgqk4?migrateFrom=2jly97

- Open the code sandbox
- Navigating to S1 with the link gets not found
- Navigating to S2 with the link works correctly (since it uses file base routing not __virtual.ts)

### Expected behavior

- S1 should resolve as well to the `foo/routes/index.tsx`

### Screenshots or Videos

n/a

### Platform

- Version: [e.g. 91.1]

## Benefit / Why do this
It allows splitting routes out by logical sections of your code and having shared layouts for those logical sections.

## Debugging / Deep dive
The `./` in the route being added comes from [here](https://github.com/TanStack/router/blob/main/packages/router-generator/src/filesystem/physical/getRouteNodes.ts#L197) which then gets added to all the virtual paths [here](https://github.com/TanStack/router/blob/main/packages/router-generator/src/filesystem/physical/getRouteNodes.ts#L85-L95). I am not sure if this is desired to indicate there is a virtual jump there, but the routes don't correctly resolve anymore in the browser due to this.

## Other Options / Work around?
Perhaps there is a better or more canonically correct way to accomplish something like this that I am unaware of?
Another approach I tried that did not work is being able to define a `__virtual.ts` at the root. This did not work either but also seems like it would be beneficial to support for something like this.

schiller-manuel:
fixes #3219

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed malformed generated paths when physical mounts target subtrees rooted by virtual layouts, particularly in cases involving nested virtual layouts with additional physical routes.


schiller-manuel:
@copilot 


In `@packages/router-generator/src/filesystem/physical/getRouteNodes.ts`:
- Around line 124-137: Concatenate and normalize route segments with cleanPath
to avoid double-slash paths: when building routePath from normalizedDir and
node.routePath, pass the combined string through cleanPath (use
cleanPath(`/${normalizedDir}${node.routePath}`)) and assign that to
node.routePath; likewise generate node.variableName using the cleaned path
(routePathToVariable(cleanPath(`/${normalizedDir}/${removeExt(node.filePath)}`)))
and, if node.originalRoutePath exists, set node.originalRoutePath =
cleanPath(`/${normalizedDir}${node.originalRoutePath}`). This mirrors the
normalization used in getRouteNodesVirtual and prevents `//foo` results.

PR #7190, opened 2026-04-15
fix(docs): remove duplicate code block in query integration guide

CBell045:
## Summary by CodeRabbit

* **Documentation**
  * Removed an unscoped example and left the framework-scoped examples in the "Using useSuspenseQuery vs useQuery" section to reduce duplication and ensure framework-specific samples are shown consistently.


schiller-manuel:
this is not the correct fix. dont add solid. remove the duplicate code block.

PR #7198, opened 2026-04-16
fix: filter route-owned chunks and CSS from root route's entry preloads/assets

### LINKED ISSUE #7197: Root route modulepreloads all route chunks in large projects (Vite 8 / Rolldown)
### Describe the bug

In large projects, the root route ends up with all route chunks in its modulepreload list, which is inherited by every SSR'd page. For a 150+ route app this means 80–150+ `<link rel="modulepreload">` tags on every server-rendered HTML response.

### Context / motivation

I ran into this while looking at SSR HTML size. Every server-rendered page carried a large block of preload tags inherited from the root route, adding a few KB of inline path strings to the critical TTFB payload regardless of which page the user actually requested.

### Minimal reproducible example

https://github.com/ZacharyL2/tsr-modulepreload-repro

### Steps to reproduce

```bash
git clone https://github.com/ZacharyL2/tsr-modulepreload-repro
cd tsr-modulepreload-repro
pnpm install
pnpm build
pnpm inspect
```

Output: `root preloads: 156`. 214 routes with typical `head()` / `loader()` — triggers the issue naturally.

### Expected behavior

Root should only carry the entry chunk and truly shared deps, not chunks owned by specific routes.

### Platform

- `@tanstack/start-plugin-core`: 1.167.35
- `@tanstack/react-start`: 1.167.41
- `@tanstack/react-router`: 1.168.22
- `vite`: 8.0.8 (Rolldown)

### Real-world observation (production app, 152 routes)

- Root preloads: 82
- `/` modulepreloads: 103

ZacharyL2:
## Summary

- In large projects (150+ routes, heavy provider chain in `__root__`), Rolldown hoists route chunks (and their transitive deps like route-data chunks) into the entry chunk's static `imports`. `buildRouteManifestRoutes()` then adds every one of them to the root route's `modulepreload` list, which is inherited by every page — defeating code splitting.
- This PR collects all preloads already owned by non-root routes, then filters them out of the root route's entry-chunk preloads. Because we collect from the already-built `routes[].preloads` (not `routeChunksByFilePath`), the filter covers transitively-reachable chunks (route data, shared chunks only used by routes) that Rolldown splits out separately.
- Only preloads are filtered. CSS assets are intentionally left untouched — genuinely shared CSS reached via entry's static imports needs to stay in root for SSR to avoid FOUC on pages that don't activate the sharing route. `dedupeNestedRouteManifestEntries` continues to handle CSS dedup.

## Test plan

- [x] Added test: filters route chunks + their transitive data chunks from root, while keeping shared vendors
- [x] Updated 2 existing tests where the new semantics move a shared chunk (`root-shared.js`) from root preloads to the routes that use it
- [x] All 32 manifestBuilder tests pass

## Verification

Verified on a production app (152 routes) and the minimal repro:

| Project | Without fix | With fix |
|---|---|---|
| Production app (152 routes) | 82 root preloads | 12 root preloads |
| Minimal repro (214 routes) | 156 root preloads | 2 root preloads |

Closes https://github.com/TanStack/router/issues/7197

Repro (natural trigger, no simulation plugin): https://github.com/ZacharyL2/tsr-modulepreload-repro

### DIFF
--- a/packages/start-plugin-core/src/start-manifest-plugin/manifestBuilder.ts
+++ b/packages/start-plugin-core/src/start-manifest-plugin/manifestBuilder.ts
@@ -395,17 +395,22 @@ export function buildRouteManifestRoutes(options: {
     }
   }
 
-  // Collect filenames of chunks that belong to specific routes so we can
-  // exclude them from the root route's entry-chunk preloads. Without this,
-  // when Rolldown moves route chunks into the entry's static `imports`
-  // (common in large projects), every route chunk leaks into root preloads
-  // and gets modulepreloaded on every page.
+  // Collect filenames of chunks and CSS that belong to specific routes so we
+  // can exclude them from the root route's entry-chunk preloads and assets.
+  // Without this, when Rolldown moves route chunks into the entry's static
+  // `imports` (common in large projects), every route chunk leaks into root
+  // preloads and every route-owned CSS leaks into root assets.
   const routeChunkFileNames = new Set<string>()
   for (const chunks of options.routeChunksByFilePath.values()) {
     for (const chunk of chunks) {
       routeChunkFileNames.add(
         options.assetResolvers.getAssetPath(chunk.fileName),
       )
+      for (const cssFile of chunk.css) {
+        routeChunkFileNames.add(
+          options.assetResolvers.getAssetPath(cssFile),
+        )
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


schiller-manuel:
this does not solve the actual issue, tracking upstream in rolldown https://github.com/rolldown/rolldown/pull/9154

PR #7208, opened 2026-04-16
fix(start): use rsbuild merge api to preserve user config

dake3601:
- Replace custom merge with rsbuild's built-in `mergeRsbuildConfig` in `modifyRsbuildConfig` hook and `createRsbuildEnvironmentPlan`
- Fixes user-defined environment config (e.g. `tools.rspack`) being silently overwritten by the plugin


schiller-manuel:
merged

PR #7209, opened 2026-04-17
fix(router-core): commit pending matches to avoid SSR deadlock

birkskyum:
Resolves
- https://discord.com/channels/719702312431386674/1494384628956401825/1494384628956401825

After SSR hydration, navigating to a route with a loader could leave the router stuck with `status: 'pending'` and `isLoading: true`. `onResolved` never fired.

The commit path was `setMatches(pendingMatches.get()) + setPending([])`, which takes a plain snapshot of the pending-pool values and then drops the pending stores. If a loader completion landed on a pending store after the snapshot was read, that write was lost: the new active store held the stale `'pending'` value and the Transitioner's `hasPending` gate stayed pinned.

The fix is a new `stores.commitPending(nextMatches)` that hands pending stores to the active pool by reference and reads their latest value instead of the snapshot. `hasPending` now also reads the pending pool so a stale active status can't keep it stuck.

Only Solid reproduces this in practice. Solid's `useTransition` defers writes, which opens the window a loader can land in. React's `startTransition` is synchronous, so the same navigation doesn't race. I attempted to repro the same test in `e2e/react-start/query-integration` and it didn't trigger.

### Tests

- 5 new unit tests in `packages/router-core/tests/granular-stores.test.ts`. Each fails if its piece of the fix is reverted.
- Playwright repro in `e2e/solid-start/solid-query-layout-suspense/tests/app.spec.ts`. Deadlocks on `main`, passes here.

## Summary by CodeRabbit

* **New Features**
  * Added a dynamic "/:slug" route with layout and index pages, plus a link from the home page.
  * Home page now uses an async loader to fetch and display its message.

* **Bug Fixes**
  * Fixed an SSR deadlock by atomically committing pending route matches.

* **Tests**
  * Added E2E and unit tests covering SSR, loader data, and dynamic route navigation.


packages/router-core/src/router.ts
```diff
@@ -2474,8 +2474,7 @@ export class RouterCore<
                      * or reloads re-run their loaders instead of reusing the failed/not-found data.
                      */
                     if (mountPending) {
-                      this.stores.setMatches(pendingMatches)
-                      this.stores.setPending([])
```
Sheraff, line 2478: If i understand correctly, the goal is to make this operation "atomic"?
> Prefer the store's current value over the caller's snapshot:
> a write may have landed after the snapshot was taken.

but I don't understand how "a write may have landed after the snapshot was taken". The `pendingMatches` snapshot in question come from `const pendingMatches = this.stores.pendingMatches.get()`, all code is synchronous between there and here, and this is a signal (so it should always be reading the latest value, at least in Solid v1, right?).

How could it not be the latest value?

packages/router-core/src/stores.ts
```diff
@@ -286,6 +289,57 @@ export function createRouterStores<TRouteTree extends AnyRoute>(
     )
   }
 
+  function commitPending(nextMatches: Array<AnyRouteMatch>) {
```
Sheraff, line 292: Just to make sure we're aligned, this function is basically
- a copy of `reconcileMatchPool`
- with the `pendingMatchStores.delete(id)` loop at the end
- and an added section for `if (pendingStore) {`

right?

birkskyum: Yes, I've refactored commitPending to call reconcileMatchPool instead of duplicating it, so it's a bit easier to read.

It moves pending stores into matchStores, passes their current values to reconcileMatchPool, then clears anything still left in the pending pool.

packages/router-core/src/router.ts
birkskyum: The commit block already is synchronous as you pointed out, but the issue is that the snapshot can lag behind the store, and the old path threw away the store identity.

The writes that get things out of sync come from earlier: `updateMatch` during `await loadMatches(...)` wraps every `pendingMatch.set(...)` in `router.startTransition`, which on the Solid adapter is `Solid.startTransition` + `useTransition`. Those writes are deferred, and can flush between the `pendingMatches.get()` read and the commit, and then the snapshot isn't up to date anymore.

The old `setMatches(snapshot) + setPending([])` path created a fresh active store from the stale snapshot and dropped the pending store reference, so the deferred write landed on an orphaned signal, active pool stuck at `status: 'pending'` causing a deadlock.

Reparenting the pending store into the active pool and reading its current value at commit avoids that. React doesn't hit it because `React.startTransition` doesn't defer signal updates the same way.

Sheraff: ok i understand, thank you

This seems problematic though, the issue could come up in more places than just the one you just fixed. And with Solid v2 it's going to be worse, because all signals behave like this, right? (not just in `startTransition` anymore).

Could we imagine `flush`ing in strategic places, or narrowing the scope of the `startTransition` block?

birkskyum: > And with Solid v2 it's going to be worse, because all signals behave like this, right? 

Yes, it theoretically would require flushing in strategic places, but we already have pretty well-behaved solid v2 version, so maybe not that many places are needed.

> This seems problematic though, the issue could come up in more places than just the one you just fixed. 

Right, I think this PR target the most problematic place, but there are certainly other places with similar pattern.

birkskyum: I reckon the structural solution would be to a larger refactor to generalise the reparent pattern into `setMatches`/`setPending`/`setCached` themselves so they always preserve identity and read store values at reconcile time.

Sheraff: yeah i definitely want to rewrite the whole load-matches thing at some point

birkskyum: I've made a tracking ticket here, with other places that appear to follow the same pattern
- https://github.com/TanStack/router/issues/7211

I think this pr still does fix the reported bug which i believe is the most problematic case, so the question is if we can resolve that for now, and then refactor since, or if it's best to go straight for the refactor.

birkskyum:
The ticket this was supposed to resolve
- https://discord.com/channels/719702312431386674/1494384628956401825/1494384628956401825

Has been confirmed to [no longer be an issue](https://discord.com/channels/719702312431386674/1494384628956401825/1495435309901418750) with latest published versions.

Closing.

PR #7219, opened 2026-04-17
fix(rsc): remove outdated stale hmr css references

jantimon:
follow-up to https://github.com/vitejs/vite-plugin-react/pull/1188 on the TanStack Router side where @schiller-manuel was already mentioned

While debugging CSS HMR in TanStack Start RSC (https://tanstack.com/blog/react-server-components) I noticed a second, independent problem: `<link rel="stylesheet">` tags pile up in `<head>` across HMR edits.. Every save appends another `<link>` and none of them get cleaned up

`CompositeComponent` and `RscNodeRenderer` both walk the `cssHrefs` set reported by the RSC Flight stream and call:

```ts
ReactDOM.preinit(href, { as: 'style', precedence: 'high' })
```

Obviously this is a dev only problem. Vite appends an HMR cache-bust (`?t=<lastHMRTimestamp>`) to every CSS href on every edit (see vite-plugin-react [#1188](https://github.com/vitejs/vite-plugin-react/pull/1188). Each edit therefore calls `preinit` with a **different** href which adds a  **new** `<link>` to `<head>` instead of replacing the previous one

A simple possible fix would be to skip these HMR css requests by excluding css files with `/[?&]t=\d+/` as Vite only emits that query in dev so:

- **Prod path** is unchanged
- **Dev path** cache-busted hrefs are ignored 

One open question: can we somehow add HMR e2e tests to verify:

1. changing a css prop updates with hmr
2. removing a css prop reverts the css value to the default

## Summary by CodeRabbit

* **Refactor**
  * Streamlined stylesheet pre-initialization: production preinit retained; dev skips preinit to avoid stale CSS during HMR.

* **New Features**
  * Added example app with routes, navigation, server-rendered demo cards (global CSS and CSS Modules), dev tooling, and Vite/RSC config.
  * Added ESLint config and .gitignore for the example.

* **Tests**
  * Added Playwright e2e tests for CSS HMR, hydration helper, and test setup/teardown.

* **Chores**
  * Bumped dev tool version and added a changeset entry.


schiller-manuel:
does this still update css then in dev upon hmr? if yes, how?

packages/react-start-rsc/src/preinitCssHrefs.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Tells React 19 Float to start fetching each CSS file early and put a
+ * <link rel="stylesheet" data-precedence="high"> in <head> during production builds
+ */
+export function preinitCssHrefs(cssHrefs: Iterable<string> | undefined): void {
+  if (!cssHrefs) return
+  for (const href of cssHrefs) {
+    // Skip Vite's HMR cache-bust query so that no stale <link>s pile up
+    // in <head> during dev from HMR edits. In prod builds, there are no `?t=` hrefs
+    if (HMR_CACHE_BUST.test(href)) continue
```
schiller-manuel: this should be guarded by a process.env.NODE_ENV check

schiller-manuel:
w.r.t. to e2e tests:
we already have hmr e2e tests in e2e/react-start/e2e
a similar setup should work here as well

packages/react-start-rsc/src/preinitCssHrefs.ts
jantimon: done - I switched entirely to your solution as it also fixed another bug

jantimon:
> does this still update css then in dev upon hmr? if yes, how?

`@vitejs/plugin-rsc` `<Resources>` emits a `<link rel="stylesheet" data-rsc-css-href="...">` directly in the Flight stream **for every CSS dependency**

`preinit` was adding a second `<link>` for preloading

jantimon:
The HMR e2e tests in this branch are blocked by https://github.com/vitejs/vite-plugin-react/pull/1188

@hi-ogawa could you please take a look at both PRs and let me know what you think about the fix and the e2e tests?

schiller-manuel:
@jantimon  can you please test with the vite-rsc-plugin preview build in https://github.com/vitejs/vite-plugin-react/pull/1188#issuecomment-4311493996

jantimon:
@hi-ogawa did a great job this weekend and finished, merged and relased my vite-plugin-react pr:
[vite-plugin-react changelog](https://github.com/vitejs/vite-plugin-react/blob/plugin-rsc@0.5.25/packages/plugin-rsc/CHANGELOG.md#0525-2026-04-27) 🎉

**vite-plugin-react v0.5.25**
>  rsc: fix server css hmr with cssLinkPrecedence: false (#1188) (f4647c4)


I will retest tanstack start again with this version and try to simplify the e2e tests today

jantimon:
Okay so I verified that `@vitejs/plugin-rsc` `0.5.25` + `vite` `8.0.10` + this PR fixes the HMR and **all new e2e tests pass**

I also **verified this PR is still needed**

Removing `if (process.env.NODE_ENV !== 'production') return` in https://github.com/jantimon/tanstack-router/blob/1d97f8323f4f686b357530ea0fe7dc37dc1bee83/packages/react-start-rsc/src/preinitCssHrefs.ts#L11 will break the e2e test which verifies that no stale styles survive a delete:

```
 tests/rsc-hmr-css.spec.ts:113:5 › rsc css hmr › global .css: removing a css property hot-updates the style
 Expected: "none"
 Received: "uppercase"
```

e2e/react-start/rsc-hmr/tests/rsc-hmr-css.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      await editCss(c, (source) =>
+        source.replace('text-transform: uppercase;', ''),
+      )
+
+      await expect(page.getByTestId(c.titleTestId)).toHaveCSS(
+        'text-transform',
+        'none',
+      )
+    })
```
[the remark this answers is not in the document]
jantimon: this is tested explicitly by removing a css prop

jantimon:
> w.r.t. to e2e tests: we already have hmr e2e tests in e2e/react-start/e2e a similar setup should work here as well

I looked into consolidating these tests but there are some blockers because of **rspack** rsc compatibility:

1. The following import fails because there is no `@rspack/core/rsc/ssr`:
 https://github.com/TanStack/router/blob/b0024d6310f123736ea18a7f8692b45265cdee74/packages/react-start-rsc/src/rsbuild/ssr-decode.ts#L8
2. If we exclude the rsc routes for rspack `routeTree.gen.ts` flip-flops between toolchains
3. Types don't match for rspack and rsc so I ran into:  
`src/routes/rsc-hmr-css-modules.tsx(4,38): error TS2345: Argument of type '"/rsc-hmr-css-modules"' is not assignable to parameter of type 'keyof FileRoutesByPath | undefined'`  
`src/routes/rsc-hmr-css-modules.tsx(13,11): error TS2339: Property 'Server' does not exist on type '{ crumb: string; } | undefined'`

If you like I can disable the typecheck for rspack inside `e2e/react-start/rsc-hmr`, exclude the HMR routes and add a git ignore entry for `routeTree.gen.ts`

What do you think?

schiller-manuel:
these tests should not run for rspack (yet), please only let them run for vite

jantimon:
@schiller-manuel 

agree but currently all RSC tests run for rspack and also the typecheck works

if I now add the HMR tests to the existing RSC tests and exclude rspack for theses tests I get the issues mentioned above

> If you like I can disable the typecheck for rspack inside e2e/react-start/rsc-hmr, exclude the HMR routes and add a git ignore entry for routeTree.gen.ts

is this okay?

schiller-manuel:
rsc in rsbuild is done now, so this should changed now to only apply for vite

jantimon:
@schiller-manuel I got rspack working with `source: {include: [/[\\/]node_modules[\\/]@tanstack[\\/]/]}` (btw is that the correct way?) and you are right!

rspack works with HMR out of the box without any fix needed

do you think we could fix the css invalidation totally on the vite site without any changes to tanstack start?
should I add the tests anyway?

LadyBluenotes:

PR #7233, opened 2026-04-20
fix(start-server-core): enforce payload size limit on POST server function requests

instantraaamen:
GET requests to server function endpoints already enforce a `MAX_PAYLOAD_SIZE` (1MB) check on the query string payload, but POST requests with `application/json` bodies go straight to `request.json()` with no size validation.

The GET path has this guard:

```ts
// Maximum payload size for GET requests (1MB)
const MAX_PAYLOAD_SIZE = 1_000_000

// ...

if (payloadParam && payloadParam.length > MAX_PAYLOAD_SIZE) {
  throw new Error('Payload too large')
}
```

But the POST path skips any size check:

```ts
let jsonPayload
if (contentType?.includes('application/json')) {
  jsonPayload = await request.json()
}
```

On self-hosted Node.js deployments without a reverse proxy (nginx, Cloudflare, etc.), there's nothing stopping an attacker from sending a multi-gigabyte POST body to any `/_serverFn` endpoint, which will get fully buffered into memory.

Managed platforms like Vercel, Cloudflare Workers, and AWS Lambda already have their own request body limits so they're not affected — but the gap between GET and POST seems like an oversight regardless.

This change reads the body as text first, checks its length against the same `MAX_PAYLOAD_SIZE` constant, then parses. Nothing fancy, just making the two paths consistent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added payload size validation to server functions to prevent processing of excessively large requests, improving stability and resource efficiency.

### DIFF
--- a/packages/start-server-core/src/server-functions-handler.ts
+++ b/packages/start-server-core/src/server-functions-handler.ts
@@ -150,7 +150,11 @@ export const handleServerAction = async ({
 
         let jsonPayload
         if (contentType?.includes('application/json')) {
-          jsonPayload = await request.json()
+          const text = await request.text()
+          if (text.length > MAX_PAYLOAD_SIZE) {
+            throw new Error('Payload too large')
+          }
+          jsonPayload = JSON.parse(text)
         }
 
         const payload = jsonPayload ? parsePayload(jsonPayload) : {}


instantraaamen:
I'm debating between a minimal `Content-Length` pre-check (covers all browser `fetch()` calls, ~2 lines) or the full streaming approach.
Feedback welcome from maintainers.

schiller-manuel:
what is state of the art in other frameworks here?

instantraaamen:
I dug into the source code and ran identical tests across 7 frameworks. Here's what I found:

### Test setup

All frameworks tested on Node.js v23.11.0 with the same payloads:

- **Test A** — GET 5KB query string (valid JSON, URL-encoded)
- **Test B** — POST 200KB JSON
- **Test C** — POST 2MB JSON
- **Test D** — POST 2MB JSON (verbose, to observe streaming behavior)

### GET query string limit

| Framework | GET 5KB (Test A) | Notes |
|---|---|---|
| Express | `200` | No query string size check. Node.js `--max-http-header-size` (16KB) is the ceiling. |
| Hono | `200` | Same — no framework-level GET limit |
| SvelteKit | `200` | `BODY_SIZE_LIMIT` applies to request bodies only, not query strings |
| Next.js | `200` | Server Actions are POST-only by design |
| H3 | `200` | No limit |
| React Router | `200` | No limit |
| **TanStack Start** | `200` | **`MAX_PAYLOAD_SIZE` (1MB) check exists on GET** — unique among these frameworks |

### POST body size limit

| Framework | POST 200KB (Test B) | POST 2MB (Test C) | Streaming abort? (Test D) |
|---|---|---|---|
| Express 5.2.1 (`json({ limit: '1mb' })`) | `200` | `413 Payload Too Large` | No (full upload, rejected after) |
| Hono 4.7.10 (`bodyLimit({ maxSize: 100KB })`) | `413 Payload Too Large` | `413 Payload Too Large` | Yes — aborted at 1.2MB |
| SvelteKit 2.21.4 / adapter-node 5.2.12 (`BODY_SIZE_LIMIT=512K`) | `200` | `500 Internal Server Error` | Yes — aborted at 745KB |
| Next.js 16.2.4 (`bodySizeLimit: '1 MB'`) | `200` | `500 Internal Server Error` | No (full upload, rejected after) |
| H3 2.1.1 (Nuxt/Nitro stack) | `200` | `200` — accepted | No limit |
| React Router 7.14.0 | `200` | `200` — accepted | No limit |
| **TanStack Start 1.167.42** | `200` | `200` — accepted | **No limit (this PR fixes it)** |

### Source code references

| Framework | How POST body is handled | Source |
|---|---|---|
| Express | `raw-body` — streaming + Content-Length pre-check | [`read.js#L39-L173`](https://github.com/expressjs/body-parser/blob/2322e111cc32/lib/read.js#L39-L173) |
| Hono | `ReadableStream` wrapper — Content-Length pre-check + streaming byte count | [`body-limit/index.ts#L68-L125`](https://github.com/honojs/hono/blob/cf2d2b7edcf0/src/middleware/body-limit/index.ts#L68-L125) |
| SvelteKit | Streaming `req.on('data')` — `chunk.length` count + `SvelteKitError(413)` | [`node/index.js#L38-L98`](https://github.com/sveltejs/kit/blob/6b41862871d4/packages/kit/src/exports/node/index.js#L38-L98) |
| Next.js | Streaming `Transform` — byte count per chunk, aborts via `pipeline()` | [`action-handler.ts#L898-L917`](https://github.com/vercel/next.js/blob/a6223ac95d5e/packages/next/src/server/app-render/action-handler.ts#L898-L917) |
| H3 | `event.req.text()` — no size check | [`body.ts#L23-L47`](https://github.com/h3js/h3/blob/e2f8cbd5fc41/src/utils/body.ts#L23-L47) |
| React Router | `request.body` passed through — no size check | [`single-fetch.ts#L37-L89`](https://github.com/remix-run/react-router/blob/7a0551035af8/packages/react-router/lib/server-runtime/single-fetch.ts#L37-L89) |
| **TanStack Start** | `request.json()` — no size check | [`server-functions-handler.ts#L151-L154`](https://github.com/TanStack/router/blob/3a59b7203a81/packages/start-server-core/src/server-functions-handler.ts#L151-L154) |

### How to read the attached logs

Raw `curl` output for each framework — search for:

| What you're looking for | Search string |
|---|---|
| GET test result | `Test A` |
| POST 200KB result | `Test B` |
| POST 2MB result | `Test C` |
| Streaming abort evidence | `Test D` |

### Possible approaches

| # | Approach | Pros | Cons |
|---|---|---|---|
| **A** | Apply `MAX_PAYLOAD_SIZE` (1MB) to POST body (current PR) | Simple, consistent with GET path | `await request.text()` buffers the full body before checking — a multi-GB payload still hits memory |
| **B** | Check `Content-Length` header before reading the body (GET + POST) | Rejects oversized requests instantly, zero memory cost | Chunked transfer-encoding has no `Content-Length` — but browsers always set it for `fetch()` |
| **C** | Streaming body reader with byte counting (like Next.js / SvelteKit) | True memory protection even against chunked requests | More code, more complexity |

### My take

**Option B** (`Content-Length` pre-check) is probably the sweet spot. Browser `fetch()` calls to server functions always include `Content-Length`, so it covers the realistic attack surface with minimal code. A streaming approach (Option C) would be a nice hardening step down the road, but even Next.js still has `// TODO: add body limit` on their Edge runtime.

Happy to update the PR to whichever approach you'd prefer.

### Attachment

[express.log](https://github.com/user-attachments/files/26971648/express.log)
[h3.log](https://github.com/user-attachments/files/26971649/h3.log)
[hono.log](https://github.com/user-attachments/files/26971650/hono.log)
[nextjs.log](https://github.com/user-attachments/files/26971652/nextjs.log)
[react-router-v7.log](https://github.com/user-attachments/files/26971653/react-router-v7.log)
[sveltekit.log](https://github.com/user-attachments/files/26971654/sveltekit.log)
[tanstack-start.log](https://github.com/user-attachments/files/26971655/tanstack-start.log)

PR #7237, opened 2026-04-22
fix(start-rsb): option to disable dev middleware

dake3601:
- Adding option from vite to rsbuild to disable dev server middleware for custom server.
Added it under `rsbuild.installDevServerMiddleware` to follow `vite.installDevServerMiddleware`, but it can probably now be added to `tanstackStartOptionsObjectSchema` under `dev`.
- Normalizing `virtualModuleState.manifestPath`. In windows it was not matching leading to `__TSS_START_MANIFEST_PLACEHOLDER__` in the bundle.


schiller-manuel:
merged

PR #7263, opened 2026-04-25
feat: match params

schiller-manuel:
## Summary by CodeRabbit

* **New Features**
  * Experimental route-matching controls: params.match predicate and params.matchPriority tie-breaker.

* **Removals**
  * Deprecated skipRouteOnParseError option removed.

* **Documentation**
  * RouteOptions docs updated with the new experimental matching controls.

* **Tests**
  * Extensive unit and e2e test coverage added for param-based matching and ambiguous/priority cases.


packages/router-core/src/new-process-route-tree.ts
```diff
@@ -637,12 +635,10 @@ type RouteLike = {
...
   options?: {
-    skipRouteOnParseError?: {
-      params?: boolean
-      priority?: number
-    }
     caseSensitive?: boolean
     params?: {
+      match?: (params: Record<string, string>) => boolean
+      matchPriority?: number
       parse?: (params: Record<string, string>) => any
```
Sheraff: `parse` should not be necessary here, should it?

schiller-manuel: removed

PR #7284, opened 2026-04-28
Solid v2 beta 10

brenelz:


packages/solid-start-client/src/StartClient.tsx
```diff
@@ -1,6 +1,12 @@
+import * as Solid from 'solid-js'
 import { RouterProvider } from '@tanstack/solid-router'
 import type { AnyRouter } from '@tanstack/router-core'
 
 export function StartClient(props: { router: AnyRouter }) {
-  return <RouterProvider router={props.router} />
+  const routerProvider = <RouterProvider router={props.router} />
+  Solid.createRenderEffect(
```
brenelz: this feels like the wrong fix but it works. Going to ask upstream about this

PR #7287, opened 2026-04-29
fix(start-plugin-core): sort server fn manifest entries for deterministic build output

### LINKED ISSUE #6762: Unstable build output due to unsorted manifest map keys in TanStack build artifact

DORI2001:
Fixes #6762.

## Problem

`getResolverManifestEntries` calls `Object.entries(serverFnsById)` and maps over the result without sorting. The `serverFnsById` dictionary is built by processing source files in build-system order, which is non-deterministic across machines and runs. Because JS object key iteration follows insertion order, the generated `manifest` object in the emitted `__tanstack-start-server-fn-resolver-*.mjs` artifact ends up with keys in a different order on each build — changing the file hash even when no functions have changed.

This breaks:
- Content-addressed caching (CDN, build caches)
- Reproducible builds / lockfile verification
- `git diff` noise on committed build artifacts

## Fix

Sort the entries by ID (stable alphabetic order) before mapping. One-line change in `getResolverManifestEntries` — no behavioural difference at runtime since `getServerFnById` looks up by key, not by position.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolver manifest entries are now emitted in deterministic lexicographic ID order.
  * Result: more reproducible manifests and consistent module indexing across builds and environments, reducing unexpected diffs and build variability.

### DIFF
--- a/.changeset/silver-lions-sort.md
+++ b/.changeset/silver-lions-sort.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/start-plugin-core': patch
+---
+
+Sort server function manifest entries by ID before emitting the resolver module. The `serverFnsById` map is populated in source-file scan order, which varies across machines and incremental builds, producing non-deterministic key ordering in the emitted `__tanstack-start-server-fn-resolver-*.mjs` artifact. Stable alphabetic sorting ensures reproducible builds, consistent content hashes, and clean `git diff` on committed artifacts.
--- a/packages/start-plugin-core/src/start-compiler/server-fn-resolver-module.ts
+++ b/packages/start-plugin-core/src/start-compiler/server-fn-resolver-module.ts
@@ -16,17 +16,19 @@ interface GenerateServerFnResolverModuleOptions {
 function getResolverManifestEntries(
   serverFnsById: Record<string, ServerFn>,
 ): Array<ResolverManifestEntry> {
-  return Object.entries(serverFnsById)
-    // Sort entries by ID so that the generated manifest has a stable, deterministic order. 
-    // Non-deterministic ordering causes the compiled hash of the same source file to change 
-    // between builds, breaking content-addressed caching and reproducible deployments.
-    .sort(([a], [b]) => (a < b ? -1 : a > b ? 1 : 0))
-    .map(([id, fn]) => ({
-      id,
-      functionName: fn.functionName,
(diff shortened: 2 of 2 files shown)


DORI2001:
Good point — switched to a plain lexical comparator (`a < b ? -1 : a > b ? 1 : 0`) to avoid any locale-sensitivity from `localeCompare`.

schiller-manuel:
i feel like we had this issue before but i cannot find the PR, didnt we merge something like this already?

DORI2001:
Not as far as I can tell — the closest is #7251 which fixed CSS *asset* ordering (stylesheet load sequence), but `getResolverManifestEntries` was untouched by that PR. This one sorts the *server function* entries in the resolver manifest, which comes from a separate `serverFnsById` dictionary built in source-file-scan order. Happy to dig further if you want to double-check a specific commit.

schiller-manuel:
found the PR : https://github.com/TanStack/router/pull/6763/changes

schiller-manuel:
@DORI2001  please add a changeset

DORI2001:
Added a changeset in `.changeset/silver-lions-sort.md` targeting `@tanstack/start-plugin-core` as a `patch`.

PR #7293, opened 2026-04-29
fix: disabled topLevelVar

schiller-manuel:
this will cause isServer not to be transpiled to `var` and thus rspack can properly DCE

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed build output configuration to properly preserve critical exports during dead-code elimination, ensuring correct behavior with rspack bundler.

### DIFF
--- a/packages/router-core/vite.config.ts
+++ b/packages/router-core/vite.config.ts
@@ -10,9 +10,9 @@ const config = defineConfig({
   build: {
     rolldownOptions: {
       output: {
-        topLevelVar: false
-      }
-    }
+        topLevelVar: false,
+      },
+    },
   },
   test: {
     name: packageJson.name,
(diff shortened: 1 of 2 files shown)


packages/router-core/vite.config.ts
```diff
@@ -7,6 +7,13 @@ import type { ViteUserConfig } from 'vitest/config'
 
 const config = defineConfig({
   plugins: [minifyScriptPlugin()] as ViteUserConfig['plugins'],
+  build: {
+    rolldownOptions: {
+      output: {
+        topLevelVar: false,
```
dsonet, line 13: According to the document here: https://rolldown.rs/reference/OutputOptions.topLevelVar

It is default to ‘false’.

schiller-manuel: true for rolldown alone, but vite sets it to true by default: https://github.com/vitejs/vite/blob/7c3a61f42da6445904e93f0e29e9a2a838fa684a/packages/vite/src/node/build.ts#L768

PR #7294, opened 2026-04-29
docs(start): document Register module augmentation in build-from-scratch

Pramuspl:
Hi there. When I started setting up TanStack start, I noticed that it's missing an important part. Here's the fix.

## Summary

- Adds the missing `declare module '@tanstack/react-router'` / `@tanstack/solid-router` `Register` interface augmentation step to the React and Solid **Build from Scratch** Start guides.
- Without this step, `<Link>` accepts any string and route data hooks lose inferred types, which is easy to miss when following the guide top-to-bottom.
- Includes a short note clarifying that `Register` is a TypeScript-only construct, so skipping it does not break the build (it only silently drops type-safety).

## Test plan

- [x] Docs-only change — no code paths affected
- [x] Verified the snippet matches the existing `Register` pattern used elsewhere in the docs
- [x] Both React and Solid guides updated for parity

### DIFF
--- a/docs/start/framework/react/build-from-scratch.md
+++ b/docs/start/framework/react/build-from-scratch.md
@@ -165,6 +165,19 @@ export function getRouter() {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


schiller-manuel:
this is not necessary. the routeTree.gen.ts file will have it

PR #7307, opened 2026-05-01
feat: add buildServerFnUrl

schiller-manuel:
## Summary by CodeRabbit

* **New Features**
  * Added `buildServerFnUrl` helper for constructing server-function URLs with type-safe data handling
  * GET requests now support automatic payload serialization into query parameters
  * Enhanced metadata support for server function fetchers

* **Refactor**
  * Unified payload serialization and URL construction logic for improved consistency


NawfelBgh:
This should address one issue from: [TanStack Start - Preloading with Link Tags Example](https://github.com/NawfelBgh/tanstack-start-example-ssr-cacheable-preload-dynamic).

> Server functions do not provide a way to get their URLs with given parameters, which is needed to construct preload URLs. Currently, only the `serverFn.url` attribute is provided which only works for preloading GET server functions with no parameters.

PR #7325, opened 2026-05-02
fix(start-server-core): fall back to GET handler for HEAD requests (RFC 9110 §9.3.2)

### LINKED ISSUE #7270: HEAD requests to server routes return text/html instead of handler's content-type

Start

### Describe the bug

Server route handlers (`createFileRoute` with `server.handlers.GET`) return the correct `content-type` on `GET`, but `HEAD` to the same path returns `content-type: text/html; charset=utf-8` (the SSR default from `getStartResponseHeaders`) and an empty body. Handler appears not to run for HEAD; the SSR shell takes over.

Real consumers (browsers, RSS readers, sitemap parsers) use GET so impact is small, but it breaks tools that do HEAD probes (`curl -sI`, link checkers, uptime monitors) and conflicts with the HTTP spec — RFC 9110 §9.3.2 requires HEAD to send the same header fields as GET.

Confirmed still reproducing on `@tanstack/react-start` **1.167.49** + `@tanstack/react-router` **1.168.24** (latest as of filing).

https://fasu.dev/sitemap.xml (TanStack Start on Cloudflare Workers)

Route source:

```ts
// src/routes/sitemap[.]xml.ts

export const Route = createFileRoute("/sitemap.xml")({
  server: {
    handlers: {
      GET: () =>
        new Response(buildSitemap(), {
          headers: {
            "content-type": "application/xml; charset=utf-8",
            "cache-control": "public, max-age=3600",
          },
        }),
    },
  },
});
```

1. Create a server-only file route with a `GET` handler returning `new Response(body, { headers: { "content-type": "application/xml; charset=utf-8" } })`.
2. `curl -i https://fasu.dev/sitemap.xml` → `content-type: application/xml; charset=utf-8` ✅
3. `curl -I https://fasu.dev/sitemap.xml` → `content-type: text/html; charset=utf-8` ❌

Reproduces in both `wrangler dev` and `vite preview`. Same behavior across `/sitemap.xml`, `/llms.txt`, `/rss.xml`, and a `/windsurf` shell-script proxy route.

### Expected behavior

HEAD returns the same headers as GET. Per RFC 9110 §9.3.2: _"the server SHOULD send the same header fields in response to a HEAD request as it would have sent if the request method had been GET."_

### Screenshots or Videos

```
$ curl -sI https://fasu.dev/sitemap.xml
HTTP/2 200
content-type: text/html; charset=utf-8       # ❌
...

$ curl -s -o /dev/null -D - https://fasu.dev/sitemap.xml
HTTP/2 200
content-type: application/xml; charset=utf-8  # ✅
cache-control: public, max-age=3600
...
```

### Platform

- Router / Start Version: `@tanstack/react-start` 1.167.49, `@tanstack/react-router` 1.168.24, `@tanstack/router-plugin` 1.167.27
- OS: Linux (Cloudflare Workers runtime + local repro)
- Browser: n/a (curl)
- Bundler: vite 8.0.10 + `@cloudflare/vite-plugin` 1.33.2
- Wrangler: 4.85.0

### Additional context

Tracing through the bundle: `handleServerRoutes` builds `routeMiddlewares = [handlerMiddleware, executeRouter]`. On GET, handler middleware returns a Response and `executeMiddleware` short-circuits. On HEAD, the handler appears to be skipped (or its return discarded) and `executeRouter` runs, invoking `defaultStreamHandler` whose `responseHeaders` default to `text/html; charset=utf-8`.

Likely culprit in `handleServerRoutes`:

```js
const handler = handlers[request.method.toUpperCase()] ?? handlers["ANY"];
```

HEAD has no entry, no `ANY` fallback, so the handler is never registered and the chain falls through to SSR.

Possible fixes:
- For server-only routes (no `component`), auto-fall-back HEAD → GET handler with body stripped (matches RFC 9110 and common framework behavior).
- Or document that users must explicitly add a `HEAD` handler when defining `GET`.

Zelys-DFKH:
Fixes #7270.

RFC 9110 §9.3.2 requires HEAD responses to match GET in headers and status, with no body. When a route has no explicit `HEAD` handler, the server falls back to GET and strips the body.

### What changed

- Handler priority: HEAD now tries `HEAD → GET → ANY`. Previously `ANY` beat `GET`, so a route with both a GET and an ANY handler would skip the GET for HEAD requests entirely — wrong per RFC 9110.
- Body stripping now covers the ANY fallback too, not just GET. If HEAD falls back to ANY, the body is stripped.
- Redirect before strip: the old code stripped the body before resolving redirects, converting a `RedirectResponse` to a plain `Response` before `handleRedirectResponse` could process it. The `Location` header was lost as a result. The fix resolves the redirect first, then strips.
- `statusText` is now copied to the stripped response.

### E2E coverage

Three new Playwright tests in `e2e/react-start/server-routes`:
- HEAD to a GET-only route: body empty, content-type preserved
- HEAD to an ANY-only route: body empty, `X-HANDLER: ANY` and `X-METHOD: HEAD` preserved
- HEAD to a route with both GET and ANY: GET wins, body empty

## Test plan

- [ ] `pnpm test` in `packages/start-server-core`
- [ ] E2E tests pass in `e2e/react-start/server-routes`

## Summary by CodeRabbit

* **New Features**
  * Added three API endpoints: /api/head-fallback, /api/get-and-any, and /api/head-redirect-fallback.

* **Bug Fixes**
  * HEAD requests now prefer HEAD → GET → ANY (RFC 9110), return empty bodies for HEAD fallbacks, preserve status/headers, and correctly resolve redirects.

* **Tests**
  * Added end-to-end tests validating HEAD behavior, handler selection, header preservation, and redirect flow.

### DIFF
--- a/packages/start-server-core/src/createStartHandler.ts
+++ b/packages/start-server-core/src/createStartHandler.ts
@@ -947,11 +947,7 @@ async function handleServerRoutes({
       request,
       getRouter,
     )
-    return new Response(null, {
-      status: resolved.status,
-      statusText: resolved.statusText,
-      headers: resolved.headers,
-    })
+    return new Response(null, resolved)
   }
 
   return ctx.response
--- a/packages/start-server-core/src/createStartHandler.ts
+++ b/packages/start-server-core/src/createStartHandler.ts
@@ -941,8 +941,12 @@ async function handleServerRoutes({
 
   // RFC 9110 §9.3.2: HEAD must carry the same header fields as GET but no body.
   // Resolve any redirect before stripping so the Location header survives.
-  if (isHeadFallback && ctx.response instanceof Response) {
-    const resolved = await handleRedirectResponse(ctx.response, request, getRouter)
+  if (isHeadFallback) {
+    const resolved = await handleRedirectResponse(
+      ctx.response as Response,
+      request,
+      getRouter,
+    )
     return new Response(null, {
       status: resolved.status,
       statusText: resolved.statusText,
--- a/packages/start-server-core/src/createStartHandler.ts
+++ b/packages/start-server-core/src/createStartHandler.ts
@@ -891,6 +891,7 @@ async function handleServerRoutes({
 
   // Add handler middleware if exact match
   const server = foundRoute?.options.server
+  let isHeadFallback = false
   if (server?.handlers && isExactMatch) {
     const handlers =
       typeof server.handlers === 'function'
@@ -898,7 +899,14 @@ async function handleServerRoutes({
         : server.handlers
 
     const requestMethod = request.method.toUpperCase() as RouteMethod
-    const handler = handlers[requestMethod] ?? handlers['ANY']
+    // Per RFC 9110 §9.3.2, HEAD must return the same header fields as GET.
+    // Priority for HEAD: explicit HEAD handler → GET → ANY (last resort).
+    const handler =
+      requestMethod === 'HEAD'
+        ? handlers['HEAD'] ?? handlers['GET'] ?? handlers['ANY']
+        : handlers[requestMethod] ?? handlers['ANY']
+    isHeadFallback =
+      requestMethod === 'HEAD' && handler !== undefined && !handlers['HEAD']
 
     if (handler) {
       const mayDefer = !!foundRoute.options.component
@@ -930,6 +938,17 @@ async function handleServerRoutes({
     params: routeParams,
     pathname,
   })
+
+  // RFC 9110 §9.3.2: HEAD must carry the same header fields as GET but no body.
+  // Resolve any redirect before stripping so the Location header survives.
+  if (isHeadFallback && ctx.response instanceof Response) {
+    const resolved = await handleRedirectResponse(ctx.response, request, getRouter)
+    return new Response(null, {
+      status: resolved.status,
+      statusText: resolved.statusText,
+      headers: resolved.headers,
+    })
+  }
 
   return ctx.response
 }
--- a/packages/start-server-core/src/createStartHandler.ts
+++ b/packages/start-server-core/src/createStartHandler.ts
@@ -903,8 +903,8 @@ async function handleServerRoutes({
     // Priority for HEAD: explicit HEAD handler → GET → ANY (last resort).
     const handler =
       requestMethod === 'HEAD'
-        ? handlers['HEAD'] ?? handlers['GET'] ?? handlers['ANY']
-        : handlers[requestMethod] ?? handlers['ANY']
+        ? (handlers['HEAD'] ?? handlers['GET'] ?? handlers['ANY'])
+        : (handlers[requestMethod] ?? handlers['ANY'])
     isHeadFallback =
       requestMethod === 'HEAD' && handler !== undefined && !handlers['HEAD']
 
--- a/packages/start-server-core/src/createStartHandler.ts
+++ b/packages/start-server-core/src/createStartHandler.ts
@@ -903,8 +903,8 @@ async function handleServerRoutes({
     // Priority for HEAD: explicit HEAD handler → GET → ANY (last resort).
     const handler =
       requestMethod === 'HEAD'
-        ? (handlers['HEAD'] ?? handlers['GET'] ?? handlers['ANY'])
-        : (handlers[requestMethod] ?? handlers['ANY'])
+        ? handlers['HEAD'] ?? handlers['GET'] ?? handlers['ANY']
+        : handlers[requestMethod] ?? handlers['ANY']
     isHeadFallback =
       requestMethod === 'HEAD' && handler !== undefined && !handlers['HEAD']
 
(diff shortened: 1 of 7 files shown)


packages/start-server-core/src/createStartHandler.ts
```diff
@@ -931,5 +942,14 @@ async function handleServerRoutes({
     pathname,
   })
 
+  // Strip body from HEAD responses that fell back to the GET handler.
+  // RFC 9110 §9.3.2: HEAD must send the same headers as GET, but with no body.
+  if (isHeadFallingBackToGet && ctx.response instanceof Response) {
```
schiller-manuel: the wrapped response does not preserve statusText

better:

```tsx
return new Response(null, ctx.response)
```

packages/start-server-core/src/createStartHandler.ts
```diff
@@ -891,14 +891,25 @@ async function handleServerRoutes({
...
 
     const requestMethod = request.method.toUpperCase() as RouteMethod
-    const handler = handlers[requestMethod] ?? handlers['ANY']
+    // Per RFC 9110 §9.3.2, HEAD should return the same header fields as GET.
+    // Fall back to the GET handler when no explicit HEAD or ANY handler is defined.
+    const handler =
+      handlers[requestMethod] ??
+      handlers['ANY'] ??
+      (requestMethod === 'HEAD' ? handlers['GET'] : undefined)
+    isHeadFallingBackToGet =
```
schiller-manuel: should we strip out the body in case ANY handles the HEAD?

packages/start-server-core/src/createStartHandler.ts
```diff
@@ -891,14 +891,25 @@ async function handleServerRoutes({
...
     const handlers =
       typeof server.handlers === 'function'
         ? server.handlers({ createHandlers: (d: any) => d })
         : server.handlers
 
     const requestMethod = request.method.toUpperCase() as RouteMethod
-    const handler = handlers[requestMethod] ?? handlers['ANY']
+    // Per RFC 9110 §9.3.2, HEAD should return the same header fields as GET.
+    // Fall back to the GET handler when no explicit HEAD or ANY handler is defined.
+    const handler =
```
schiller-manuel, line 904: if both GET and ANY exist, HEAD uses ANY, not GET.

could be solved via:
```tsx
const handler =
  requestMethod === 'HEAD'
    ? handlers['HEAD'] ?? handlers['GET'] ?? handlers['ANY']
    : handlers[requestMethod] ?? handlers['ANY']
```

schiller-manuel: add a separate E2E test route for this

packages/start-server-core/src/createStartHandler.ts
```diff
@@ -931,5 +942,14 @@ async function handleServerRoutes({
     pathname,
   })
 
+  // Strip body from HEAD responses that fell back to the GET handler.
+  // RFC 9110 §9.3.2: HEAD must send the same headers as GET, but with no body.
+  if (isHeadFallingBackToGet && ctx.response instanceof Response) {
+    return new Response(null, {
```
schiller-manuel: A GET-only handler returning redirect({ to: '/login' }) can  become a 307 without Location, because handleRedirectResponse() no longer sees isRedirect(response). 

either resolve redirects before stripping the body, or strip the body after handleRedirectResponse().

packages/start-server-core/src/createStartHandler.ts
Zelys-DFKH: Fixed. Handler selection is now `handlers['HEAD'] ?? handlers['GET'] ?? handlers['ANY']`, so GET always wins over ANY for HEAD requests.

Added `/api/get-and-any` with separate GET and ANY handlers, and a test that sends `HEAD /api/get-and-any` and asserts `x-handler: GET`.

packages/start-server-core/src/createStartHandler.ts
Zelys-DFKH: Fixed. `handleRedirectResponse` now runs before the body-stripping step, so Location survives.

Added `/api/head-redirect-fallback` (GET handler returns `redirect({ to: '/api/head-fallback', statusCode: 307 })`), with a test that sends HEAD and checks the browser can follow the redirect to a 200. If Location were dropped, the browser would stall on the 307 and the assertion would fail.

Also adopted your `new Response(null, resolved)` suggestion from the other thread. Cleaner.

packages/start-server-core/src/createStartHandler.ts
```diff
@@ -931,5 +939,16 @@ async function handleServerRoutes({
...
 
+  // RFC 9110 §9.3.2: HEAD must carry the same header fields as GET but no body.
+  // Resolve any redirect before stripping so the Location header survives.
+  if (isHeadFallback) {
+    const resolved = await handleRedirectResponse(
+      ctx.response as Response,
+      request,
+      getRouter,
+    )
+    return new Response(null, resolved)
```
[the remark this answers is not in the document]
schiller-manuel: this makes sense. please incorporate

schiller-manuel:
thanks for your PR!

Zelys-DFKH:
> thanks for your PR!

Thank you for allowing me to contribute!  Keep up the great work 🙏

Zelys-DFKH:
Thanks for the review and merge!

PR #7330, opened 2026-05-03
Add failing test for Invariant failed error reproduction

s3bba:
See the described issue this test reproduces https://github.com/TanStack/router/issues/7331


brenelz:
Thank you for the failing test. This should get fixed here

https://github.com/TanStack/router/pull/7607

PR #7335, opened 2026-05-03
fix: Preserve existing nested scroll positions across SPA navigations that create new restoration keys

schiller-manuel:


packages/router-core/src/scroll-restoration.ts
```diff
@@ -264,6 +287,48 @@ export function setupScrollRestoration(router: AnyRouter, force?: boolean) {
         hash &&
         hashScrollIntoViewOptions &&
         (action === 'PUSH' || action === 'REPLACE')
+      let scrollToTopElements: Array<Element> | undefined
+
+      if (
+        router.isScrollRestoring &&
```
schiller-manuel, line 293: @Sheraff

PR #7338, opened 2026-05-04
fix(plugin): faster hashing

Sheraff:
Yagiz said it was faster, let's see.

`crypto.hash(algorithm, data[, options])`
[nodejs docs](https://nodejs.org/api/crypto.html#cryptohashalgorithm-data-options)
Added in: v21.7.0, v20.12.0
History: v25.4.0 | This API is no longer experimental.

## Benchmarks

Isolation micro-benchmark (just comparing node native modules)

Screenshot 2026-05-04 at 20 59 20

Full benchmark (running full plugin with mocked filesystem)

Screenshot 2026-05-04 at 21 12 58

Screenshot 2026-05-04 at 21 12 50

## Summary by CodeRabbit

* **Chores**
  * Switched internal hash computation to use Node.js’ modern hashing API across packages and build tooling.
  * Added and published a reusable hash utility in the router utilities package.
* **Documentation**
  * Updated examples and guides (ISR ETag generation and server-function ID examples) to reflect the new hashing approach.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/router/pull/7338?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/start-plugin-core/src/start-compiler/compiler.ts
```diff
@@ -600,7 +599,7 @@ export class StartCompiler {
         })
       }
       if (!functionId) {
-        functionId = crypto.createHash('sha256').update(entryId).digest('hex')
+        functionId = hash('sha256', entryId, 'hex')
```
vietnamesekid, line 635: Since we're on a roll with modernizing things, what do you think about `functionId ??= hash('sha256', entryId, 'hex')`? I have a personal vendetta against the ! operator it looks like a tiny person shouting at my code 😂

Issue #6739, opened 2026-02-23
SSR dehydrated match IDs are URL-shaped strings — Google crawls them as phantom URLs


agry-nejo:
### Which project does this relate to?

Router

### Describe the bug

> **Disclaimer:** I'm relatively new to TanStack Start/Router and it's possible I'm missing a configuration option that addresses this. If so, I'd appreciate being pointed in the right direction. I did search existing issues and docs before filing.

TanStack Router constructs match IDs by concatenating `route.id + interpolatedPath + loaderDepsHash` ([`router.ts:1630`](https://github.com/TanStack/router/blob/main/packages/router-core/src/router.ts#L1630)). These match IDs are then serialized into SSR HTML via `dehydrateMatch()` ([`ssr-server.ts:40`](https://github.com/TanStack/router/blob/main/packages/router-core/src/ssr/ssr-server.ts#L40)) and injected into inline `<script>` tags through `crossSerializeStream`.

The resulting match IDs look like relative URLs. For example, a route `src/routes/$orgId/projects/$projectId/index.tsx` visited at `/acme/projects/dashboard` produces:

```
/$orgId/projects/$projectId//acme/projects/dashboard/{}
```

Google's crawler parses the full SSR HTML source — including `<script>` content — extracts these path-like strings, and attempts to crawl them. This produces phantom entries in Google Search Console reported as "Redirect Error" or "Not Found (404)".

This is the same class of issue as Next.js's `__NEXT_DATA__` problem ([vercel/next.js#39377](https://github.com/vercel/next.js/discussions/39377), [vercel/next.js#40143](https://github.com/vercel/next.js/issues/40143)).

### Your Example Website or App

https://stackblitz.com/edit/github-4qzy4wjc?file=src%2Froutes%2Fabout.tsx

### Steps to Reproduce the Bug or Issue

1. Create a TanStack Start app with a dynamic route (e.g., `src/routes/$orgId/projects/$projectId/index.tsx`)
2. Start the dev server
3. Visit a page that resolves the route (e.g., `/acme/projects/dashboard`)
4. View the HTML source
5. Search for `$_TSR` in the source — the dehydrated `<script>` tags contain match IDs like:
   ```
   /$orgId/projects/$projectId//acme/projects/dashboard/
   ```
   This string starts with `/`, contains path separators, and is indistinguishable from a relative URL.

### Expected behavior

[...]

PR #7343, opened 2026-05-04
fix(router-core): encode SSR route ids opaquely

Marve10s:
## Summary

- Replace dehydrated SSR route ID encoding with an opaque `__TSR__` base64url token instead of NUL separators.
- Encode/decode manifest route keys and `notFound().routeId` values in the SSR payload, in addition to match IDs and `lastMatchId`.
- Keep hydration backwards-compatible with existing NUL/replacement-character SSR payloads.

This follows #6739 and #6742. #6742 hid plain slash-shaped match IDs, but the issue can still persist when crawlers or browser parsing normalize NUL/replacement characters, and when route IDs appear through other SSR payload surfaces such as the manifest and not-found errors. This keeps those internal router identifiers from being discoverable as path-like strings in SSR HTML.

## Reproduction

- Minimal public repro: https://github.com/Marve10s/tanstack-router-ssr-route-id-repro
- Current published packages in the repro:
  - `@tanstack/react-router@1.169.1`
  - `@tanstack/react-start@1.167.63`
- Repro command path:
  - `pnpm install`
  - `pnpm build`
  - `pnpm start`
  - in another terminal: `pnpm scan`
- Current result: `pnpm scan` reports SSR HTML leaks for route IDs such as `/_marketing/blog/$slug`, `/_marketing/doc/$docId`, and `/app/_authenticated/_dashboard`.

## Real-World Impact

We found this with [Screaming Frog SEO Spider](https://www.screamingfrog.co.uk/seo-spider/) and Google Search Console. These tools are discovering TanStack internal route IDs from the rendered HTML and reporting the discovered URLs as `404` pages.

image

The discovered URLs follow patterns like this:

image

Those URLs are internal TanStack route IDs, not public application routes, so they return `404` when fetched directly. The problem is that they are serialized in a crawler-visible form.

### Testing

- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/router-core:test:unit --outputStyle=stream --skipRemoteCache -- tests/ssr-match-id.test.ts tests/ssr-route-id-dehydration.test.ts tests/ssr-server-manifest.test.ts tests/hydrate.test.ts`
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/router-core:test:eslint --outputStyle=stream --skipRemoteCache`
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/router-core:test:types --outputStyle=stream --skipRemoteCache`
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/router-core:test:unit --outputStyle=stream --skipRemoteCache`
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/router-core:test:build --outputStyle=stream --skipRemoteCache`
- Repro validation on published packages: `pnpm build` passes, and `pnpm scan` fails with the expected leak report.

## Summary by CodeRabbit

* **Bug Fixes**
  * SSR route identifiers and manifest route keys are now encoded as opaque, URL-safe tokens so crawlers won’t interpret internal route IDs or manifest keys as public paths.

* **Tests**
  * Added and updated tests to verify encoding, decoding, and hydration of SSR route identifiers and manifest keys.


schiller-manuel:
" when crawlers or browser parsing normalize NUL/replacement characters"
i can't believe this really. why would a crawler replace null chars with "/"?

Marve10s:
> " when crawlers or browser parsing normalize NUL/replacement characters" i can't believe this really. why would a crawler replace null chars with "/"?

I can agree on that, this is just one of my assumptions. Unfortunately , I didn't find yet open source SEO tool that would hit same issue. I really don't have any evidence that Googlebot or Screaming Frog specifically replace NUL characters with `/`.

The concrete reproduction here does not depend on that assumption, though. On the current published packages, the SSR payload still contains raw route IDs through other surfaces, especially `manifest.routes` keys, for example:

- `/_marketing/blog/$slug`
- `/_marketing/doc/$docId`
- `/app/_authenticated/_dashboard`

Those are already slash-shaped strings in the HTML, so a crawler can discover them directly.

The NUL/replacement-character part was not great example of me trying to explain why the previous match-id encoding still feels fragile for HTML serialization. Since `hydrateSsrMatchId` already supports both `\0` and `\uFFFD` as legacy slash encodings, this PR keeps that backwards compatibility, but moves newly dehydrated IDs to an opaque token format.

PR #7357, opened 2026-05-07
fix(start): import Register from framework package so module augmentation works

### LINKED ISSUE #7353: TanStack Start server-entry requestContext type ignores @tanstack/react-start Register augmentation

TanStack Start / React Start server entry types.

### Describe the bug

The current Server Entry Point docs say request context should be typed by augmenting `@tanstack/react-start`:

```ts
declare module '@tanstack/react-start' {
  interface Register {
    server: {
      requestContext: MyRequestContext
    }
  }
}
```

With that in place, this should typecheck:

```ts
import handler, { createServerEntry } from '@tanstack/react-start/server-entry'

export default createServerEntry({
  fetch(request, env, ctx) {
    return handler.fetch(request, {
      context: { deps: createAppDeps(env, ctx) },
    })
  },
})
```

In our app, TypeScript reports:

```text
Object literal may only specify known properties, and deps does not exist in type BaseContext.
```

The reason appears to be that `@tanstack/react-start/server-entry` types `ServerEntry` using `Register` from `@tanstack/react-router`:

```ts
import type { Register } from '@tanstack/react-router'
export type ServerEntry = { fetch: RequestHandler<Register> }
```

So the documented `@tanstack/react-start` augmentation is not visible to `serverEntry.fetch(...)`; the context type falls back to `BaseContext`.

### Current workaround

We have to duplicate the same `server.requestContext` augmentation for both modules:

```ts
declare module '@tanstack/react-start' {
  interface Register {
    server: { requestContext: MyRequestContext }
  }
}

  interface Register {
    server: { requestContext: MyRequestContext }
  }
}
```

This typechecks, but it seems inconsistent with the public docs and leaks Start-specific server context into the router module augmentation.

### Expected behavior

The documented `@tanstack/react-start` augmentation should be sufficient for `handler.fetch(request, { context })` / `createServerEntry(...)`.

Alternatively, the docs should mention that `@tanstack/react-router` also needs to be augmented for the server-entry fetch type.

### Versions

```text
@tanstack/react-start: 1.167.59
@tanstack/start-server-core: 1.167.26
@tanstack/start-client-core: 1.168.1
@tanstack/react-router: 1.169.1
TypeScript: 5.9.3
```

Zelys-DFKH:
Fixes #7353.

`createServerEntry` in all three framework adapters imported `Register` from the router package (`@tanstack/react-router`, `@tanstack/solid-router`, `@tanstack/vue-router`), while the docs tell users to augment their start package:

```ts
declare module '@tanstack/react-start' {
  interface Register {
    server: { requestContext: { userId: string } }
  }
}
```

TypeScript module augmentation is module-scoped, so augmenting `@tanstack/react-start` had no effect on a `Register` imported from `@tanstack/react-router`. The typed `context` parameter just silently didn't show up — no error, just no types, which makes it genuinely hard to debug.

The fix follows the same pattern as the existing `router` registration (you augment `@tanstack/react-router` and import from it). All three packages already re-export `Register` through `@tanstack/start-client-core`, and the server-entry files already self-import from their own packages (e.g. `from '@tanstack/react-start/server'`). Changing the `Register` import to match the module users are told to augment is all it takes.

### Changes

- `packages/react-start/src/default-entry/server.ts`: `from '@tanstack/react-router'` → `from '@tanstack/react-start'`
- `packages/solid-start/src/default-entry/server.ts`: `from '@tanstack/solid-router'` → `from '@tanstack/solid-start'`
- `packages/vue-start/src/default-entry/server.ts`: `from '@tanstack/vue-router'` → `from '@tanstack/vue-start'`

No runtime behavior changes.

### How to verify

Add this augmentation in your project:

```ts
declare module '@tanstack/react-start' {
  interface Register {
    server: { requestContext: { userId: string } }
  }
}
```

Before: `context` on `createServerEntry({ fetch })` is optional and typed as `BaseContext` regardless of the augmentation. After: `context` is required and typed as `{ userId: string } & { nonce?: string }`.

## Release Notes

* **Refactor**
  * Updated type imports in React Start, Solid Start, and Vue Start framework integration packages to improve internal consistency and organization.

### DIFF
--- a/packages/react-start/src/default-entry/server.ts
+++ b/packages/react-start/src/default-entry/server.ts
@@ -2,7 +2,7 @@ import {
   createStartHandler,
   defaultStreamHandler,
 } from '@tanstack/react-start/server'
-import type { Register } from '@tanstack/react-router'
+import type { Register } from '@tanstack/react-start'
 import type { RequestHandler } from '@tanstack/react-start/server'
 
 const fetch = createStartHandler(defaultStreamHandler)
--- a/packages/solid-start/src/default-entry/server.ts
+++ b/packages/solid-start/src/default-entry/server.ts
@@ -2,7 +2,7 @@ import {
   createStartHandler,
   defaultStreamHandler,
 } from '@tanstack/solid-start/server'
-import type { Register } from '@tanstack/solid-router'
+import type { Register } from '@tanstack/solid-start'
 import type { RequestHandler } from '@tanstack/solid-start/server'
 
 const fetch = createStartHandler(defaultStreamHandler)
--- a/packages/vue-start/src/default-entry/server.ts
+++ b/packages/vue-start/src/default-entry/server.ts
@@ -2,7 +2,7 @@ import {
   createStartHandler,
   defaultStreamHandler,
 } from '@tanstack/vue-start/server'
-import type { Register } from '@tanstack/vue-router'
+import type { Register } from '@tanstack/vue-start'
 import type { RequestHandler } from '@tanstack/vue-start/server'
 
 const fetch = createStartHandler(defaultStreamHandler)


Zelys-DFKH:
The NxCloud failure in the comment above ran on an older commit (`1674edd`, 2026-05-07). A follow-up push landed on 2026-06-10 (`9c2de22`) but it appears CI did not re-trigger on that commit. Could a maintainer re-run CI or approve the workflow run? The change itself is three one-line import swaps with no logic changes, so I would not expect any test or lint impact.

PR #7364, opened 2026-05-08
fix(start-client-core): allow middleware to return custom error structures

### LINKED ISSUE #7238: Middleware error handling: returning custom values from catch blocks doesn't reach the client
## Problem

We use function middleware to catch errors and convert them to structured error responses so the client can branch on error codes (show field-level validation errors, redirect on auth failures, show billing-specific messages, etc.).

```typescript
export const errorHandler = createMiddleware({ type: 'function' }).server(async ({ next }) => {
  try {
    return await next()
  } catch (error) {
    if (error instanceof DomainError) {
      return { success: false, error: { code: 'BILLING_ERROR', message: error.message } }
    }
    return { success: false, error: { code: 'INTERNAL_SERVER_ERROR', message: 'Something went wrong' } }
  }
})
```

Registered globally:
```typescript
createStart(() => ({
  functionMiddleware: [errorHandler],
}))
```

**Expected:** Client receives `{ success: false, error: { code: 'BILLING_ERROR', ... } }` as the resolved value.

**Actual:** Client promise rejects with just the inner `error` object `{ code: 'BILLING_ERROR', ... }`. The `{ success: false }` wrapper is stripped.

## Root cause

`executeMiddleware` uses `{ result, error }` internally (`ServerFnMiddlewareResult`). When middleware returns an object with an `.error` property, line 140 of `createServerFn.ts` treats it as a middleware-chain error and throws it:

```typescript
if (result.error) throw result.error
```

Any return value from middleware that has a truthy `.error` field collides with this internal protocol. The middleware's return is never delivered to the client as a resolved value.

## What we need

We want a global error boundary for server functions — similar to FastAPI exception handlers or Express error middleware. The goal is:

1. **Classified errors** (auth, billing, validation) — send a structured error code to the client so it can react (redirect, show field errors, show billing prompts)
2. **Unexpected errors** — log server-side for debugging, send a generic "Something went wrong" to the client
3. **Single boundary** — handlers just throw, middleware handles classification

## Questions

1. Is returning custom values from middleware catch blocks intended to work? The docs only show `return next()` or `throw` from middleware — never custom returns.
2. If not, what's the recommended pattern for global error classification in server functions? Should middleware `throw` a structured error object and let the client catch the rejection?
3. Would TanStack consider documenting an official error-handling middleware pattern?

## Environment

- `@tanstack/react-start` v1.167.17
- `@tanstack/start-client-core` v1.167.17

Zelys-DFKH:
Fixes #7238

## Problem

Middleware that wanted to return structured error responses from catch blocks would have those errors thrown immediately to the client, preventing global error classification middleware from working:

```typescript
// Before: result.error gets thrown, { success: false } is lost
try {
  await next()
} catch (e) {
  return { success: false, error: { code: 'BILLING_ERROR', message: e.message } }
}
```

The middleware protocol was using `.error` property for two incompatible purposes:
1. **Framework errors**: Thrown exceptions from middleware, validation failures (instances of Error)
2. **Application errors**: Custom error structures returned by middleware (plain objects)

This collision meant middleware couldn't distinguish between "I'm returning an error as application data" and "this is a framework-level failure that should be thrown."

## Solution

Use JavaScript's type system to distinguish errors:
- Only throw \`Error\` instances (actual exceptions)
- Pass through non-Error values in \`result.error\` as application data

This allows middleware to return custom error structures while preserving proper error propagation for real framework errors.

```typescript
// After: Custom error structure is returned intact
try {
  await next()
} catch (e) {
  return { success: false, error: { code: 'BILLING_ERROR', message: e.message } }
}
// Client receives: { success: false, error: { code: 'BILLING_ERROR', ... } }
```

## Changes

- Line 165: \`if (result.error)\` → \`if (result.error instanceof Error)\`
- Line 305: \`if (result.error)\` → \`if (result.error instanceof Error)\`

## Breaking Change (Minor)

Middleware that throw non-Error values will now have those values captured in \`result.error\` instead of being thrown to the client.

**Impact**: Only affects middleware using non-standard error throwing patterns. Best practice is to throw \`Error\` instances.

## Summary by CodeRabbit

* **New Features**
  * Added a new page with a button to invoke a server call and display a structured non-error payload when returned.
* **Bug Fixes**
  * Client fetcher and middleware now preserve non-exception structured payloads (returned to callers) while still throwing real exceptions and redirect/not-found signals.
* **Tests**
  * Added an end-to-end test validating the UI shows the expected structured payload.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/router/pull/7364?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/e2e/react-start/server-functions/src/routeTree.gen.ts
+++ b/e2e/react-start/server-functions/src/routeTree.gen.ts
@@ -47,6 +47,7 @@ import { Route as MiddlewareMiddlewareFactoryRouteImport } from './routes/middle
 import { Route as MiddlewareFunctionMetadataRouteImport } from './routes/middleware/function-metadata'
 import { Route as MiddlewareClientMiddlewareRouterRouteImport } from './routes/middleware/client-middleware-router'
 import { Route as MiddlewareCatchHandlerErrorRouteImport } from './routes/middleware/catch-handler-error'
+import { Route as MiddlewareNonErrorPayloadRouteImport } from './routes/middleware/non-error-payload'
 import { Route as MethodNotAllowedMethodRouteImport } from './routes/method-not-allowed/$method'
 import { Route as CookiesSetRouteImport } from './routes/cookies/set'
 import { Route as AbortSignalMethodRouteImport } from './routes/abort-signal/$method'
@@ -251,6 +252,12 @@ const MiddlewareCatchHandlerErrorRoute =
     path: '/middleware/catch-handler-error',
     getParentRoute: () => rootRouteImport,
   } as any)
+const MiddlewareNonErrorPayloadRoute =
+  MiddlewareNonErrorPayloadRouteImport.update({
+    id: '/middleware/non-error-payload',
+    path: '/middleware/non-error-payload',
+    getParentRoute: () => rootRouteImport,
+  } as any)
 const MethodNotAllowedMethodRoute = MethodNotAllowedMethodRouteImport.update({
   id: '/method-not-allowed/$method',
   path: '/method-not-allowed/$method',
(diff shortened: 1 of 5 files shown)


Zelys-DFKH:
CI note: the two failing tests in the Test job are pre-existing failures unrelated to this PR's changes.

**Failing tests:**
- `tests/raw-stream.spec.ts:625` - Stream error: propagates error to client (hangs)
- `e2e/react-start/hmr/tests/app.spec.ts:1157` - server-fn-hmr-error shows "none" instead of error message

These exact failures appear identically in #7373 (feat: CSRF middleware), which was merged on 2026-05-09 with the same Test: fail status. Neither test is related to the middleware error propagation code changed in this PR.

@coderabbitai resume

PR #7381, opened 2026-05-11
fix(router-core): fix search middleware type to use schema output

sprusr:
## Summary

The type for `search` passed to a route's search middleware is currently set to the input of whatever is passed to `validateSearch`. In reality this value is the output of the search validator.

In the majority of cases these types are the same, but it becomes a pronounced issue with asymmetric schemas, for example with Zod codecs:

```ts
// input is number, output is BigInt
const numberToBigInt = z.codec(z.int(), z.bigint(), {
  decode: (num) => BigInt(num),
  encode: (bigint) => Number(bigint),
});

const searchSchema = z.object({
  codec: numberToBigInt.optional(),
})
```

The value that we get in the search middleware is a `BigInt` but it's typed as a `number`.

## Changes

Simple switch from `ResolveFullSearchSchemaInput` to `ResolveFullSearchSchema`.

## Reproduction

You can find here a small TanStack Start project with a demo of the mismatch: https://github.com/sprusr/tanstack-router-search-middleware-types

## Summary by CodeRabbit

* **Refactor**
  * Search middleware now consistently receives the validated search shape (middleware input types aligned with validator output).
  * Helper that strips defaulted search parameters accepts validator-shaped defaults for consistent typing.
* **Tests**
  * Added type-level tests to verify middleware and validators produce matching search shapes and stripping behavior.
* **Chores**
  * Recorded a patch changeset for the router-core package.

### DIFF
--- a/.changeset/quiet-plums-help.md
+++ b/.changeset/quiet-plums-help.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
+fix(router-core): use search validator output type for search middleware context
--- a/packages/router-core/src/route.ts
+++ b/packages/router-core/src/route.ts
@@ -1274,9 +1274,7 @@ export interface UpdatableRouteOptions<
   preloadGcTime?: number
   search?: {
     middlewares?: Array<
-      SearchMiddleware<
-        ResolveFullSearchSchemaInput<TParentRoute, TSearchValidator>
-      >
+      SearchMiddleware<ResolveFullSearchSchema<TParentRoute, TSearchValidator>>
     >
   }
   /** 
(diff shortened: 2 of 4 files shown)


sprusr:
Long time TanStack Router fan, first time contributor. I spent a bit of time validating my theory with this one, so hopefully haven't jumped to the wrong conclusion.

As an aside, I would love to see more comprehensive support of asymmetrical schemas particularly when it comes to search params. An example use case: I'm unhappy with the appearance of how arrays of strings are encoded in the url by default, and would prefer to have them as space-separated values - being able to do this with Zod codecs would be wonderful (I am already, but it relies on own middleware).

sprusr:
In fact if maintainers agree with conclusions here then I think `buildLocation` has wrong sided types too, which affects `Link` and `navigate` at least.

sprusr:
@chorobin wondering if you'd have a chance to take a look at this. Not sure what the situation is now with external contributors, but it would be great to have this types issue fixed in some form or another. Thank you!

schiller-manuel:
can you please add a type test?

sprusr:
Thank you @schiller-manuel for pushing this over the line

PR #7389, opened 2026-05-12
ci: zizmor

Sheraff:


.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,25 @@
+name: GitHub Actions Security Analysis
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: ['**']
+
+permissions: {}
```
SeanCassiere, line 9: I think we should explicitly define the runners permissions as content read-only.

Sheraff: If i'm not mistaken, this is even *less* than `contents: read`. It means no permissions at all. 

Source: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions
> You can use the following syntax to disable permissions for all of the available permissions:
> ```yml
> permissions: {}
> ```

SeanCassiere: I suppose I'm a bit confused if it doesn't have read permission but still able to do security analysis 🤔

SeanCassiere: Either ways, if its working across the other repos then its cool - not going to block from me. Just thought it probably make sense for it to be `contents: read`.

PR #7510, opened 2026-05-30
fix(solid-router): prevent HeadContent hydration warnings 

brenelz:
Fix two related issues with HeadContent in solid-router:

1. Hydration: render bare element templates with one getNextElement per invocation (switch instead of if/return), call hydration-affecting hooks unconditionally, and capture the relocated node via the JSX return value instead of a ref attribute so SSR attributes on void meta/link elements are preserved. Eliminates "unclaimed server-rendered node" warnings.

2. FOUC: stabilize head tag object identity per-key across renders so <For> (which keys by reference) reconciles unchanged tags in place instead of remounting every head node on navigation. Previously a single changed tag (e.g. title) invalidated the whole array, remounting the app stylesheet link and briefly detaching it, causing a flash of unstyled content in dark mode.

### DIFF
--- a/packages/solid-router/src/Asset.tsx
+++ b/packages/solid-router/src/Asset.tsx
@@ -73,9 +73,7 @@ function HeadElement(props: {
         innerHTML:
           typeof props.children === 'string' ? props.children : undefined,
       }
-      element = (
-        <style {...attrs} />
-      ) as unknown as Element
+      element = (<style {...attrs} />) as unknown as Element
       break
     }
     case 'meta':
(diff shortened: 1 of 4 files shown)


packages/solid-router/src/Asset.tsx
```diff
@@ -25,89 +25,103 @@ export function Asset({
...
+// On the client, relocate a rendered head element into document.head so head
+// tags end up in the right place even when <HeadContent /> is rendered in
+// <body>. The *same* node that Solid rendered/hydrated is moved — never
+// recreated — so the server-rendered node stays claimed (its hydration id is
+// preserved) and stylesheets/scripts are not refetched or re-executed.
+//
+// When <HeadContent /> is placed in <head> (the SSR/hydration case), the node
+// is already in document.head, so this is a no-op and the element is left
+// exactly where Solid hydrated it.
+function useRelocateToHead(getEl: () => Node | undefined) {
```
birkskyum, line 37: We could consider being strict here like react and just requiring the HeadContent to be in head

PR #7533, opened 2026-06-03
Revert #7524

birkskyum:
Revert 
- #7524

## Summary by CodeRabbit

* **Bug Fixes**
  * Immediate cleanup of transient stream state when server streams end, ensuring prompt resource release and preventing lingering client-side state.

* **Tests**
  * Added automated tests validating client teardown and correct timing of stream cleanup across different document loading states.


packages/router-core/tests/tsr-script-teardown.test.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// Assign `self.$_TSR`.
+function installBootstrap(): TsrBootstrap {
+  new Function(minifiedTsrBootStrapScript)()
+  return (window as any).$_TSR
+}
+
+function setReadyState(value: DocumentReadyState) {
+  Object.defineProperty(document, 'readyState', {
```
schiller-manuel: why do we need to test for `readyState`? streaming is independent from that

birkskyum: we don't need it now that it's immediate again - I've removed it.

PR #7571, opened 2026-06-07
fix: resolve MatchRoute return types for pathless (layout) routes

### LINKED ISSUE #6011: router.matchRoute doesn't resolve TS type with pathless (layout) routes

EduardF1:
## Description
Fixes #6011

Currently, `router.matchRoute` returns a type of `false` when matching against paths contained within pathless (layout) routes (e.g. `_pathless/nested/$id`). The runtime behavior is correct, but the generic `TResolved` resolution fails to extract the underlying string type when passing through `NoInfer` via the Layout masking.

This patch adds a string union extraction step on `TResolved`, ensuring the generic constraint successfully unwraps the pathless definition and provides proper TS intellisense (`ResolveParams`) without evaluating to `never` or `false`.

## Testing
Verified TS compiler locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type safety for route matching results by allowing arbitrary string destinations.
  * Corrected route parameter inference for routes nested under pathless (layout) routes, so matching results now expose the expected parameters instead of incorrectly resolving to no parameters.
  * Updated `MatchRoute` and `useMatchRoute` type behavior to resolve parameters from the matched route path.


EduardF1:
Friendly ping — this PR is still mergeable and all checks are green. Anything I can do to help it land? Happy to rebase onto the latest base or adjust the scope if the approach has changed since it was opened.

EduardF1:
Thanks for your patience. Re-checking the type-level repro against the latest code, the previous `RouteById` + `Extract<TResolved, string>` approach still resolved to `never`: `RouteById` is keyed by route *id*, but `to` resolves to the route *fullPath* (the pathless segment is stripped), so the lookup missed.

Switched the `MatchRouteFn` return type to `RouteByPath` (fullPath-keyed) and added `matchRoute.test-d.ts` asserting pathless-route params resolve (and that the old id-keyed lookup is `never`). Verified with the tsc repro and the router-core typecheck.

PR #7572, opened 2026-06-07
feat(start): support serving assets via relative base urls

### LINKED ISSUE #5966: support serving via a Relative base url

Start

### Describe the bug

I have an app that is running in an iFrame within another app ([home assistant](https://www.home-assistant.io/blog/2019/04/15/hassio-ingress/)). 

Because of this I need to use relative URLs for my assets otherwise things will not resolve because they hit the server of the host app.

I've tried setting `base: "./",` in my vite config, and also I have SPA enabled to keep things as simple as possible. When I check my `_start.html` my links look like the following:

`href="/./assets/main-aAavoY_p.js"`

so I don't get a relative url and the page breaks.

na

1. create a default app
2. serve within an iframe of a host app
3. 😭

### Expected behavior

I would love to be able to use this router to build my addon (and in fact already have ;)) however I'm not able to deploy or ship it because I'm having a hard time getting it to run in the only environment its possible for it to run in.

### Screenshots or Videos

_No response_

### Platform

- Bundler: [e.g. vite]

### Additional context

_No response_

EduardF1:
## Description
Fixes #5966

When TanStack Start is configured with a relative base path (`base: './'`) to be hosted inside sub-directory setups or iFrames (e.g., Home Assistant add-ons), the asset loader was incorrectly discarding the relative dot and prepending a root slash.

This patch normalizes relative base paths in `manifestBuilder.ts` without overriding the `./` prefix, so manifest asset paths stay relative and assets load from the expected location.

## Testing
Verified output generated by the standard build command.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed asset loading for applications served from relative base paths such as `.` or `./`.
  * Relative base paths now remain relative, ensuring manifest asset URLs resolve correctly.


EduardF1:
Thanks for the review. Tracing the actual data flow, the manifest-builder branch turned out to be dead code: `ufo` 1.6.4's `joinURL('./', fileName)` already returns `./fileName`, and the real problem is upstream — `normalizePublicBase('./')` normalizes the relative base to `/./` before it ever reaches the manifest builder, so the `basePath === './'` check could never fire.

Fixed `normalizePublicBase` to preserve `'.'` / `'./'` and reverted the manifestBuilder hunk. Verified in Docker: `normalizePublicBase('./') === './'` and `joinURL('./', 'assets/app.js') === './assets/app.js'`.

PR #7586, opened 2026-06-08
fix(router-core): avoid null bytes in dehydrated SSR match ids

### LINKED ISSUE #7581: SSR produces invalid HTML (Null Byte U+0000)

Router

### Describe the bug

When TanStack Start/Router performs Server-Side Rendering, it produces multiple Null Bytes (U+0000) inside the `"$tsr-stream-barrier"` script tag of the server-generated HTML.
According to the HTML Living Standard specification, this character is forbidden from appearing in input streams. Specifically, section "13.2.2 Parse errors" lists the "control-character-in-input-stream" error (https://html.spec.whatwg.org/multipage/parsing.html#parse-errors).

It appears that browsers and other clients simply ignore this; however, when using an HTML validator for frontend projects (like we do at our company), this issue appears consistently. When validating ssr-produced HTML with https://validator.w3.org the error `Saw U+0000 in stream` comes up.

https://github.com/TanStack/router/tree/main/examples/react/start-basic

1. Create any TanStack Start project that uses SSR. This is easily reproducible with the `start-basic` example from this repo.  
2. Clone this example and install dependencies: `https://github.com/TanStack/router/tree/main/examples/react/start-basic`  
3. Build the project: `npm run build`  
4. Start the server: `node .output/server/index.mjs`  
5. Get the SSR-produced HTML: `curl http://localhost:3000/ > index.html`  
6. Visit the W3 markup validation service and upload the file. (https://validator.w3.org/#validate_by_upload)  
7. Observe the error: `Saw U+0000 in stream.`

This issue is NOT exclusive to the `start-basic` example. From my testing, it occurs every time Router/Start performs server-side rendering and includes the `"$tsr-stream-barrier"` script tag.

### Expected behavior

I expect TanStack's SSR feature to abide by HTML standards and not produce HTML validation errors by default when validating SSR-produced HTML on services like https://validator.w3.org/.

### Screenshots or Videos

Null Bytes occur in these 5 places. First occurence is after `__root__`. 

Image

### Platform

- Router / Start Version: react-router: 1.170.15, react-start: 1.168.24
- OS: Debian Linux
- Browser: Firefox (not relevant to this issue)
- Browser Version: 150.0.1
- Bundler: vite
- Bundler Version: vite 8.0.14

(See the `start-basic` example)

### Additional context

The null bytes seem to appear at five very specific locations in the script tag (as shown in the screenshot). From my testing, this behaviour (of exactly 5) is very consistent, no matter the project or served content.

VihAMBR:
## Summary

`dehydrateSsrMatchId` encodes match IDs for the SSR hydration payload by replacing `/` with `\0`. That encoding came from #6739, so the dehydrated ids stop looking like relative URLs that crawlers pick up as phantom pages.

The catch is that U+0000 is forbidden in the HTML input stream (the `control-character-in-input-stream` parse error in the HTML spec). The dehydrated ids are inlined into the `$tsr-stream-barrier` `<script>`, so every SSR response ends up with raw null bytes and fails markup validation. `validator.w3.org` reports `Saw U+0000 in stream` (#7581). It only works in browsers today because the parser silently rewrites those null bytes to U+FFFD, which is exactly why `hydrateSsrMatchId` already carries a `� -> /` fallback.

This swaps the delimiter from `\0` to `�` (U+FFFD REPLACEMENT CHARACTER):

- it still contains no `/`, so the #6739 crawler fix holds
- U+FFFD is a valid HTML character, so the payload validates and the delimiter survives transport intact instead of depending on the parser to mangle it
- `hydrateSsrMatchId` already decodes `� -> /`, so the client side needs no change

The existing `\0 -> /` decode branch is left in place so any payload still carrying a null byte keeps round-tripping.

## Testing

Added a codec test asserting the dehydrated id has no C0 control characters. It fails on `main` and passes with this change.

- `nx run @tanstack/router-core:test:unit` (39 files, 1179 passed)
- `nx run @tanstack/router-core:test:types`
- `nx run @tanstack/router-core:test:eslint`
- `nx run @tanstack/router-core:build`

Fixes #7581

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server-side rendering hydration payload encoding by replacing null-byte delimiters with a safer sentinel character, preventing invalid U+0000 characters in inline markup and improving validator compatibility.

* **Tests**
  * Added new assertions to confirm legacy null-byte hydration payloads are still decoded for backward compatibility, and that newly generated payloads never include null bytes or other control characters that could break inline SSR content.

### DIFF
--- a/.changeset/ssr-match-id-null-byte.md
+++ b/.changeset/ssr-match-id-null-byte.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
+Encode dehydrated SSR match IDs with the replacement character instead of a null byte, so the inlined hydration payload no longer contains U+0000 (which is invalid in HTML and rejected by markup validators)
--- a/packages/router-core/src/ssr/ssr-match-id.ts
+++ b/packages/router-core/src/ssr/ssr-match-id.ts
@@ -1,5 +1,5 @@
 export function dehydrateSsrMatchId(id: string): string {
-  return id.replaceAll('/', '\0')
+  return id.replaceAll('/', '\uFFFD')
 }
 
 export function hydrateSsrMatchId(id: string): string {
--- a/packages/router-core/tests/ssr-match-id.test.ts
+++ b/packages/router-core/tests/ssr-match-id.test.ts
@@ -23,4 +23,21 @@ describe('ssr match id codec', () => {
   it('decodes browser-normalized replacement chars back to slashes', () => {
     expect(hydrateSsrMatchId('\uFFFDposts\uFFFD1')).toBe('/posts/1')
   })
+
+  it('does not emit control characters that are invalid in SSR HTML', () => {
+    const dehydratedId = dehydrateSsrMatchId(
+      '/$orgId/projects/$projectId//acme/projects/dashboard/{}',
+    )
+
+    // U+0000 and the other C0 control characters trigger a
+    // control-character-in-input-stream parse error when the dehydrated id is
+    // inlined into the SSR <script> payload, so the codec must never emit them.
+    const nullChar = String.fromCharCode(0)
+    expect(dehydratedId).not.toContain(nullChar)
+
+    const hasControlChar = dehydratedId
+      .split('')
+      .some((char) => char.charCodeAt(0) <= 0x1f)
+    expect(hasControlChar).toBe(false)
+  })
 })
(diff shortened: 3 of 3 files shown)


VihAMBR:
Reopening from my main account (VihaanAgarwal).

VihAMBR:
Reopened as #7654.

PR #7588, opened 2026-06-08
response reconciliation

### LINKED ISSUE #5107: `setResponseStatus` does not work anymore with `throw new Error('')`

Start

### Describe the bug

Classic code I have in my better auth with TS project for a long time : 

```ts
export const authUserRequiredMiddleware = createMiddleware({
  type: 'function'
}).server(async ({ next }) => {
  const { headers } = getWebRequest()

  const session = await auth.api.getSession({
    headers,
    query: {
      // ensure session is fresh
      // https://www.better-auth.com/docs/concepts/session-management#session-caching
      disableCookieCache: true
    }
  })

  if (!session) {
    setResponseStatus(401)
    throw new Error('Unauthorized')
  }

  return next({ context: { user: session.user } })
})
```

I'm pretty sure I had the correct status code (401) with my `Unauthorized` message.

Now, I got a 500 in any case.. I would like to throw with a custom status code. Not sure if there is a regression or is it me ?

https://codesandbox.io/p/devbox/bold-johnson-3ly6mc?workspaceId=ws_VUPqXCcptK4abDwy6H7mEq

1. Go to `https://codesandbox.io/p/devbox/bold-johnson-3ly6mc?workspaceId=ws_VUPqXCcptK4abDwy6H7mEq`
2. Click on the button
3. Check network tab
4. Check that we got 500 instead of 401

### Expected behavior

As a user (not logged), I should get a 401 status instead of 500. 

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.131.36
- OS: Mac
- Browser: Chrome 
- Browser Version: 139.0.7258.155
- Bundler: Vite
- Bundler Version: 7.0.0

### Additional context

I think i worked a few weeks ago. 80% sure.

### LINKED ISSUE #5407: `setResponseHeaders` and `setResponseStatus` don't work in global middleware

### LINKED ISSUE #5464: Bug: Only the last setCookie takes effect on the server route

Start

### Describe the bug

When trying to set multiple cookies in a server handler using setCookie, only the last cookie gets actually set in the browser. All previous setCookie calls seem to be overridden or ignored.

https://github.com/Mr-Alexandre/bug-repro-tanstack-start-cookie

1. Start the dev server (e.g., `pnpm dev`).
2. Open your browser and go to http://localhost:3000/api/set-cookies.
3. Open DevTools → Network tab → Select the request to `/api/set-cookies`.
4. Check the "Set-Cookie" headers in the Response Headers.
5. You will see that only the last cookie is set.

### Expected behavior

As a user, I expected multiple Set-Cookie headers to be sent and both cookies to be set, but I am seeing only the last one is applied. The previous one is missing.

### Screenshots or Videos

https://github.com/user-attachments/assets/cbbc9ff2-c8c7-44d3-9c4b-76c4471191c7

### Platform

- OS: [macOS]
- Browser: [Chrome]
- Browser Version: [140.0.7339.214]
- Bundler: [vite]
- Bundler Version: [7.1.7]

### Additional context

This bug happens consistently and prevents setting multiple cookies for auth/session scenarios. Possibly related to how Response headers are managed in TanStack Start server handlers.

schiller-manuel:
fixes #5107
fixes #5464
fixes #5407

## Summary by CodeRabbit

* **New Features**
  * Improved response reconciliation across server routes, SSR, and server functions (cookies, headers, status).
  * Enhanced session handling and cookie helpers for robust persistence and recovery.
  * Added default error/catch UI and route utilities for E2E apps.

* **Bug Fixes**
  * Safer server-entry error recovery and protected header handling to avoid overwrites.
  * Fixed middleware/response ordering and multi-cookie preservation across redirects/replacements.

* **Documentation**
  * Updated server entry guide and added reconciliation docs.

* **Tests**
  * New comprehensive E2E and unit tests covering reconciliation and session behavior.


packages/start-server-core/src/headers.ts
```diff
@@ -0,0 +1,38 @@
...
+  for (const [name, value] of source) {
+    if (name !== 'set-cookie') {
+      target.set(name, value)
+    }
+  }
+  for (const cookie of getSetCookieValues(source)) {
+    target.append('set-cookie', cookie)
+  }
+}
+
```
Sheraff, line 32: In theory there are other headers than `set-cookie` that you are allowed to set multiple times (for example on a Response that would be `Link`, `Vary`, ...). I'm not sure this method is correctly handling that.

packages/start-server-core/src/internal-request-response.ts
```diff
@@ -0,0 +1,1233 @@
...
+
+function normalizeHeaderName(name: string): string {
+  return name.toLowerCase()
+}
+
+function sanitizeStatusMessage(statusMessage = ''): string {
+  let sanitized = ''
+  for (const char of statusMessage) {
+    const code = char.charCodeAt(0)
+    if (code === 0x09 || (code >= 0x20 && code <= 0x7e)) {
```
Sheraff, line 86: more of a nitpick for my puny human mind: it's nice to leave comments on char code values to make it easier to know which char they are referring to.

packages/start-server-core/src/internal-request-response.ts
```diff
@@ -0,0 +1,1233 @@
...
+function sanitizeStatusCode(
+  statusCode: string | number | undefined,
+  defaultStatusCode: number | undefined = 200,
+): number {
+  const fallbackStatusCode = defaultStatusCode
+  if (!statusCode) {
+    return fallbackStatusCode
+  }
+  const code = typeof statusCode === 'string' ? Number(statusCode) : statusCode
+  if (!Number.isInteger(code) || code < 200 || code > 599) {
```
Sheraff, line 102: idk if it matters here, but some services send some quite exotic status codes (https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)
- shopify => 783 Unexpected Token
- linkedin => 999 Request denied

maybe users of Start also want to do weird things like this?

packages/start-server-core/src/internal-request-response.ts
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+function applyHeaderState(
+  target: Headers,
+  eventHeaders: Headers,
+  meta: ResponseMeta,
+  protectedHeaders?: ProtectedHeaders,
+): void {
+  if (meta.clearHeaders) {
+    for (const name of Array.from(target.keys())) {
```
Sheraff, line 345: silly AI doesn't know javascript
```suggestion
    for (const name of target.keys()) {
```

packages/start-server-core/src/internal-request-response.ts
```diff
@@ -0,0 +1,1233 @@
...
+ *
+ * If `xForwardedHost` is `true`, it will use the `x-forwarded-host` header if it exists.
+ *
+ * If no host header is found, it will default to "localhost".
+ */
+export function getRequestHost(opts?: { xForwardedHost?: boolean }) {
+  const headers = getRequestHeaders()
+  if (opts?.xForwardedHost) {
+    const forwardedHost = headers.get('x-forwarded-host')
+    const host = forwardedHost?.split(',')[0]?.trim()
```
Sheraff, line 843: Nit: if you know you only need the 1st segment, limit `.split()` to that segment
```suggestion
    const host = forwardedHost?.split(',', 1)[0]?.trim()
```

packages/start-server-core/src/internal-request-response.ts
schiller-manuel: Careful — this one is load-bearing. Headers iteration is live; deleting while iterating skips entries:

```js
const h = new Headers()
h.set('a','1'); h.set('b','2'); h.set('c','3'); h.set('d','4')
for (const name of h.keys()) h.delete(name)
console.log([...h.keys()]) // [ 'b', 'd' ] — half survive!
```

The `Array.from` snapshot is required. Added a comment in code so this doesn't get "simplified" later.

packages/start-server-core/src/headers.ts
schiller-manuel: Multi-value headers are actually handled losslessly here: iterating a `Headers` object yields repeated values as one comma-combined string (`[...h]` for two `Link` values gives `'<a>, <b>'`), and per RFC 9110 §5.3 that combined form is semantically equivalent — so `set(name, combinedValue)` preserves them. `Set-Cookie` is the one header where comma-joining is unsafe (Expires contains commas), which is why it's special-cased via `getSetCookie()`. Added a comment in code explaining this.

packages/start-server-core/src/internal-request-response.ts
schiller-manuel: Good question, but those exotic codes structurally can't flow through Start: the entire server pipeline (server routes, SSR documents, and server fns alike) uses Fetch `Response` objects as its currency, and the constructor throws outside 200-599:

```js
new Response(null, { status: 999 })
// RangeError: init["status"] must be in the range of 200 to 599, inclusive.
```

So LinkedIn's 999 can't even be *returned* from a handler, let alone reconciled. The clamp just mirrors the platform constraint (and old h3-based behavior, which silently fell back too). New in this PR: `setResponseStatus` now warns in development when a value gets ignored, so this no longer fails silently. Also added a comment above `sanitizeStatusCode` documenting the constraint.

packages/start-server-core/src/internal-request-response.ts
schiller-manuel: Done — added char comments (tab / printable ASCII space–'~').

packages/start-server-core/src/internal-request-response.ts
schiller-manuel: Applied — also to the same pattern in getRequestIP and getRequestProtocol.

e2e/react-start/response-reconciliation/.gitignore
```diff
@@ -0,0 +1,22 @@
...
+.cache
+.tanstack
+.env
+.vercel
+.output
+/build/
+/api/
+/server/build
+/public/build# Sentry Config File
+.env.sentry-build-plugin
```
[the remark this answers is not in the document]
schiller-manuel: Fixed in both new e2e apps.

e2e/react-start/session-handling/src/routes/api/session-seal.ts
```diff
@@ -0,0 +1,20 @@
+import { createFileRoute } from '@tanstack/react-router'
+import { sealSession, useSession } from '@tanstack/react-start/server'
+import { readJson, sessionConfig } from '~/session'
+
+export const Route = createFileRoute('/api/session-seal')({
+  server: {
+    handlers: {
+      POST: async ({ request }) => {
+        const config = sessionConfig({ cookie: false })
+        const session = await useSession<Record<string, any>>(config)
```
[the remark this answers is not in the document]
schiller-manuel: Fixed — concrete session data types where serialization validation applies, Record<string, unknown> elsewhere.

e2e/react-start/response-reconciliation/package.json
```diff
@@ -0,0 +1,84 @@
...
+    "@tanstack/react-router": "workspace:*",
+    "@tanstack/react-start": "workspace:*",
+    "react": "^19.2.3",
+    "react-dom": "^19.2.3"
+  },
+  "devDependencies": {
+    "@playwright/test": "^1.57.0",
+    "@rsbuild/core": "^2.0.8",
+    "@rsbuild/plugin-react": "^2.0.0",
+    "@tanstack/router-e2e-utils": "workspace:^",
```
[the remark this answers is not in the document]
schiller-manuel: Keeping workspace:^ — 37 of 40 e2e apps use it for @tanstack/router-e2e-utils; matching the dominant local convention.

e2e/react-start/response-reconciliation/src/routes/server-functions.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const resultWithResponse = result as typeof result & { result?: unknown }
+  if (!(resultWithResponse.result instanceof Response)) {
+    throw new Error('Expected replacement test to observe a Response')
+  }
+  resultWithResponse.result.headers.set('x-function-a', 'yes')
+  setResponseHeader('x-function-helper-delta', 'yes')
+  return new Response('function-b', {
+    headers: { 'x-function-b': 'yes' },
+  }) as any
+})
```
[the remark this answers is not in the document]
schiller-manuel: Replaced with a documented `as never` escape: function middleware types don't model raw Response replacement, but the middleware executor supports it and this fixture deliberately exercises that runtime path.

e2e/react-start/session-handling/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "@tanstack/react-router": "workspace:*",
+    "@tanstack/react-start": "workspace:*",
+    "react": "^19.2.3",
+    "react-dom": "^19.2.3"
+  },
+  "devDependencies": {
+    "@playwright/test": "^1.57.0",
+    "@rsbuild/core": "^2.0.8",
+    "@rsbuild/plugin-react": "^2.0.0",
+    "@tanstack/router-e2e-utils": "workspace:^",
```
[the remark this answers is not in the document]
schiller-manuel: Keeping workspace:^ — matches the convention used by 37 of 40 existing e2e apps.

hokkyss:
Thanks for accomodating my request 🙏🙏

hokkyss:
> Use `handleStartError(error)` when you want custom top-level logic, such as logging or reporting, but still want Start to produce the same error response as the generated server entry.
> 
> `handleStartError(error)` is useful because the error is caught outside the active Start handler. It recovers the Start request state associated with non-primitive thrown errors and reconciles that state onto the error response. Primitive throws cannot carry that association.
> 
> If you do not want to preserve response helpers that ran before the error, do not call `handleStartError`. Return your own response instead:

I am curious, is it possible to customize the error response, but still using Start's reconciliation behavior?

schiller-manuel:
@hokkyss can you explain what you want to do exactly?

hokkyss:
My idea was:
1. My server functions/`getEnv` throws an HttpError with status codes, error message, etc
2. Show custom response depending on the error

Now that I think about it, if I want to customize the error response and still use Start's reconciliation, then middleware is the answer (?)

schiller-manuel:
also need to add an e2e test for https://github.com/TanStack/router/issues/7647

alvarogomezj:
hi! Really hoping this gets merged so I can try out the framework, as this was a blocker for me. Any news?

alvarogomezj:
@schiller-manuel any possibility about this being merged any time soon?

PR #7590, opened 2026-06-09
fix(start): align Rsbuild SSR asset URLs for css?url imports

### LINKED ISSUE #7543: [Start + rsbuild] CSS emitted into both client and server dist, both URLs leak into SSR'd <head>

Router

## Summary

Using the rsbuild, importing CSS via `?url` emits the CSS file **twice** — once in `dist/client/assets/css/` and again in `dist/server/static/css/` — and both URLs end up in the SSR-rendered `<head>`.

- **dev**: both requests succeed (rsbuild's dev server serves both paths), but the page loads the same stylesheet twice with two HTTP round-trips.
- **prod**: the `/static/...` URL 404s under the standard `srvx --prod -s dist/client …` start command, because only `dist/client` is exposed.

Reproduces with or without `rsc.enabled`.

**After rsbuild build**:

dist/client/assets/css/styles.<hash>.css   ✅ byte-identical
dist/server/static/css/styles.<hash>.css   ✅ byte-identical, unused at runtime

**Rendered HTML**:

```html
<link rel="stylesheet" href="/static/css/styles.<hash>.css" data-precedence="default">
<link rel="stylesheet" href="/assets/css/styles.<hash>.css" data-precedence="default">
```

https://github.com/ploskovytskyy/reproduce-start-rsbuild

1. clone the repo and `pnpm install`
2. run `pnpm dev:rsbuild`
3. open localhost and check <head> in browser inspector, there are two <link> for styles

Optional:
1. run `pnpm build:rsbuild`
2. run `pnpm start` - it starts prod server
3. now in the browser second styles request will fail with 404 since we only expose client assets

### Expected behavior

A single `<link rel="stylesheet">` in the rendered head, pointing at a file served from the client assets directory in both dev and prod.

### Screenshots or Videos

_No response_

### Platform

- Bundler: [e.g. vite]

## Root cause

createRsbuildEnvironmentPlan in [packages/start-plugin-core/src/rsbuild/planning.ts](https://github.com/TanStack/router/blob/main/packages/start-plugin-core/src/rsbuild/planning.ts) defines two independent Rspack compilations:

Client env (lines 121-138) explicitly sets distPath.{css, cssAsync, svg, font, wasm, image, media, assets} under RSBUILD_CLIENT_ASSETS_DIR and assetPrefix: opts.publicBase.

Server env (lines 168-177) sets only distPath.root. Everything else falls back to rspack defaults — cssFilename = static/css/[name].[contenthash:8].css and no assetPrefix.

Each compilation runs its own CSS extract pipeline and produces its own ?url resolution. Since the source CSS is identical, the content hashes match, but the URLs differ (/assets/css/... vs /static/css/...). SSR picks up the server URL; client hydration produces a fresh <link> for the client URL.

This is rspack-specific. Affects any asset imported via ?url in code that runs on both server and client (CSS, SVG, fonts, images). CSS is just the most visible because of the broken <link> in prod and the doubled payload in dev.

## Workaround
A top-level `output.distPath` override cascades into both environments and aligns the URLs:

```ts
// rsbuild.config.ts
export default defineConfig({
  output: {
    distPath: {
      css: "assets/css",
      // cssAsync: "assets/css/async", // if you use async CSS chunks
    },
  },
  plugins: [pluginReact(), tanstackStart()],
});
```

The duplicate file still gets emitted in dist/server/assets/css/ (unused), but the SSR'd URL and the client URL match.

## Proposed fix

Apply the same mirroring inside createRsbuildEnvironmentPlan's server env.
Leaves the duplicate emit on disk but removes the visible bug.

For example, here is my branch with the patch and a draft commit: https://github.com/ploskovytskyy/router/tree/fix/start-rsbuild-client-assets-duplicate

Image

SyMind:
Fix https://github.com/TanStack/router/issues/7543

- Align Rsbuild server public asset output paths with the client build so SSR `css?url` imports resolve to client-served asset URLs.
- Add an `rsc-rsbuild` e2e route that imports a stylesheet with `?url` and verifies the actual element style is applied.
- Polish the `rsc-rsbuild` fixture homepage and node_modules client component page to match the main RSC fixture’s server/client visual style.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SSR stylesheet asset URLs for `?url` imports so server-rendered pages load the same public assets as the client.
* **New Features**
  * Added an end-to-end route showcasing CSS imported via `?url` (including injection into the document head).
  * Enhanced the demo UI with richer styling, navigation, and clearer server/client/SSR sections.
* **Tests**
  * Added Playwright coverage to validate page rendering and confirm SSR-applied `?url` styles.
* **Chores**
  * Updated TypeScript typings for `*.css?url` imports.


ploskovytskyy:
Thank you very much for such a quick fix!

e2e/react-start/rsc-rsbuild/src/vite-env.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*?url' {
```
schiller-manuel: should we rename that file? :)

schiller-manuel: or even better, could rsbuild provide client types like vite does? 

https://vite.dev/guide/features#client-types

PR #7594, opened 2026-06-10
fix(router-core): encode URL-unsafe ASCII chars in encodePathLikeUrl to break infinite redirect on `<test`-style paths

### LINKED ISSUE #7587: Infinite redirect loop ("ERR_TOO_MANY_REDIRECTS") caused by encoded unsafe characters in URL pathname

Start

### Describe the bug

Hey!

When a URL pathname contains encoded unsafe characters (e.g `<`, `>`, `"`, `//`, ...), an infinite redirect loop is started and the browser crashes with an `ERR_TOO_MANY_REDIRECTS` error.

This happens in both basic paths and paths with parameters.

Any blank project initialized with @tanstack/cli is sufficient for reproducing the issue—no additional code is needed

1. Create a TanStack Start project from scratch. You don't need to add anything; just start the dev server and proceed to step 2
3. Open ``http://localhost:3000/<test``

### Expected behavior

I think it should throw a 404 error

### Screenshots or Videos

Image
Image

### Platform

- Router / Start Version: Router v1.170.15 / React-Start v1.168.25
- OS: Windows 11 - Ubuntu WSL2
- Browser: Brave (chrominium)
- Browser Version: v1.91.168
- Bundler: Vite
- Bundler Version: v8.0.16

### Additional context

_No response_

spokodev:
## What

Requests whose pathname contains encoded URL-unsafe ASCII characters (e.g. `<`, `>`, `"`, `` ` ``, `{`, `}`) crashed with `ERR_TOO_MANY_REDIRECTS` because the router kept rewriting them between encoded and decoded forms. Visiting `http://localhost:3000/<test` on any fresh TanStack Start project would loop until the browser gave up. This PR makes `encodePathLikeUrl` percent-encode the WHATWG URL "path percent-encode set" so its output is a fixed point of `decodePath`, and the SSR redirect comparator no longer fires on these inputs.

Closes #7587.

## Root cause

The SSR pipeline does three things to incoming URLs:

1. `decodePath` (in `packages/router-core/src/utils.ts`) strips control characters and decodes percent sequences via `decodeURI`. For `/%3Ctest` this returns `"/<test"`.
2. `parseLocation` (router fast path) keeps `publicHref` as the raw pathname (`"/%3Ctest"`) and stores the decoded form in `pathname` (`"/<test"`).
3. `buildLocation` rebuilds the URL from the decoded `pathname` and runs the assembled path through `encodePathLikeUrl`, which previously only re-encoded whitespace and non-ASCII characters. For `<` (an ASCII char in the WHATWG path percent-encode set) it returned the input verbatim, so the new `publicHref` was `"/<test"`.

The redirect detector in `router.ts` then compared:

```ts
if (this.latestLocation.publicHref !== nextLocation.publicHref) {
  throw redirect({ href, ... })
}
```

`"/%3Ctest" !== "/<test"` so the router threw a redirect to `/<test`. The browser percent-encoded the response Location header back to `/%3Ctest` and re-requested, which produced the same diff. The loop terminated only when Chrome stopped at `ERR_TOO_MANY_REDIRECTS`.

The clean fix is to make `encodePathLikeUrl` invertible against `decodePath` for the characters `decodePath` decodes. The WHATWG URL spec defines the path percent-encode set as: C0 control + space + `"` + `<` + `>` + `` ` `` + `{` + `}`. Whitespace and the control range were already covered; this PR adds `<`, `>`, `"`, `` ` ``, `{`, `}`. `?` and `#` are deliberately not added because the function is called on the combined `pathname + search + hash` string where those characters are the component separators.

## Tests added

A new `describe('URL-unsafe ASCII chars (WHATWG path percent-encode set)')` block in `packages/router-core/tests/utils.test.ts`. Each `it` targets a specific contract:

- **`encodes < > " ` { } in path segments`** — happy path for every character in the added encode set. Without the fix, all five assertions fail with `expected "/<test" to be "/%3Ctest"` etc.
- **`round-trips with decodePath for unsafe ASCII chars`** — guards the actual invariant used by the SSR comparator (`encodePathLikeUrl(decodePath(raw).path) === raw`) across three inputs.
- **`does not double-encode already-percent-encoded sequences`** — negative regression so the new regex does not match `%` literals.
- **`preserves URL component separators (?, #, /)`** — confirms `?`, `#`, and `/` still pass through, so search/hash survive when `encodePathLikeUrl` is called on a combined path string.

The existing `encodePathLikeUrl` happy-path tests (unicode, spaces, emoji, `[1]` brackets) still pass unchanged, so the fix does not widen the encode set beyond what the WHATWG spec mandates.

Verified the failing-test -> fix -> passing-test loop locally:

- Before the fix: 3 of the new tests fail with the exact mismatch above.
- After the fix: all 116 tests in `packages/router-core/tests/utils.test.ts` pass (`113 passed | 3 expected fail`, no regressions in any existing case).

## Repro

Before the fix, on any TanStack Start project (the issue's reproducer):

1. `npx @tanstack/cli@latest new start-basic && cd start-basic`
2. `pnpm dev`
3. Visit `http://localhost:3000/<test`
4. Browser crashes with `ERR_TOO_MANY_REDIRECTS` within seconds.

Direct repro at the utility layer (no Start app required):

```ts
import { decodePath, encodePathLikeUrl } from '@tanstack/router-core'

const raw = '/%3Ctest'
const decoded = decodePath(raw).path           // "/<test"
const reencoded = encodePathLikeUrl(decoded)   // before: "/<test"   after: "/%3Ctest"
console.log(reencoded === raw)                 // before: false      after: true
```

The `reencoded === raw` invariant is exactly what the SSR redirect comparator relies on - the boolean flip is what stops the redirect loop.

## Validation

```bash
pnpm vitest run tests/utils.test.ts                           # 113 passed | 3 expected fail, 0 failures
pnpm vitest run tests/utils.test.ts -t "encodePathLikeUrl"    # 10 passed (6 existing + 4 new)
```

Pre-existing import resolution errors in unrelated test files (`build-location.test.ts`, `callbacks.test.ts` etc. fail to resolve `@tanstack/history` regardless of this branch — confirmed by running them on `main` with the change stashed).

Changeset added as a `patch` bump on `@tanstack/router-core`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an infinite redirect loop that occurred when request URLs contained special ASCII characters such as `<`, `>`, `"`, `` ` ``, `{`, and `}`.


spokodev:
Closing this - measured against current main, it no longer fixes anything, and it would introduce a new mismatch.

#7587 was fixed differently: `sanitizePathSegment` re-encodes `" < > ` { }` after `decodeSegment`, so `decodePath('/%3Ctest')` now returns `/%3Ctest` unchanged and the round trip already holds. With this PR applied a **literal** `<` goes the other way - `decodePath('/<test')` -> `/<test`, then `encodePathLikeUrl` -> `/%3Ctest` - so it breaks a case that currently works.

Diagnosis for #8041 in a separate comment there.

PR #7608, opened 2026-06-11
fix(solid-router): hydration mismatch for ssr:false routes with pendingComponent

### LINKED ISSUE #7283: [Solid] ssr:false route with pendingComponent fails during dev hydration
### Summary

`ssr: false` routes with a `pendingComponent` still fail during dev hydration in the Solid Start selective SSR example.

This looks closely related to #7085 / #7266. #7266 fixed the `ssr: 'data-only'` case, but the same class of failure still reproduces for `ssr: false` in dev mode.

### Reproduction

MRE branch:

https://github.com/ljho01/router/tree/codex/solid-ssr-false-pending-mre

Relevant added route:

https://github.com/ljho01/router/blob/codex/solid-ssr-false-pending-mre/e2e/solid-start/selective-ssr/src/routes/ssr-false-pending-component.tsx

The route is intentionally minimal:

```tsx
export const Route = createFileRoute('/ssr-false-pending-component')({
  ssr: false,
  loader: async () => {
    await new Promise((resolve) => setTimeout(resolve, 1500))
    return { loadedAt: new Date().toISOString() }
  },
  pendingComponent: () => (
    <div data-testid="ssr-false-pending-component-pending" />
  ),
  component: SsrFalsePendingComponentRoute,
})
```

I tested against upstream main at `d6decca41807e9ca28279e2db6640e7a8bdc1229`.

### Steps

```sh
pnpm install --frozen-lockfile
VITE_SERVER_PORT=57731 pnpm --dir e2e/solid-start/selective-ssr dev:e2e --host localhost --port 57731 --strictPort
```

Then open:

```txt
http://localhost:57731/ssr-false-pending-component
```

### Actual behavior

The route never reaches the loaded component. It stays around the root HTML and the browser console reports:

```txt
Warning: The following error wasn't caught by any route! At the very least, consider setting an 'errorComponent' in your RootRoute!
Warning: template is not a function
```

In my Playwright check, the pending node was present but the ready label timed out after 10s:

```json
{
  "pending": 1,
  "readyText": "TIMEOUT: locator.textContent: Timeout 10000ms exceeded.",
  "bodyText": "Selective SSR E2E Test\nHome\nroot\nssr: \"undefined\"\nexpected data location execution:\nloader: server\ncontext: server"
}
```

The existing `ssr: 'data-only'` route in the same example works correctly under the same dev server.

### Expected behavior

The `ssr: false` route should behave like the fixed `ssr: 'data-only'` case: hydrate without adding a mismatched outer pending fallback and eventually render the loaded component with no browser errors.

### Extra notes

The production e2e target passes, so this appears to be dev-mode specific:

```sh
CI=1 NX_DAEMON=false pnpm nx run tanstack-solid-start-e2e-selective-ssr:test:e2e --outputStyle=stream --skipRemoteCache -- tests/pending-component-hydration.spec.ts
# 2 passed
```

The suspicious branch is in `packages/solid-router/src/Match.tsx`:

```tsx
const resolvedNoSsr =
  currentMatchState().ssr === false ||
  currentMatchState().ssr === 'data-only'

const shouldSkipSuspenseFallback =
  (isServer ?? router.isServer)
    ? resolvedNoSsr
    : currentMatchState().ssr === 'data-only'
```

Locally, changing the client-side condition to also use `resolvedNoSsr` made both the `data-only` and `ssr:false` MRE routes pass in dev:

```tsx
const shouldSkipSuspenseFallback = resolvedNoSsr
```

I did not include that fix in the MRE branch; the branch is left failing for reproduction.

### Environment

- OS: macOS
- Node: v22.17.1
- pnpm: 10.28.0
- Browser check: Playwright Chromium

brenelz:
Fixes #7283

## Problem

`ssr: false` routes with a `pendingComponent` fail during dev hydration with `template is not a function` and never render the loaded component.

Selective SSR routes render the pending component *inside* the suspense boundary on the server and skip the suspense fallback. On the client, #7266 only skipped the fallback for `ssr: 'data-only'`, so hydrating an `ssr: false` route rendered a suspense fallback with no matching server markup — a hydration mismatch.

## Solution

The naive fix suggested in the issue (skipping the fallback for all selective SSR routes on the client too) regresses client-side navigation: the existing unit test `Router.preload="intent", pendingComponent renders during unresolved route loader` fails because the pending UI never shows when navigating to an `ssr: false` route in the browser.

Instead, skip the fallback only when it would conflict with server markup: on the server, and on the client *while hydrating* (detected via solid-js's `sharedConfig.context`, which is only set during hydration). Client-side navigations and pure CSR keep their pending fallback. As a side effect this also restores the client-nav pending fallback for `data-only` routes, which #7266 had over-broadly removed.

## Testing

- Reproduced the bug in the selective-ssr example dev server before the fix; after the fix both the `ssr: false` and `data-only` pending-component routes hydrate cleanly with zero browser errors and reach the loaded state.
- All 813 solid-router unit tests pass, including the client-nav pending test that the naive fix broke.
- Full selective-ssr e2e suite passes (10/10), including a new `ssr: false` regression test mirroring the issue's MRE.

Note: the e2e suite runs against a production build where the original bug didn't reproduce, so the new test guards the markup contract; the dev-mode failure was verified manually as described above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed hydration mismatches for routes configured with `ssr: false` and `pendingComponent`. Routes now skip the suspense fallback during server-side hydration while properly displaying the pending component on client-side navigation.

* **Tests**
  * Added end-to-end test coverage for pending component hydration behavior on selective SSR routes.

### DIFF
--- a/.changeset/lucky-windows-shake.md
+++ b/.changeset/lucky-windows-shake.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/solid-router': patch
+---
+
+Fix hydration mismatch for `ssr: false` routes with a `pendingComponent`. The suspense fallback is now skipped only while hydrating server-rendered markup, so selective SSR routes hydrate cleanly while client-side navigations keep showing their pending component.
--- a/e2e/solid-start/selective-ssr/src/routeTree.gen.ts
+++ b/e2e/solid-start/selective-ssr/src/routeTree.gen.ts
@@ -9,11 +9,18 @@
 // Additionally, you should also exclude this file from your linter and/or formatter to prevent it from being checked or modified.
 
 import { Route as rootRouteImport } from './routes/__root'
+import { Route as SsrFalsePendingComponentRouteImport } from './routes/ssr-false-pending-component'
 import { Route as PostsRouteImport } from './routes/posts'
 import { Route as DataOnlyPendingComponentRouteImport } from './routes/data-only-pending-component'
 import { Route as IndexRouteImport } from './routes/index'
 import { Route as PostsPostIdRouteImport } from './routes/posts.$postId'
 
+const SsrFalsePendingComponentRoute =
+  SsrFalsePendingComponentRouteImport.update({
+    id: '/ssr-false-pending-component',
+    path: '/ssr-false-pending-component',
+    getParentRoute: () => rootRouteImport,
+  } as any)
 const PostsRoute = PostsRouteImport.update({
   id: '/posts',
   path: '/posts',
@@ -40,12 +47,14 @@ export interface FileRoutesByFullPath {
   '/': typeof IndexRoute
   '/data-only-pending-component': typeof DataOnlyPendingComponentRoute
   '/posts': typeof PostsRouteWithChildren
+  '/ssr-false-pending-component': typeof SsrFalsePendingComponentRoute
   '/posts/$postId': typeof PostsPostIdRoute
 }
 export interface FileRoutesByTo {
   '/': typeof IndexRoute
   '/data-only-pending-component': typeof DataOnlyPendingComponentRoute
   '/posts': typeof PostsRouteWithChildren
+  '/ssr-false-pending-component': typeof SsrFalsePendingComponentRoute
   '/posts/$postId': typeof PostsPostIdRoute
 }
 export interface FileRoutesById {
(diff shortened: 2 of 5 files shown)


birkskyum (approved):

brenelz:
Superseeded by this PR and confirmed with AI it will be fixed

https://github.com/TanStack/router/pull/7744

PR #7612, opened 2026-06-12
fix(react-router): re-export PathParamError from framework packages

### LINKED ISSUE #7545: `error.routerCode` removed and not a clear replacement for `error.routerCode === 'PARSE_PARAMS'` or `error.routerCode === 'VALIDATE_SEARCH'`

Router

### Describe the bug

In commit [`2f53749`](https://github.com/TanStack/router/commit/2f5374945e2138559a51464f45a5152eae67e1dd) (shipped in `@tanstack/react-router@1.170.10`), `error.routerCode` was removed from the error passed to a route's `onError` handler. Previously, search/param validation failures could be detected like this:

```ts
  component: RouteComponent,
  params: pathParamsSchema, // Using Zod (v4) schema
  validateSearch: searchParamsSchema, // Using Zod (v4) schema
  ....
  onError: (error) => {
    if (error.routerCode === 'PARSE_PARAMS') throw redirect({ to: '/parse-error' })
    if (error.routerCode === 'VALIDATE_SEARCH') throw redirect({ to: '/validation-error' })
    throw error
  }
}
```

This now silently breaks because `routerCode` no longer exists. I would say this change is more of a breaking change as the contract has changed, but I'm unsure if it is intentional or just a bug.

Looking at the code base (guessing the change is intentionally) a possible solution to have the minimum changes and exact functionality could be:

```ts
onError: (error) => {
  if (error instanceof PathParamError) throw redirect({ to: '/' })
  if (error instanceof SearchParamError) throw redirect({ to: '/' })
  throw error
}
```

However,`@tanstack/react-router` only re-exports `SearchParamError`, not `PathParamError` it is only exported from `@tanstack/router-core`, forcing consumers to import from a transitive dependency (maybe is not but most likely yes):

```ts
import { SearchParamError } from '@tanstack/react-router'; // ✅ works
import { PathParamError } from '@tanstack/react-router';   // ❌ not exported
import { PathParamError } from '@tanstack/router-core';    // works, but most likely reaching into a transitive dep
```

### Expected behavior

Either have `routerCode` back or re-export `PathParamError` from `@tanstack/react-router` alongside `SearchParamError`, so consumers don't have to import from `@tanstack/router-core` (which most likely is going to be a transitive dep).

If that's not correct then docs should explain the proper way to differentiate between a parse error and a validation error is provided.

### Platform

- Router / Start Version: `@tanstack/react-router` 1.170.11 (`@tanstack/router-core` 1.171.9, `@tanstack/router-plugin` 1.168.14)
- OS: Any
- Browser: Any
- Browser Version: Any
- Bundler: vite
- Bundler Version: 8.0.16

### Additional context

Maybe the usage of `error instanceOf PathParamError` or `error instanceOf SearchParamError` is not a good solution so in that case what would be a good solution?

Side note: the documentation still references the removed `routerCode` in [Search Params guide](https://tanstack.com/router/latest/docs/framework/react/guide/search-params#validating-search-params)

LeSingh1:
## Summary

Closes #7545.

`PathParamError` is exported from `@tanstack/router-core` but was absent from the re-export lists in `@tanstack/react-router`, `@tanstack/solid-router`, and `@tanstack/vue-router`. `SearchParamError` was already re-exported on that same line — `PathParamError` was simply never added alongside it.

The practical impact: users who want to distinguish path-param validation failures from other errors in an `onError` handler or `errorComponent` had to write:

```ts
import { PathParamError } from '@tanstack/router-core' // transitive dep, not a public package
```

After this fix they can write:

```ts
import { PathParamError } from '@tanstack/react-router'
// or @tanstack/solid-router / @tanstack/vue-router
```

## Changes

- `packages/react-router/src/index.tsx` — add `PathParamError` to the existing `SearchParamError` re-export
- `packages/solid-router/src/index.tsx` — same
- `packages/vue-router/src/index.tsx` — same
- `packages/react-router/tests/router.test.tsx` — new test verifying that a `params.parse` failure surfaces as `PathParamError` in the route's `errorComponent`, imported directly from `@tanstack/react-router`

## Test plan

- `pnpm --filter @tanstack/react-router test:unit` — all 44 test files pass (913 tests, +1 from baseline)
- No type errors

## Summary by CodeRabbit

* **New Features**
  * `PathParamError` is now exported from React, Solid, and Vue router packages, enabling developers to handle path parameter validation errors more effectively.

* **Tests**
  * Added test coverage for path parameter validation error handling.

### DIFF
--- a/packages/react-router/src/index.tsx
+++ b/packages/react-router/src/index.tsx
@@ -275,7 +275,7 @@ export type {
 
 export { createRouter, Router } from './router'
 
-export { lazyFn, SearchParamError } from '@tanstack/router-core'
+export { lazyFn, SearchParamError, PathParamError } from '@tanstack/router-core'
 
 export { RouterProvider, RouterContextProvider } from './RouterProvider'
 export type { RouterProps } from './RouterProvider'
--- a/packages/react-router/tests/router.test.tsx
+++ b/packages/react-router/tests/router.test.tsx
@@ -12,6 +12,7 @@ import { composeRewrites, notFound } from '@tanstack/router-core'
 import {
   Link,
   Outlet,
+  PathParamError,
   RouterProvider,
   SearchParamError,
   createBrowserHistory,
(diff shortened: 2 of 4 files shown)


yunarch:
@LeSingh1 still working on this? or is ready to be merged? I can take over if not 👍

Issue #7618, opened 2026-06-12
TanStack Start (RSC): route component does not HMR when its file has a co-located createServerFn


sreetamdas:
### Which project does this relate to?

Start

### Describe the bug

In RSC mode, a route component does **not** Hot Module Reload when its route file *also* contains a co-located `createServerFn` (e.g. a loader server function in the same file as the route component).

On editing the route component's JSX, the terminal shows only:

```
(rsc) hmr update /src/routes/index.tsx?tss-serverfn-split
```

…with **no client update**, so the DOM never changes — and there is no full reload either. Moving the `createServerFn` into a separate file restores Fast Refresh, which is the tell that this is specifically about *co-location*.

I've traced the root cause to `@vitejs/plugin-rsc` and opened a proposed fix upstream: **vitejs/vite-plugin-react#1248**. Filing here for visibility/tracking (so Start users hitting "my route won't HMR" find the cause + workaround, and so the `@vitejs/plugin-rsc` bump is tracked once it lands).

**Root cause (short version):** the route file's `createServerFn` puts the file in the `rsc` module graph. `@vitejs/plugin-rsc`'s client `hotUpdate` returns `[]` for any file in the rsc graph that isn't inside a `"use client"` boundary (a guard to avoid full reloads from server-only files watched as style deps, e.g. Tailwind/`addWatchFile`). The route component is a genuine client module (imported by the client route tree) but isn't a `"use client"` reference, so its client HMR is wrongly suppressed.

**Workaround:** move the `createServerFn`/server code into a separate (e.g. `*.server.ts`) module and import it into the route file.

### Complete minimal reproducer

https://github.com/sreetamdas/tanstack-rsc-hmr-co-located-serverfn-repro

### Steps to Reproduce the Bug

1. `pnpm install && pnpm dev` and open http://localhost:3000.
2. Click **Increment** once (so `Count: 1`) to seed client state.
3. In `src/routes/index.tsx`, edit the `<h1 data-testid="marker">` text (`edit-me-baseline` → anything) and save.
4. Observe: the heading does **not** update; the terminal logs only `(rsc) hmr update …?tss-serverfn-split`; no full reload.
5. For contrast: move `getServerData` into a separate file and import it — HMR now works (heading updates, counter preserved).

### Expected behavior

Editing a route component's JSX should Fast Refresh the component in dev, regardless of whether the route file also defines a `createServerFn`. Server-only code co-located in the file should not strip the route component's client HMR boundary.

### Platform

[...]

PR #7621, opened 2026-06-12
test(rsc): HMR e2e for a route component co-located with a server fn (canary for #7618 / vitejs/vite-plugin-react#1248)

sreetamdas:
### Description

Adds a dev-mode HMR e2e for a TanStack Start **RSC** route whose file **co-locates a `createServerFn` with the route component**, asserting the component Fast Refreshes on edit (and preserves local state).

This is intentionally opened as a **draft / CI canary**: the test **fails on `main` today** because it reproduces a real bug — editing such a route component emits only an `(rsc) hmr update …?tss-serverfn-split` with no client update, so the DOM never changes. It will go **green** once the upstream `@vitejs/plugin-rsc` fix lands and the dependency is bumped. So the PR's CI status is a direct signal for "is the issue resolved yet?".

- Bug + analysis: **#7618**
- Standalone minimal repro: https://github.com/sreetamdas/tanstack-rsc-hmr-co-located-serverfn-repro
- Root cause + proposed upstream fix: **vitejs/vite-plugin-react#1248** (the client `hotUpdate` guard returns `[]` for any file in the rsc graph that isn't a `"use client"` boundary — which wrongly catches a genuine client route component co-located with a server fn).

**What's here**
- `e2e/react-start/rsc/src/routes/rsc-hmr-serverfn.tsx` — fixture: `createServerFn` + route component (marker + counter) in one file.
- `e2e/react-start/rsc/tests-hmr/rsc-hmr.spec.ts` — edits the marker, asserts it updates and the counter survives (HMR, not reload).
- `e2e/react-start/rsc/playwright.hmr.config.ts` — a **dev-mode** Playwright lane (the existing rsc e2e runs the built/prod server, which can't exercise HMR), plus a `test:e2e:hmr` script.

**Needs guidance**
The new lane isn't wired into the nx/CI e2e matrix yet — HMR needs a dev server, and the existing dev-mode `hmr` e2e is non-RSC. Happy to integrate it wherever you'd prefer (extend the `hmr` suite to RSC, or register this as its own CI e2e project). Flagging since it won't run in CI as-is.

### DIFF
--- a/e2e/react-start/rsc/package.json
+++ b/e2e/react-start/rsc/package.json
@@ -13,7 +13,8 @@
     "build:vite": "vite build && tsc --noEmit",
     "preview": "vite preview",
     "start": "node server.js",
-    "test:e2e-full": "pnpm build && sh -c 'rm -f \"port-${E2E_PORT_KEY:-$npm_package_name}.txt\" \"port-${E2E_PORT_KEY:-$npm_package_name}-external.txt\"' && playwright test --project=chromium"
+    "test:e2e-full": "pnpm build && sh -c 'rm -f \"port-${E2E_PORT_KEY:-$npm_package_name}.txt\" \"port-${E2E_PORT_KEY:-$npm_package_name}-external.txt\"' && playwright test --project=chromium",
+    "test:e2e:hmr": "playwright test --config playwright.hmr.config.ts --project=chromium"
   },
   "nx": {
     "metadata": {
--- a/e2e/react-start/rsc/playwright.hmr.config.ts
+++ b/e2e/react-start/rsc/playwright.hmr.config.ts
@@ -1,0 +1,40 @@
+import { defineConfig, devices } from '@playwright/test'
+import { getTestServerPort } from '@tanstack/router-e2e-utils'
+import packageJson from './package.json' with { type: 'json' }
+
+// Dedicated HMR test lane for the RSC app. Unlike playwright.config.ts (which
+// runs the built server via `pnpm start`), HMR must run against the dev server.
+const e2ePortKey = process.env.E2E_PORT_KEY ?? `${packageJson.name}-hmr`
+const PORT = await getTestServerPort(e2ePortKey)
+const baseURL = `http://localhost:${PORT}`
+
+export default defineConfig({
+  testDir: './tests-hmr',
+  workers: 1,
+  reporter: [['line']],
+
+  use: {
+    baseURL,
+  },
+
+  webServer: {
+    command: `pnpm dev:vite --port ${PORT}`,
+    url: baseURL,
+    reuseExistingServer: !process.env.CI,
+    stdout: 'pipe',
+    timeout: 120_000,
+    env: {
+      PORT: String(PORT),
+      E2E_PORT_KEY: e2ePortKey,
+    },
+  },
+
+  projects: [
+    {
+      name: 'chromium',
+      use: {
+        ...devices['Desktop Chrome'],
+      },
+    },
+  ],
+})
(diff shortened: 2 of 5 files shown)


schiller-manuel:
hmr is notorously hard to test in CI, so please check out how the `e2e/react-start/hmr` ensures that CI env with low compute and high latencies still passes.. we might just create a new e2e/react-start/rsc-hmr project for this  

would be best if you create multiple routes, one with css modules, one without any css, etc.
just to ensure a broad coverage of possible scenarios

also please configure this to run in both vite and rsbuild so we ensure both build tool adapters work as expected

sreetamdas:
@schiller-manuel thank you for the response! I'll see how the test can be updated - in the meantime, I wanted to clarify that the real fix here is https://github.com/vitejs/vite-plugin-react/pull/1248, this PR is really just to highlight the current breakage. 

Once the fix is merged upstream, the idea is to see the test here turn green — hope that makes sense! 

btw I checked locally on the rsbuild setup, and rsbuild **doesn't** seem to have this issue

sreetamdas:
Status update — the upstream fix has landed, so this canary has served its purpose.

vitejs/vite-plugin-react#1248 was merged and released in [`@vitejs/plugin-rsc@0.5.30`](https://www.npmjs.com/package/@vitejs/plugin-rsc/v/0.5.30). hi-ogawa replaced my e2e with his own `examples/client-first` + `e2e/client-first.test.ts`, which he described as closer to TanStack's client-first RSC architecture.

I've closed #7618 with the root-cause writeup, and opened #7900 to raise the `@tanstack/react-start-rsc` peer floor to `>=0.5.30`. That matters more than it looks: `pnpm-lock.yaml` here pinned `0.5.20`, so the workspace was installing an affected version regardless of the `>=0.5.20` range.

@schiller-manuel on your review — I'd like to do it properly rather than reshape this branch, and a few things you flagged turned out better than expected:

- **Dual toolchain is viable.** I verified rsbuild + RSC + a dev server does Fast Refresh (state preserved, and a `window` global set before the edit survived, so it was HMR and not a reload). Worth noting the vite lane is the only one that can regress *this* bug — `@vitejs/plugin-rsc` is Vite-only, and the rsbuild path goes through Rspack/`react-server-dom-rspack` — so the rsbuild lane is adapter-parity coverage rather than a guard on vitejs/vite-plugin-react#1248.
- **`e2e/react-start/hmr` is the right harness template**, as you suggested: it's already `vite-ssr` / `rsbuild-ssr` dual-toolchain in dev mode, and `waitForServer` + `preOptimizeDevServer` + `createHmrFileEditor` (with `normalizeSource` self-healing dirty fixtures) are exactly the low-compute/high-latency protections you pointed me at. `e2e/react-start/rsc` supplies the `rsc: { enabled: true }` configs. So `rsc-hmr` = `hmr`'s harness + `rsc`'s configs, serial (`workers: 1`, no shards, since the fixtures are mutated on disk).
- **Fixture matrix** I'm planning, per your ask: co-located server fn with no CSS (the original bug), co-located + CSS module, a separate-file control, and a CSS-only edit asserting styles still hot-swap without a full reload. That last pair is the important one — the upstream guard is built entirely on classifying importers as CSS vs non-CSS, so that's the interaction most likely to regress. I hand-checked both against 0.5.30 and they pass.

One question before I put it up, since it affects how you'd want it wired: the nx playwright plugin's mode allowlist is `['ssr','spa','prerender','preview']` with no `dev`, which is why `hmr` declares `mode: 'ssr'` and then starts a dev server and ignores the built output. Happy to follow that same convention for `rsc-hmr`, or wire it differently if you'd prefer.

I'll open `rsc-hmr` as a fresh PR against current `main` rather than force-pushing over this branch — it's six weeks stale and conflicts now, and none of its files survive the rewrite. Closing this one once that's up, unless you'd rather I kept the thread here.

sreetamdas:
Superseded by #7901, which is the `e2e/react-start/rsc-hmr` project from @schiller-manuel's review above: separate project, `hmr`'s CI hardening, a CSS-module route and a no-CSS route, and both the vite and rsbuild adapters.

Closing this rather than force-pushing over it — the branch is six weeks stale and none of its files survived the rewrite.

Worth recording the result from #7901 here, since it is what this canary was originally meant to demonstrate. Pinning that project to the broken `@vitejs/plugin-rsc@0.5.27` fails exactly the two co-located component-edit tests, with the original symptom (only `(rsc) hmr update …?tss-serverfn-split`, no client update), while the CSS-only edit and the separate-file control stay green. On `0.5.30` all four pass on both adapters.

PR #7661, opened 2026-06-21
feat: add `isBot` option to `renderRouterToStream` for streaming SSR bot detection

### LINKED ISSUE #7660: Streaming SSR: make the `isbot` "buffer full HTML for bots" behavior configurable (opt-in / opt-out)

Router — the logic lives in `@tanstack/react-router`'s SSR server (`renderRouterToStream`) and is used by `defaultStreamHandler` / `createStartHandler`, so Start is affected too.

### Summary

`renderRouterToStream` branches on `isbot(User-Agent)` with no way to override or disable it ([`packages/react-router/src/ssr/renderRouterToStream.tsx`](https://github.com/TanStack/router/blob/main/packages/react-router/src/ssr/renderRouterToStream.tsx#L67-L68), and L153–160):

```ts
// renderToReadableStream branch
if (isbot(request.headers.get('User-Agent'))) {
  await waitForReadyOrAbort(stream.allReady, request.signal)
}

// renderToPipeableStream branch
...(isbot(request.headers.get('User-Agent'))
  ? { onAllReady() { pipeable.pipe(reactAppPassthrough) } }
  : { onShellReady() { pipeable.pipe(reactAppPassthrough) } })
```

Any UA `isbot` matches gets the **buffered** path (whole document withheld until React's `allReady`); everyone else gets **progressive streaming**. Reasonable as a default, but it's **undocumented and not configurable** — neither `defaultStreamHandler` nor `createStartHandler(...)(defaultStreamHandler)` expose a hook, and `isbot` is a hard dependency.

### Why this is a problem

`isbot` classifies **synthetic performance tools** as bots — Lighthouse (`Chrome-Lighthouse`), PageSpeed Insights (`pagespeed`), WebPageTest (`PTST`), Pingdom, SpeedCurve. So:

1. **Lab audits measure a different code path than real users.** Lighthouse/PSI hit the buffered `onAllReady` branch; real browsers hit `onShellReady`. The lab score stops representing the streaming experience real users (and field/CrUX) get.
2. **TTFB inflation on the buffered path.** The response is withheld until the slowest loader resolves, so FCP/LCP/Speed Index in those tools are gated behind full data resolution — frequently scoring *worse* than reality.
3. **No way to pick a different policy.** Teams may legitimately want to always stream (their crawlers are JS-capable and they prefer fast TTFB), exclude only perf tools, or supply a custom predicate (custom UA list, verified-crawler header, IP tags).

### Current workaround (hacky)

The only opt-out today is wrapping the handler and stripping the UA so `isbot` returns false:

```ts
const streamHandler: typeof defaultStreamHandler = (opts) =>
  defaultStreamHandler({ ...opts, request: stripUserAgent(opts.request) }) // isbot(null) === false
```

This couples app code to an internal heuristic and breaks silently if detection changes.

### Proposed solution

A configurable hook, defaulting to today's behavior (no breaking change). e.g. on `renderRouterToStream`, surfaced through `defaultStreamHandler` / `createStartHandler`:

```ts
renderRouterToStream({
  request, router, responseHeaders, children,
  // default: (req) => isbot(req.headers.get('User-Agent'))
  isBot: false, // or a predicate: (req) => isbot(...) && !isPerfTool(req)
})
```

Alternatives: a narrower `waitForAllReady?: boolean | ((request: Request) => boolean)`, or a router-level SSR option set once. Either way, keep the `isbot` default but let apps override/disable it.

### Additional context

- Source: `packages/react-router/src/ssr/renderRouterToStream.tsx` (present on `main`; observed in `@tanstack/react-router@1.170.16`).
- Happy to open a PR once a maintainer signals the preferred API shape.

anonrig:
Closes #7660

## Summary

Streaming SSR waits for the entire document (React's `allReady`) for requests that `isbot` flags by their `User-Agent`, and streams the shell first for everyone else. It's a reasonable default, but it isn't configurable — and `isbot` classifies synthetic performance tools (Lighthouse, PageSpeed Insights, WebPageTest, Pingdom, …) as bots, so lab audits get measured on the buffered path instead of the streaming path real users actually receive.

This PR adds an `isBot` option to `renderRouterToStream` so the decision is made in the **server request handler**, **defaulting to the current behavior** (no breaking change):

```tsx
import {
  createRequestHandler,
  renderRouterToStream,
  RouterServer,
} from '@tanstack/react-router/ssr/server'

createRequestHandler({ request, createRouter })(
  ({ request, router, responseHeaders }) =>
    renderRouterToStream({
      request,
      router,
      responseHeaders,
      children: <RouterServer router={router} />,
      // undefined (default): use the built-in isbot(User-Agent) check
      // boolean: force the bot (true) / non-bot (false) path
      // (request) => boolean: custom predicate
      isBot: false,
    }),
)
```

### Why the request handler, not `createRouter`

An earlier revision put this on the router's `ssr` options. That's the wrong home: `createRouter` runs isomorphically, so an `isBot` predicate (and anything it imports) would be bundled into the **client**. `renderRouterToStream` lives in `ssr/server` and only ever runs on the server, so the option — and the user's predicate — stay server-side. It's also available in the Start path, whose `defaultStreamHandler` calls `renderRouterToStream` too.

## Changes

- `@tanstack/react-router`, `@tanstack/solid-router`, `@tanstack/vue-router`: `renderRouterToStream` accepts `isBot?: boolean | ((request: Request) => boolean)`, falling back to `isbot(request.headers.get('User-Agent'))` when unset.
- Tests: new cases in `react-router/tests/renderRouterToStream.test.tsx` covering the default, `true`, `false`, and predicate forms.
- Changeset (minor) for the three adapter packages.

## Verification

- `build` — router-core + all three adapters
- `test:unit` — react/solid/vue
- `test:types` — ts5.5 → ts6.0 (all three adapters)
- `test:eslint` — 0 errors

## New Features

* Added new `isBot` configuration option for streaming server-side rendering to customize how bot detection works
* Override default automatic User-Agent detection with a boolean flag to force specific behavior globally
* Alternatively, supply a custom request-scoped predicate function for flexible per-request bot determination
* Existing default behavior remains fully backward compatible with all applications

### DIFF
--- a/.changeset/ssr-isbot-option.md
+++ b/.changeset/ssr-isbot-option.md
@@ -1,16 +1,34 @@
 ---
-'@tanstack/router-core': minor
 '@tanstack/react-router': minor
 '@tanstack/solid-router': minor
 '@tanstack/vue-router': minor
 ---
 
-feat: add `ssr.isBot` router option to configure streaming SSR bot detection
+feat: add `isBot` option to `renderRouterToStream` to configure streaming SSR bot detection
 
 Streaming SSR waits for the full document (React's `allReady`) for requests that `isbot` flags by their `User-Agent`, and streams the shell first for everyone else. This was hardcoded, so performance auditors (Lighthouse, PageSpeed Insights, WebPageTest, …) — which `isbot` classifies as bots — were measured on the buffered path instead of the streaming path real users get.
 
-The new `ssr.isBot` router option overrides that decision while keeping the current behavior by default:
+`renderRouterToStream` now accepts an `isBot` option so the decision can be made in the server request handler (it lives in `ssr/server` and never ships to the client). The default is unchanged.
 
-- `undefined` (default): unchanged — use the built-in `isbot` User-Agent check.
-- `boolean`: force the bot (`true`) or non-bot (`false`) path for every request.
+- `undefined` (default): use the built-in `isbot` User-Agent check.
+- `boolean`: force the bot (`true`) or non-bot (`false`) path for the request.
 - `(request) => boolean`: provide a custom predicate.
+
+```tsx
+import {
+  createRequestHandler,
+  renderRouterToStream,
+  RouterServer,
+} from '@tanstack/react-router/ssr/server'
+
+createRequestHandler({ request, createRouter })(
+  ({ request, router, responseHeaders }) =>
+    renderRouterToStream({
+      request,
+      router,
+      responseHeaders,
+      children: <RouterServer router={router} />,
+      isBot: false,
+    }),
+)
+```
--- a/.changeset/ssr-isbot-option.md
+++ b/.changeset/ssr-isbot-option.md
@@ -1,0 +1,16 @@
+---
+'@tanstack/router-core': minor
+'@tanstack/react-router': minor
+'@tanstack/solid-router': minor
+'@tanstack/vue-router': minor
+---
+
+feat: add `ssr.isBot` router option to configure streaming SSR bot detection
+
+Streaming SSR waits for the full document (React's `allReady`) for requests that `isbot` flags by their `User-Agent`, and streams the shell first for everyone else. This was hardcoded, so performance auditors (Lighthouse, PageSpeed Insights, WebPageTest, …) — which `isbot` classifies as bots — were measured on the buffered path instead of the streaming path real users get.
+
+The new `ssr.isBot` router option overrides that decision while keeping the current behavior by default:
+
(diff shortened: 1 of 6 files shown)


tannerlinsley:
This uncovered a deeper opportunity for extended control of streaming features across the board. We’re superseding this concept with this PR here: https://github.com/TanStack/router/pull/7714

anonrig:
> This uncovered a deeper opportunity for extended control of streaming features across the board. We’re superseding this concept with this PR here: #7714

Thank you for the update. Can't wait!

PR #7663, opened 2026-06-21
perf(start-client-core): O(1) buffer drain in client frame decoder

anonrig:
## What

Replace the O(n) `bufferList.shift()` in the client-side frame decoder (`packages/start-client-core/src/client-rpc/frame-decoder.ts`) with an O(1) head pointer.

`extractFlattened()` dropped each fully-consumed chunk from the front of `bufferList` with `shift()`. When a single large frame (e.g. a big `RawStream` payload) is assembled from many small network reads, the extract loop calls `shift()` once per chunk — and each `shift()` re-indexes the whole array, so reassembly degrades to **O(n²)**.

This PR tracks the first un-consumed chunk with a `bufferHead` index and advances it in O(1) instead of shifting. `readHeader()` reads from `bufferHead` as well. Consumed slots are released for GC, and the array is compacted:

- **fully drained** (`bufferHead === bufferList.length`) → reset in O(1) (the common terminal state), or
- once the consumed prefix grows past a small threshold → `splice()` it off (amortized O(1) per consumed chunk).

This mirrors the index-pointer approach already used in `transformStreamWithRouter` for the same reason.

## Why

Same hot path as the sibling zero-copy PR: decoding streamed server-function responses and `RawStream` payloads. The O(n²) bites specifically when one frame spans many buffered chunks.

Standalone micro-benchmark (Node, median of 12 runs), draining N buffered chunks:

| Chunks | `shift()` | head pointer | Speedup |
|---|---|---|---|
| 200 | 26.7 ms | 2.5 ms | **10.6x** |
| 1000 | 173 ms | 15 ms | **11.5x** |

## Tests

- Existing `frame-decoder` suite passes.
- Added two tests for the changed paths:
  - a 200-byte `CHUNK` payload delivered **one byte at a time** (forces the header slow path + many whole-chunk consumptions + the fully-drained reset),
  - 100-byte frames fed in 7-byte reads that **never align** with frame boundaries, so the head pointer climbs past the compaction threshold repeatedly (exercises the `splice()` prefix drop).

```
nx run @tanstack/start-client-core:test:unit   -> 20 passed
nx run @tanstack/start-client-core:test:types  -> no errors
```

## Notes

- Independent of the sibling zero-copy PR. Both touch `extractFlattened`, so whichever merges second will need a trivial rebase.
- Includes a changeset (`@tanstack/start-client-core` patch).

## Summary by CodeRabbit

* **Performance**
  * Optimized frame decoder to more efficiently handle large frames when data arrives as multiple small reads, reducing processing overhead.

* **Tests**
  * Added test coverage for frame decoder with fragmented input streams and non-aligned chunk boundaries.


Sheraff:
Can you fix the conflicts on this one? And then I'll merge it

Sheraff:
GitHub could not update this private-fork branch because of the conflict with #7662, so I resolved it on current `main` in replacement PR #8009. The original commit authorship is preserved there.

PR #7664, opened 2026-06-21
perf(router-core): faster enumerable-key collection in structural sharing

anonrig:
## What

Speed up `getEnumerableOwnKeys` — the helper that backs `replaceEqualDeep` (structural sharing) in `packages/router-core/src/utils.ts`.

It previously called `Object.getOwnPropertyNames(o)` and then invoked `propertyIsEnumerable` **once per key** to confirm every own string property is enumerable. This swaps that O(n) loop of JS method calls for `Object.keys(o)` (native, enumerable-only) plus a single length comparison against `getOwnPropertyNames(o)` to detect non-enumerable own string props. Symbol handling is unchanged.

```diff
-  const names = Object.getOwnPropertyNames(o)
-  for (const name of names) {
-    if (!isEnumerable.call(o, name)) return false
-  }
+  const keys = Object.keys(o)
+  if (keys.length !== Object.getOwnPropertyNames(o).length) return false
```

## Why

`replaceEqualDeep` runs on **every selector result on every state update** when `defaultStructuralSharing` is enabled, so `getEnumerableOwnKeys` is called for every plain object it walks. For selector-heavy apps this is a hot client path (it short-circuits on the server, so this is purely a client-side win).

Benchmark on the real function (Vitest `bench`, node env where `isServer` is false), deeply-equal new state object shaped like typical router state (search / params / nested loaderData / context):

| | ops/s |
|---|---|
| before | ~560K |
| after | ~762K |

≈ **1.36× faster** in-situ (an isolated A/B of just the key-collection step measured ~1.55×). The win grows with object key count, since the old path made one `propertyIsEnumerable` call per key.

## Correctness

Behavior is identical:
- Same returned keys in the same order. `Object.keys` and `Object.getOwnPropertyNames` follow the same `[[OwnPropertyKeys]]` ordering (integer indices ascending, then string keys in insertion order), and in the non-bail case every string key is enumerable, so the two orderings coincide.
- Still bails (returns `false`) for objects with any non-enumerable own string property (length mismatch) or any non-enumerable own symbol (existing per-symbol check).

Verified by:
- The existing `getEnumerableOwnKeys behavior (via replaceEqualDeep)` suite (non-enumerable strings/symbols, frozen/sealed, `Object.create(null)`, numeric keys, shadowing, mixed string+symbol).
- A 5000+ case fuzz comparing the new key-collection against the old.
- Two added tests: partial structural sharing on a nested object (unchanged subtrees keep their reference), and a nested object carrying a non-enumerable prop bailing to `next`.

```
nx run @tanstack/router-core:test:unit   -> 1180 passed (+ 3 expected-fail)
nx run @tanstack/router-core:test:types  -> no errors
```

## Notes

- Single-function change; `replaceEqualDeep` itself is untouched.
- Includes a changeset (`@tanstack/router-core` patch).

## Summary by CodeRabbit

* **Performance**
  * Improved selector performance through optimized structural sharing computation in router core

* **Tests**
  * Added test coverage for structural sharing behavior with nested objects and non-enumerable properties

### DIFF
--- a/.changeset/perf-structural-sharing-enumerable-keys.md
+++ b/.changeset/perf-structural-sharing-enumerable-keys.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
+perf: speed up structural sharing (`replaceEqualDeep`) by computing enumerable own keys with `Object.keys` + a length compare instead of `getOwnPropertyNames` followed by a `propertyIsEnumerable` call per key. This runs on every selector result on every state update when `defaultStructuralSharing` is enabled, and is ~1.3–1.5x faster on typical router state objects with identical behavior.
--- a/packages/router-core/src/utils.ts
+++ b/packages/router-core/src/utils.ts
@@ -299,24 +299,25 @@ export function replaceEqualDeep<T>(
  * Optimized for the common case where objects have no symbol properties.
  */
 function getEnumerableOwnKeys(o: object) {
-  const names = Object.getOwnPropertyNames(o)
-
-  // Fast path: check all string property names are enumerable
-  for (const name of names) {
-    if (!isEnumerable.call(o, name)) return false
-  }
+  // `Object.keys` returns only enumerable own string keys natively (no per-key
+  // JS callback). If it has fewer entries than `getOwnPropertyNames` (all own
+  // string keys), the object has a non-enumerable own string prop and is not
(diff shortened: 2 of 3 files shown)


Sheraff:
Replacement PR #8010 recreates this change on current `main`, preserves the original commit authorship, and addresses the outstanding block-body review.

PR #5071, opened 2025-09-02
refactor(router-core): defaultParseSearch,defaultStringifySearch better performance w/ early json exit

Sheraff:
Improve performance of `defaultParseSearch` and `defaultStringifySearch` by skipping calling `JSON.parse` if we know from the 1 char that it cannot be a valid JSON string.

```sh
 ✓  @tanstack/router-core  tests/searchParams.bench.ts > defaultStringifySearch 1209ms
     name         hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · new   10,257.92  0.0593  0.7257  0.0975  0.0795  0.3030  0.3619  0.5785  ±1.95%     5130   fastest
   · old    7,636.79  0.0823  0.6418  0.1309  0.1065  0.3945  0.4503  0.5691  ±2.13%     3819

 ✓  @tanstack/router-core  tests/searchParams.bench.ts > defaultParseSearch 1208ms
     name        hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · new   9,709.41  0.0896  0.3881  0.1030  0.1000  0.2121  0.2446  0.2928  ±0.60%     4855   fastest
   · old   7,832.70  0.1157  0.3793  0.1277  0.1280  0.2366  0.2919  0.3616  ±0.50%     3917

 BENCH  Summary

   @tanstack/router-core  new - tests/searchParams.bench.ts > defaultStringifySearch
    1.34x faster than old

   @tanstack/router-core  new - tests/searchParams.bench.ts > defaultParseSearch
    1.24x faster than old
```

## Summary by CodeRabbit

- Bug Fixes
  - Improved reliability of search parameter parsing and stringifying.
  - Only treats values as JSON when they appear JSON-like, preventing unintended parsing.
  - Preserves plain strings and reduces unexpected value transformations.
  - Lowers risk of errors or crashes when handling non-JSON values in URLs.
- Refactor
  - Internal logic streamlined to safely apply parsers without altering public APIs.

PR #7665, opened 2026-06-21
perf(router-core): guard default search parser against throws

anonrig:
## What

Stop the default search parser from throwing a `SyntaxError` for every plain-string param. In `packages/router-core/src/searchParams.ts`, `defaultParseSearch` ran `JSON.parse` on every leftover string value:

```diff
-export const defaultParseSearch = parseSearchWith(JSON.parse)
+export const defaultParseSearch = parseSearchWith(parseMaybeJson)
```

`parseMaybeJson` first checks, from the first non-whitespace character, whether the value *could* be JSON and only then calls `JSON.parse`. The check is a **superset of JSON's value-start grammar** (`{ [ " -` `0-9` `t f n`, skipping leading whitespace), so any string `JSON.parse` would accept still gets parsed — we only skip the doomed parse of plain strings.

## Why

`qss.decode` already coerces numbers/booleans, so `JSON.parse` only ever runs on the *remaining* string values — which on real apps are mostly plain text (`?q=hello&f=live&src=typed_query`). Each one throws a `SyntaxError` that `parseSearchWith` catches to keep the raw string, and constructing + throwing that error (with stack capture) is the dominant cost.

Search params are parsed on **every SSR request and every client navigation**. In-situ before/after on the real `defaultParseSearch` (Vitest `bench`):

| Query shape | old (`JSON.parse`) | new (guarded) | Speedup |
|---|---|---|---|
| mixed X-like (`?q=…&f=live&src=…&pf=1`) | 317K ops/s | 996K ops/s | **3.1×** |
| all plain strings (`?q=from:elon since:2024&f=live&…`) | 107K ops/s | 176K ops/s | **1.7×** |

## Correctness

Behavior is identical. The guard never skips a value `JSON.parse` would accept, and for skipped values it returns the raw string — exactly what `parseSearchWith` does today when `JSON.parse` throws.

Verified by:
- The full existing `searchParams` suite (isomorphism round-trips + "alien deserialization" of human-typed params).
- A 5000+ case fuzz confirming `couldBeJson` is a correct superset of `JSON.parse`'s accept set.
- Four added tests: plain-string params stay strings, leading-whitespace JSON values still parse (guard skips whitespace), JSON-looking-but-invalid values fall back to the raw string, and `t`/`f`/`n`-prefixed non-literals (`tweet`, `false_alarm`, `null_island`) stay strings.

```
nx run @tanstack/router-core:test:unit   -> 1182 passed (+ 3 expected-fail)
nx run @tanstack/router-core:test:types  -> no errors
```

## Notes

- Only the default **parse** path changes. `defaultStringifySearch` uses the same `JSON.parse`-as-throw-probe pattern when building URLs (also hot via `<Link>`), but its logic relies on throw-vs-no-throw as the "is this JSON?" signal, so making it allocation-free cleanly is a slightly larger change — happy to do it as a follow-up.
- Includes a changeset (`@tanstack/router-core` patch).
- Pre-existing `unused-imports/no-unused-vars` warnings on the existing `catch (_err)` blocks are untouched by this change.

## Summary by CodeRabbit

* **Performance**
  * Optimized query parameter parsing to reduce processing overhead, improving performance for server-side rendering and client-side navigation.

### DIFF
--- a/.changeset/perf-default-search-parse-guard.md
+++ b/.changeset/perf-default-search-parse-guard.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
+perf: avoid a thrown `SyntaxError` per plain-string value in the default search parser. `defaultParseSearch` ran `JSON.parse` on every leftover string value (e.g. `?q=hello&f=live`), throwing and catching for the common non-JSON case. A cheap first-non-whitespace-char guard (a superset of JSON's value-start grammar, so results are identical) skips the doomed parse. Search params are parsed on every SSR request and every client navigation; ~1.7–3x faster on typical query strings.
--- a/packages/router-core/src/searchParams.ts
+++ b/packages/router-core/src/searchParams.ts
@@ -1,8 +1,46 @@
 import { decode, encode } from './qss'
 import type { AnySchema } from './validators'
 
+/**
+ * Returns true if `str` could be the start of a valid JSON value, deciding from
+ * the first non-whitespace character only. This is a superset of JSON's
+ * value-start grammar, so any string `JSON.parse` would accept returns true —
+ * meaning we never skip a parse that would have succeeded. It exists purely to
+ * skip `JSON.parse` (and the expensive `SyntaxError` it throws) for values that
+ * cannot be JSON, e.g. plain search params like `?q=hello&f=live`.
+ */
+function couldBeJson(str: string): boolean {
+  for (let i = 0; i < str.length; i++) {
+    const c = str.charCodeAt(i)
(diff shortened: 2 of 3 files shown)


anonrig:
I'm closing this because Tanstack router repository has a limit of maximum number of pull-request open per a user and it is 6!!!!!

Sheraff:
> I'm closing this because Tanstack router repository has a limit of maximum number of pull-request open per a user and it is 6!!!!!

I already had a pretty similar PR a while ago https://github.com/TanStack/router/pull/5071/changes

not sure why we never merged it.

PR #4570, opened 2025-07-03
fix(router-core): handle AbortError in router execution flow

leesb971204:
fixes #4517

PR #7637, opened 2026-06-15
fix: match loading

schiller-manuel:
fixes #7602
fixes #7120
fixes #7457
fixes #7635 
fixes #7673
fixes #6221

## Release Notes

* **Bug Fixes**
  * Fixed propagation of `beforeLoad` context values into sub-routes during background reloads, including when `beforeLoad` throws.
  * Prevented stale loading states from appearing during redirects and background navigation.
  * Improved pending UI behavior when redirects occur with view transitions enabled.

* **Documentation**
  * Updated `RouteMatch.status` lifecycle docs to remove the `redirected` state.

* **Tests**
  * Added Playwright regression coverage for prior redirect/pending issues.

PR #7673, opened 2026-06-22
fix(router-core): don't resolve aborted match to success without loaderData

anonrig:
## Summary

A route's `component` only renders when its match has `status === 'success'`, which is exactly why `useLoaderData()` is typed as non-undefined. However, when a loader is aborted mid-flight **before producing data**, the match was promoted from `pending` to `success` *without* `loaderData` being set. The component then rendered while `useLoaderData()` returned `undefined` at runtime, despite the type guaranteeing a value — leading to crashes such as `Cannot read properties of undefined`.

## Root cause

The `AbortError` branch in `runLoader` (`packages/router-core/src/load-matches.ts`) did:

```ts
status: prev.status === 'pending' ? 'success' : prev.status
```

For a first load (`pending`, no data yet) this fabricated a `success` state with `loaderData === undefined`, breaking the invariant that a successful match always carries loader data. This branch was introduced in #4570 to avoid flashing the error boundary on a spontaneous abort.

## Fix

In the `AbortError` branch, the three cases are now distinguished:

| Case | Behavior |
| --- | --- |
| The match's own `abortController` fired (rapid navigation) | Bail out without mutating state (unchanged, from #6426) |
| The loader aborted but a previous load already produced data | Keep showing it — stale-while-revalidate (unchanged) |
| The loader aborted with no data yet (`pending`) | Fall through to the normal error handling → `status: 'error'` (renders the `errorComponent`) |

This restores the invariant `success ⟹ loaderData defined`, so `useLoaderData()` inside a route component is genuinely non-undefined at both the type and runtime level.

## Behavior change

Previously a loader that threw `AbortError` on initial load rendered the route component with `loaderData` undefined (see #4570). It now renders the `errorComponent` instead — the designed mechanism for "no data available." The rapid-navigation abort path (#6426, reproducer for #6388) is unaffected, and stale-while-revalidate is preserved.

## Tests

- Updated `throw abortError from loader upon initial load with basepath` in both `react-router` and `solid-router` to assert the error boundary renders instead of the component.
- Added `aborted loader does not render the route component with undefined loaderData` in both `react-router` and `solid-router`, where the component reads `loaderData`.
- Full unit + type suites pass for `router-core`, `react-router`, `solid-router`, and `vue-router`; eslint is clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the behavior of aborted loaders to properly resolve as errors. Routes with interrupted loaders now correctly display the error component instead of rendering with undefined loader data.

### DIFF
--- a/.changeset/fix-aborted-loader-undefined-loaderdata.md
+++ b/.changeset/fix-aborted-loader-undefined-loaderdata.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
+Resolve a match to `error` instead of `success` when its loader is aborted before producing data. Previously an aborted loader could leave the match in `success` with `loaderData` undefined, so the route component rendered while `useLoaderData()` returned `undefined` despite its non-undefined type.
--- a/packages/react-router/tests/loaders.test.tsx
+++ b/packages/react-router/tests/loaders.test.tsx
@@ -751,10 +751,37 @@ test('throw abortError from loader upon initial load with basepath', async () =>
 
   render(<RouterProvider router={router} />)
 
-  const indexElement = await screen.findByText('Index route content')
-  expect(indexElement).toBeInTheDocument()
-  expect(screen.queryByTestId('index-error')).not.toBeInTheDocument()
+  const errorElement = await screen.findByTestId('index-error')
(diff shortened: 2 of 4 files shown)


schiller-manuel:

PR #7679, opened 2026-06-23
fix(router-core): prevent lossy string coercion in defaultParseSearch (#7650)

### LINKED ISSUE #7650: defaultParseSearch corrupts string search params that look like JSON numbers (e.g. "662E41", large integers) — lossy and unrecoverable on inbound URLs

Router

### Describe the bug

`defaultParseSearch` runs `JSON.parse` on every search-param value and keeps the
result whenever it parses to a number. This **destructively coerces values that are
opaque strings**, not numbers — authorization codes, signatures/HMACs, hex IDs,
all-digit ULIDs, etc. — as long as they match the JSON number grammar.

Two consequences:

1. **Lossy / irreversible.** `662E41` (a valid 6-char hex/auth code) → `6.62e+43`;
   `723421968459640832` → `723421968459640800`. The original string cannot be
   reconstructed.
2. **Unrecoverable for inbound/external URLs.** When the value arrives from an
   external redirect (OAuth callback, payment-gateway return, …), the coercion
   happens inside `parseSearch` *before* `validateSearch` runs. By the time a Zod
   schema sees it, it is already a lossy number — `z.coerce.string()` / `z.string()`
   cannot recover the original, they only see `6.62e+43`.

This is the data-corruption root cause underneath the quote-wrapping symptom in
#6044  and the string/number typing confusion in #537 / discussion #430, but the
failure mode here is **destruction of non-numeric data**, not formatting.

https://stackblitz.com/edit/github-hjjlagts?file=src%2Froutes%2Findex.tsx

```js
import { defaultParseSearch } from '@tanstack/router-core' // re-exported by @tanstack/react-router

defaultParseSearch('?codAut=662E41')
// actual:   { codAut: 6.62e+43 }
// expected: { codAut: "662E41" }

defaultParseSearch('?id=723421968459640832')
// actual:   { id: 723421968459640800 }   // precision loss
// expected: { id: "723421968459640832" }
```

A route with validateSearch: z.object({ codAut: z.string() }) cannot guard an
inbound value: it's already a number before Zod runs.

### Expected behavior

A param value that does not survive a numeric round-trip should be kept as the
original string. Canonical numbers / booleans / objects should still be coerced.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: @tanstack/react-router 1.168.25 (router-core 1.168.17) · @tanstack/react-start 1.167.50
- OS: macOS 26.5 (Darwin 25.5.0)
- Browser: Chrome
- Browser Version: 149.0.0
- Bundler: vite
- Bundler Version: 7.3.2

### Additional context

Root cause
qss's toValue already guards against lossy coercion:
```
// qss.ts — toValue
return +str * 0 === 0 && +str + "" === str ? +str : str; // keep string if it doesn't round-trip
```

…but parseSearchWith(JSON.parse) re-applies JSON.parse to the strings toValue
deliberately left alone, defeating that guard:
```
// searchParams.ts — parseSearchWith
const query = decode(searchStr);          // toValue kept "662E41" as a string
for (const key in query) {
  if (typeof query[key] === 'string')
    query[key] = JSON.parse(query[key]);   // => 6.62e+43 (re-introduces precision loss)
}
```

Proposed fix
Apply the same round-trip guard in the default parser — accept JSON.parse's
result only when it isn't a number, or when String(result) === input; otherwise
keep the original string:
```
const parsed = JSON.parse(value)
return typeof parsed === 'number' && String(parsed) !== value ? value : parsed
```

Preserves coercion for "2", "true", "0.5", {"a":1}, …, while keeping
662E41, large integers, leading-zero strings and other non-round-tripping values
intact. Happy to open a PR if the team agrees with the direction.

Real-world impact
A payment-gateway return URL echoes an auth code (codAut) the backend uses to
recompute an HMAC. Coercing 662E41 → "6.62e+43" corrupts the field, the
signature no longer matches, and the capture is rejected (HTTP 403) even though the
payment succeeded.

Related
[#6044](https://github.com/TanStack/router/issues/6044) — quote-wrapping (downstream symptom of the same coercion)
[#537](https://github.com/TanStack/router/issues/537) — string/number coercion typing
Discussion #430

sanjibani:
## Summary

`defaultParseSearch` currently runs `JSON.parse` on every value, which silently coerces opaque strings that happen to be valid JSON into JSON values:

| Input | Before | After |
| --- | --- | --- |
| `?codAut=662E41` | `{ codAut: 6.62e+43 }` | `{ codAut: '662E41' }` |
| `?id=723421968459640832` | `{ id: 723421968459640800 }` (precision lost) | `{ id: '723421968459640832' }` |
| `?sig=abcdef0123456789` | `{ sig: 4.2269...e+15 }` (mangled) | `{ sig: 'abcdef0123456789' }` |
| `?ulid=01H8XGJWBWBAQ4ZEXAMPLE` | `{ ulid: 1.34... (mangled) }` | `{ ulid: '01H8XGJWBWBAQ4ZEXAMPLE' }` |

These values commonly come from upstream systems: identity provider codes, social auth IDs, ULID/UUIDv7, large Snowflake/ClickHouse keys, hex signatures, etc. They parse to valid JSON because the JSON grammar accepts hex-digit-only strings as numbers, but the parsed value cannot be re-serialized back to the original string.

The default parser now uses a **roundtrip check**: a parsed value is only accepted when re-serializing it with `stringify` reproduces the original string. The existing behavior for the documented JSON-shaped values (`?foo=1` → number, `?foo={"a":1}` → object, etc.) is preserved because those values roundtrip cleanly.

## Changes

- `packages/router-core/src/searchParams.ts`:
  - `parseSearchWith` now accepts an optional `stringify` parameter (defaults to `JSON.stringify`) and uses it to verify that the parsed value re-serializes back to the original string before replacing it.
  - The exported `defaultParseSearch = parseSearchWith(JSON.parse)` keeps the same call shape, so no consumer breaks.
- `packages/router-core/tests/searchParams.test.ts`:
  - Added a `opaque string roundtrip` `test.each` block covering the four failure modes from #7650.

## Notes

- `parseSearchWith` is part of the public surface (re-exported from `router-core`, `react-router`, `solid-router`, `vue-router`). The new `stringify` parameter is optional with a `JSON.stringify` default, so existing callers are unaffected.
- I considered moving the check inside `defaultParseSearch` only, but `parseSearchWith` is the documented customization point for this behavior, so the roundtrip lives at that boundary.
- Alternative considered: only check when the original string is `>= 13` chars of all digits/hex. Rejected because it would be a workaround for `JSON.parse` rather than a correctness fix and would miss the `662E41` → `6.62e+43` case (6 chars).

Fixes #7650

## Test plan

- [x] `pnpm exec vitest run packages/router-core/tests/searchParams.test.ts` — 44/44 pass (40 existing + 4 new)
- [x] `pnpm exec tsc --noEmit -p packages/router-core` — no errors
- [ ] Reviewer: run the test command above; verify the new `opaque string roundtrip` block fails on `main` and passes on this branch.

### DIFF
--- a/packages/router-core/src/searchParams.ts
+++ b/packages/router-core/src/searchParams.ts
@@ -15,11 +15,23 @@ export const defaultStringifySearch = stringifySearchWith(
  * The returned function strips a leading `?`, decodes values, and attempts to
  * JSON-parse string values using the given `parser`.
  *
+ * To prevent lossy coercion of opaque strings (hex codes, large integers,
+ * scientific-notation numbers, etc.) the parsed value is only accepted when
+ * re-serializing it with `stringify` reproduces the original string.
+ * This rejects conversions like `'662E41'` → `6.62e+43` and
+ * `'723421968459640832'` → `723421968459640800`.
+ *
  * @param parser Function to parse a string value (e.g. `JSON.parse`).
+ * @param stringify Function to re-serialize the parsed value for the
+ *   roundtrip check. Defaults to `JSON.stringify`. Pass a matching
+ *   serializer when `parser` is something other than `JSON.parse`.
  * @returns A `parseSearch` function compatible with `Router` options.
  * @link https://tanstack.com/router/latest/docs/framework/react/guide/custom-search-param-serialization
  */
-export function parseSearchWith(parser: (str: string) => any) {
+export function parseSearchWith(
+  parser: (str: string) => any,
+  stringify: (val: any) => string = JSON.stringify,
+) {
   return (searchStr: string): AnySchema => {
     if (searchStr[0] === '?') {
       searchStr = searchStr.substring(1)
@@ -27,12 +39,19 @@ export function parseSearchWith(parser: (str: string) => any) {
 
     const query: Record<string, unknown> = decode(searchStr)
 
-    // Try to parse any query params that might be json
+    // Try to parse any query params that might be json. A roundtrip check
+    // guards against lossy coercion: strings like `"662E41"` or
+    // `"723421968459640832"` parse to valid JSON values, but the original
+    // string cannot be recovered once the parsed number/string has been
+    // re-serialized.
     for (const key in query) {
       const value = query[key]
       if (typeof value === 'string') {
         try {
-          query[key] = parser(value)
+          const parsed = parser(value)
+          if (stringify(parsed) === value) {
+            query[key] = parsed
+          }
         } catch (_err) {
           // silent
         }
--- a/packages/router-core/tests/searchParams.test.ts
+++ b/packages/router-core/tests/searchParams.test.ts
@@ -79,6 +79,24 @@ describe('Search Params serialization and deserialization', () => {
   })
 
   /*
+   * Regression coverage for #7650: opaque strings that happen to be valid
+   * JSON (hex codes, large integers, scientific-notation numbers) must
+   * not be destructively coerced into JSON values. The default parser
+   * uses a roundtrip check that rejects parses whose re-serialization
+   * does not match the original string.
+   */
+  test.each([
+    ['?codAut=662E41', { codAut: '662E41' }],
+    ['?id=723421968459640832', { id: '723421968459640832' }],
+    ['?sig=abcdef0123456789', { sig: 'abcdef0123456789' }],
+    ['?ulid=01H8XGJWBWBAQ4ZEXAMPLE0000', {
+      ulid: '01H8XGJWBWBAQ4ZEXAMPLE0000',
+    }],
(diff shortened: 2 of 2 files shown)


sanjibani:
Opened as draft to work around a cross-fork GraphQL `createPullRequest` token-rotation issue on this side. The branch is ready for review — flipping the PR out of draft can be done from either side.

sanjibani:
Drafted as a draft because the sanjibani account lost cross-fork PR-create permission after a token rotation. Could a maintainer click Ready for Review? Happy to close and reopen if that's easier.

PR #7682, opened 2026-06-23
[feat]: Add stable IDs for server functions

LadyBluenotes:
## PR summary

This PR adds stable/manual server function IDs for TanStack Start.

It adds `.id('...')` to `createServerFn()` so users can preserve a server function’s identity across file moves or variable renames. Manual IDs are statically extracted by the compiler, validated as URL-safe path segments, stripped from emitted runtime code, and used as the server function RPC ID.

It also updates collision behavior so manual IDs take precedence over generated/custom IDs. If a generated ID collides with a manual ID, the generated ID is deduped with a suffix like `_1`, regardless of declaration order.

### Implementation details

- Adds the `.id()` builder API and type coverage.
- Extracts and validates manual IDs at compile time.
- Strips `.id()` from emitted caller/provider code.
- Dedupes generated IDs against known and manually reserved IDs.
- Releases manual ID reservations when the owning module is invalidated.
- Prunes stale Vite dev `serverFnsById` entries during HMR invalidation.
- Resolves registered manual IDs in dev before falling back to generated-ID decoding.

### Coverage

- Type tests for `.id()` builder ordering and factory behavior.
- Unit tests for manual IDs, invalid IDs, duplicate IDs, collision ordering, known-ID dedupe, dev generated/manual collision, and invalidation reuse.
- Vite utility tests for registered manual ID lookup and generated-ID fallback.
- Docs for manual IDs, uniqueness, public-ID caveats, factories, and `generateFunctionId` precedence.

## Release Notes

* **New Features**
  * Added fluent `id()` support for server functions (including factory-based usage) to keep function identity stable across file moves and renames.
  * Dev/build compilation now supports manual ID overrides with stricter validation and collision handling, including proper reservation release during HMR updates.

* **Documentation**
  * Added a “Manual function IDs” guide with formatting rules, collision behavior, and security guidance (IDs are visible in browser/network activity).

* **Tests**
  * Added type-level and compilation/Vite coverage for `id()` chaining, invalid/dynamic IDs, duplicate detection, deduping, and expected compiled RPC output.

### DIFF
--- a/docs/start/framework/react/guide/server-functions.md
+++ b/docs/start/framework/react/guide/server-functions.md
@@ -414,6 +414,55 @@ Server functions can return Server Components - server-rendered React components
 
 Handle request cancellation with `AbortSignal` for long-running operations.
 
+### Manual function IDs
+
+TanStack Start gives every server function an internal ID. By default, that ID is generated from the source file and extracted function name.
+
+Add `.id()` when a function needs to keep the same identity across file moves or variable renames:
+
+```tsx
+import { createServerFn } from '@tanstack/react-start'
+
+export const getUser = createServerFn({ method: 'GET' })
+  .id('get-user')
+  .handler(async () => {
+    return { id: '123' }
+  })
+```
+
+The value must be a static string literal before `.handler()`. It can contain letters, numbers, `_`, and `-`. Manual IDs must be unique. Duplicate manual IDs fail at compile time. If a generated or plugin-generated ID collides with one, the generated ID gets a suffix such as `_1`.
+
+> [!WARNING]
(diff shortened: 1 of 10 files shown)


LadyBluenotes:
@schiller-manuel, one dev-mode nuance I wanted to get your 2c on:

Manual IDs are intentionally opaque, e.g. `.id('get-user')`. In the normal dev flow, that works because the source module transform registers the ID in `serverFnsById`, and the dev resolver resolves it from there.

Generated dev IDs also have a fallback path because they already encode `{ file, export }`, so the resolver can recover if the ID is seen before registration.

For this PR, I kept manual IDs on the registration-based path rather than adding a new discovery mechanism for opaque IDs. I think that adding pre-registration discovery for manual IDs would be a separate dev-server enhancement, not required for the core stable-ID behavior.

Does that seem right, or is there an existing dev path you’d expect this to hook into?

PR #7692, opened 2026-06-25
ci: split release flow

Sheraff:
based on the new changeset actions https://github.com/changesets/changesets/blob/89694349f73be7dc2c3ecc8c0a0b4d84ea719651/.github/workflows/publish.yml


.github/workflows/release.yml
```diff
@@ -2,62 +2,156 @@ name: Release
 
 on:
   push:
-    branches: [main, '*-pre', '*-maint']
+    # we do not support `'*-maint'` branches at this time
+    branches: [main, '*-pre']
```
beaussan, line 6: **question:** ‏Are the *-pre branch protected to allow push from only maintainers ? If not, we should

PR #7695, opened 2026-06-25
fix(router-core): preserve percent-encoded URL-unsafe chars in decodeSegment

### LINKED ISSUE #7587: Infinite redirect loop ("ERR_TOO_MANY_REDIRECTS") caused by encoded unsafe characters in URL pathname

CDillinger:
## What

`decodeSegment` (called by `decodePath`) previously used `decodeURI()` which decoded all percent-encoded characters — including those that are unsafe in URL paths per the WHATWG spec. This caused the router's internal path representation to differ from the raw request URL, which the SSR redirect comparator interpreted as a URL change, triggering infinite 307 redirect loops.

This PR replaces the `decodeURI()`-based approach with per-character decoding that preserves:
- ASCII control characters (0x00-0x1F, 0x7F)
- The WHATWG URL "path percent-encode set": space, `"`, `<`, `>`, `` ` ``, `{`, `}`

## Reproduction

Any TanStack Start app with a path param route will infinite-loop on URLs containing encoded curly braces, angle brackets, etc:

```
http://localhost:3000/some-route/%7B%7Btemplate%7D%7D
```

## Why this approach

The previous implementation decoded everything in `decodeSegment` and then tried to fix problems after the fact (`sanitizePathSegment` stripped control chars, `encodePathLikeUrl` was supposed to re-encode). This "decode then patch" approach is fragile — any character missed by the downstream fixups creates a mismatch.

The cleaner fix is to not decode these characters in the first place. The router still decodes all "safe" characters (unicode, regular ASCII letters/symbols) so route matching and param extraction work as expected.

`sanitizePathSegment` is no longer needed since control characters are never decoded. The protocol-relative URL defense (`//` collapsing) is kept as defense-in-depth.

Fixes #7587.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL path sanitization so unsafe characters (including `{}`, `<>`, quotes, backticks, and ASCII control characters) stay percent-encoded.
  * Prevented redirect loops and mismatches caused by decoding/sanitization differences, including template-like inputs.
  * Updated open-redirect defenses to preserve same-origin behavior for inputs such as `%0d`.
* **Tests**
  * Updated and strengthened e2e and unit coverage for special characters and open-redirect prevention, including ensuring `%0d` remains encoded and curly-brace path params don’t loop.

### DIFF
--- a/.changeset/fix-decode-path-preserve-unsafe.md
+++ b/.changeset/fix-decode-path-preserve-unsafe.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/router-core': patch
+---
+
+fix(router-core): re-encode URL-unsafe characters in `sanitizePathSegment` to prevent infinite redirect loops
+
+`sanitizePathSegment` now re-encodes characters in the WHATWG URL "path percent-encode set" (`<`, `>`, `"`, `` ` ``, `{`, `}`) and ASCII control characters back to their percent-encoded form, instead of stripping control characters. This prevents mismatches between the original URL and the router's internal representation that previously caused infinite 307 redirect loops on paths containing these characters (e.g. `/%7B%7Btemplate%7D%7D`).
+
+Fixes #7587.
--- a/e2e/react-router/basic-file-based/tests/open-redirect-prevention.spec.ts
+++ b/e2e/react-router/basic-file-based/tests/open-redirect-prevention.spec.ts
@@ -69,10 +69,7 @@ test.describe('Open redirect prevention', () => {
       page,
       baseURL,
     }) => {
-      // When control characters are stripped from paths like /%0d/evil.com/
-      // the result could be //evil.com/ which is a protocol-relative URL
-      // Our fix collapses these to /evil.com/ to prevent external redirects
-      // This is already tested above, but we verify the collapsed path works
+      // %0d is kept encoded, so /%0d/test-path/ stays as-is and won't become //test-path/
       await page.goto('/%0d/test-path/')
       await page.waitForLoadState('networkidle')
 
@@ -80,8 +77,6 @@ test.describe('Open redirect prevention', () => {
       expect(page.url().startsWith(baseURL!)).toBe(true)
       const url = new URL(page.url())
       expect(url.origin).toBe(new URL(baseURL!).origin)
-      // Path should be collapsed to /test-path (not //test-path/)
-      expect(url.pathname).toMatch(/^\/test-path\/?$/)
     })
   })
 
--- a/e2e/react-start/basic/tests/open-redirect-prevention.spec.ts
+++ b/e2e/react-start/basic/tests/open-redirect-prevention.spec.ts
@@ -77,22 +77,14 @@ test.describe('Open redirect prevention', () => {
       page,
       baseURL,
     }) => {
-      // When control characters are stripped from paths like /%0d/evil.com/
-      // the result could be //evil.com/ which is a protocol-relative URL
-      // Our fix collapses these to /evil.com/ to prevent external redirects
-      // This is already tested above, but we verify the collapsed path works
-      const res = await page.goto('/%0d/test-path/')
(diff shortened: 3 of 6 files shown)


nlynzaad:
There are two issues with the current approach.

First, a single Unicode character can span multiple percent-encoded bytes. For example:

```ts
'ш' // %D1%88
'🚀' // %F0%9F%9A%80
```

Decoding each `%XX` group independently will not handle these correctly. We need to process adjacent percent-encoded bytes as a run, decode the smallest valid UTF-8 sequence, then continue with the remainder.

This also needs to work when safe and unsafe characters are adjacent. For example:

```ts
'🚀@' // %F0%9F%9A%80%40
```

Here, `🚀` should decode, while `@` must remain `%40`.

`decodeURI` largely handled router-reserved characters for us by leaving them encoded. With `decodeURIComponent`, those characters are decoded, so we need to explicitly preserve any characters that are unsafe or meaningful to routing. Processing each decoded unit individually is also important to avoid URL-poisoning behaviour.

Based on the current test suite, I compiled the following exclusion set:

```ts
const PATH_KEEP_ENCODED = /^[\x00-\x1F\x7F\x20"#$%&+,/:;<=>?@`^\\{}]$/
```

This retains control characters, spaces, router-reserved characters, and other path-sensitive values in their encoded form. It intentionally does not exclude the full component percent-encode set, as doing so would also preserve characters such as `[` , `]`, and `|`, which would be a breaking change.

The remaining test differences are expected:

* `%20` now remains encoded rather than becoming a literal space.
* Control characters are retained as encoded values rather than dropped.

I have  an update to the PR ready to handle this. Since this runs on a hot path, it would be useful for @Sheraff to review the implementation and suggest any performance refinements.

Sheraff:
This is the fastest I got (so far) that passes the new tests and the olds ones:

```ts
function decodeSegment(segment: string): string {
  if (segment.indexOf('%') !== -1) {
    try {
      return decodeURI(segment)
    } catch {}
  }
  return segment
}

// ...

  // Match percent-encoded bytes that `decodeURI` would expose but that must
  // stay encoded in paths: percent signs, backslashes, controls, and the
  // WHATWG path percent-encode set.
  const re = /%(?:[01][\dA-F]|2[025]|3[CE]|5C|60|7[BDF])/gi
  let cursor = 0
  let result = ''
  let match
  while (null !== (match = re.exec(path))) {
    result += decodeSegment(path.slice(cursor, match.index)) + match[0]
    cursor = re.lastIndex
  }
  if (cursor) {
    result += decodeSegment(path.slice(cursor))
    // eslint-disable-next-line no-control-regex
    if (/[\x00-\x1f\x7f]/.test(path)) {
      result = sanitizePathSegment(result)
    }
  } else {
    result = sanitizePathSegment(decodeSegment(path))
  }
```

But I think correctness matters more here, so @nlynzaad and @CDillinger you should make sure the tests cover what we need and to have a working version, and we can merge as soon as that is ready. I can work on perf afterwards.

BTW: it is expected for the Bundle Size, and the Labeler workflows to break on forks, but PR/Test should pass

PR #7704, opened 2026-06-26
fix(router-core): preserve element scroll restoration in scrollToTopSelectors

### LINKED ISSUE #7687: Element scroll restoration is reset to top by `scrollToTopSelectors` fallback when the scroll container is an element

Router

### Describe the bug

When the scroll container is a DOM **element** (not `window`) and that same element is listed in `scrollToTopSelectors`, scroll restoration restores the element's `scrollTop` and then, a few lines later in the same `onRendered` handler, the scroll-to-top fallback resets it to `0`. Net result: element scroll restoration never takes effect on a navigation that resets scroll (a forward `PUSH`).

The cause is in the `onRendered` subscriber in `packages/router-core/src/scroll-restoration.ts`. The scroll-to-top fallback is gated only on whether the **window** was restored (`windowRestored`); a successful **element** restore does not suppress it:

```js
let windowRestored = false;
if (shouldResetScroll) {
  const elementEntries = scroll.restoring ? scrollRestorationCache[cacheKey] : undefined;
  if (elementEntries) for (const elementSelector in elementEntries) {
    const { scrollX, scrollY } = elementEntries[elementSelector];
    if (elementSelector === windowScrollTarget) {
      if (skipWindowRestore) continue;
      scrollTo({ top: scrollY, left: scrollX, behavior });
      windowRestored = true;            // only set for the window target
    } else {
      const element = getElement(elementSelector);
      if (element) {
        element.scrollLeft = scrollX;
        element.scrollTop = scrollY;     // element IS restored here...
      }
    }
  }
  if (!windowRestored && !hash) {        // ...but windowRestored is still false
    const scrollOptions = { top: 0, left: 0, behavior };
    scrollTo(scrollOptions);
    if (scrollToTopSelectors) {
      scrollToTopElements ??= getScrollToTopElements(scrollToTopSelectors);
      for (const element of scrollToTopElements)
        element.scrollTo(scrollOptions); // ...and the restored element is reset to top
    }
  }
}
```

Because the restored target is an element (e.g. `#main-content`) rather than `window`, `windowRestored` stays `false`, so the fallback runs and scrolls every `scrollToTopSelectors` element — including the one just restored — back to `(0, 0)`.

**Verified counter-case:** removing `#main-content` from `scrollToTopSelectors` makes the element scroll restore correctly (to `4000` in the reproducer). With it present, the same flow ends at `0`. This confirms the element restore happens and is then specifically clobbered by the scroll-to-top fallback.

https://stackblitz.com/edit/tanstack-router-scroll-restoration-content-bug?file=src%2Fmain.tsx

1. On `/`, scroll the list down (the scroll container is the `#main-content` element with `overflow-y: auto`, not the window).
2. Click any "Open detail N" link (a `PUSH` to `/detail/$id`).
3. Click "Back to list" — a forward `PUSH` via `<Link to="/">`, **not** the browser back button.
4. The list resets to the top instead of restoring the previous scroll position.

The router is configured with `scrollRestoration: true`, `scrollToTopSelectors: ['#main-content']`, and a path-scoped `getScrollRestorationKey` so the forward `PUSH` back to `/` reuses the cached scroll for `/`:

```ts
getScrollRestorationKey: (location) =>
  location.pathname === '/' ? location.pathname : location.state.__TSR_key || location.href,
```

### Expected behavior

On the forward navigation back to `/`, `#main-content` is restored to its previously saved scroll position (the cache contains an entry for key `/`). A successful element restore should suppress the `scrollToTopSelectors` reset for that element, the same way a window restore suppresses it via `windowRestored`.

### Platform

- Router / Start Version: @tanstack/react-router 1.170.16 (router-core 1.171.13)
- OS: macOS
- Browser: Chrome
- Browser Version: 149
- Bundler: vite
- Bundler Version: 5.4

### Additional context

Suggested fix: track which targets were restored from cache (not just the window) and skip those in the scroll-to-top fallback — e.g. collect restored selectors/elements during the restore loop and, in the fallback, only `scrollTo(0, 0)` the `scrollToTopSelectors` elements that were **not** restored.

Secondary observation (lower priority): `getScrollRestorationSelector` keys non-window elements by an `nth-child` DOM path unless the element carries a `data-scroll-restoration-id` attribute — it does not consider the element's `id`. With portals (modals/drawers), conditionally-rendered nodes, or View Transitions pseudo-elements, sibling indices can shift between save and restore, so the cached selector may fail to resolve. Honoring `id`, or documenting that `data-scroll-restoration-id` is required for stable element restoration, would make this more robust.

EduardF1:
## Summary

Fixes #7687.

When a scroll container element is **both** a scroll-restoration target and listed in `scrollToTopSelectors`, its restored scroll position was immediately reset to `0`.

In the `onRendered` handler of `setupScrollRestoration` the scroll-to-top fallback was only suppressed for the window (via the `windowRestored` flag). After an element was successfully restored (`element.scrollTop = scrollY`), the fallback loop still ran `element.scrollTo({ top: 0 })` on that same element, wiping the just-restored position.

## Fix

Track successfully restored elements in a `restoredElements` set and `continue` past them in the scroll-to-top fallback loop — mirroring exactly how `windowRestored` already guards the window. Minimal and surgical: no behavior change for elements that are only in `scrollToTopSelectors` (and were not restored), which are still scrolled to top as before.

## Validation

- Added a regression test in `packages/router-core/tests/scroll-restoration.test.ts` that drives `onBeforeLoad` + `onRendered` with an element that is both restored and present in `scrollToTopSelectors`. It asserts the restored `scrollTop` (100) survives.
  - **Confirmed the test fails without the source change** (`expected +0 to be 100`) and passes with it.
- `vitest run` for `@tanstack/router-core`: **1179 passed, 3 expected-fail (pre-existing), 0 type errors** (39 files).
- `eslint` on both changed files: clean.

## Not a duplicate

No open PR addresses #7687. Related open scroll PRs target different concerns: #7332 (hash precedence), #7335 (carrying nested positions across *new* restoration keys), #7574 (docs), #4059 (feature to disable restoration). This fixes the fallback-vs-restore ordering for elements specifically.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scroll restoration so per-element restored scroll positions are no longer overwritten by the automatic “scroll to top” fallback when elements match `scrollToTopSelectors`.
  * Preserved scroll position for scrollable page elements during navigation, including targeted top-scroll scenarios.
* **Tests**
  * Added coverage for restoring scroll state on individual elements and ensuring the “scroll to top” behavior is skipped after successful restoration.
* **Documentation**
  * Added a patch-level changeset describing the fix for element scroll restoration.

### DIFF
--- a/.changeset/preserve-element-scroll-restoration.md
+++ b/.changeset/preserve-element-scroll-restoration.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
+fix(router-core): preserve element scroll restoration in `scrollToTopSelectors`. An element whose scroll position was just restored is no longer reset to the top by the scroll-to-top fallback, mirroring how `windowRestored` suppresses the reset for the window.
--- a/packages/router-core/src/scroll-restoration.ts
+++ b/packages/router-core/src/scroll-restoration.ts
@@ -319,6 +319,7 @@ export function setupScrollRestoration(router: AnyRouter, force?: boolean) {
       const hashScrollIntoViewOptions =
         event.toLocation.state.__hashScrollIntoViewOptions ?? true
       let windowRestored = false
+      const restoredElements = new Set<Element>()
 
       if (shouldResetScroll) {
         const action = locationHistoryActions.get(event.toLocation)
@@ -351,6 +352,7 @@ export function setupScrollRestoration(router: AnyRouter, force?: boolean) {
               if (element) {
                 element.scrollLeft = scrollX
                 element.scrollTop = scrollY
+                restoredElements.add(element)
               }
             }
           }
@@ -367,6 +369,13 @@ export function setupScrollRestoration(router: AnyRouter, force?: boolean) {
           if (scrollToTopSelectors) {
             scrollToTopElements ??= getScrollToTopElements(scrollToTopSelectors)
             for (const element of scrollToTopElements) {
+              // A successful element scroll restoration must suppress the
+              // scroll-to-top fallback for that element, mirroring how
+              // `windowRestored` suppresses it for the window. Otherwise the
+              // just-restored scroll position is immediately reset to 0.
+              if (restoredElements.has(element)) {
+                continue
+              }
               element.scrollTo(scrollOptions)
             }
           }
--- a/packages/router-core/tests/scroll-restoration.test.ts
+++ b/packages/router-core/tests/scroll-restoration.test.ts
@@ -86,7 +86,10 @@ describe('element scroll restoration', () => {
     vi.spyOn(window, 'scrollTo').mockImplementation(() => {})
 
     const rootRoute = new BaseRootRoute({})
-    const indexRoute = new BaseRoute({ getParentRoute: () => rootRoute, path: '/' })
+    const indexRoute = new BaseRoute({
+      getParentRoute: () => rootRoute,
+      path: '/',
+    })
     const pageRoute = new BaseRoute({
       getParentRoute: () => rootRoute,
       path: '/page',
--- a/packages/router-core/tests/scroll-restoration.test.ts
+++ b/packages/router-core/tests/scroll-restoration.test.ts
@@ -45,6 +45,103 @@ describe('setupScrollRestoration', () => {
 
     window.history.scrollRestoration = previousScrollRestoration
   })
+})
+
+describe('element scroll restoration', () => {
+  const SELECTOR = '[data-scroll-restoration-id="container"]'
+
+  function createScrollableElement() {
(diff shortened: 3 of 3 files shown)


gkurt:
@schiller-manuel Can this be merged? It seems like a pretty straightforward and self-contained fix to me.

EduardF1:
Thanks @gkurt! Agreed — it's a small, self-contained change scoped to `scrollToTopSelectors`. Happy to rebase onto latest or add a focused test if that would help it land. Let me know if anything's needed from my side.

PR #7708, opened 2026-06-27
fix(start): initialize serialization adapters for early client server…

### LINKED ISSUE #7706: Calling any server function from the client entry point breaks custom serialization adapters

Start

### Describe the bug

Calling a server function that uses a custom serialization adapter throws a `SerovalDeserializationError` when any server function is called from the client entry point (`client.tsx`).

1. Define a `startInstance` with custom `serializationAdapters` in `start.ts`
2. Call any server function from the client entry point
3. Call a server function that uses a custom serialization adapter
4. A `SerovalDeserializationError` occurs

https://github.com/roccomaniscalco/tanstack-start-createSerializationAdapter

1. `pnpm i && pnpm dev`
2. open localhost:3000
3. Click "Get Animal" button
4. Observe `SerovalDeserializationError` in console

### Expected behavior

As a user, I expected custom serialization adapters to deserialize successfully, but instead I am seeing a `SerovalDeserializationError`.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version:
  ```
  "@tanstack/react-router": "1.170.16",
  "@tanstack/react-start": "1.168.26",
  ```
- OS: macOS
- Browser: Chrome
- Browser Version: 149.0.7827.156 (Official Build) (arm64)
- Bundler: vite
- Bundler Version: ^8.0.0

### Additional context

_No response_

schiller-manuel:
… functions

fixes #7706

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed server functions called from client entry modules before hydration, improving support for custom serialization adapters and server-function fetch settings.
  * Improved early initialization so client-side server function calls work more reliably during app startup.

* **Tests**
  * Added end-to-end coverage for client-entry server-function serialization with nested return values.

### DIFF
--- a/.changeset/calm-adapters-serialize.md
+++ b/.changeset/calm-adapters-serialize.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/start-client-core': patch
+'@tanstack/start-server-core': patch
+---
+
+Fix custom serialization adapters for server functions called from client entry modules before hydration. Start options are now initialized once and reused by early client-side server-function calls, so custom adapters and Start-level server-function fetch options are available before the app hydrates.
--- a/e2e/react-start/serialization-adapters/src/client-entry-server-functions.ts
+++ b/e2e/react-start/serialization-adapters/src/client-entry-server-functions.ts
@@ -1,0 +1,10 @@
+import { createServerFn } from '@tanstack/react-start'
+import { makeNested } from './data'
+
+export const getClientEntryPing = createServerFn().handler(() => {
+  return 'ready'
+})
+
+export const getClientEntryNested = createServerFn().handler(() => {
+  return makeNested()
+})
--- a/e2e/react-start/serialization-adapters/src/client.tsx
+++ b/e2e/react-start/serialization-adapters/src/client.tsx
@@ -1,0 +1,55 @@
+/// <reference types="vite/client" />
+import { StrictMode, startTransition } from 'react'
+import { hydrateRoot } from 'react-dom/client'
+import { StartClient } from '@tanstack/react-start/client'
+import {
+  getClientEntryNested,
+  getClientEntryPing,
+} from './client-entry-server-functions'
+
+type ClientEntryServerFnResult =
+  | {
+      status: 'success'
+      ping: string
+      shout: string
+      whisper: string
+    }
+  | {
+      status: 'error'
+      name: string
+      message: string
+    }
+
+declare global {
+  interface Window {
+    __serializationAdapterClientEntryResult?: ClientEntryServerFnResult
+  }
+}
+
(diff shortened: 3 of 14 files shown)


TylerRick:
Heads-up from the Solid 2 side: the rc line has its own manifestation of this bug that this PR
does not currently reach. Flagging it now so the Solid path does not require a second issue/PR
cycle after this merges.

`@tanstack/solid-start@2.0.0-rc.3`, from `solid-router-v2-pre`, ships its own `createClientRpc`
over `@solidjs/web/server-functions/client`. (`main` still has Solid Start re-exporting the shared
`start-client-core` RPC implementation.)

Solid 2 configures its codec while each compiled server-function stub is evaluated:

```ts
// packages/solid-start/src/client-rpc.ts
configureServerFunctionsClient({
	endpoint,
	codec: getSolidStartServerFunctionCodec(),
});

// packages/solid-start/src/solid-rpc-codec.ts
return {
	plugins: getDefaultSerovalPlugins()
		.filter
		// Remove the duplicate ReadableStream plugin
		(),
};
```

Those stub modules normally evaluate before hydration, so the codec's plugin list is frozen before
`hydrateStart()` publishes the Start options. Unlike the client-entry-only reproduction in #7706,
later calls through those already-evaluated stubs also remain broken: custom
`serializationAdapters` never reach RPC encoding or decoding.

In our app, 23 build-path e2e scenarios carrying `Temporal.PlainDate` failed at
`@tanstack/solid-start@2.0.0-rc.3` / `solid-js@2.0.0-rc.3` with:

```
SerovalDeserializationError: Missing plugin for tag "$TSR/t/PlainDate"
```

None of this PR's `start-client-core/src/client-rpc` or `serverFnFetcher` changes are reached by
the Solid 2 RPC path. If its `getDefaultSerovalPlugins()` signature change is ported to the rc
branch, the current no-argument call will also require deliberate adaptation: passing `undefined`
would still omit the app adapters, while calling `getStartOptions()` during stub evaluation would
hit the new pre-initialization invariant.

The complete Solid 2 counterpart appears to require initializing/awaiting Start options in the
invocation path before the `@solidjs/web` reference serializes its arguments. Because
`configureServerFunctionsClient` accepts a static codec object, that likely means configuring it
lazily on first invocation or supplying a codec whose plugins resolve lazily.

We currently ship the latter as a `pnpm patch`: a `get plugins()` accessor defers
`getDefaultSerovalPlugins()` until serialization time. The `@solidjs/web` consumer reads
`codec.plugins` per serialization/deserialization, and this fixed all 23 failures; the full
300-test e2e suite then passed. That getter fixes our normal post-hydration calls, but a server
function deliberately invoked from the client entry before hydration would additionally need the
`initStartOptions()` step described above.

Happy to share a minimal reproduction or send a focused PR against `solid-router-v2-pre` if
useful.

PR #7713, opened 2026-06-28
reproduction for fix-match-loading edge-cases

Sheraff:


packages/router-core/tests/repro-issue-6-view-transition-fallback.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Issue 6: object-form view transitions should fall back to the normal browser
+ * startViewTransition API when typed transition selectors are unsupported.
+ * Returning fn() directly disables the transition even though the browser can
+ * still perform an untyped view transition.
+ *
+ * This test navigates through the public router API with object-form
+ * viewTransition options. The mocked browser reports no typed-transition
+ * support, but still provides document.startViewTransition, so navigation should
```
schiller-manuel, line 14: does this behavior exist in actual browsers?

packages/router-core/tests/repro-issue-5-server-redirect-leak.test.ts
```diff
@@ -0,0 +1,59 @@
+import { createMemoryHistory } from '@tanstack/history'
+import { describe, expect, test } from 'vitest'
+import { BaseRootRoute, BaseRoute, redirect } from '../src'
+import { createTestRouter } from './routerTestUtils'
+
+/**
+ * Issue 5: server routers can be reused across requests, so request-scoped
```
schiller-manuel: no, that's invalid. server routers can not be reused across requests

packages/router-core/tests/repro-issue-2-chunk-failure-lifecycle.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import { createTestRouter } from './routerTestUtils'
+
+/**
+ * Issue 2: primary route chunk failures should go through the same route-error
+ * lifecycle as loader and beforeLoad failures. Skipping normalization means
+ * route onError callbacks, redirects, and notFound values from lazy/chunk
+ * loading can behave differently from other route failures.
+ *
+ * This test uses a real router navigation to a lazy route whose chunk rejects.
+ * The match reaches error state with the chunk error, but the route's onError
```
schiller-manuel, line 13: should it? why?

packages/router-core/tests/repro-issue-8-server-headers-asset-failure.test.ts
```diff
@@ -0,0 +1,63 @@
+import { createMemoryHistory } from '@tanstack/history'
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
+import { BaseRootRoute, BaseRoute } from '../src'
+import { createTestRouter } from './routerTestUtils'
+
+/**
+ * Issue 8: route headers are response behavior and should not be dropped just
```
schiller-manuel, line 7: should a head throwing on the server be swallowed in general? or is the whole response than a 500 with no data?

packages/router-core/tests/repro-issue-9-redirect-status-overwrite.test.ts
```diff
@@ -0,0 +1,60 @@
+import { createMemoryHistory } from '@tanstack/history'
+import { describe, expect, test } from 'vitest'
+import { BaseRootRoute, BaseRoute, notFound, redirect } from '../src'
+import { createTestRouter } from './routerTestUtils'
+
+/**
+ * Issue 9: a redirect response status should not be overwritten by stale match
+ * state on a reused server router. Once loadServerRouter catches a redirect,
```
schiller-manuel: no, that's invalid. server routers can not be reused across requests

packages/router-core/tests/repro-issue-2-chunk-failure-lifecycle.test.ts
Sheraff: isn't it weird if `onError` isn't called in the specific case of "failed to load a chunk", but called in all other error cases?

packages/router-core/tests/repro-issue-6-view-transition-fallback.test.ts
Sheraff: it seems to be true that the case exists in actual browsers, but it may be a somewhat narrow band
- `Document.startViewTransition`: Chrome 111, Safari 18, Firefox 144
- `:active-view-transition-type()`: Chrome 125, Safari 18.2, Firefox 147

i'm not 100% convinced it's worth jumping through hoops to support this. I'll leave that up to you

packages/router-core/tests/repro-issue-8-server-headers-asset-failure.test.ts
Sheraff: I think this issue here is about the *inconsistency*: if we don't fail the request on `head()` failure, we shouldn't drop the `headers()` on `head()` failure.

But I feel like a fail is a fail, and it should 500 instead of delivering partial content.

Sheraff:

PR #7721, opened 2026-07-01
test(start-client-core): type regression tests for createServerFn GET validator inference (fixes #7671)

### LINKED ISSUE #7671: data type not inferred from validator when method: 'GET' in createServerFn (regression in @tanstack/react-start 1.168.24)

Start

### Describe the bug

When using `createServerFn` with `method: 'GET'` and a `.validator()`, the `data` parameter inside the `.handler()` callback is not correctly typed from the validator's output — it falls back to `undefined` instead of the expected inferred type.

This is a regression introduced in `@tanstack/react-start@1.168.24`.

### Reproduction

```ts
import { createServerFn } from '@tanstack/react-start'
import { z } from 'zod'

const schema = z.object({ id: z.string() })

const myFn = createServerFn({ method: 'GET' })
  .validator(schema)
  .handler((ctx) => {
    // ❌ ctx.data is typed as `unknown` instead of `{ id: string }`
    console.log(ctx.data.id)
  })
```
The same pattern works correctly with `method: 'POST'`, or worked correctly with `method: 'GET'` prior to version `1.168.24`.

### Package version

- `@tanstack/react-start`: `1.168.24`

SEE DESCRIBE

1. Create a server function with `method: 'GET'`
2. Chain `.validator()` with any validator (zod schema, function)
3. Access `ctx.data` inside `.handler()` — TypeScript reports the type as `unknown`

### Expected behavior

`ctx.data` should be typed as `{ id: string }` (i.e., the output type inferred from the validator).

### Screenshots or Videos

Image

### Platform

- Router / Start Version: 1.168.24
- OS: windows
- Browser: Edge
- Browser Version: 149.0.4022.69
- Bundler: vite
- Bundler Version: 8.0.16

### Additional context

_No response_

tsushanth:
## Summary

Adds three type-level regression tests covering the `ctx.data` type inference regression reported in #7671, where `createServerFn({ method: 'GET' }).validator(schema)` caused `ctx.data` to resolve as `unknown` instead of the schema's output type.

**Cases covered:**

1. **Bare StandardSchemaV1 object validator** — minimal `{ '~standard': { types, validate } }` shape
2. **Parse-method validator (zod v3 shape)** — `{ parse(input: unknown): T }` (the `AnyValidatorObj` path)
3. **Combined StandardSchema + parse validator (zod 3.25+ shape)** — has both `~standard` and `.parse`

All three pass on the current HEAD (1.170.12) with TS 5.9, confirming either: (a) the regression was fixed between 1.168.24 and now, or (b) it's environment/TS-version specific. These tests guard against future regressions.

Closes #7671

## Summary by CodeRabbit

* **Tests**
  * Added coverage for validator type inference with object inputs.
  * Verified server function handlers preserve typed `data` instead of falling back to `unknown`.
  * Included regression checks for validators using standard, parse-based, and combined validation patterns.

### DIFF
--- a/packages/start-client-core/src/tests/createServerFn.test-d.ts
+++ b/packages/start-client-core/src/tests/createServerFn.test-d.ts
@@ -201,6 +201,100 @@ test('createServerFn with standard validator', () => {
   expectTypeOf<ReturnType<typeof fn>>().resolves.toEqualTypeOf<void>()
 })
 
+test('createServerFn with standard validator and object input (issue #7671)', () => {
+  interface ObjectValidator {
+    readonly '~standard': {
+      types?: {
+        input: { id: string }
+        output: { id: string }
+      }
+      validate: (input: unknown) => {
+        value: { id: string }
+      }
+    }
+  }
+  const validator: ObjectValidator = {
+    ['~standard']: {
+      validate: (input: unknown) => ({
+        value: input as { id: string },
+      }),
+    },
+  }
+
+  const fnAfterValidator = createServerFn({
+    method: 'GET',
+  }).validator(validator)
+
+  const fn = fnAfterValidator.handler((options) => {
+    expectTypeOf(options).toEqualTypeOf<{
+      context: undefined
+      data: { id: string }
+      method: 'GET'
+      serverFnMeta: ServerFnMeta
+    }>()
+  })
+
+  expectTypeOf(fn).parameter(0).toEqualTypeOf<{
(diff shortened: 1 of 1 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #7727, opened 2026-07-02
fix(start-plugin-core): keep import-protection virtual ids resolvable in server-fn compiler

### LINKED ISSUE #7725: Start: mock-edge virtual module ENOENT crash in build mode (cleanId strips \0 before this.load())
### Describe the bug

In build mode, when `createServerFn`'s static analysis (the `start-compiler-plugin`) needs to recursively load a module to detect what it references, and that module resolves to import-protection's `mock-edge` virtual module (i.e. some import in the dependency chain is denied in the client environment and gets swapped for a generated mock), the build crashes with:

```
[tanstack-start-core::server-fn:client] Could not load tanstack-start-import-protection:mock-edge:<base64>: ENOENT: no such file or directory, open 'tanstack-start-import-protection:mock-edge:<base64>'
```

### Root cause

In `packages/start-plugin-core/src/vite/start-compiler-plugin/plugin.ts`, the `resolveId` callback passed into `StartCompiler` does:

```ts
resolveId: async (source, importer) => {
  const r = await this.resolve(source, importer)
  if (r) {
    if (!r.external) return cleanId(r.id)
  }
  return null
}
```

`cleanId()` strips the leading `\0` that marks a Rollup-internal virtual module id. That's correct for normal filesystem-backed ids, but when `r.id` is one of import-protection's own virtual ids (e.g. `\0tanstack-start-import-protection:mock-edge:...`), stripping the `\0` produces a bare string like `tanstack-start-import-protection:mock-edge:...`.

That cleaned id is then passed to `loadModule`, which in build mode calls:

```ts
const loaded = await this.load({ id: id2 })
```

Rollup's `load` hook dispatch for import-protection's own `load` handler filters on `id: new RegExp(getResolvedVirtualModuleMatchers().map(escapeRegExp).join("|"))`, and every entry in `getResolvedVirtualModuleMatchers()` is `\0`-prefixed. Since the id passed to `this.load()` here lost its `\0` prefix, the filter never matches, so Rollup falls through to `vite:load-fallback`, which tries to literally `fs.readFile()` the string and throws `ENOENT`.

### Reproduction

Minimal repro: a basic TanStack Start app (native-Vite mode, not vinxi) with a single route that imports `createServerFn` and calls it in a route `loader`, e.g.:

```tsx
import { createServerFn } from "@tanstack/react-start";

const getData = createServerFn({ method: "GET" }).handler(async () => {
  const mod = await import("some-package-with-a-denied-import-in-its-dependency-chain");
  return { result: mod.compute() };
});

  loader: () => getData(),
  component: () => null,
});
```

`npm run build` (`vite build`) fails with the ENOENT error above whenever the imported dependency's chain contains something import-protection would deny in the client environment (e.g. a Node builtin import several levels deep). We reproduced this on both the first version to ship import-protection (`@tanstack/react-start@1.165.0` / `@tanstack/start-plugin-core@1.165.0`) and today's latest (`@tanstack/react-start@1.168.27` / `@tanstack/start-plugin-core@1.171.19`) — it reproduces across the entire version range that has import-protection.

### Workaround

We patched `resolveId` to skip `cleanId` specifically for import-protection's own virtual ids:

```diff
 resolveId: async (source, importer) => {
   const r = await this.resolve(source, importer)
   if (r) {
-    if (!r.external) return cleanId(r.id)
+    if (!r.external) return r.id.startsWith("\0tanstack-start-import-protection:") ? r.id : cleanId(r.id)
   }
   return null
 }
```

This fixes the build without disabling import-protection — the mock module resolves and loads correctly, exactly as it's designed to.

N/A — minimal repro described above, happy to provide a full reproduction repo if useful.

1. Scaffold a TanStack Start app using native-Vite mode (not vinxi) with `@tanstack/react-start` >= 1.165.0.
2. Add a `createServerFn().handler()` whose handler references (even transitively) something import-protection would mark as a denied import in the client environment.
3. Run `vite build`.
4. Observe the `Could not load tanstack-start-import-protection:mock-edge:...: ENOENT` crash.

### Expected behavior

The build should succeed, with the denied import correctly replaced by import-protection's generated mock module (as it does for the analogous cases that don't happen to route through the server-fn compiler's internal `resolveId`/`loadModule` static-analysis path).

### Screenshots or Videos

No response

### Platform

- OS: Linux (NixOS 25.11)
- Node: v22 / v26
- @tanstack/react-start: 1.165.0 through 1.168.27 (reproduces across this whole range)
- vite: 7.3.6
- Build mode: native Vite (not vinxi)

### Additional context

Related-but-distinct: the *older* pre-import-protection bug class (#4022, #3990, #2783 — "Readable is not exported by `__vite-browser-external`") is still present in versions *before* import-protection shipped (confirmed at 1.144.0/1.145.4 and 1.160.0). So currently there's no published version of `@tanstack/react-start` that builds a `createServerFn`-using native-Vite app cleanly without either this bug or that one, unless the fix above (or something equivalent) is applied.

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/TanStack/query/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://tanstack.com/start/latest/docs/framework/react/overview).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

rodriguescarson:
Fixes #7725

## Problem

In build mode, when `createServerFn`'s static analysis (the server-fn `start-compiler-plugin`) recursively loads a module whose dependency chain resolves to import-protection's `mock-edge` virtual module, the build crashes:

```
[tanstack-start-core::server-fn:client] Could not load tanstack-start-import-protection:mock-edge:<base64>: ENOENT: no such file or directory
```

## Root cause

In `packages/start-plugin-core/src/vite/start-compiler-plugin/plugin.ts`, the `resolveId` callback ran every resolved id through `cleanId()`:

```ts
if (!r.external) return cleanId(r.id)
```

`cleanId()` strips the leading `\0` that marks a Rollup-internal virtual module. That's correct for filesystem-backed ids, but import-protection's own ids (e.g. `\0tanstack-start-import-protection:mock-edge:...`) need that `\0`: the cleaned id is passed to `loadModule` → `this.load({ id })`, and import-protection's `load` handler filters on the `\0`-prefixed matchers (`getResolvedVirtualModuleMatchers()`). Without the prefix the filter never matches, so Rollup falls through to `vite:load-fallback`, which `fs.readFile()`s the bare string and throws `ENOENT`.

## Fix

Preserve the `\0` prefix for import-protection virtual ids; everything else is cleaned as before:

```ts
if (r.id.startsWith('\0tanstack-start-import-protection:')) {
  return r.id
}
return cleanId(r.id)
```

This mirrors the discriminator already used elsewhere in the plugin (`import-protection-plugin/plugin.ts` special-cases `\0tanstack-start-import-protection:` too), and it fixes the build without disabling import-protection — the mock module resolves and loads exactly as designed.

Credit to the detailed root-cause analysis in #7725.

## Verification

- `nx run @tanstack/start-plugin-core:test:build` (package + 12 dependency tasks) green.
- `test:types` (tsc) clean.
- ESLint clean on the changed file.
- `vitest run`: 486/486 tests pass, no type errors.

The crash itself only reproduces in a full `vite build` of a Start app with a denied import in a server-fn's dependency chain (per #7725's repro), so it isn't covered by a unit test here; happy to add an e2e case if you'd like one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of special internal module references so they continue to resolve correctly instead of falling through to normal file loading.
  * This helps prevent build or import issues in cases involving protected virtual modules.

### DIFF
--- a/packages/start-plugin-core/src/vite/start-compiler-plugin/plugin.ts
+++ b/packages/start-plugin-core/src/vite/start-compiler-plugin/plugin.ts
@@ -364,6 +364,15 @@ export function startCompilerPlugin(
 
                 if (r) {
                   if (!r.external) {
+                    // Keep import-protection's own virtual ids (e.g. mock-edge)
+                    // `\0`-prefixed: `loadModule`/`this.load()` routes them to the
+                    // import-protection `load` handler by matching on the `\0`
+                    // prefix. Stripping it via cleanId() makes the load fall
+                    // through to vite:load-fallback, which fs.readFile()s the
+                    // bare id and throws ENOENT in build mode (#7725).
+                    if (r.id.startsWith('\0tanstack-start-import-protection:')) {
+                      return r.id
+                    }
                     return cleanId(r.id)
                   }
                 }


LadyBluenotes:
Maintainers have upcoming compiler work planned, so I'm going to close this in favour of that work

rodriguescarson:
Makes sense, thanks for the heads-up @LadyBluenotes. Glad it'll be handled properly in the compiler work. Happy to help test once that lands.

PR #7728, opened 2026-07-02
fix(start-plugin-core): re-try failed paths correctly

roopakv:
retryCount / retryDelay on the prerender don't actually do anything right now.

We had a race condition where the service wasnt up yet so fetch failed, so adding a retry was a simple fix. The retry branch calls addCrawlPageTask(page) again, but that path is already in the seen set from the first attempt, so addCrawlPageTask early-returns and the task is never re-queued. The page just gets dropped. We never reach the failOnError throw, so the failure is swallowed and the build finishes with "Prerendered 0 pages" and exit code is 0 though some pages failed.

I ran into this in SPA mode, the shell prerender occasionally races the preview server and fails with ECONNREFUSED, and setting retryCount did nothing.

Fix is to drop the path from seen before re-queuing so the retry actually re-fetches. retriesByPath still bounds it, so it stops after retryCount attempts and falls through to failOnError.

Added a test that fails the first fetch with retryCount: 1 and checks the page is fetched twice.

Note: I used AI to debug and come up with this fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved prerender retry handling so pages that fail temporarily can be crawled again on retry.
  * Fixed an issue where retried pages could be skipped after an initial failed fetch.

* **Tests**
  * Added coverage for prerender retries to verify failed requests are retried successfully.

### DIFF
--- a/packages/start-plugin-core/src/prerender.ts
+++ b/packages/start-plugin-core/src/prerender.ts
@@ -219,6 +219,7 @@ export async function prerender({
             )
             await new Promise((resolve) => setTimeout(resolve, retryDelay))
             retriesByPath.set(page.path, retries + 1)
+            seen.delete(page.path)
             addCrawlPageTask(page)
           } else if (prerenderOptions.failOnError ?? true) {
             throw error
--- a/packages/start-plugin-core/tests/prerender-retry.test.ts
+++ b/packages/start-plugin-core/tests/prerender-retry.test.ts
@@ -1,0 +1,68 @@
+import { describe, expect, it, vi } from 'vitest'
+import { prerender } from '../src/prerender'
+
+vi.mock('../src/utils', async () => {
+  const actual = await vi.importActual<any>('../src/utils')
+  return {
+    ...actual,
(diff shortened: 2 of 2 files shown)


roopakv:
@schiller-manuel any chance you could take a look at this one? This is the last thing keeping us from switching to tanstack start entirely :)

Sheraff:
Hi @roopakv sorry we haven't yet had the time to get to this issue. But out of curiosity, why did you decide to close it? Has this been otherwise fixed?

roopakv:
Hi @Sheraff sorry I should have posted what I did instead. 

turns out this is still an issue, but my above fix is not the right one.

What I ended up doing is having the preview server always bind to 127.0.0.1, i think at times it was binding to the ipv6 maybe which is why it was flaky

```
preview: {
    host: '127.0.0.1',
  },
```

Since the above change yesterday we have had 0 flakes. 🤞 this the right fix. 

The other solution could be to have the pre-renderer try the right IP that the preview server binds to, but I want to look into that after I confirm that this is fix

Issue #7457, opened 2026-05-21
Uncaught undefined in in the <MatchInnerImpl> component


ryanagillie:
### Which project does this relate to?

Router

### Describe the bug

This is a bit of a complicated one, but I think I've found a race condition where an undefined is thrown inside the \<MatchInnerImpl\> component after a beforeLoad redirect that causes the routes to fail to display.

It only fails on first load, and usually a refresh or HMR will fix it.

I only wrote this because some of these exact conditions were met in a private work repository and it took me 2 hours to create a stable reproduction.

<details>

<summary>Stack Trace</Summary>

```
defaultOnUncaughtError          @ react-dom_client.js?v=be4189c8:5256
logUncaughtError                @ react-dom_client.js?v=be4189c8:5286
runWithFiberInDEV               @ react-dom_client.js?v=be4189c8:850
lane.callback                   @ react-dom_client.js?v=be4189c8:5314
callCallback                    @ react-dom_client.js?v=be4189c8:4094
commitCallbacks                 @ react-dom_client.js?v=be4189c8:4102
runWithFiberInDEV               @ react-dom_client.js?v=be4189c8:850
commitLayoutEffectOnFiber       @ react-dom_client.js?v=be4189c8:6985
flushLayoutEffects              @ react-dom_client.js?v=be4189c8:8670
commitRoot                      @ react-dom_client.js?v=be4189c8:8583
commitRootWhenReady             @ react-dom_client.js?v=be4189c8:8078
performWorkOnRoot               @ react-dom_client.js?v=be4189c8:8050
performWorkOnRootViaSchedulerTask @ react-dom_client.js?v=be4189c8:9058
performWorkUntilDeadline        @ react-dom_client.js?v=be4189c8:35

<MatchInnerImpl>
exports.jsx                     @ react_jsx-runtime.js?v=eeebbf31:192
MatchView                       @ @tanstack_react-rout….js?v=9b2a07b8:6384
react_stack_bottom_frame        @ react-dom_client.js?v=be4189c8:12867
renderWithHooksAgain            @ react-dom_client.js?v=be4189c8:4267
renderWithHooks                 @ react-dom_client.js?v=be4189c8:4218
updateFunctionComponent         @ react-dom_client.js?v=be4189c8:5568
beginWork                       @ react-dom_client.js?v=be4189c8:6139
runWithFiberInDEV               @ react-dom_client.js?v=be4189c8:850
[...]

Issue #7367, opened 2026-05-08
Pending spinner + throw redirect = `MatchInnerImpl` error


piscopancer:
### Describe the bug

Hey, I want to know why there's component race condition for a pending component when i throw redirects. The logic's simple - show a spinner for good looks right away, but check the user role in the background - then redirect to their dedicated routes

```ts
// __root.tsx
export const Route = createRootRouteWithContext<RouterContext>()({
  component: RootComponent,
  pendingMs: 0, // show spinnner immediately - error when `beforeLoad` is over. Set it to `1000` - no error
  pendingMinMs: 500,
  pendingComponent: () => {
    return <SpinnerScreen />;
  },
  errorComponent: (props) => <pre>{JSON.stringify(props.error, null, 2)}</pre>,
  async beforeLoad({ context, location }) {
    const user = await context.qc.ensureQueryData(sessionQueryOptions);
    if (location.pathname.startsWith(WelcomeRoute.to)) return;
    const redir = match(user.type)
      .with("edit_op", () =>
        !location.pathname.startsWith(EditSalesRoute.to)
          ? redirect({
              to: EditSalesRoute.to,
              replace: true,
            })
          : undefined,
      )
      .with(...)
      .exhaustive();
    if (redir) throw redir;
  },
});
```

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/f71838ac-67c1-4d97-ada0-c5961320e0a3" />

and you can see a white screen
when i reload the page it will be normal
when i set pendingMs to a 1000, a spinner will not show up but the error will be gone, luckily
though i need a spinner right away

### Your minimal, reproducible example

-

### Steps to reproduce

-

### Expected behavior

No react dom error must appear

### How often does this bug happen?

None

### Platform

Windows 10, Zen browser latest

### Tanstack Query adapter

None

### TanStack Query version

5.90.12

### TypeScript version

5.9.3

codeart1st:
It's probably the same as my issue #7120

piscopancer:
> It's probably the same as my issue #7120 

And did you find a solution? Currently, I think it's a bug and I even wanted to update the package. 

Also I don't find any other neat way to render a spinner because beforeLoad prevents react from rendering

piscopancer:
any update on this?

melv-n:
#7158 aims to address #7120

piscopancer:
> #7158 aims to address #7120

So there is no fix yet?

[...]

PR #7158, opened 2026-04-11
fix: handle stale redirected matches during view transitions

schiller-manuel:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a client-side crash where a redirect during navigation could blank the UI when view transitions and lazy routes were enabled.
  * Ensured pending UI displays reliably during such redirects instead of leaving the screen empty.

* **Tests**
  * Added an end-to-end reproduction test and fixture to prevent regressions and validate the pending UI/redirect behavior.

PR #7741, opened 2026-07-03
fix(react-router): Catch thrown nullish value and prevent nullish

joseinaqa:
loadPromise from being thrown

The issue happens in circumstances where match.status === 'redirected' but the loadPromise is nulled.
Reproduction is possible by using
https://github.com/joseinaqa/redirect-race-app and after this change does not happen anymore.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling by normalizing falsy “thrown” values so error boundaries always receive a real `Error`.
  * Preserved component stack trace details when available, improving diagnostics for caught errors.
  * Fixed Suspense/loading behavior so “pending” rendering doesn’t break when match resolution can be temporarily unavailable.
  * Ensured client-side rethrows always raise a concrete error instead of a potentially missing value.

### DIFF
--- a/packages/react-router/src/CatchBoundary.tsx
+++ b/packages/react-router/src/CatchBoundary.tsx
@@ -53,14 +53,18 @@ class CatchBoundaryImpl extends React.Component<{
     return { resetKey }
   }
   static getDerivedStateFromError(error: Error) {
-    return { error }
+    return { error: error || new Error('A falsy value was thrown') }
   }
   reset() {
     this.setState({ error: null })
   }
   componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+    if (error == null && this.state.error) {
+      this.state.error.stack =
+        errorInfo.componentStack ?? this.state.error.stack
+    }
     if (this.props.onCatch) {
-      this.props.onCatch(error, errorInfo)
+      this.props.onCatch(this.state.error ?? error, errorInfo)
     }
   }
   render() {
--- a/packages/react-router/src/Match.tsx
+++ b/packages/react-router/src/Match.tsx
@@ -287,6 +287,8 @@ function OnRendered() {
   return null
 }
 
+const permaPendingPromise = new Promise(() => {})
+
 export const MatchInner = React.memo(function MatchInnerImpl({
   matchId,
 }: {
@@ -338,15 +340,17 @@ export const MatchInner = React.memo(function MatchInnerImpl({
     const out = Comp ? <Comp key={key} /> : <Outlet />
 
     if (match._displayPending) {
-      throw getMatchPromise(match, 'displayPendingPromise')
+      throw (
+        getMatchPromise(match, 'displayPendingPromise') ?? permaPendingPromise
+      )
     }
 
     if (match._forcePending) {
-      throw getMatchPromise(match, 'minPendingPromise')
+      throw getMatchPromise(match, 'minPendingPromise') ?? permaPendingPromise
     }
 
     if (match.status === 'pending') {
-      throw getMatchPromise(match, 'loadPromise')
+      throw getMatchPromise(match, 'loadPromise') ?? permaPendingPromise
     }
 
     if (match.status === 'notFound') {
@@ -368,7 +372,7 @@ export const MatchInner = React.memo(function MatchInnerImpl({
 
         invariant()
       }
-      throw getMatchPromise(match, 'loadPromise')
+      throw getMatchPromise(match, 'loadPromise') ?? permaPendingPromise
     }
 
     if (match.status === 'error') {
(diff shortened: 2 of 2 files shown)


tristansinclair:
We hit this exact crash in production and independently arrived at essentially this same diff before finding this PR — so consider this a field test of the approach. Stack: React 19.2 SPA (no SSR), Vite, `@tanstack/react-router` 1.169.1 + TanStack Query, Clerk for auth, Sentry for error reporting. Sharing the evidence since the PR has no review yet:

**Debugger-confirmed throw site.** Sentry only showed a stackless `Uncaught undefined` (thrown `undefined` has no stack; the single frame is React's `defaultOnUncaughtError` → `reportError`). With CDP `Debugger.setPauseOnExceptions('all')` filtered to exception values of type `undefined`, every pause landed on `throw getMatchPromise(match, 'loadPromise')` in `MatchInnerImpl` — in our runs specifically the `status === 'pending'` branch — 16 throws in a single page load. The thrown `undefined` then escaped every `CatchBoundary` (falsy `if (error)` check), React unmounted the root, and users got a permanent white screen.

**Deterministic reproduction shape.** Cold load at `/` where `beforeLoad` awaits async work for a few hundred ms (for us: waiting on clerk-js readiness, then an API fetch) and finally `throw redirect(...)`, with `defaultPendingComponent` set, `defaultPendingMs: 200`, `defaultPendingMinMs: 300`. Driving that flow headlessly against our real app: **6/6 cold loads crashed** on stock 1.169.1.

**This PR's approach fixes it.** We shipped the equivalent change as a dist patch (below) and re-ran the same harness: **0/13 crashes** across the same conditions and harsher variants (CPU-throttled, delayed auth responses, multi-hop redirect chains), with all navigations completing normally.

Also verified the affected code is byte-identical in the latest release (1.170.17), so this isn't fixed by upgrading.

One correctness nit: `getDerivedStateFromError` normalizes with `error ?? new Error(...)`, but the render callback checks truthiness (`if (error)`). A thrown `0`, `''`, `false`, or `NaN` is non-nullish yet falsy, so it still slips through the boundary and unmounts the app the same way. `error || new Error(...)` closes the gap completely — an error boundary can't meaningfully render a falsy error anyway, so there's no behavior worth preserving there.

For anyone needing a stopgap while this PR is pending, here's the dist-level patch we're running in production via `bun patch` (same idea works with `pnpm patch` / `patch-package`):

<details>
<summary><code>@tanstack/react-router@1.169.1</code> dist patch (esm + cjs)</summary>

```diff
diff --git a/dist/esm/Match.js b/dist/esm/Match.js
--- a/dist/esm/Match.js
+++ b/dist/esm/Match.js
@@ -128,7 +128,7 @@ function OnRendered({ resetKey }) {
 var MatchInner = React$1.memo(function MatchInnerImpl({ matchId }) {
 	const router = useRouter();
 	const getMatchPromise = (match, key) => {
-		return router.getMatch(match.id)?._nonReactive[key] ?? match._nonReactive[key];
+		return router.getMatch(match.id)?._nonReactive[key] ?? match._nonReactive[key] ?? new Promise(() => {});
 	};
 	if (isServer ?? router.isServer) {
 		const match = router.stores.matchStores.get(matchId)?.get();
diff --git a/dist/esm/CatchBoundary.js b/dist/esm/CatchBoundary.js
--- a/dist/esm/CatchBoundary.js
+++ b/dist/esm/CatchBoundary.js
@@ -30,7 +30,7 @@ var CatchBoundaryImpl = class extends React$1.Component {
 		return { resetKey };
 	}
 	static getDerivedStateFromError(error) {
-		return { error };
+		return { error: error || new Error(`Unhandled falsy error: ${String(error)}`) };
 	}
 	reset() {
 		this.setState({ error: null });
diff --git a/dist/cjs/Match.cjs b/dist/cjs/Match.cjs
--- a/dist/cjs/Match.cjs
+++ b/dist/cjs/Match.cjs
@@ -130,7 +130,7 @@ function OnRendered({ resetKey }) {
 var MatchInner = react.memo(function MatchInnerImpl({ matchId }) {
 	const router = require_useRouter.useRouter();
 	const getMatchPromise = (match, key) => {
-		return router.getMatch(match.id)?._nonReactive[key] ?? match._nonReactive[key];
+		return router.getMatch(match.id)?._nonReactive[key] ?? match._nonReactive[key] ?? new Promise(() => {});
 	};
 	if (_tanstack_router_core_isServer.isServer ?? router.isServer) {
 		const match = router.stores.matchStores.get(matchId)?.get();
diff --git a/dist/cjs/CatchBoundary.cjs b/dist/cjs/CatchBoundary.cjs
--- a/dist/cjs/CatchBoundary.cjs
+++ b/dist/cjs/CatchBoundary.cjs
@@ -32,7 +32,7 @@ var CatchBoundaryImpl = class extends react.Component {
 		return { resetKey };
 	}
 	static getDerivedStateFromError(error) {
-		return { error };
+		return { error: error || new Error(`Unhandled falsy error: ${String(error)}`) };
 	}
 	reset() {
 		this.setState({ error: null });
```

Patching `getMatchPromise` itself covers all four throw sites (`displayPendingPromise`, `minPendingPromise`, and both `loadPromise` branches) with one hunk per module format. A missing promise means the match is being torn down mid-transition, so suspending forever is safe — React discards the suspended subtree as soon as the router commits the new matches.

</details>

Related reports this would resolve: #7753, #7457, #7367 (and it overlaps the scenario in #7120 / #7158).

Issue #7638, opened 2026-06-15
Crash with router.invalidate() inside startTransition


valerii15298:
### Which project does this relate to?

Start

### Describe the bug

When using `router.invalidate()` inside `startTransition` the app crashes. The necessary precondition is that some nested route should error out with showing `errorComponent`

### Complete minimal reproducer

https://github.com/valerii15298/tss-repros/tree/invalidate-transition-crash

### Steps to Reproduce the Bug

```bash
git clone https://github.com/valerii15298/tss-repros
cd tss-repros
git switch invalidate-transition-crash
pnpm install
pnpm dev
```

Open the *http://localhost:3000/*
Click [Go to Test](http://localhost:3000/test) link
Click *invalidate* button
Wait up to 5 seconds
The app crashes with white screen and error:
```
Error: Rendered more hooks than during the previous render.
```

### Expected behavior

The app should not crash

### Platform

- Router / Start Version: latest at the time of this issue(look into package json)
- OS: macOs
- Browser: any 
- Bundler: Vite
- Bundler Version: *vite@8.0.16* will be updated inside repro package json

valerii15298:
It took me a while to create this minimal repro. The first time I encountered it with using form actions with react, which I later abandoned.
First time I crate this discord question and reproduction attached with form actions: https://discord.com/channels/719702312431386674/1478737651116015730

Related twit: https://x.com/vpetryniak/status/2049533818030145684?s=20

Since then I abandoned using react form actions, startTransition, router.invalidate and router loaders because of it. But needed one small thing from it and was able to reproduce it.

pupuking723:
I'd like to investigate this. I'll reproduce the router.invalidate() + startTransition crash from the minimal repo on current main first, then add a regression case around the nested errorComponent path before changing implementation. If this overlaps with ongoing Start transition/error-boundary work, I'll report that here before opening a PR.

Issue #7379, opened 2026-05-11
`matches` in `head` property does not have `loaderData` or `context` present


mwkcoding:
### Which project does this relate to?

Router

### Describe the bug

When navigating to a route, `matches` does not have contain the `loaderData` or `context` properties the first time head is called. So in my StackBlitz example, if you navigate directly to the route, the title doesn't update because `loaderData` is not defined.

### Complete minimal reproducer

https://stackblitz.com/edit/github-qkpldktp

### Steps to Reproduce the Bug

1. Click "Open preview in new tab"
2. Click "Connect to Project"
3. Go to `nested/even/more/nested`
4. See the console logs that loaderData is not present on any of the matches

### Expected behavior

I expect that the `loaderData` and/or `context` property is present on `matches` from the head argument. That should also cause the title of the browser tab to become: "Nested - More - Even - Nested - Company Inc."

### Platform

- Router / Start Version: 1.169.2
- OS: macOS
- Browser: Chrome
- Browser Version: 148.0.7778.97 (Official Build) (arm64)
- Bundler: Vite
- Bundler Version: 8.0.12

### Additional context

My use case is that I may want some value from some query data that I'm loading in the `loader` property to show in the title and I want to dynamically "stack" to make a sort of breadcrumb in the title.

schiller-manuel:
> See the console logs that loaderData is not present on any of the matches

cannot reproduce

<img width="771" height="685" alt="Image" src="https://github.com/user-attachments/assets/12f944e1-5eef-4eb7-81c4-336c34082d6b" />

please be specific what you are missing

mwkcoding:
@schiller-manuel the bug occurs when directly navigating to the route. Not by clicking the link to it in the app. But navigating directly.

seankrossing:
@schiller-manuel It works as expected when you navigate between routes, the issue specifically occurs on the first page "load" so to speak

Issue #6371, opened 2026-01-12
React: using validateSearch with zod schema with default values makes the abortController abort the request on the loader: signal is aborted without reason


edcaron:
### Which project does this relate to?

Router

### Describe the bug

When using validateSearch with zod schema with default values makes the abortController abort the fetch request on the loader with "signal is aborted without reason" when entering the route via typing on the url without the default route parameters on the url. This only happens when directly entering the route via url. Navigating programmatically using Link works normally.

### Your Example Website or App

https://codesandbox.io/p/sandbox/tanstack-router-cancelation-issue-forked-fg3s4z

### Steps to Reproduce the Bug or Issue

1. go to the repo: https://codesandbox.io/p/sandbox/tanstack-router-cancelation-issue-forked-fg3s4z
2. open the preview
3. open the chrome devtools for the preview
4. click on the link "about"
5. edit the preview url and remove everything after /about and hit enter
6. check the  chrome devtools for the preview that will have the error

### Expected behavior

The behavior would be that the default route params would be added to the url and the content would be loaded using the fetch on the loader with no error.

### Screenshots or Videos

https://github.com/user-attachments/assets/3bf1de2f-778e-4245-8446-82c4f0a35c13

### Platform

- Router / Start Version: 1.147.3
- OS: linux
- Browser: Chromium
- Browser Version: : 143.0.7499.169
- Bundler: vite
- Bundler Version: 7.1.7

### Additional context

to make it happen: 

- must use zod with default values for props
- must use validateSearch with the zod schema
- must use fetch on the loader function and add the abortController.signal to the signal param on the fetch function

PR #7744, opened 2026-07-03
fix: match loading — review fixes and #7713 edge cases

Sheraff:
Review follow-up for #7637: fixes every issue found in review (including all edge cases reproduced in #7713) so that branch + this PR is the ready-to-merge version of `fix-match-loading`.

## Tests imported from #7713

All unit repros land as regular regression tests (descriptive names, internal issue numbering stripped), plus the `hydration-capped-assets` e2e app/spec. The view-transition repro (issue 6) was deliberately **not** imported: disabling typed view transitions on browsers without `:active-view-transition-type` support is accepted behavior.

## Fixes

| Area | Problem | Fix |
| --- | --- | --- |
| Load promise contract | `await router.load()`/`invalidate()` settled while a superseding load was still in flight; preloads borrowing a pending-published owner were dropped (`joinPreloadedActiveMatch` saw a stuck `'pending'` snapshot) | Superseded/redirected loads await the navigation chain before settling; the borrow protocol waits for the foreground commit when the re-read owner is a render-ready pending publication |
| Solid same-route pending | `pendingReplacement()` swapped committed content for the pending fallback (or `null`!) at `t=0`, bypassing `pendingMs`, and stamped `pendingUntil` at navigation start (delaying fast commits by up to `pendingMinMs`) | Branch removed; core's pending publication drives pending UI visibility, same as React |
| Vue pending crash | `MatchInner` dev-invarianted on a pending match whose local `loadPromise` was settled (hydration display matches, `cancelMatches`); main's `_displayPending` guard had been dropped | A settled promise is a legal stale-snapshot state; `pendingUntil` stamps only a still-pending local promise; `_displayPending` early-return restored |
| `cancelMatches` scope | Every active match's controller was aborted at load start — killing deferred/streamed data tied to a success stay-match's loader signal on any child navigation or `invalidate()` | Only `status === 'pending' \|\| isFetching === 'loader'` matches are cancelled (pending-pool cancellation unchanged), matching the documented signal contract |
| Error observability | The outer catch swallowed all non-redirect errors (incl. user `onEnter`/`onStay`/`onLeave` throws via the awaited view-transition callback); React's `isTransitioning` could stick forever | Non-sentinel failures re-surface as unhandled rejections; each lifecycle hook is isolated so one throw can't skip the rest; `startTransition` clears `isTransitioning` in `finally` |
| Pending-publication cache loss | Publication evicted the exiting lane from every pool; a superseded (back) navigation re-ran fresh within-`staleTime` loaders | `commitReady` preserves exiting success matches in `cachedMatches` (deduped in `commitFinalMatches`) |
| Asset projection | Server committed head/scripts/headers as one unit (a `head()` failure dropped route **headers**); background reloads published fresh `loaderData` under stale meta when projection failed; hydration executed `head()`/`scripts()` for matches the server omitted | Per-kind server commit (sync throws still don't block the response); background publication is atomic for data+assets (projection failure keeps the old lane); hydration skips assets for `ssr: false` matches |
| Route chunks | Targeted boundary requests awaited (and inherited failures from) the whole-route `_componentsPromise`; rejected preloads were cached forever; primary chunk failures bypassed `onError`/redirect/notFound | Per-component-type in-flight tracking (targeted requests join their own type only; full preloads never couple to a stale generation); rejections evicted for retry; primary chunk failures run `normalizeRouteFailure` (boundary-chunk failures still commit directly as the recursion guard) |
| Dehydrated boundaries (root cause of #7713 issue 7) | The follow-up client load after hydration skipped a dehydrated parent's `beforeLoad`, so children the server omitted at a notFound/error boundary were loaded, committed, and asset-projected with missing loader data | A dehydrated notFound/error boundary replays as the pass's serial failure, capping the lane exactly like the server did |

## Behavior decisions

- **Preload adoption restored (data-only)** — a navigation that clicks through an in-flight hover preload adopts its successful loader run: the loader executes once. Unlike the old join (which this PR's base had removed because preload-computed redirects could hijack navigations — main's own `skip if pending preload (redirect)` test codified landing on the wrong page — plus the #1886 undefined-context class), adoption is data-only: the navigation always runs its own `beforeLoad`, and a preload ending in redirect/notFound/error is never adopted — the navigation retries with its own loader. Joining is deadlock-safe by construction (gated on the donor's loader being in flight, past the borrow-protocol serial phase). Concurrent preloads of the same route dedupe to one loader run.
- **All-or-nothing preload caching replaced with successful-prefix caching** — a failed descendant no longer discards expensive successful ancestor loads; cache invariants hold (only owned success snapshots are cached).
- **Typed view-transition fallback unchanged** — accepted as-is.

## Cleanup

Byte-identical twin helpers (`commitMatch`, `getLoader`) moved to shared `load-matches.ts`; dead client-only fields no longer copied into the server load context; duplicated status-code derivation removed from `loadServerRouter`'s catch. INTERNALS.md and the changeset updated to match all semantic changes.

## Verification

- All repro tests from #7713 pass (unit + e2e); each new fix has a test that failed before the fix.
- Full unit + lint suites green: router-core (1452), react (931), solid (818), vue (789), start-server-core.
- `react-start/basic` e2e green in all three modes (SSR/prerender/SPA).
- Bundle size vs `main`: net **smaller** on every target except react-start rsbuild (+150–250 B); fixes cost ~250–620 B over the raw PR.
- Local SSR benchmarks vs `main`: −4 % to −17 % on document scenarios (no regressions beyond noise); client-nav −2 to −4.5 % vs main.

## Community issues addressed (issue/PR sweep)

A triage of open issues and PRs overlapping this work. Every integrated issue has a regression test asserting the desired behavior.

**Already fixed by this branch — now pinned with tests** (each `issue-<n>-*.test` file):
#7602, #7457, #7367, #7635, #4684, #6221, #7379, #4696, #4444, #6107, #5106, #7638, #4572, #3179, #4112, #2980, #2905, #2182, #3928, #6371 (#7120 is covered by the base PR's dedicated e2e suite).

**Fixed in this PR:**
- Fixes #4759 — `pendingMs: 0` deferred pending publication to a macrotask, painting one blank frame on bare initial loads; publication is now synchronous when nothing is rendered yet.
- Fixes #7283 — Solid dev-mode hydration mismatch with `ssr: false`/`data-only` + `pendingComponent`: the `_displayPending` branch rendered the fallback at a different structural depth than the server. It now renders through the shared `ClientOnly` slot (this also repairs a data-only regression the base PR introduced vs main). Covered by a dev-server e2e spec.

Fixes #7367, fixes #4684, fixes #7379, fixes #4696, fixes #4444, fixes #6107, fixes #5106, fixes #7638, fixes #4572, fixes #3179, fixes #4112, fixes #2980, fixes #2905, fixes #2182, fixes #3928, fixes #6371.

**Known gap, documented as an expected-fail test (follow-up needed):**
- #5778 — a preload borrowing an active ancestor derives context from its committed snapshot, not live `router.options.context`. Fixing requires recomputing borrowed ancestors' contexts from their stored layers — a change to the read-only borrow protocol that deserves its own PR. The documented `invalidate()` workaround is pinned as passing.

**External PRs:**
- Superseded by this branch, intent pinned with ported tests: #7741, #7051, #7003 (also covers #7006), #7114 (pinned via the #7110 repro), #6645.
- Judged not applicable: #6049 (scenario no longer reproduces under the rewritten serial phase).
- Coordinate separately (real, but out of scope here): #6598 (context invalidation vs background loading), #7608 (hydration pending — partially covered here, needs rebase), #7311 (deferred head loading feature), #7588 (response/status/header reconciliation).


packages/react-router/src/Transitioner.tsx
```diff
@@ -1,36 +1,108 @@
...
+
+    const subscriptions = [
+      router.stores.isLoading.subscribe(emitEdges),
+      router.stores.hasPending.subscribe(emitEdges),
+    ]
+    return () => {
+      for (const subscription of subscriptions) {
+        subscription.unsubscribe()
+      }
+    }
```
Sheraff:
```suggestion
    const loadingSub = router.stores.isLoading.subscribe(emitEdges)
    const pendingSub = router.stores.hasPending.subscribe(emitEdges)
    return () => {
      loadingSub.unsubscribe()
      pendingSub.unsubscribe()
    }
```

Sheraff:

PR #7746, opened 2026-07-04
chore(solid-router-v2-pre): merge main

brenelz:
Merges `main` into `solid-router-v2-pre` (22 commits, including the pnpm catalog migration and `minimumReleaseAge` supply-chain settings).

## Conflict resolutions

- **Solid packages/examples**: kept the v2 beta versions from this branch; carried over main's real changes hidden among the version bumps (`@rsbuild/core` `^2.0.8` → `^2.0.11` in `solid-start` and the rspack example).
- **`pnpm-workspace.yaml`**: adopted main's catalog + `minimumReleaseAge: 1440`, kept this branch's `minimumReleaseAgeExclude` (solid-js, `@solidjs/*`, babel-preset-solid, babel-plugin-jsx-dom-expressions) and scoped `babel-preset-solid` overrides. Dropped main's `solid-js: 1.9.12` override (would downgrade the Solid v2 toolchain) and set the catalog's `@tanstack/solid-query` to `^6.0.0-beta.5`.
- **Root `package.json`**: adopted `catalog:` references for the query deps.
- **`benchmarks/client-nav/package.json`**: kept both `@platformatic/flame` (this branch) and `@tanstack/router-plugin` (main).
- **`pnpm-lock.yaml`**: regenerated with `pnpm install --lockfile-only`. The `solid-js@1.9.12` entries that remain are the pre-existing intentional ones for the `vite-plugin-solid@2.x` consumers (devtools-core etc.).

## Verification

- `pnpm install` resolves cleanly
- `@tanstack/solid-router` builds
- solid-router unit tests: 813 passed, 2 skipped (46 files), no type errors


brenelz:
Two rounds of CI fixes pushed:

**\`01dbb52\` — bundle-size build failure** (\`Unmatched version @rspack/core@2.0.8 and @rspack/binding@2.0.0\`): regenerating the lockfile with \`pnpm install --lockfile-only\` on darwin-arm64 dropped every non-local platform binary from \`@rspack/binding@2.0.8\`'s \`optionalDependencies\`, so Linux CI fell back to a stale 2.0.0 binding. Restored the full platform list to match main's lockfile. Note: pnpm 11.9.0 reproducibly re-prunes these entries on any local re-resolve, so this may bite again on future lockfile updates on this branch.

**\`26d94d9\` — solid CodSpeed job failures**: main's new client-nav scenarios and memory benchmarks are written for Solid 1.x. Adapted them to this branch's Solid 2 toolchain (`@solidjs/web` imports, v2 deps in the memory manifests, `Suspense` → `Loading`, `createResource` → router `Await`). All solid benchmark builds and benches now pass locally, **except one**:

⚠️ **\`streaming-peak\` (solid) is skipped, and it exposes what looks like a real issue on this branch**: Solid 2's \`renderToStream\` emits \`</body></html>\` at the end of the initial shell and streams deferred \`Loading\`-boundary HTML *after* it. \`transformStreamWithRouter\` assumes deferred content arrives before \`</body>\` (true for React/Vue), so it tail-buffers all post-shell Solid chunks — capped at \`MAX_TAIL_CHARS = 64 kB\` — and throws \`SSR stream tail exceeded maximum buffer\` once the scenario's ~200 kB of deferred sections arrive. Smaller payloads don't crash but are held until stream end, so Solid streaming SSR is effectively buffered rather than progressive through the transformer. Verified empirically against the built scenario server: shell + fallbacks flush, seroval payload scripts flush, then the transformer dies at the third deferred app chunk.

brenelz:
Re-ran the failed \`Test\` job. The only real failure was \`tanstack-solid-start-e2e-basic:build:vite:prerender\`, which died with \`UND_ERR_HEADERS_TIMEOUT\` after prerendering 46/52 pages — one page's SSR (looks like \`/deferred\`) never returned response headers within undici's 5-minute cap.

Analysis suggests this is a **pre-existing load-sensitive race, not a merge regression**:

- The merge changes nothing in the vite prerender path — no \`start-plugin-core/prerender\`, no \`router-core/ssr\`, no solid-start source changes, and identical \`srvx\`/\`h3\`/\`isbot\`/\`seroval\` versions in the lockfile.
- \`/deferred\` responds reliably in isolation (15/15 runs, ~2 s each) against the built \`dist-vite-prerender\` server.
- Locally the full prerender is fine on an idle machine but intermittently hangs or drops pages (47/52) when the machine is under heavy load — consistent with a slow CI agent hitting it.

Mechanism worth noting: prerender requests carry no User-Agent, so \`isbot\` treats them as bots and \`renderRouterToStream\` awaits Solid's full-ready promise before sending headers. If Solid 2's all-ready promise loses a race under load, the response never starts and the whole build times out. Possibly related to the streaming/tail issue described above.

PR #7747, opened 2026-07-04
fix(router-core): keep match.pathname consistent across nested matches with param parsing

### LINKED ISSUE #7731: match.pathname is inconsistent across nested matches with parsed path params

Router

### Describe the bug

When a route uses `params.parse` and `params.stringify` to map a plural public URL slug to a singular internal value, nested/index matches can expose inconsistent `match.pathname` values.

The layout match keeps the raw URL segment, while the nested index/page match can use the parsed param value.

### Reproduction

```ts
const postTypeRoute = createFileRoute("/$postType")({
  params: {
    parse: ({ postType }) => ({
      postType: postType === "articles" ? "article" : postType,
    }),
    stringify: ({ postType }) => ({
      postType: postType === "article" ? "articles" : postType,
    }),
  },
})

const postTypeIndexRoute = createFileRoute("/$postType/")({})
```

Visit:

```txt
/articles
```

Observed via `useMatches()`:

```txt
/$postType   pathname: /articles   params.postType: article
/$postType/  pathname: /article/   params.postType: article
```

### Expected behavior

I would expect `match.pathname` values in the same match chain to be consistent. Ideally both should represent the current public URL or canonical stringified URL:

```txt
/articles
/articles
```

At minimum, it would be helpful to document that `match.pathname` can contain raw or parsed params depending on match depth when path params are parsed.

### Impact

This can cause duplicate or incorrect route-derived UI when using `useMatches()`, for example breadcrumbs, analytics paths, document metadata, or active navigation state.

### Current workaround

```ts
const location = router.buildLocation({
  to: match.fullPath,
  params: match.params,
})

const canonicalHref = location.publicHref
```

### Version

Observed with `@tanstack/react-router` 1.170.16.

### Question

Is this inconsistency expected, or should child/index matches use the same raw/stringified path param representation as parent matches?

momomuchu:
`match.pathname` is inconsistent across nested matches when `params.parse`/`params.stringify` are used: a parent route keeps the raw value in its own pathname, but its parsed value leaks into child routes' pathnames.

In `matchRoutesInternal` (`packages/router-core/src/router.ts`), each match's pathname is built with `interpolatePath` from a single shared `routeParams` object. After building a route's pathname, `Object.assign(routeParams, strictParams)` wrote that route's `params.parse` result back into the same shared object, which is then used to interpolate the child routes' pathnames.

This keeps `routeParams` raw for all `interpolatePath` calls and accumulates parsed params separately (`accumulatedStrictParams`), used only for `match.params`. Parsed-param inheritance semantics are preserved.

Added a `match-params.test.ts` case asserting pathname consistency across nested matches with parse/stringify. It fails without the fix and passes with it. router-core plus the react/solid/vue router downstream suites pass, tsc clean.

Closes #7731

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nested route parameter propagation so parsed parameter values are consistently available across all matched routes.
  * Fixed an issue where strict/parsed params could unintentionally influence later pathname interpolation, ensuring `match.pathname` stays aligned with the original URL segments.
  * Resolved consistency problems for routes that transform dynamic segments using `params.parse`/`params.stringify`.

* **Tests**
  * Added coverage to verify nested pathname and parameter consistency when dynamic segments are remapped via `params.parse`/`params.stringify`.

### DIFF
--- a/.changeset/lazy-poems-repeat.md
+++ b/.changeset/lazy-poems-repeat.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
+fix: keep match.pathname consistent across nested matches when a route uses params.parse
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -1448,6 +1448,13 @@ export class RouterCore<
     const { foundRoute, routeParams } = matchedRoutesResult
     let { matchedRoutes } = matchedRoutesResult
     let isGlobalNotFound = false
+    // Accumulates each route's strict (params.parse-applied) params as we walk
+    // down the match chain, so that `match.params` on a child match includes
+    // its ancestors' parsed params. Kept separate from `routeParams`, which
+    // must stay as the raw, un-parsed params used to interpolate `match.pathname`
+    // for every route in the chain -- otherwise a parent route's params.parse
+    // would leak its parsed value into the pathname of routes matched afterward.
+    const accumulatedStrictParams: Record<string, unknown> = Object.create(null)
 
     // Check to see if the route needs a 404 entry
     if (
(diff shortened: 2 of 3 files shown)


LadyBluenotes:
Could you add a regression for nested `params.parse` composition? The current test proves `pathname` consistency, but it does not prove a child `params.parse` still receives its parent’s parsed value.

PR #7751, opened 2026-07-07
fix(router-generator): make buildRouteTree route ordering deterministic

### LINKED ISSUE #7750: buildRouteTree sorts sibling routes with an inconsistent comparator ((d) => d), producing non-deterministic routeTree.gen.ts ordering
**Which project does this relate to?**

`@tanstack/router-generator` (affects `@tanstack/react-start` / `@tanstack/router-plugin` codegen). Reproduced on `main` and in the latest release (`1.167.18`).

**Describe the bug**

The generated `routeTree.gen.ts` reorders sibling routes non-deterministically across machines/CI, even with identical route files. Two routes that tie on segment-count and index-token (e.g. `/account/beta-features` and `/account/busy-guard`) swap places between generations, dirtying the committed route tree with semantically-identical churn. In a shared repo this repeatedly leaks unrelated route-tree diffs into PRs, since each author's machine may regenerate a different order than what's committed.

It is **not** filesystem-read-order dependent (a stable read order still hits it) — the cause is a broken comparator, so the resulting order depends on V8's sort internals, which vary by Node version and by the surrounding set of routes.

**Root cause**

In `packages/router-generator/src/generator.ts`, `buildRouteTree` sorts the route nodes with `multiSortBy`, whose **final** accessor is `(d) => d`:

```ts
const sortedRouteNodes = multiSortBy(acc.routeNodes, [
  (d) => (d.routePath?.includes(`/${rootPathId}`) ? -1 : 1),
  (d) =>
    d.routePath === undefined
      ? undefined
      : countSlashSeparatedParts(d.routePath),
  (d) => {
    const segments = d.routePath?.split('/').filter(Boolean) ?? []
    const last = segments[segments.length - 1] ?? ''
    return indexTokenSegmentRegex.test(last) ? -1 : 1
  },
  (d) => d,            // <-- compares whole route-node objects
])
```

`multiSortBy`'s comparator ends with `return ao > bo ? 1 : -1`. When `ao`/`bo` are route-node **objects**, both coerce to the string `"[object Object]"`, so `ao === bo` is `false` (distinct references) and `ao > bo` is `false` — meaning it returns `-1` for **both** `(a, b)` and `(b, a)`. That is an inconsistent comparator, so `Array.prototype.sort` leaves routes that tie on the earlier keys in an order that depends on the engine's sort implementation, the input order, and the array size — hence the cross-machine churn.

Note the **pre-sort** in the same file ends with `(d) => d.routePath` (a proper string total order), which is almost certainly what `buildRouteTree`'s sort intended as well.

**Your Example Website or App / Steps to Reproduce**

Any file-based project with two sibling routes that share a parent and segment depth (e.g. `src/routes/account.beta-features.tsx` and `src/routes/account.busy-guard.tsx`). Generate the route tree on two different Node versions (or after adding/removing an unrelated route) and diff `routeTree.gen.ts` — the two entries swap places with no other change.

**Expected behavior**

A fresh generation is byte-identical regardless of Node version or route-set composition, so the committed `routeTree.gen.ts` stays stable.

**Proposed fix**

Change the final accessor in `buildRouteTree`'s `multiSortBy` from `(d) => d` to `(d) => d.routePath`, mirroring the pre-sort and giving the tiebreaker a total order. PR incoming.

**Platform**

- router-generator: `1.167.18` (and `main`)
- Reproduced across Node 20 / 22 / 24

0x80:
Fixes #7750.

### Problem

`buildRouteTree` sorts route nodes with `multiSortBy`, whose **final** accessor was `(d) => d` — comparing whole route-node objects:

```ts
const sortedRouteNodes = multiSortBy(acc.routeNodes, [
  (d) => (d.routePath?.includes(`/${rootPathId}`) ? -1 : 1),
  (d) => (d.routePath === undefined ? undefined : countSlashSeparatedParts(d.routePath)),
  (d) => { /* index-token check */ },
  (d) => d,            // <-- whole object
])
```

`multiSortBy`'s comparator ends with `return ao > bo ? 1 : -1`. For two distinct route-node objects, both coerce to `"[object Object]"`, so `ao === bo` is `false` and `ao > bo` is `false` → it returns `-1` for **both** `(a, b)` and `(b, a)`. That's an inconsistent comparator, so `Array.prototype.sort` leaves routes that tie on the earlier keys (same segment-count, same index-token status — e.g. `/account/beta-features` vs `/account/busy-guard`) in an order that depends on the engine's sort implementation, the input order, and the array size. The result: `routeTree.gen.ts` reorders non-deterministically across machines and Node versions, dirtying committed route trees with semantically-identical churn.

### Fix

Change the final tiebreaker from `(d) => d` to `(d) => d.routePath`, mirroring the pre-sort earlier in the same file (which already ends with `(d) => d.routePath`). `routePath` is unique per node, so this gives a total order and makes generation byte-identical regardless of Node version or route-set composition.

A changeset is included (`@tanstack/router-generator`, patch).

### Tests

The `buildRouteTree` tiebreaker sort is extracted into an exported pure `sortRouteNodes(routeNodes, indexTokenSegmentRegex)` helper (behavior-identical to the previous inline `multiSortBy` call) so it can be unit-tested directly. New tests in `router-generator/tests/utils.test.ts`:

- **Stable ordering** — feeds a route set that ties on every key before `routePath` (e.g. `/account/beta-features` vs `/account/busy-guard`, `/about` vs `/posts`) through **every** input permutation and asserts identical output, proving the sort is independent of input order, engine sort implementation, and array size. This test fails against the old `(d) => d` tiebreaker and passes against `(d) => d.routePath`.
- **Tiebreaker + precedence** — asserts tied siblings order by `routePath`, and that root / segment-count / index-token precedence still apply ahead of the tiebreaker.

### Regenerated snapshots

The deterministic tiebreaker changes the order in which route nodes are emitted, so the 38 committed `routeTree.snapshot.*` files under `tests/generator/` are regenerated. This is expected: the previous snapshots captured the *arbitrary* order the inconsistent comparator happened to produce on the machine that generated them. The buggy sort was stable run-to-run for a fixed input order and engine, so CI kept passing — the non-determinism only appeared *across* machines/Node versions (issue #7750).

The snapshot diffs are **pure emission-order reordering**: every route's `id`, `path`, and `parentRoute` is unchanged, and the route-tree structure (built from `acc.routeTree`, not the sorted list) is untouched — only the order of imports and route declarations shifts. In other words, the regenerated snapshots are exactly the semantically-identical churn this fix eliminates going forward, applied once.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed non-deterministic route-tree generation when routes are equivalent, ensuring consistent ordering across runs and machines.
  * Stabilized tie-breaking so generated route outputs no longer vary based on engine/input quirks.
* **Tests**
  * Added new unit tests to validate deterministic sorting across permutations and tie scenarios.
  * Updated generated route-tree snapshots to match the new stable ordering.

### DIFF
--- a/packages/router-generator/src/index.ts
+++ b/packages/router-generator/src/index.ts
@@ -23,7 +23,6 @@ export {
   removeUnderscores,
   resetRegex,
   multiSortBy,
-  sortRouteNodes,
   writeIfDifferent,
   format,
   removeExt,
--- a/packages/router-generator/src/index.ts
+++ b/packages/router-generator/src/index.ts
@@ -23,6 +23,7 @@ export {
   removeUnderscores,
   resetRegex,
   multiSortBy,
+  sortRouteNodes,
   writeIfDifferent,
   format,
   removeExt,
--- a/packages/router-generator/tests/utils.test.ts
+++ b/packages/router-generator/tests/utils.test.ts
@@ -981,7 +981,9 @@ describe('sortRouteNodes', () => {
 
   /** Every distinct ordering of `arr` (Heap-free recursive permutation). */
   function permutations<T>(arr: Array<T>): Array<Array<T>> {
-    if (arr.length <= 1) return [arr]
+    if (arr.length <= 1) {
+      return [arr]
+    }
     const result: Array<Array<T>> = []
     arr.forEach((item, i) => {
       const rest = [...arr.slice(0, i), ...arr.slice(i + 1)]
--- a/packages/router-generator/tests/utils.test.ts
+++ b/packages/router-generator/tests/utils.test.ts
@@ -6,6 +6,7 @@ import {
   createRoutePathSegmentMetadata,
   createRouteNodesByFullPath,
   createRouteNodesByTo,
+  createTokenRegex,
   determineInitialRoutePath,
   hasEscapedLeadingUnderscore,
   hasEscapedTrailingUnderscore,
@@ -21,6 +22,7 @@ import {
   removeUnderscores,
   removeUnderscoresWithEscape,
   routePathToVariable,
+  sortRouteNodes,
 } from '../src/utils'
 import type { ImportDeclaration, RouteNode } from '../src/types'
 
@@ -963,5 +965,85 @@ describe('RoutePrefixMap', () => {
 
       expect(map.findParent('/users')).toBeNull()
     })
+  })
+})
+
+describe('sortRouteNodes', () => {
+  const indexTokenSegmentRegex = createTokenRegex('index', { type: 'segment' })
+
+  const node = (routePath: string): RouteNode =>
+    ({ routePath }) as unknown as RouteNode
+
+  const sortPaths = (routePaths: Array<string>): Array<string | undefined> =>
+    sortRouteNodes(routePaths.map(node), indexTokenSegmentRegex).map(
+      (n) => n.routePath,
+    )
+
+  /** Every distinct ordering of `arr` (Heap-free recursive permutation). */
+  function permutations<T>(arr: Array<T>): Array<Array<T>> {
+    if (arr.length <= 1) return [arr]
+    const result: Array<Array<T>> = []
+    arr.forEach((item, i) => {
+      const rest = [...arr.slice(0, i), ...arr.slice(i + 1)]
+      for (const perm of permutations(rest)) {
+        result.push([item, ...perm])
+      }
+    })
+    return result
+  }
+
+  it('orders tied siblings by routePath', () => {
+    // Same segment count, both non-index: they tie on every key before routePath.
+    const expected = ['/account/beta-features', '/account/busy-guard']
+    expect(
+      sortPaths(['/account/busy-guard', '/account/beta-features']),
+    ).toEqual(expected)
+    expect(
+      sortPaths(['/account/beta-features', '/account/busy-guard']),
+    ).toEqual(expected)
+  })
+
+  it('applies root, segment-count and index precedence before the routePath tiebreaker', () => {
+    const sorted = sortPaths([
+      '/posts',
+      '/account/busy-guard',
+      '/index',
+      '/__root',
+      '/account/beta-features',
+      '/about',
+      '/posts/index',
+    ])
+
+    expect(sorted).toEqual([
+      '/__root', // root always first
+      '/index', // depth 1, index segment before non-index
+      '/about', // depth 1, non-index, routePath tiebreak
+      '/posts', // depth 1, non-index, routePath tiebreak
+      '/posts/index', // depth 2, index segment before non-index
+      '/account/beta-features', // depth 2, non-index, routePath tiebreak
+      '/account/busy-guard', // depth 2, non-index, routePath tiebreak
+    ])
+  })
+
+  it('is deterministic for every input permutation (stable sort)', () => {
+    // Includes several sets that tie on all keys before routePath (the two
+    // `/account/*` leaves, `/about` vs `/posts`) — exactly the case that the
+    // former whole-object tiebreaker left in an engine- and input-order-
+    // dependent order. `sortRouteNodes` must map every permutation to the
+    // same output.
+    const routePaths = [
+      '/__root',
+      '/index',
+      '/about',
+      '/posts',
+      '/account/beta-features',
+      '/account/busy-guard',
+    ]
+
+    const reference = sortPaths(routePaths)
+
+    for (const permutation of permutations(routePaths)) {
+      expect(sortPaths(permutation)).toEqual(reference)
+    }
   })
 })
--- a/.changeset/deterministic-route-tree-sort.md
+++ b/.changeset/deterministic-route-tree-sort.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/router-generator": patch
+'@tanstack/router-generator': patch
 ---
 
 Make `buildRouteTree` route ordering deterministic. Its final sort tiebreaker compared whole route-node objects (`(d) => d`), which coerce to `"[object Object]"` and so never order — an inconsistent comparator that left routes tying on segment-count/index-token in an engine- and input-dependent order, producing non-deterministic `routeTree.gen.ts` diffs across machines. The tiebreaker now compares `routePath`, mirroring the pre-sort and giving a stable total order.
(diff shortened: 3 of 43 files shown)


schiller-manuel:
can you please add unit tests that prove stable sorting?

0x80:
@schiller-manuel added — pushed in 85b7e95.

To make the sort testable I pulled the `buildRouteTree` tiebreaker out into an exported pure `sortRouteNodes(routeNodes, indexTokenSegmentRegex)` (behavior-identical to the old inline `multiSortBy` call) and covered it in `router-generator/tests/utils.test.ts`. The stable-sort test builds a route set that ties on every key before `routePath` (`/account/beta-features` vs `/account/busy-guard`, `/about` vs `/posts`, …), runs it through **every** input permutation, and asserts identical output — so it's independent of input order, engine sort implementation, and array size. It fails against the old `(d) => d` tiebreaker and passes against `(d) => d.routePath`. A second test pins the `routePath` tiebreak plus the root / segment-count / index-token precedence ahead of it.

Heads-up: the deterministic order also changes route emission order, so I had to regenerate the 38 `routeTree.snapshot.*` fixtures under `tests/generator/`. The diffs are pure reordering — every route's `id`/`path`/`parentRoute` and the tree structure are unchanged. The old snapshots captured the arbitrary order the inconsistent comparator produced on whatever machine generated them (stable run-to-run for a fixed input, which is why CI stayed green — the non-determinism only showed up across machines, #7750). More detail in the updated PR description.

packages/router-generator/tests/utils.test.ts
```diff
@@ -965,3 +967,83 @@ describe('RoutePrefixMap', () => {
...
+    ({ routePath }) as unknown as RouteNode
+
+  const sortPaths = (routePaths: Array<string>): Array<string | undefined> =>
+    sortRouteNodes(routePaths.map(node), indexTokenSegmentRegex).map(
+      (n) => n.routePath,
+    )
+
+  /** Every distinct ordering of `arr` (Heap-free recursive permutation). */
+  function permutations<T>(arr: Array<T>): Array<Array<T>> {
+    if (arr.length <= 1) return [arr]
```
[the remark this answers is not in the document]
0x80: Good catch — fixed in e83e34b to match the AGENTS.md control-statement convention.

packages/router-generator/src/index.ts
```diff
@@ -23,6 +23,7 @@ export {
   removeUnderscores,
   resetRegex,
   multiSortBy,
+  sortRouteNodes,
```
LadyBluenotes: Did you intend to make this a public API?  Tests already import it from utils so if extraction is only for unit testing, can we keep it internal and omit this re-export?

0x80: No, that was unintentional — the extraction was only for unit testing, and both the test and generator.ts import it directly from ./utils. Dropped the re-export in a07c93cf02 so it stays internal.

LadyBluenotes (approved):
LGTM. Going to wait for @schiller-manuel to give the final sign off :)

0x80:
### Note on the failing checks

All three red checks are environmental artifacts of this PR coming from a fork, not defects in the change. The code checks — `Test`, `Benchmark PR`, all CodSpeed benchmark *runs*, CodeRabbit, Socket, zizmor, Preview — are green.

**`labeler`** and **`Upsert PR Comment`** both fail with `403 Resource not accessible by integration`. On a fork PR, GitHub hands the workflow a read-only `GITHUB_TOKEN` (`Contents: read, PullRequests: read`), so neither setting labels nor posting the benchmark comment is permitted. These pass once run with write scope in the upstream context; there is nothing in the diff to change.

**CodSpeed Performance Analysis** flags a single regressed benchmark, `client-route-tree-scale navigation loop (react)`, at 74.2 ms → 77.2 ms (-3.87%) — against 3 improved benchmarks, 176 untouched, and a net +51% overall. This is almost certainly measurement variance: the only change here is the tiebreaker in the generator's `sortRouteNodes` (`(d) => d` → `(d) => d.routePath`), which alters the textual order of route declarations in `routeTree.gen.ts`. The runtime router re-sorts routes by score at load time, so generator ordering has no effect on the navigation matching loop this benchmark exercises. Happy to acknowledge it on CodSpeed.

schiller-manuel:
@0x80 please don't copy LLM output here. this is not helping. it makes me want to close this PR instantly

0x80:
@schiller-manuel I think my LLM understood and explained things better than I would, so I thought it was helpful, and I didn't want to spend time rewriting its conclusions in my own words just to hide that I'm using an LLM. Feel free to close this if you think LLMs are not good enough for making PRs on your repo. 

Without it, I wouldn't never found the time to file a PR for this.

LadyBluenotes:
> @schiller-manuel I think my LLM understood and explained things better than I would, so I thought it was helpful, and I didn't want to spend time rewriting its conclusions in my own words just to hide that I'm using an LLM. Feel free to close this if you think LLMs are not good enough for making PRs on your repo.
> 
> Without it, I wouldn't never found the time to file a PR for this.

It's not that we have an issue with people using LLMs. Most of us use them in one form or another.

The issue is when LLM output is copied verbatim into a PR without sufficient review, validation, or evidence that the author understands and agrees with the conclusions being presented (we don't know whether you read it, whether a bot just put it in, etc). When someone submits a change, we would like to have the confidence that they can explain and understand the reasoning behind it

Framing it as "I didn't want to spend time rewriting it" is also part of the concern. Maintainers/reviewers are expected to spend their time understanding, validating, and maintaining contributions. It's difficult to justify asking reviewers to invest that effort when the PR author is saying that even a minimal level of review/writing something in their own words wasn't worth their own time.

Using an LLM to help analyze an issue is completely fine. We just expect contributors to take ownership of the final content they submit.

0x80:
@LadyBluenotes I understand your frustration, and I've experienced similar tension myself on open source projects I maintain. Whatever I can do with an LLM you can do too, so next time I will only report the issue and leave the PR/solution to you if I feel that I don't have enough time to really take ownership.

I tried to turn off external PRs on my projects recently, because I got onto an argument about PR ownership, and I'd rather have people only report issues, but Github doesn't let you turn off just external contributor PRs as far as I know.

schiller-manuel:
reporting issues with minimal reproducers is actually way more helpful in general, so yes that's probably a better way then just let an agent fix it

Issue #7213, opened 2026-04-17
Server functions that are stored in a separate file cause middlewares to throw error


petervmeijgaard:
### Which project does this relate to?

Start

### Describe the bug

When a middleware is using a server function that's placed in a different file, using said middleware in a different server function will cause the following error:

> Server function info not found for [ID]

It's working fine when the server function is defined in the same file as the middleware

### Your Example Website or App

- [GitHub](https://github.com/petervmeijgaard/tanstack-start-reproduction)
- [Stackblitz](https://stackblitz.com/~/github.com/petervmeijgaard/tanstack-start-reproduction)

### Steps to Reproduce the Bug or Issue

1. Add server function in a new file (i.e. `./server-fns/get-foo-bar.ts`)

```ts
import { createServerFn } from "@tanstack/react-start";

export const getFooBar = createServerFn({ method: "GET" }).handler(
  async () => {
    return { foo: "bar" };
  },
);
```

2. Use this server function in a middleware that's stored in a separate file (i.e. `./middlewares/foo-bar-middleware`)

```ts
import { createMiddleware } from "@tanstack/react-start";

import { getFooBar } from "#/server-fns/get-foo-bar";

export const fooBarMiddleware = createMiddleware({ type: "function" }).server(
  async ({ next }) => {
    const context = await getFooBar();

    return next({ context });
  },
);

```

3. Use the middleware in a server function that's used within a route (i.e. `./routes/index.tsx`)

```tsx
import { createFileRoute } from "@tanstack/react-router";

import { fooBarMiddleware } from "#/middleware/foo-bar-middleware";
import { createServerFn } from "@tanstack/react-start";

const getFooBar = createServerFn({ method: "GET" })
  .middleware([fooBarMiddleware])
  .handler(async ({ context }) => {
    console.log(context);

    return context.foo;
  });

export const Route = createFileRoute("/broken")({
  component: RouteComponent,
  loader: async () => {
    const foo = await getFooBar();

    return { foo };
  },
});

function RouteComponent() {
  const foo = Route.useLoaderData({
    select: (ctx) => ctx.foo,
  });

  return <h1>{foo}</h1>;
}
```

4. Start the development server and visit the route
5. See that the page is working as expected
6. Build the application
7. Start the production server and visit the same page
8. See that the page is broken and the following error is shown:

> Server function info not found for [ID]

### Expected behavior

As a user, I expected that this would work and that no error was thrown. However, this results in an production breaking exception

[...]

PR #7752, opened 2026-07-07
Fix/server fn cross file manifest

plauclair-nesto:
# Fix: server functions missing from the production manifest (TanStack Start, issue #7213)

**Repository:** fork of `TanStack/router`
**Branch:** `fix/server-fn-cross-file-manifest`
**Commits:** `8b53769ee4` (Playwright bump) · `d33b368de7` (the fix)
**Upstream issue:** https://github.com/TanStack/router/issues/7213

---

## 1. Executive summary

TanStack Start lets you write **server functions** — functions that run on the server and are callable from the browser over HTTP (an RPC). In production builds, a whole category of server functions was silently left out of the server's lookup table ("the manifest"), so calling them crashed at runtime with:

> `Server function info not found for <id>`

Crucially, this only happened in **production builds**, not in development — so it was the kind of bug that passes local testing and breaks in production.

The affected category: any server function that is only ever reached from **server-side code** — for example, one called inside a middleware, or called by another server function. Organising code into separate files (a completely normal practice) reliably triggered it.

**The fix** makes the build finish discovering these server functions before it writes out the manifest. It is a **one-file change** to the build tooling (`@tanstack/start-plugin-core`), shipped as a patch. We validated it with new automated tests and the full existing test suite across all three supported frameworks (React, Solid, Vue). Build-time impact is negligible (single-digit milliseconds, within measurement noise). There is no security or API change.

---

## 2. Background

A server function is declared with `createServerFn().handler(...)`. At build time the tooling splits each one into two pieces: a small **client stub** (shipped to the browser, makes an HTTP call) and the **real implementation** (kept on the server). The server keeps a **manifest** — a map from a function's ID to its implementation — so that when a request arrives (or when server code calls the function in-process), it can find and run the right implementation. If a function is missing from that manifest, any attempt to resolve it throws `Server function info not found`.

---

## 3. The bug

**Symptom.** Production builds throw `Server function info not found for <id>` at runtime. Development works fine. Single-file setups work fine. The error appears once server functions and the code that uses them are split across files.

**What triggers it.** A server function that is reached **only through server-only code**. Two common shapes:
- A **middleware** whose `.server()` handler calls a server function defined in another file.
- A **server function whose handler calls another server function** in another file.

**Why it matters.** Both shapes are ordinary, encouraged patterns (shared middleware, shared server-side helpers, code organised into modules). The failure is production-only, so it evades local development and typically surfaces only after deploy. The error message is opaque (an internal ID, no source location).

**Where the error originates.** `packages/start-plugin-core/src/start-compiler/server-fn-resolver-module.ts:53` — the generated resolver throws when an ID is absent from the manifest.

---

## 4. Root cause

Three facts combine into a timing race.

**(a) A server function is registered only when its file is compiled.**
The build registers a server function into the shared registry (`serverFnsById`) as a side effect of compiling the file that defines it — see `packages/start-plugin-core/src/start-compiler/handleCreateServerFn.ts:350` and `:476`. If a definition file never gets compiled before the manifest is written, the function is absent.

**(b) The client build cannot see server-only functions.**
The browser build strips out server-only code: a middleware's `.server()` body is removed (`packages/start-plugin-core/src/start-compiler/handleCreateMiddleware.ts:68`), and a server function's handler body is replaced by a client stub. Dead-code elimination then drops the now-unused imports. So a function reached *only* from that stripped code never enters the client build's module graph, and is never registered during the client pass.

**(c) The manifest is generated once, mid-build, possibly too early.**
The manifest lives in a "resolver" virtual module that is generated exactly once, the first time the server runtime imports it — `packages/start-plugin-core/src/vite/start-compiler-plugin/plugin.ts:750` (the `load()` hook) calling `generateServerFnResolverModule(...)` at `:792`. The build order is: **client environment fully, then server environment** (`packages/start-plugin-core/src/vite/planning.ts:204` then `:208`). The resolver is imported early in the server build — *before* the server graph has transitively compiled every server-only definition file.

**Net effect:** the manifest ends up containing essentially "every server function the client references." Server-only functions are systematically excluded, because the client can't see them (b) and the server writes the manifest before it has finished finding them (c).

**Why dev works.** In development there is no pre-baked manifest; the dev resolver decodes the target from the request and imports it on demand, so nothing needs to be discovered ahead of time.

**Why it's specific to the default setup.** The bug is confined to the configuration where the SSR environment is also the server-function "provider" (`ssrIsProvider`, the Vite default). In the alternative setup (a separate provider environment, e.g. RSC), that environment builds *last* (`planning.ts:211+`), so discovery is already complete by the time it writes the manifest. We confirmed the reference `rsbuild` bundler is unaffected — it regenerates the resolver *after* the graph is fully built, which Vite did not do.

**Figure 1 — why the function goes missing:**

```mermaid
flowchart TB
    A["Server function in its own file,<br/>reached only from server-only code"]

    subgraph CB["Client build — runs first, in full"]
      C1["Strips server-only code, so this function<br/>is never compiled or registered"]
    end

    subgraph SB["Server build — runs second"]
      S1["Manifest generated once, early —<br/>before this function's file is compiled"]
    end

    A --> C1
    A --> S1
    C1 --> MISS["Function absent from the manifest"]
    S1 --> MISS
    MISS --> ERR["Production runtime error:<br/>Server function info not found"]

    classDef err fill:#fdecea,stroke:#d1352c,color:#611000
    class ERR err
```

---

## 5. The fix

**Idea:** before the server build writes the manifest, force the reachable server module graph to finish compiling, so discovery is complete.

All changes are in `packages/start-plugin-core/src/vite/start-compiler-plugin/plugin.ts`.

**Gate (only runs where the bug exists).** `plugin.ts:781` — the discovery step runs only when `ssrIsProvider`. For the separate-provider/RSC case it is skipped, because that environment already builds after discovery completes. This also avoids doing redundant work (and avoids running during RSC's throwaway "scan" build pass).

**Two-phase discovery** — `ensureServerModuleGraphTransformed(...)` at `plugin.ts:270`, awaited at `:782`:

- **Phase 1 — bounded walk of the app's own source** (`plugin.ts:278`). Starting from the app's already-known modules, it loads each one and follows its imports, which forces them to compile (and thus register their server functions). This is deliberately **bounded to the project's own source** via `shouldWalkForServerFns` (`plugin.ts:239`) — it excludes `node_modules`, virtual modules, and anything outside the project root. That bound is essential: it stops the walk from wandering into the framework runtime (which is large and, importantly, is what imports the resolver module we are in the middle of generating — walking into it would deadlock the build).

- **Phase 2 — location-agnostic sweep of provider modules** (`plugin.ts:317`). A function called from inside another function's *handler* is dropped from the normal (caller) module, so Phase 1 alone misses it. But its implementation lives in a dedicated "provider" module whose handler body is intact. Phase 2 loads the provider module of **every currently-known server function** and lets loading pull in its direct dependencies — registering the nested functions those handlers call. It repeats until no new functions appear. Because a provider module is *always* a user/shared server-function implementation (never the framework runtime), it is safe to load regardless of where it lives — which is what lets this correctly discover server functions defined in **separate packages**, without reintroducing the framework-walk problem that Phase 1's bound guards against.

**One comment refresh.** The surrounding comment at `plugin.ts:792` was updated to reflect that, in this configuration, server functions are intentionally callable both in-process on the server and directly from the client (see §6, Security), matching the reference `rsbuild`/RSC behavior.

**Figure 2 — the fix: finish discovery before writing the manifest:**

```mermaid
flowchart TB
    L["Resolver load(): before writing the manifest"]
    L --> G{"Is SSR the provider?<br/>(Vite default)"}
    G -->|"No — provider env builds last,<br/>already complete"| GEN
    G -->|Yes| P1

    subgraph P1["Phase 1 — bounded walk of app source"]
      P1a["Compile app modules and follow imports;<br/>skip framework runtime and node_modules"]
      P1b["Registers functions reached from intact<br/>server-only bodies, e.g. middleware .server()"]
      P1a --> P1b
    end

    subgraph P2["Phase 2 — provider-module fixpoint"]
      P2a["Load every known function's provider module<br/>(handler body intact; any location)"]
      P2b["Registers handler-nested and<br/>separate-package functions"]
      P2c["Repeat until no new functions appear"]
      P2a --> P2b --> P2c
    end

    P1 --> P2
    P2 --> GEN["Generate manifest — now complete"]

    classDef ok fill:#eaf7ea,stroke:#2e7d32,color:#143314
    class GEN ok
```

---

## 6. Implications

**Behavior — previously-broken patterns now work.** Server functions reached only via a middleware, only via another server function's handler, arbitrarily nested chains, and functions defined in a **separate shared package** are all now present in the production manifest and resolve correctly at runtime. Nothing that worked before changes.

**Security — no change; matches the reference implementation.** Server functions were already, by design, callable from the client (directly by ID, or via serialization — see `createSsrRpc.ts:10` for the in-process path and the resolver's client-origin gate at `server-fn-resolver-module.ts:42`). In the default `ssrIsProvider` configuration this gate is intentionally off in **both** Vite and the reference `rsbuild` bundler, and `rsbuild` already includes these functions in its manifest. Our fix brings Vite in line with that existing, intended behavior — it does not newly expose anything that `rsbuild` didn't already expose. (The library's documented intent: server functions are callable anywhere — over HTTP from the client, in-process on the server, skipping the network hop. Our change only fixes discovery; it does not touch dispatch, the client stub, or the origin gate.)

**Performance — negligible.** We measured production SSR build time with the discovery step ON vs OFF, several runs each: `website` ≈ +3ms, `basic` ≈ +27ms (~8%, within noise), `server-functions` (largest, 263 SSR modules) ≈ +10ms (~2%). The step largely front-loads work the build already does (module loads are cached), so its apparent internal cost (~400ms) is overwhelmingly overlapping, not additive. Caveat: our test apps top out at ~260 SSR modules; if a very large app ever shows cost, an easy follow-up is to parallelize the walk.

**Cross-framework.** The change is in the framework-agnostic core, so it applies to React, Solid, and Vue Start identically. All three were validated (see §7).

**Scope / limitations.** The change is confined to the `ssrIsProvider` Vite build path; dev, the RSC/custom-provider path, and the `rsbuild` bundler are untouched by construction and were verified unaffected.

---

## 7. Testing & validation

**New automated regression tests** (in `e2e/react-start/server-functions/tests/server-functions.spec.ts`), each reproducing a shape that was broken:
- `:721` — server function reached only through a cross-file **middleware**.
- `:750` — server function called only from **another server function's handler**.
- `:777` — a **multi-level cross-file chain** (A → B → C), which specifically exercises the Phase-2 loop iterating more than once.
- `:797` — a server function in a **separate package** whose handler calls a nested server-only function in that same package. Backed by a small fixture package, `e2e/react-start/server-functions-shared-lib/` (see `src/index.ts:7` and `src/internal.ts:7`).

All four fail on the original code (with the exact `Server function info not found` error) and pass with the fix.

**Full suite — all green:**
- Package gate (`nx affected`): `test:eslint`, `test:types`, `test:unit`, `test:build` — all pass.
- React Start e2e: `server-functions` 57, `rsc` 258, `basic` 134, `server-functions-global-middleware` 7, `static-server-functions` 5, `server-routes` 6, `serialization-adapters` 8.
- Solid Start e2e: `server-functions` 29, `server-routes` 2.
- Vue Start e2e: `server-functions` 27, `server-routes` 2.

No `Server function info not found` anywhere. The `serialization-adapters` suite passing is notable — it exercises the "return a server function to the client and call it" path, confirming the fix doesn't disturb the callable-anywhere behavior.

---

## 8. Risk assessment

- **Blast radius:** one file of build tooling; runtime dispatch, client code, and the public API are untouched.
- **Not a regression risk:** before the fix these functions were entirely missing (broken); the change strictly adds correct entries. Everything that already worked still works (verified by the full suite).
- **Deadlock/perf risks considered and closed:** Phase 1's project-root bound prevents walking into the framework runtime (which would deadlock on the in-progress resolver); Phase 2 only loads known server-function implementations, keeping it bounded. Both were validated on real builds, including the multi-environment RSC build.
- **Confined by a feature gate:** only the affected configuration runs the new code path.

---

## 9. How it's packaged

Two clean, separated commits on `fix/server-fn-cross-file-manifest`:

1. `8b53769ee4` — **`chore(deps)`**: bumps `@playwright/test` to `^1.61.0` across the workspace. Independent of the fix; resolves intermittent test-runner stalling. (117 files: package manifests + lockfile.)
2. `d33b368de7` — **`fix(start-plugin-core)`**: the one-file fix, a Changesets entry (`@tanstack/start-plugin-core` patch, `.changeset/fix-server-fn-manifest-cross-file.md`), the four regression tests, and the fixture package. (17 files.)

The Changesets entry means the fix will be released as a patch version of `@tanstack/start-plugin-core`, cascading patch bumps to the React/Solid/Vue Start packages automatically.

---

## 10. Key file reference (quick index)

| Concern | Location |
|---|---|
| Error thrown when a fn is missing | `packages/start-plugin-core/src/start-compiler/server-fn-resolver-module.ts:53` |
| Client-origin access gate | `.../server-fn-resolver-module.ts:42` |
| Where a server fn is registered | `.../start-compiler/handleCreateServerFn.ts:350`, `:476` |
| `.server()` body stripped on client | `.../start-compiler/handleCreateMiddleware.ts:68` |
| Build order (client → server) | `.../vite/planning.ts:204`, `:208`, `:211` |
| In-process server-to-server call | `packages/start-server-core/src/createSsrRpc.ts:10` |
| **The fix — discovery gate** | `.../vite/start-compiler-plugin/plugin.ts:781` |
| **The fix — two-phase discovery** | `.../start-compiler-plugin/plugin.ts:270` (Phase 1 `:278`, Phase 2 `:317`) |
| **The fix — walk bound** | `.../start-compiler-plugin/plugin.ts:239` |
| Manifest generation | `.../start-compiler-plugin/plugin.ts:792` |
| Regression tests | `e2e/react-start/server-functions/tests/server-functions.spec.ts:721,750,777,797` |
| Separate-package fixture | `e2e/react-start/server-functions-shared-lib/src/{index,internal}.ts` |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved production reliability of server-function manifest generation, preventing missing-function runtime errors.
  * Fixed edge cases where server functions reachable only through middleware, nested server-function chains, or separate shared packages could fail to be discovered/loaded.

* **Tests**
  * Added end-to-end regression coverage for middleware-invoked and nested/shared server functions in production scenarios.

* **Chores**
  * Updated Playwright (`@playwright/test`) to a newer version across the end-to-end test projects and workspace configuration.

### DIFF
--- a/.changeset/fix-server-fn-manifest-cross-file.md
+++ b/.changeset/fix-server-fn-manifest-cross-file.md
@@ -1,0 +1,19 @@
+---
+"@tanstack/start-plugin-core": patch
+---
+
+fix(start-plugin-core): register server functions reachable only from server-only code in the production manifest
+
+When SSR is the server-function provider (the Vite default), a server function
+reachable only from server-only code — e.g. a middleware `.server()` body or
+another server function's handler — could be omitted from the generated server
+function manifest. Calls then failed at runtime in production with
+`Server function info not found for <id>`, while dev builds worked. This happened
+because the resolver module is generated once and could be produced before those
+functions had been discovered.
+
+The server-function resolver now completes discovery of the reachable server
+module graph (including provider modules, so handler-nested and external-package
+server functions are covered) before generating the manifest.
+
+Fixes #7213.
--- a/e2e/react-router/basepath-file-based/package.json
+++ b/e2e/react-router/basepath-file-based/package.json
@@ -18,7 +18,7 @@
     "react-dom": "^19.0.0"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/basic-esbuild-file-based/package.json
+++ b/e2e/react-router/basic-esbuild-file-based/package.json
@@ -19,7 +19,7 @@
     "zod": "^4.4.3"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/basic-file-based-code-splitting/package.json
+++ b/e2e/react-router/basic-file-based-code-splitting/package.json
@@ -21,7 +21,7 @@
     "zod": "^4.4.3"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/basic-file-based/package.json
+++ b/e2e/react-router/basic-file-based/package.json
@@ -23,7 +23,7 @@
     "zod": "^4.4.3"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/basic-react-query-file-based/package.json
+++ b/e2e/react-router/basic-react-query-file-based/package.json
@@ -24,7 +24,7 @@
     "zod": "^4.4.3"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/basic-react-query/package.json
+++ b/e2e/react-router/basic-react-query/package.json
@@ -22,7 +22,7 @@
     "tailwindcss": "^4.2.2"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/basic-scroll-restoration/package.json
+++ b/e2e/react-router/basic-scroll-restoration/package.json
@@ -21,7 +21,7 @@
     "tailwindcss": "^4.2.2"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/basic-virtual-file-based/package.json
+++ b/e2e/react-router/basic-virtual-file-based/package.json
@@ -23,7 +23,7 @@
     "zod": "^4.4.3"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/basic-virtual-named-export-config-file-based/package.json
+++ b/e2e/react-router/basic-virtual-named-export-config-file-based/package.json
@@ -23,7 +23,7 @@
     "zod": "^4.4.3"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/basic/package.json
+++ b/e2e/react-router/basic/package.json
@@ -20,7 +20,7 @@
     "tailwindcss": "^4.2.2"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/escaped-special-strings/package.json
+++ b/e2e/react-router/escaped-special-strings/package.json
@@ -18,7 +18,7 @@
     "react-dom": "^19.0.0"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
+    "@playwright/test": "^1.61.0",
     "@tanstack/router-e2e-utils": "workspace:^",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
--- a/e2e/react-router/generator-cli-only/package.json
+++ b/e2e/react-router/generator-cli-only/package.json
@@ -21,7 +21,7 @@
     "tailwindcss": "^4.2.2"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.1",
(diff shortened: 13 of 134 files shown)


LadyBluenotes:
Thanks for the detailed investigation and the cross-file regression coverage. This change modifies how the compiler discovers server functions and builds the production manifest, which overlaps with broader compiler work planned by the maintainers. We aren’t accepting compiler-related PRs right now, so we’re going to close this one, but we’ll keep #7213 open to track the underlying issue

PR #7762, opened 2026-07-08
fix(router-core): guard against evicted matches during in-flight preloads

### LINKED ISSUE #7759: TypeError: Cannot read properties of undefined (reading '_nonReactive') from preloadRoute when an in-flight preload's match is evicted

Router

### Describe the bug

When a route is being preloaded (e.g. `defaultPreload: 'intent'`) and its cached match is evicted while the preload's async work is still in flight — most commonly because the user clicks the link and actually navigates mid-preload, or because `router.invalidate()` / `clearExpiredCache()` runs — the preload pipeline crashes with an unhandled rejection:

```
TypeError: Cannot read properties of undefined (reading '_nonReactive')
    at Re (Match-uwMccdg0.js:1:14474)
    at async Promise.all (/index 1)
    at async ze (Match-uwMccdg0.js:1:15299)
    at async Ka.preloadRoute (Match-uwMccdg0.js:1:36895)
```

(minified production stack; the frames are `load-matches.ts` internals called from `preloadRoute`)

The root cause is that several code paths in `packages/router-core/src/load-matches.ts` re-look up the match with `inner.router.getMatch(matchId)` after an `await` and read `._nonReactive` without guarding against the match having been removed from `state.matches` / `pendingMatches` / `cachedMatches` in the meantime, e.g. the non-null assertions here (current `main`):

- `preBeforeLoadSetup`: `const existingMatch = inner.router.getMatch(matchId)!` followed by `existingMatch._nonReactive.beforeLoadPromise`
- `executeBeforeLoad`: `const match = inner.router.getMatch(matchId)!` followed by `match._nonReactive.loadPromise`
- similar unguarded reads in the loader-run/cleanup paths

For a preload, the match only lives in `cachedMatches`, so any eviction during the in-flight async work (lazy chunk load, `beforeLoad`, loader) makes the next `getMatch(matchId)` return `undefined` and the next `._nonReactive` access throw.

Navigation itself still succeeds (the real navigation builds its own matches), so the impact is an unhandled rejection that pollutes the console and error tracking (Sentry etc.) on every hover-preload-then-click race.

### Steps to reproduce

1. Create a router with `defaultPreload: 'intent'` and a route whose loader (or lazy component chunk) takes a moment to resolve.
2. Hover a `<Link>` to that route so `preloadRoute` starts.
3. Before the preload settles, click the link (or call `router.invalidate()`).
4. Observe the unhandled `TypeError: Cannot read properties of undefined (reading '_nonReactive')` in the console.

The regression tests added in #7003 / #7006 (`clearExpiredCache()` / `invalidate()` while a `preloadRoute` loader is unresolved) reproduce the same race deterministically.

### Expected behavior

An in-flight preload whose match has been evicted should clean up quietly (resolve/clear its controlled promises and pending timeout) instead of throwing an unhandled TypeError.

Reproducible in production on x.com (hover the login link on the logged-out homepage, then click it while the preload is in flight).

### Platform

- OS: macOS
- Browser: Chrome
- Version: @tanstack/react-router 1.170.17 / @tanstack/router-core 1.171.14 (unguarded reads still present on `main`)

### Additional context

There are two open PRs that appear to address exactly this: #7003 and #7006. Filing this issue since neither PR links to a tracking issue and the crash is otherwise undocumented (closest existing report is #7753, which is a different symptom of a cleared `loadPromise`).

thribhuvan003:
when a preload's cached match gets evicted while its async work is still in flight (cache cleared, gc'd, or the user actually navigates mid-preload), the load pipeline re-looks the match up after each `await` with a non-null assertion, so the next `._nonReactive` read throws `TypeError: Cannot read properties of undefined (reading '_nonReactive')` as an unhandled rejection.

this guards those re-lookups the same way `executeHead` and `runLoader`'s catch already do. the post-load cleanup falls back to the matched object instead of bailing — `_nonReactive` keeps its identity across store updates, so resolving the promises on it means a second load awaiting the first one's `loaderPromise` still settles instead of hanging forever (that hang is real: the second regression test times out without the fix).

added two deterministic tests (evict mid-loader, and a second preload joining the in-flight load). both fail without the fix, and router-core (1186) + react-router (912) suites pass with it.

fixes #7759

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened route preloading and loader flow to safely handle cached match eviction during in-flight operations.
  * Prevented crashes and ensured preload/loader promises settle cleanly even when the underlying match disappears mid-load.

* **Tests**
  * Added coverage for cache eviction during active preloads, including repeated preload calls, verifying both promises settle without error logging.

### DIFF
--- a/.changeset/fix-preload-evicted-match-guards.md
+++ b/.changeset/fix-preload-evicted-match-guards.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/router-core': patch
+---
+
+fix(router-core): guard against evicted matches during in-flight preloads
+
+When a preload's cached match is evicted while its async work is still in flight (e.g. the cache is cleared, garbage-collected, or the user navigates mid-preload), the load pipeline re-looked-up the match after each `await` with a non-null assertion and crashed with `TypeError: Cannot read properties of undefined (reading '_nonReactive')`. The re-lookups are now guarded, and the post-load cleanup falls back to the matched object so in-flight promises still settle and concurrent loads awaiting them don't hang.
+
+Fixes #7759.
--- a/packages/router-core/src/load-matches.ts
+++ b/packages/router-core/src/load-matches.ts
@@ -370,7 +370,9 @@ const preBeforeLoadSetup = (
   setupPendingTimeout(inner, matchId, route, existingMatch)
 
   const then = () => {
-    const match = inner.router.getMatch(matchId)!
+    const match = inner.router.getMatch(matchId)
+    // in case the match was evicted while awaiting the previous beforeLoad
+    if (!match) return
     if (
       match.preload &&
       (match.status === 'redirected' || match.status === 'notFound')
@@ -391,7 +393,10 @@ const executeBeforeLoad = (
   index: number,
   route: AnyRoute,
 ): void | Promise<void> => {
-  const match = inner.router.getMatch(matchId)!
+  const match = inner.router.getMatch(matchId)
+  // in case the match was evicted while `preBeforeLoadSetup` was awaiting the
+  // previous beforeLoad
+  if (!match) return
 
   // explicitly capture the previous loadPromise
   let prevLoadPromise = match._nonReactive.loadPromise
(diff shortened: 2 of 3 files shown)


LadyBluenotes:
Closing this in favor of #7805, which includes #7759 in its scope and replaces the `load-matches.ts` path changed here. If that rewrite does not land, we can revisit this fix

PR #7765, opened 2026-07-09
fix(start-server-core): copy immutable Response.redirect() headers when merging pending set-cookie values

### LINKED ISSUE #7755: [start] Unhandled 500 (TypeError: immutable) when a server route returns Response.redirect() while a set-cookie is pending on the event

Start

### Describe the bug

When a server route handler returns a `Response` whose headers are immutable — most commonly `Response.redirect(url)`, but also a passed-through `fetch()` response — and a cookie is pending on the h3 event (via `setCookie()`, or indirectly via better-auth's `tanstackStartCookies()` plugin which calls `setCookie` after every auth API call), the request fails with an unhandled 500 instead of returning the redirect:

```json
{"status":500,"unhandled":true,"message":"HTTPError"}
```

Root cause is `mergeEventResponseHeaders` in `@tanstack/start-server-core` (`src/request-response.ts`):

```js
function mergeEventResponseHeaders(response, event) {
  if (response.ok) return;
  const eventSetCookies = getSetCookieValues(event.res.headers);
  if (eventSetCookies.length === 0) return;
  const responseSetCookies = getSetCookieValues(response.headers);
  response.headers.delete("set-cookie"); // ← throws TypeError: immutable
  ...
}
```

A 3xx redirect is not `response.ok`, so the merge path runs, and `Response.redirect()` returns a response with an immutable headers guard per the Fetch spec. The `TypeError` propagates as an unhandled `HTTPError` 500.

This is easy to hit in production without realizing: any authenticated server route that ends in `Response.redirect(...)` breaks as soon as the auth library refreshes a session cookie during the request (better-auth with `cookieCache` does this on every `getSession`). The handler completes successfully — side effects and DB writes land — and then the framework 500s while post-processing the response, which makes it look like a routing/dispatch failure. Meanwhile unauthenticated requests to the same route (no pending cookie → early return) work fine, which is very confusing to debug.

Minimal repro route below (any Start app):

```ts
// src/routes/api/redirect-repro.ts
import { setCookie } from "@tanstack/react-start/server";

export const Route = createFileRoute("/api/redirect-repro")({
  server: {
    handlers: {
      GET: () => {
        setCookie("example", "value"); // anything that puts a pending cookie on the event
        return Response.redirect("https://example.com/", 302);
      },
    },
  },
});
```

1. Add the route above to a Start app.
2. `curl -i http://localhost:3000/api/redirect-repro`
3. Observe `500` with body `{"status":500,"unhandled":true,"message":"HTTPError"}` instead of a `302`.

Also reproducible directly against the dist internals:

```js
import { requestHandler, setCookie } from "@tanstack/start-server-core/dist/esm/request-response.js";

const handler = requestHandler(async () => {
  setCookie("example", "value");
  return Response.redirect("https://example.com/", 302);
});
const res = await handler(new Request("http://localhost/api/x"));
console.log(res.status); // 500, cause: TypeError: immutable at mergeEventResponseHeaders
```

### Expected behavior

The redirect should be returned (with the pending set-cookie headers merged). When the returned response has immutable headers, `mergeEventResponseHeaders` could clone it instead of mutating in place, e.g.:

```js
// new Response(response.body, response) yields mutable headers
```

or guard the mutation with a try/catch fallback to a cloned response.

### Platform

- OS: macOS / Cloudflare Workers (workerd) — reproduces on both Node (undici) and workerd, since both enforce the immutable headers guard
- `@tanstack/react-start`: 1.168.27
- `@tanstack/start-server-core`: 1.169.16

### Additional context

Workaround for anyone hitting this: return `new Response(null, { status: 302, headers: { location: url } })` instead of `Response.redirect(url)` from server route handlers.

Possibly related to the change that introduced event-cookie merging for non-2xx responses (#7189 / #5464) — the merge itself is desirable, it just can't assume the handler's response headers are mutable.

thribhuvan003:
returning `Response.redirect()` from a handler after calling `setCookie()` crashes the whole request — `mergeEventResponseHeaders` calls `.delete('set-cookie')` on the redirect's headers, which are immutable per spec, so the merge throws `TypeError: Headers are immutable` and the user gets a 500 instead of their redirect + cookie.

fixed by copying the response (same status/statusText/body, fresh mutable headers) when the in-place delete throws, and merging into the copy. mutable responses keep the exact in-place behavior they had before.

added two tests: the redirect case fails without the fix and passes with it, and a mutable non-ok response still merges the same way. the three test files already failing in this package (`createCsrfMiddleware`, `createStartHandler`, `frame-protocol`) fail identically on a clean checkout of main, so they're unrelated.

fixes #7755

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed redirects and other non-OK responses so cookies set during a request are preserved instead of causing a server error.
  * Improved handling of responses with immutable headers, preventing failed handler requests and keeping redirect destinations intact.

* **Tests**
  * Added coverage for cookie merging with redirect responses and other responses that already include headers.

### DIFF
--- a/.changeset/redirect-immutable-headers.md
+++ b/.changeset/redirect-immutable-headers.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/start-server-core': patch
+---
+
+fix(start-server-core): merge pending set-cookie headers into `Response.redirect()` without throwing
+
+`Response.redirect()` returns a response whose headers are immutable per spec, so merging pending event cookies into it crashed the request with `TypeError: Headers are immutable`. The merge now copies the response when its headers can't be modified, so redirects returned from handlers keep their cookies instead of turning into a 500.
+
+Fixes #7755.
--- a/packages/start-server-core/src/request-response.ts
+++ b/packages/start-server-core/src/request-response.ts
@@ -78,24 +78,40 @@ function getSetCookieValues(headers: Headers): Array<string> {
   return value ? [value] : []
 }
 
-function mergeEventResponseHeaders(response: Response, event: H3Event): void {
+function mergeEventResponseHeaders(
+  response: Response,
+  event: H3Event,
+): Response {
   if (response.ok) {
-    return
+    return response
   }
 
   const eventSetCookies = getSetCookieValues(event.res.headers)
   if (eventSetCookies.length === 0) {
-    return
+    return response
   }
 
   const responseSetCookies = getSetCookieValues(response.headers)
-  response.headers.delete('set-cookie')
+  let target = response
+  try {
+    response.headers.delete('set-cookie')
+  } catch {
+    // Response.redirect() responses have immutable headers per spec,
+    // so merge into a mutable copy instead
+    target = new Response(response.body, {
+      status: response.status,
+      statusText: response.statusText,
+      headers: response.headers,
+    })
+    target.headers.delete('set-cookie')
+  }
   for (const cookie of responseSetCookies) {
-    response.headers.append('set-cookie', cookie)
+    target.headers.append('set-cookie', cookie)
   }
   for (const cookie of eventSetCookies) {
-    response.headers.append('set-cookie', cookie)
+    target.headers.append('set-cookie', cookie)
   }
+  return target
 }
 
 function attachResponseHeaders<T>(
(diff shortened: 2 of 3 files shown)


LadyBluenotes:
Closing this PR for now because it was opened before #7755 was triaged by a maintainer. We'll keep the issue opened, but please wait til a maintainer has had time to go over it

PR #7770, opened 2026-07-10
fix(start-plugin-core): merge user router.plugins into Start generator plugins

### LINKED ISSUE #7768: Start: user router.plugins never reach the route generator (start-plugin-core overwrites them)

Start

### Describe the bug

Generator plugins passed as `tanstackStart({ router: { plugins: [...] } })` never reach the route generator — no hook (`init`, `onRouteTreeChanged`, `afterTransform`) is ever invoked, in dev or build, with no error or warning.

Cause: `tanStackStartRouter` builds its internal generator-plugin list and places it **after** `...routerConfig` in the config it hands to `tanstackRouterGenerator`, so the user's `plugins` key is overwritten:

https://github.com/TanStack/router/blob/4eed408f12e0dee55708b1e2b53fefb3441c414d/packages/start-plugin-core/src/vite/start-router-plugin/plugin.ts#L147-L159

```ts
tanstackRouterGenerator(() => {
  const routerConfig = getConfig().startConfig.router
  const plugins = [clientTreeGeneratorPlugin, routesManifestPlugin()]
  if (startPluginOpts.prerender?.enabled === true) {
    plugins.push(prerenderRoutesPlugin())
  }
  return {
    ...routerConfig,          // includes the user's routerConfig.plugins…
    target: corePluginOpts.framework,
    routeTreeFileFooter: getRouteTreeFileFooter,
    plugins,                  // …which is clobbered here
  }
}, routerPluginContext)
```

The same options work correctly with the plain `tanstackRouter` Vite plugin (no Start), where user config flows through `getConfig` untouched — so this is Start-specific.

Suggested fix (verified locally via a patched package):

```ts
if (routerConfig.plugins) plugins.push(...routerConfig.plugins)
```

```ts
// vite.config.ts — minimal repro

export default defineConfig({
  plugins: [
    tanstackStart({
      router: {
        plugins: [
          {
            name: 'probe',
            init: () => console.error('[probe] init'),
            onRouteTreeChanged: () => console.error('[probe] onRouteTreeChanged'),
          },
        ],
      },
    }),
  ],
})
```

1. Add the probe generator plugin above to any TanStack Start app's `vite.config.ts`
2. Run `vite dev` or `vite build`
3. Neither `[probe]` line is printed; with the plain `tanstackRouter()` plugin (non-Start), both print

### Expected behavior

User-supplied `router.plugins` are merged with Start's internal generator plugins and their hooks fire on generator runs — or, if generator plugins are intentionally unsupported under Start, a loud error instead of silent acceptance.

### Platform

- @tanstack/react-start: 1.168.27 (start-plugin-core 1.171.19; bug still present on `main`)
- OS: macOS

### Additional context

Footnote once plugins do run: `createRouterGeneratorPlugin` wraps `generator.run()` in `catch (e) { console.error(e) }`, so a generator plugin that throws (e.g. to enforce a project invariant) can log but never fail the build. May be intentional, but it compounds the silence here.

thribhuvan003:
`tanstackStart({ router: { plugins: [...] } })` was silently dropping generator plugins. Start builds its own internal plugin list and was setting `plugins` after spreading `routerConfig`, so the user key got overwritten.

fix appends user plugins after Start's internals on both the vite and rsbuild paths. added a small merge helper + unit tests (3/3).

Fixes #7768.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved user-configured router generator plugins when using TanStack Start.
  * User plugins are now combined with Start’s built-in plugins instead of being overwritten.

* **Tests**
  * Added coverage for plugin merging, including empty configurations and protection against modifying existing plugin lists.

### DIFF
--- a/.changeset/start-merge-user-generator-plugins.md
+++ b/.changeset/start-merge-user-generator-plugins.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/start-plugin-core': patch
+---
+
+fix(start-plugin-core): merge user `router.plugins` into Start's generator plugins instead of overwriting them
+
+`tanstackStart({ router: { plugins: [...] } })` was dropping user generator plugins because Start built its internal plugin list after spreading `routerConfig`, which clobbered the user's `plugins` key. User plugins are now appended after Start's internals (vite + rsbuild).
+
+Fixes #7768.
--- a/packages/start-plugin-core/src/rsbuild/start-router-plugin.ts
+++ b/packages/start-plugin-core/src/rsbuild/start-router-plugin.ts
@@ -7,6 +7,7 @@ import {
 import { routesManifestPlugin } from '../start-router-plugin/generator-plugins/routes-manifest-plugin'
 import { prerenderRoutesPlugin } from '../start-router-plugin/generator-plugins/prerender-routes-plugin'
 import { buildRouteTreeFileFooterFromConfig } from '../start-router-plugin/route-tree-footer'
+import { mergeStartGeneratorPlugins } from '../start-router-plugin/merge-generator-plugins'
 import { RSBUILD_ENVIRONMENT_NAMES } from './planning'
 import type { RsbuildPluginAPI } from '@rsbuild/core'
 import type { GetConfigFn, TanStackStartCoreOptions } from '../types'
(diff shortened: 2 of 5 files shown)


schiller-manuel:
router plugins are not a public API

PR #7776, opened 2026-07-10
fix(router-core): support parsed params in matchRoute

### LINKED ISSUE #2450: MatchRoute component fails with custom parse/stringify for numeric route params

Router

### Describe the bug

There's an issue with the `MatchRoute` component when used in conjunction with routes that have custom `parse`/`stringify` functions for handling numeric parameters. Specifically, the component fails to match when provided with a number, but works correctly when given a string.

https://stackblitz.com/github/tanstack/router/tree/main/examples/react/kitchen-sink?embed=1&theme=dark&preset=node&file=src/main.tsx

1. Define a route with a custom `parse`/`stringify` for a numeric parameter, e.g.:
   ```typescript
   const invoiceRoute = createRoute({
     // ...other config
     path: '$invoiceId',
     params: {
       parse: (params) => ({
         invoiceId: z.number().int().parse(Number(params.invoiceId)),
       }),
       stringify: ({ invoiceId }) => ({ invoiceId: `${invoiceId}` }),
     },
     // ...rest of the route config
   })
   ```

2. Use the `MatchRoute` component with a numeric value:
   ```jsx
   <MatchRoute
     to={invoiceRoute.to}
     params={{
       invoiceId: 123, // Using a number
     }}
     pending
   >
     {(match) => <Spinner show={!!match} wait="delay-50" />}
   </MatchRoute>
   ```

3. Observe that the match fails and the component doesn't render as expected.

4. Change the `invoiceId` to a string:
   ```jsx
   <MatchRoute
     to={invoiceRoute.to}
     params={{
       invoiceId: "123", // Using a string
     }}
     pending
   >
     {(match) => <Spinner show={!!match} wait="delay-50" />}
   </MatchRoute>
   ```

5. Observe that the match now succeeds and the component renders correctly.

### Expected behavior

The `MatchRoute` component should work correctly with both numeric and string values when a custom `parse`/`stringify` is defined for the route parameter.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 1.58.16

### Additional context

This issue appears to be related to how the `MatchRoute` component interacts with the custom `parse`/`stringify` functions defined for route parameters. It seems that the component might be bypassing or incorrectly applying these functions when attempting to match the route.

LadyBluenotes:
`matchRoute` now resolves matches through the processed route tree.

- Applies route `params.parse` results before comparing or returning params.
- Preserves parsed parent params for child parsers and handles repeated param names.
- Uses route precedence and parser rejection during matching, so lower-priority or rejected routes no longer match through raw path templates.
- Replaces `findSingleMatch` and its cache with `findRouteMatch`.
- Normalizes exact trailing-slash matching while leaving fuzzy matching unchanged.
- Keeps `caseSensitive`, basepath, search, pending, optional-param, and wildcard behavior covered.

`matchRoute` now returns `false` for destination templates that are not registered in the route tree.

Includes the reproducer from #2460.

Closes #2450.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed route matching to fully respect custom typed `params.parse`/`params.stringify`, including proper numeric conversions.
  * `matchRoute` now returns parsed parameters (not raw strings) and correctly fails when typed parsing is rejected or throws.
  * Improved matching consistency across nested routes, including `search` and trailing-slash behavior.
* **Tests**
  * Added/expanded `matchRoute` and `useMatchRoute` test coverage for React, Solid, and Vue, covering typed params, fuzzy matching, and mismatch cases.
* **Chores**
  * Added a patch changeset documenting the matching fix.


packages/router-core/src/router.ts
```diff
@@ -2994,19 +2994,43 @@ export class RouterCore<
...
+            const parsedParams = parseParams(params as Record<string, string>)
+            if (parsedParams === false) {
+              return false
+            }
+            Object.assign(params, parsedParams)
+          }
+        }
+      } catch {
+        return false
+      }
```
Sheraff: I think this could be shortened using the `extractStrictParams` method that exists in this file
```ts
      for (const matchedRoutes of this.getRouteBranch(destinationRoute)) {
        try {
          extractStrictParams(route, params)
        } catch {
          return false
        }
      }
```

Sheraff (commented):
I think it makes sense to fix this. But also `router.matchRoute` is broken more generally because it does not take the route tree into account. It could return `true` for the *specific* route you're asking for, but when actually compared against the entire tree, another route would match because it had higher matching priority.

packages/router-core/src/router.ts
LadyBluenotes: kk I updated to use it

LadyBluenotes:
> I think it makes sense to fix this. But also `router.matchRoute` is broken more generally because it does not take the route tree into account. It could return `true` for the _specific_ route you're asking for, but when actually compared against the entire tree, another route would match because it had higher matching priority.

Added a regression for `osts/edit` versus `posts/$postId`. So when full-tree matching identifies `/posts/edit` as the winner, `matchRoute({ to: '/posts/$postId' })` now returns `false`. The existing fallback remains for route branches that cannot be resolved through parser-aware tree matching :))

LadyBluenotes:
@Sheraff made the changes we talked about in discord

Sheraff:
Can you remind me why we still need the `findSingleMatch` method if it now respects the tree? Why not use the same `findRouteMatch` as we use for regular route matching?

Sheraff:
does this collide with https://github.com/TanStack/router/pull/8327/changes?

LadyBluenotes:
> does this collide with https://github.com/TanStack/router/pull/8327/changes?

there's conflicts in files, if that's what youre referring to. I can stack this on manuel's?

PR #7803, opened 2026-07-13
fix(prerender): ensure failed prerenders are retried

wildlyinaccurate:
Right now if a page fails to prerender, it is not retried. I see things like this in my logs:

```
[prerender] Crawling: /site/klg75hzo
[prerender] Encountered error, retrying: /site/klg75hzo in 10000ms
```

but that URL `/site/klg75hzo` is never retried.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved prerendering retry behavior so page requests are retried until a successful response is received or the configured retry limit is reached.
  * Collected asynchronous queue task failures and surfaced the first relevant error after prerender completes.
  * Maintained existing error-handling semantics, including `failOnError: false` stopping retries without throwing.
* **Tests**
  * Added Vitest coverage for retry success, retry-limit stop behavior, and promise rejection/throwing based on `failOnError`.

### DIFF
--- a/packages/start-plugin-core/src/prerender.ts
+++ b/packages/start-plugin-core/src/prerender.ts
@@ -89,6 +89,10 @@ export async function prerender({
     const concurrency = startConfig.prerender?.concurrency ?? os.cpus().length
     logger.info(`Concurrency: ${concurrency}`)
     const queue = new Queue({ concurrency })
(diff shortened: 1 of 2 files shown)


wildlyinaccurate:
Fixes from CodeRabbit's first review have been applied and squashed.

PR #7804, opened 2026-07-13
fix(sitemap): exclude failed prerender pages from the sitemap

wildlyinaccurate:
I have a large prerendered static site. Sometimes pages fail to prerender, but they are still included in the sitemap. This causes crawlers like Google to crawl URLs that ultimately 404.

## Summary by CodeRabbit

* **Bug Fixes**
  * Retried prerender pages are now properly re-queued, even if they were previously marked as seen.
  * Pages that fail after exhausting their configured retry attempts are now excluded from the sitemap, including pages discovered via link crawling.
  * Successfully rendered pages remain eligible for sitemap inclusion.
  * When fail-on-error is enabled, the failure is still surfaced after the page has been excluded from the sitemap.

### DIFF
--- a/packages/start-plugin-core/src/prerender.ts
+++ b/packages/start-plugin-core/src/prerender.ts
@@ -108,12 +108,12 @@ export async function prerender({
 
     return Array.from(prerendered)
 
-    function addCrawlPageTask(page: Page) {
-      if (seen.has(page.path)) return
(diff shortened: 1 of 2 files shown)


wildlyinaccurate:
Fixes from CodeRabbit's first review have been applied and squashed.

PR #7806, opened 2026-07-13
docs(start): clarify verbatimModuleSyntax must be disabled

galaxykhh:
## Summary

- Update the React and Solid **Build from Scratch** guides to explicitly set `"verbatimModuleSyntax": false` in the example `tsconfig.json`
- Clarify that TypeScript's **compiler default** is still `false`, while `tsc --init` (TypeScript 5.9+ and 6.0) generates `"verbatimModuleSyntax": true` as a recommended option
- Align the `start-counter` examples (linked from the React guide) and the Start core skill doc with the same guidance

## Problem

The build-from-scratch guides warn that enabling `verbatimModuleSyntax` can cause server-only code to leak into client bundles, but they:

1. Do not show how to disable it in the example `tsconfig.json`
2. Do not mention that `tsc --init` now generates `"verbatimModuleSyntax": true` (microsoft/TypeScript#61813)

This is especially relevant for TypeScript 6 users who run `tsc --init` before following the guide, since the generated config opts into `verbatimModuleSyntax` even though the compiler default remains `false`.

## Changes

| File | Change |
|------|--------|
| `docs/start/framework/react/build-from-scratch.md` | Add `"verbatimModuleSyntax": false` to example tsconfig; expand NOTE |
| `docs/start/framework/solid/build-from-scratch.md` | Same |
| `examples/react/start-counter/tsconfig.json` | Add `"verbatimModuleSyntax": false` |
| `examples/solid/start-counter/tsconfig.json` | Same |
| `packages/start-client-core/skills/start-core/SKILL.md` | Mention `tsc --init` behavior |

## Test plan

- [x] Docs-only change; no runtime code modified
- [ ] Verified example `tsconfig.json` files remain valid JSON
- [ ] Reviewed rendered NOTE callout for clarity on tanstack.com docs preview (if applicable)

## Summary by CodeRabbit

* **Documentation**
  * Expanded guidance across React and Solid “build from scratch” and “common mistakes” docs on `verbatimModuleSyntax`.
  * Clarified that TanStack Start doesn’t support enabling it, and explained that while TypeScript defaults to `false`, `tsc --init` may set it to `true` (TypeScript 5.9+/6.0), so `tsconfig.json` should explicitly set `"verbatimModuleSyntax": false`.

* **Configuration**
  * Updated React and Solid start-counter example `tsconfig.json` files to set `verbatimModuleSyntax` to `false`.


docs/start/framework/react/build-from-scratch.md
```diff
@@ -39,13 +39,16 @@ We highly recommend using TypeScript with TanStack Start. Create a `tsconfig.jso
...
-    "strictNullChecks": true
+    "strictNullChecks": true,
+    "verbatimModuleSyntax": false
   }
 }
 ```
 
 > [!NOTE]
-> Enabling `verbatimModuleSyntax` can result in server bundles leaking into client bundles. It is recommended to keep this option disabled.
+> TanStack Start does not support `verbatimModuleSyntax` enabled — it can cause server-only code to leak into client bundles.
```
schiller-manuel: it's not that we dont "support" it. it can be used but could cause those leaks if types are imported from modules that have server only code

galaxykhh: You're right, it's not that Start doesn't support it.

Updated the wording in 3162d78 to reflect that it can be used, but may leak server-only code into client bundles when types are imported from modules that also contain server-only code. Kept the recommendation to leave it disabled.

PR #7808, opened 2026-07-13
fix(router-core): fire onBeforeNavigate after beforeLoad redirect

### LINKED ISSUE #3920: `onBeforeNavigate` not invoked after `redirect` within `beforeLoad` method was invoked

Router

### Describe the bug

When using throw redirect() within a beforeLoad function, subsequent router.subscribe("onBeforeNavigate") events may fail to trigger, causing potential data loss for components that rely on this event for cleanup or persistence operations.

https://github.com/dschu-lab/tanstack-router-onbeforeload-redirect-issue

- Start with an unauthenticated state
- Authenticate user (which triggers a redirect() in beforeLoad)
- Navigate to a page with onBeforeNavigate subscriptions
- Attempt to navigate away from this page
- Expected onBeforeNavigate event is not triggered

Subscription to the event is implemented at:
https://github.com/dschu-lab/tanstack-router-onbeforeload-redirect-issue/blob/2eb4bd98e7be565983eb7bac8c889859e51da8f3/src/components/PersistForm/PersistForm.tsx#L44-L51

### Expected behavior

The `onBeforeNavigate` event should consistently trigger regardless of whether a `redirect(`) was previously thrown in a `beforeLoad` or `loader` function.

### Platform

- OS: macOS 15.3.2
- Browser: Chrome
- Version: 134.0

### Additional context

_No response_

ChihebBENCHEIKH1:
## Problem

`onBeforeNavigate` was gated on `!this.stores.redirect.get()` in `router-core`'s `load()`. `stores.redirect` is populated whenever `beforeLoad` throws `redirect(...)` (line 2629) and never cleared, so every navigation after the first in-`beforeLoad` redirect silently skipped the `onBeforeNavigate` emit. The neighbouring `onBeforeLoad` emit has no such gate and kept firing correctly.

Real-world impact: Sentry's TanStack Router tracing integration stops producing navigation spans once any route ever redirects in `beforeLoad` (getsentry/sentry-javascript#21966), and userland patterns that hook `onBeforeNavigate` (e.g. form-state persistence) go silent after the first redirect.

## Fix

Remove the `state.redirect` gate around the `onBeforeNavigate` emit so it matches `onBeforeLoad` and fires on every navigation. `stores.redirect` is still populated for `createRequestHandler` (SSR redirect headers) — nothing else reads it.

## Tests

Added `fires on subsequent navigations after a redirect in beforeLoad (#3920)` in `packages/router-core/tests/callbacks.test.ts`: sets up `/foo` whose `beforeLoad` throws `redirect({ to: '/bar' })`, then navigates `/foo` → `/baz` and asserts `onBeforeNavigate` still fires on `/baz`. Fails on `main`, passes with this change. Full `@tanstack/router-core` and `@tanstack/react-router` unit + type + eslint suites are green.

Fixes #3920.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed navigation lifecycle events not firing after a route redirect.
  * `onBeforeNavigate` now consistently runs for every navigation, including navigations following a `beforeLoad` redirect.
  * Improved compatibility with integrations that rely on navigation events, such as tracing and monitoring tools.

### DIFF
--- a/.changeset/fix-on-before-navigate-after-redirect.md
+++ b/.changeset/fix-on-before-navigate-after-redirect.md
@@ -1,0 +1,11 @@
+---
+'@tanstack/router-core': patch
+---
+
+fix(router-core): fire `onBeforeNavigate` on every navigation, including those following a `beforeLoad` `redirect(...)`
+
+`onBeforeNavigate` was gated on `stores.redirect` being unset, but that store was populated whenever `beforeLoad` threw `redirect(...)` and never cleared. As a result, every navigation after the first in-`beforeLoad` redirect silently skipped the `onBeforeNavigate` emission (`onBeforeLoad` next to it never had this gate). This broke Sentry's TanStack Router tracing integration and any userland pattern (e.g. form-state persistence) that hooks into `onBeforeNavigate`.
+
+The gate is removed so `onBeforeNavigate` now matches `onBeforeLoad` and fires on every navigation.
+
+Fixes #3920.
--- a/packages/router-core/src/router.ts
+++ b/packages/router-core/src/router.ts
@@ -2483,12 +2483,10 @@ export class RouterCore<
           const prevLocation = this.stores.resolvedLocation.get()
           const locationChangeInfo = getLocationChangeInfo(next, prevLocation)
 
-          if (!this.stores.redirect.get()) {
-            this.emit({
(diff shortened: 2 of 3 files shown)


Sheraff:
This issue was fixed as part of https://github.com/TanStack/router/pull/7805

PR #7817, opened 2026-07-15
feat(router-core): expose destination location to route lifecycle hooks

### LINKED ISSUE #7778: feat(router-core): expose destination location to route lifecycle hooks
## Problem

Route lifecycle hooks receive the route match but not the resolved destination location. Consumers need the destination URL and search state in `onEnter`, `onStay`, and `onLeave`.

## Proposal

Add a second callback argument:

- `onEnter(match, { location })`
- `onStay(match, { location })`
- `onLeave(match, { location })`

Supersedes stale #3622 - which targeted an older lifecycle implementation, lacked runtime coverage for `onStay` and `onLeave`, and left `onLeave` typing unresolved.

FrancoKaddour:
Closes #7778

onEnter, onStay, and onLeave now receive a second argument exposing the resolved
destination location of the navigation that triggered them:

- onEnter(match, { location })
- onStay(match, { location })
- onLeave(match, { location })

Implementation
- router.ts: the hook dispatch loop in load() passes { location: next }, where next
  is the resolved destination location of the current navigation.
- route.ts: the three UpdatableRouteOptions signatures accept the new ctx argument,
  typed { location: ParsedLocation } following the file's existing convention of not
  supplying the search schema in route-option locations.
- For onLeave, the received location is the destination being navigated to, not the
  location being left — the match itself already carries the departed state.
- Backward compatible: existing callbacks that only take match keep working.

Tests
- Updated existing callbacks tests to assert the new argument.
- Added a dedicated test verifying all three hooks receive the destination pathname
  and search state.
- Full router-core suite passes (1194 tests, vitest typecheck clean, tsc 5.9 clean).

Docs
- RouteOptionsType.md: updated the three signatures and documented ctx.location.

## Summary by CodeRabbit

* **New Features**
  * Route lifecycle callbacks (`onEnter`, `onStay`, `onLeave`) now receive a second argument containing the resolved destination `location` (path, search parameters, and `href`).
* **Documentation**
  * Updated route options documentation to describe the new callback context and `location` details.
* **Tests**
  * Expanded runtime and type-level coverage to verify the second argument’s `location` contents across navigation scenarios.

### DIFF
--- a/docs/router/api/router/RouteOptionsType.md
+++ b/docs/router/api/router/RouteOptionsType.md
@@ -266,21 +266,24 @@ type loaderDeps = (opts: { search: TFullSearchSchema }) => Record<string, any>
 
 ### `onEnter` property
 
-- Type: `(match: RouteMatch) => void`
+- Type: `(match: RouteMatch, ctx: { location: ParsedLocation }) => void`
 - Optional
 - A function that will be called when a route is matched and loaded after not being matched in the previous location.
+- `ctx.location` is the resolved destination [`ParsedLocation`](./ParsedLocationType.md) of the navigation that triggered the hook.
 
 ### `onStay` property
 
-- Type: `(match: RouteMatch) => void`
+- Type: `(match: RouteMatch, ctx: { location: ParsedLocation }) => void`
 - Optional
 - A function that will be called when a route is matched and loaded after being matched in the previous location.
+- `ctx.location` is the resolved destination [`ParsedLocation`](./ParsedLocationType.md) of the navigation that triggered the hook.
 
 ### `onLeave` property
 
-- Type: `(match: RouteMatch) => void`
+- Type: `(match: RouteMatch, ctx: { location: ParsedLocation }) => void`
 - Optional
 - A function that will be called when a route is no longer matched after being matched in the previous location.
+- `ctx.location` is the resolved destination [`ParsedLocation`](./ParsedLocationType.md) of the navigation being made, not the location being left.
 
 ### `onCatch` property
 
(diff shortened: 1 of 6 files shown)


LadyBluenotes (changes requested):
This is looking good overall. The shared router-core implementation is a nice fit, and the runtime coverage looks solid.

Could you also add direct type coverage for the new callback argument in:

- `packages/react-router/tests/route.test-d.tsx`
- `packages/solid-router/tests/route.test-d.tsx`

Each lifecycle callback could accept `ctx` and assert that `ctx.location` is `ParsedLocation`. It would also be useful to pin the deliberately unparameterized `ctx.location.search` type.

The existing type suites pass, but those callbacks currently only inspect `match`, so this would make sure the new public API stays covered.

FrancoKaddour:
Thanks for the review, @LadyBluenotes! Added the type coverage in 59505d5 — the `onEnter`/`onStay`/`onLeave` callbacks in the existing lifecycle type tests (both `react-router` and `solid-router`) now accept `ctx`, assert that `ctx.location` is `ParsedLocation`, and pin the deliberately unparameterized `ctx.location.search` type. `test:types` passes on TS 5.5–6.0 in both packages.

LadyBluenotes:
Thanks for the work you’ve put into this. I just want to be transparent and set expectations that it may be a while before we can give it a proper review. The API needs broader consideration, and we’re waiting on a few upcoming releases before introducing further API changes, especially with stable approaching.

FrancoKaddour:
Totally understandable — thanks for the transparency! No rush at all. I'll keep the branch around and I'm happy to adapt the implementation whenever the API discussion settles after the upcoming releases. Good luck with the road to stable! 🙌

FrancoKaddour:
Closing this for now since the underlying API is still in flux — no point keeping a PR open against a moving target. Happy to revisit once the direction settles. Thanks for the heads-up earlier!

PR #7832, opened 2026-07-16
fix(history): respect ignoreBlocker in history.go

### LINKED ISSUE #4867: `history.go` does not respect `ignoreBlocker` option

xianjianlf2:
### What & why

`history.go(n, { ignoreBlocker: true })` still fires navigation blockers, unlike `back`/`forward` which honor the option (#4867). The internal history adapter's `go` signature was `(n: number) => void`, so `createHistory.go` dropped `ignoreBlocker` when calling `opts.go(index)`, and the browser `go` only set `nextPopIsGo` without setting `skipBlockerNextPop`. The resulting `popstate` therefore never bypassed blockers in `onPushPopEvent`.

### Fix (`packages/history/src/index.ts`, 3 lines)

- Widen the internal adapter signature to `go: (n: number, ignoreBlocker: boolean) => void`.
- `createHistory.go` forwards the option: `opts.go(index, navigateOpts?.ignoreBlocker ?? false)`.
- Browser `go` sets `skipBlockerNextPop = true` when `ignoreBlocker` — mirroring the existing `back`/`forward` lines exactly.

### Tests

New `packages/history/tests/createBrowserHistory.test.ts` sets up two real jsdom session entries, registers a blocker, and drives an actual `go(-1)` popstate — asserting the blocker runs for `go(-1)` and is skipped for `go(-1, { ignoreBlocker: true })`. This deliberately reaches the browser `popstate` handler (the gap called out on the prior PR #4872). Verified red→green (without the fix: 1 failed; with: 2 passed); full `@tanstack/history` suite 27 passed, `tsc --noEmit` clean.

Closes #4867

## Summary by CodeRabbit

* **New Features**
  * Added an option to bypass navigation blockers when moving through browser history.
  * Standard history navigation continues to check registered blockers by default.

* **Bug Fixes**
  * Ensured blocker checks are consistently applied during normal history navigation.

* **Tests**
  * Added coverage for both blocked navigation and navigation with blocker checks bypassed.


xianjianlf2:
Addressed the CodeRabbit style follow-up in `ef44d1e` by wrapping the `ignoreBlocker` branch in `history.go` with braces.

Local check:

```sh
pnpm --filter @tanstack/history exec vitest run tests/createBrowserHistory.test.ts
```

Result: 1 file / 2 tests passed, no type errors.

xianjianlf2:
Merged latest `origin/main` into `fix/history-go-ignore-blocker-4867` and resolved the add/add history test conflict by keeping the new browser-history queue tests plus the `go(..., { ignoreBlocker: true })` blocker coverage.

Validation:
- `pnpm install --frozen-lockfile`
- `pnpm exec prettier --check packages/history/src/index.ts packages/history/tests/createBrowserHistory.test.ts`
- `pnpm --dir packages/history run build`
- `pnpm --dir packages/history exec vitest run tests/createBrowserHistory.test.ts` (6 passed, no type errors)

Current head: `576b442e31b1c0060f5e6c603fcba7f2c1d8a7ee`

xianjianlf2:
Closing this because #8264 has now landed on `main` with the same `go(..., { ignoreBlocker: true })` fix and broader handling for `back`, `forward`, and `beforeunload`. I synced this branch with the current `main` and verified the `@tanstack/history` unit, type, build, and ESLint targets.

PR #7818, opened 2026-07-15
fix(router-core): use configured search serializer for loader dependency keys

gonzoblasco:
## Description

The `loaderDepsHash` was hardcoded to `JSON.stringify`, ignoring the router's configured `stringifySearch` option. This meant custom search serializers that support values `JSON.stringify` cannot handle (e.g. `bigint`, `Set`, `Map`) would fail or produce incorrect hashes when those values were used in `loaderDeps`.

## Root Cause

In `packages/router-core/src/router.ts`, line 1538:

```ts
const loaderDepsHash = loaderDeps ? JSON.stringify(loaderDeps) : ''
```

`JSON.stringify` is used regardless of the `stringifySearch` option configured on the router. If a user configures a custom `stringifySearch` to support non-JSON-serializable values in search params, those same values cannot be used in `loaderDeps` because the hash will throw or produce incorrect results.

## Fix

Replace `JSON.stringify(loaderDeps)` with `this.options.stringifySearch(loaderDeps)` so the hash uses the same serializer configured for URL search state.

```diff
- const loaderDepsHash = loaderDeps ? JSON.stringify(loaderDeps) : ''
+ const loaderDepsHash = loaderDeps
+   ? this.options.stringifySearch(loaderDeps)
+   : ''
```

This continues the work started in #4713.

## Tests

Added `packages/router-core/tests/loader-deps-serializer.test.ts` with two regression tests:

1. **Custom serializer with bigint support** — verifies that a router configured with a custom `stringifySearch` supporting `bigint` can use `bigint` values in `loaderDeps` without throwing (which `JSON.stringify` would do with `TypeError: Do not know how to serialize a BigInt`).

2. **Different deps produce different match IDs** — verifies that changing `loaderDeps` values produces different match IDs, confirming the hash is still deterministic and distinguishable.

## Verification

- ✅ New tests pass
- ✅ Existing `callbacks.test.ts` (9 tests) — all pass
- ✅ Existing `match-params.test.ts` (23 tests) — all pass
- ⚠️ `load.test.ts` — 1 pre-existing flaky test fails (`skip if pending preload (redirect)`) — confirmed failing on `main` without this change

Fixes #7787

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved route matching for loader dependencies by deriving the match identity from the router’s configured search serializer.
  * Prevented navigation failures when loader dependencies include values such as `bigint`.
  * Ensured route loaders refresh correctly when relevant search filter values change.
* **Tests**
[...]

PR #7834, opened 2026-07-16
fix(router-core): use safeStringify for loader dependency hash keys

### LINKED ISSUE #7787: fix(router-core): use configured search serializer for loader dependency keys
## Problem

Custom `stringifySearch` is used for URL search state but not for loader dependency cache keys.

Routes using custom search serialization can therefore support values in search params that fail when used in `loaderDeps`.

## Scope

- Use the configured search serializer when deriving loader dependency cache keys.
- Add a regression test using a custom serializer and a dependency value that `JSON.stringify` cannot handle.
- Verify loader dependency cache identity changes when the serialized dependency value changes.

## Related Work

This continues the work started in #4713:
https://github.com/TanStack/router/pull/4713

gonzoblasco:
## Description

Replace `JSON.stringify` with `safeStringify` in `loaderDepsHash` computation to handle types that `JSON.stringify` cannot serialize (bigint, Set, Map, circular references, functions, symbols, etc.).

## Problem

`loaderDepsHash` was hardcoded to `JSON.stringify`, which throws on `bigint` values and cannot handle `Set`, `Map`, circular references, or symbols. Users with custom search serializers that support these types would get runtime errors when using them in `loaderDeps`.

## Previous attempt

PR #7818 attempted to use the configured `stringifySearch` serializer instead. As @schiller-manuel correctly pointed out, loader deps are not necessarily search params and should not be tied to the search stringifier.

## This approach

Introduce `safeStringify` — a zero-dependency utility that safely serializes any value to a deterministic JSON string:

- **bigint** → `"123n"` (string representation)
- **Set** → sorted array of serialized values
- **Map** → array of `[key, value]` entry arrays
- **undefined** → empty string
- **Circular references** → `"[Circular]"` (detected via WeakSet)
- **Functions** → `"[Function]"`
- **Symbols** → description string
- **Date** → ISO string
- **Plain objects** → keys sorted for deterministic output

No API changes. No new dependencies. No coupling between loader deps and search params.

## Tests

Added 12 unit tests covering all edge cases above.

## Verification

- ✅ `tests/utils.test.ts` — 127 passed, 3 expected fail (pre-existing)
- ✅ `tests/searchParams.test.ts` — all passed
- ✅ Only 3 files changed (utils.ts, router.ts, utils.test.ts)

Fixes #7787

## Summary by CodeRabbit

* **New Features**
  * Added safe serialization for complex loader dependency values.
  * Added support for customizing loader dependency serialization.

* **Bug Fixes**
  * Improved route matching for dates, maps, sets, bigints, circular references, and nested values.
  * Improved consistency when dependency properties are supplied in different orders.
  * Reduced unnecessary route rematches for equivalent dependency data.

* **Tests**
  * Added coverage for varied value types, custom serialization, and circular or nested structures.


gonzoblasco:
Gentle reminder on this one - the safeStringify fix for loader dependency hash keys is ready and mergeable. Happy to rebase if needed.

schiller-manuel:
we wont merge this as is, this too niche and its bundlesize impact is not warranted for all users. possibly we could allow letting a user configure a custom stringifier for loaderDeps.

gonzoblasco:
Thanks @schiller-manuel , that makes sense - forcing safeStringify into the bundle for everyone is not worth it.

I've reworked the PR: instead of using it inline, `loaderDepsHash` now uses a new `stringifyLoaderDeps` router option that defaults to `JSON.stringify` (no behavior change, no added bundle weight). Users with loader deps that JSON.stringify can't handle (bigint, Set, Map, circular refs, etc.) can opt in by passing `safeStringify`, which I've kept exported from `@tanstack/router-core`.

Added two tests covering the new option: one that verifies the custom stringifier is actually used for the hash, and one that confirms passing `safeStringify` produces a stable hash (loader cached, not re-run) for bigint deps.

gonzoblasco:
Addressed the CodeRabbit findings - the collision issue was a real bug, thanks for catching it.

**Collision-proof safeStringify (major fix)**
`safeStringify({ id: 123n })` and `safeStringify({ id: '123n' })` no longer collide. Special values now use a reserved NUL-tag prefix (bigint, Set, Map, undefined, Date, functions, symbols, circular refs), and user strings/keys that start with the prefix are escaped by doubling it. Added regression tests for each collision class: bigint vs string, undefined vs empty string, Date vs ISO string, Set vs array, and user strings starting with the reserved prefix.

**Robust default**
The `stringifyLoaderDeps` default is now `defaultStringifyLoaderDeps`: JSON.stringify for plain deps (no bundle weight in the common case) with a safeStringify fallback when serialization throws (e.g. bigint), so the default no longer crashes on bigint loader deps.

**Type safety**
Swapped `any` for `unknown`/typed loader-deps in the serializer and tests.

Full router-core suite passes (1632 tests), types and lint clean.

PR #7838, opened 2026-07-16
fix(start-plugin-core): keep prerender trailing slash off the query string

### LINKED ISSUE #7837: Prerender crawler corrupts query strings and loops forever when links carry search params
# Prerender crawler corrupts query strings and loops forever when links carry search params

## Which project does this relate to?

Start

## Describe the bug

The static prerender crawler in `@tanstack/start-plugin-core` applies `withTrailingSlash` to the **entire request path including the query string**. `withTrailingSlash` from `ufo`, called without its `respectQueryAndFragment` option, appends the slash to the end of the string, so the slash lands **inside the last search-param value** instead of on the path.

`packages/start-plugin-core/src/prerender.ts` (line ~140):

```ts
const res = await requestWithRedirects(
  withTrailingSlash(withBase(page.path, routerBasePath)),
  // ...
)
```

Demonstration in isolation:

```ts
import { withTrailingSlash, withBase } from 'ufo'

withTrailingSlash(withBase('/posts?page=2', '/'))
// => "/posts?page=2/"          ❌  slash appended to the query value

withTrailingSlash(withBase('/posts?page=2', '/'), true)
// => "/posts/?page=2"          ✅  slash on the path, query preserved
```

When `crawlLinks` is enabled (the default), this turns into an infinite loop:

1. A page renders an anchor whose href carries a search param, such as `/posts?page=2`.
2. The crawler requests `withTrailingSlash("/posts?page=2")` => `/posts?page=2/`.
3. The server parses `page=2/` (the trailing slash is now part of the value) and renders a page whose own links serialise `page=2%2F`.
4. `extractLinks` picks that up, the crawler appends another slash => `page=2%2F%2F`, and so on.

Every URL is unique, so the crawler's `seen` set never dedups. The queue grows without bound and the build hangs until it runs out of memory.

This bites any app that has a search param on a route reachable through a crawled `<a href>`. A common trigger is a root-level search param such as a currency or locale selector rendered in shared page chrome, so every prerendered page emits a self-link carrying it.

## Your Example Website or App

Minimal reproduction: any TanStack Start app with `prerender.enabled` and a route that renders an `<a>` whose href includes a query string, such as a `<Link to="." search={{ page: 2 }}>`.

## Steps to Reproduce the Bug or Issue

1. Create a TanStack Start app with `tanstackStart({ prerender: { enabled: true } })`.
2. Add a search param to a route and render a `<Link>` to that route with the search param set (so the prerendered HTML contains `<a href="/route?param=value">`).
3. Run `pnpm build`.
4. Observe the prerender log crawling the same path with an ever-growing `%2F` suffix on the param value, and the build never completing.

```
[prerender] Crawling: /route?param=value
[prerender] Crawling: /route?param=value%2F
[prerender] Crawling: /route?param=value%2F%2F
...
```

## Expected behavior

The trailing slash should be applied to the path only, leaving the query string intact:

```
/route?param=value   →   /route/?param=value
```

The crawler should then converge and dedup normally rather than generating an unbounded stream of unique URLs.

## Screenshots or Videos

_N/A_

## Platform

- Affected package: `@tanstack/start-plugin-core`, reproduced on 1.171.20 with the same code on `main`
- Reproduced on Linux with Node 24

## Additional context

The fix is a one-line change: pass `respectQueryAndFragment: true` to `withTrailingSlash`. Happy to open a PR (with a changeset).

cjsewell:
# fix(start-plugin-core): keep prerender trailing slash off the query string

## Summary

The prerender crawler applied `withTrailingSlash` to the full request path including the query string. Without `ufo`'s `respectQueryAndFragment` option, `withTrailingSlash` appends the slash to the end of the string, so it lands inside the last search-param value rather than on the path:

```ts
withTrailingSlash('/posts?page=2')       // "/posts?page=2/"   (before)
withTrailingSlash('/posts?page=2', true) // "/posts/?page=2"   (after)
```

With `crawlLinks` enabled, this corrupts any crawled link that carries a search param and turns it into an infinite loop. Each round appends one more slash to the value. Because every URL is unique the crawler's `seen` set never dedups, so the queue grows without bound and the build hangs.

## Change

`packages/start-plugin-core/src/prerender.ts`:

```diff
-          const res = await requestWithRedirects(
-            withTrailingSlash(withBase(page.path, routerBasePath)),
+          const res = await requestWithRedirects(
+            withTrailingSlash(withBase(page.path, routerBasePath), true),
```

## Test plan

- Prerender an app whose crawled HTML contains an `<a href="/route?param=value">`.
- Before: the crawler loops on `/route?param=value%2F%2F…` and never finishes.
- After: the request goes to `/route/?param=value` with the query preserved. The crawl then converges and dedups.

Closes #7837

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed prerendering so trailing slashes remain part of the URL path instead of being appended to query parameters.
  * Prevented potential infinite crawl loops when link crawling is enabled.

* **Release**
  * Included in a patch release of the prerendering plugin.

### DIFF
--- a/.changeset/prerender-trailing-slash-query.md
+++ b/.changeset/prerender-trailing-slash-query.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/start-plugin-core': patch
+---
+
+fix(start-plugin-core): keep the prerender trailing slash off the query string
+
+The crawler applied `withTrailingSlash` to the full path including the query string, so the slash landed inside the last search-param value and, with `crawlLinks` enabled, sent the crawl into an infinite loop. Passing ufo's `respectQueryAndFragment` flag keeps the slash on the path.
--- a/packages/start-plugin-core/src/prerender.ts
+++ b/packages/start-plugin-core/src/prerender.ts
@@ -137,7 +137,7 @@ export async function prerender({
 
         try {
           const res = await requestWithRedirects(
-            withTrailingSlash(withBase(page.path, routerBasePath)),
+            withTrailingSlash(withBase(page.path, routerBasePath), true),
             {
               headers: {
                 ...(prerenderOptions.headers ?? {}),


LadyBluenotes:
We only review implementation PRs after the underlying issue has been reproduced, triaged, and the scope has been agreed with a maintainer. Please include a minimal reproduction in the PR and wait for maintainer confirmation before submitting a fix. I’m closing this PR for now; we can revisit it once the issue is confirmed

PR #7848, opened 2026-07-18
fix(router-core): isolate client loading lane ownership

Sheraff:
## Summary

This follow-up makes client loading ownership explicit: foreground transactions, preload lanes, loader flights, background reloads, and HMR no longer borrow authority from one another implicitly. It preserves accepted loader data while preventing stale context, work, and projections from crossing lane boundaries.

### 1. Loader-flight ownership is lane-local

- **Global per-match joining mixed independent lanes.** The router-wide `_flights` registry allowed unrelated navigation, preload, and background lanes to join an individual loader by match id even when their parent generation or complete lane semantics differed. The registry is removed; each loader invocation belongs to one lane, and active sharing happens only through complete-lane adoption.
- **Cancellation and cache cleanup could affect the wrong owner.** A lane releasing shared work could abort a loader still represented by an accepted active or cached match. Match-local leases now survive settlement and transfer only when the semantic generation is accepted; losing, trimmed, expired, and cache-rejected matches release exactly their own lease.
- **Adopted controllers had ambiguous lifetime.** Superseded claims and fatal post-adoption failures could retain the preload controller, while HMR could release an inherited flight before preload execution acquired it. Controller ownership now transfers directly to the claiming transaction and every unsuccessful path performs one authoritative cleanup.

### 2. Preloads are shared and claimed as complete lanes

- **Sharing only a descendant loader produced incoherent parent/context generations.** Per-loader borrowing could combine cached parent data, active background work, and a child from another lane. `_preloads` now tracks complete lanes; identical concurrent preloads and navigations adopt the whole serial-plus-loader pipeline or run independently.
- **Match ids alone were not an exact adoption boundary.** Lanes with different parsed params or search objects could be treated as compatible despite producing different route semantics. Adoption now requires the same route sequence, match ids/loader deps, exact parsed params, and exact search.
- **Completed preload context leaked into navigation.** Cached `beforeLoad` output could reuse stale authorization or route context even though only loader data is safe to retain. Completed preloads now clear `beforeLoad` context before cache publication; navigation reruns context while reusing fresh loader data.
- **`preload: false` could be bypassed by a claim.** A navigation could adopt a speculative lane where a required loader had intentionally been skipped. Such lanes are no longer claimable, so navigation performs its blocking load.
- **Failed or terminal preload claims could partially become navigation state.** Root global-not-found, failed, redirected, canceled, or otherwise invalid lanes could leave descendants or partial work in the claiming transaction. Rejected adoption discards the donor and reruns the complete navigation lane in blocking mode.
- **Stale cache generations could win a late claim or publication.** Cache clearing or replacement after preload planning could still allow old work to publish. Claims validate the captured committed base and cache-array identity, and cache publication uses per-match identity compare-and-swap checks.
- **A superseded navigation could remain blocked on an independent preload.** Navigation now races its wait against its own transaction signal, allowing the navigation to settle without canceling unrelated speculative work.

### 3. Foreground transaction installation preserves the current writer

- **Planning could retire the current transaction before a replacement existed.** If matching failed or reentered, the router could lose its authoritative foreground owner. `_preflight` now guards synchronous planning, while `_tx` remains current until a successfully planned replacement transaction is installed.
- **Resource handoff was coupled to promise settlement instead of publication authority.** Currentness checks now guard commit, view-transition work, and completion publication; a losing lane releases its private matches without mutating committed state.

### 4. Background reloads publish detached generations

- **Stale projection mutated committed matches before compare-and-swap checks.** Async `head` and `scripts` hooks ran on objects shared with the committed lane, so a superseded background reload could alter visible metadata even when publication was rejected. Projection now runs on detached, flight-free copies and publishes only while both the owning transaction and exact committed base remain current.
- **Untouched matches gained false fetches during projection-only publication.** Detached copies now preserve `fetchCount` for unchanged matches.
- **Trimmed or rejected background work leaked loader ownership.** Failure boundaries clone shared matches before mutation, discarded descendants release private resources, and successful publication transfers loader-flight ownership into the detached projected lane.
- **Successful background reloads redundantly waited for normal route chunks.** The committed route chunk is already loaded, so successful background work skips it while error and not-found boundaries still load their required chunks.
- **Projection continued invoking hooks after cancellation.** Projection awaits are raced against the lane controller and iteration stops once the lane is aborted.

### 5. Development route refresh uses the normal transaction protocol

- **HMR eagerly aborted the foreground transaction before rematching succeeded.** Refresh now retires preload/cache generations first, plans synchronously with scoped reuse disabled for the changed subtree, and lets ordinary replacement installation retire the previous foreground transaction.
- **Aborted HMR preloads could hang forever in projection or route-chunk readiness.** Development readiness and projection waits observe the preload signal, so refresh settles even when obsolete `head`, `scripts`, or route-chunk promises never resolve.
- **Old route code could republish through speculative cache work.** HMR globally aborts active preload lanes and clears cached leases before rematerializing the changed route and descendants.

### 6. Exact object equality has an explicit mode

- **Exact lane comparison still ignored `undefined` keys.** The default deep-equality behavior treats missing and `undefined` keys as equivalent, which is unsafe for parsed params and search lane identity. `deepEqual(a, b, false)` now requests exact object keys recursively, including keys whose value is `undefined`.

## Documentation

- Updates the public preloading guide to clarify that completed preload context is rerun for navigation while fresh loader data remains reusable.
- Expands `router-core/INTERNALS.md` with the transaction, lane, lease, cache, background, route-chunk, and HMR ownership protocols.

## Validation

- `@tanstack/router-core:test:unit`: 93 files, 1,337 passed, 3 expected failures
- `@tanstack/router-core:test:types`: TypeScript 5.5 through 6.0
- `@tanstack/router-core:test:eslint`: passed with 20 existing warnings
- Full 17-scenario bundle-size build
- `react-router.minimal`: gzip 89,443 bytes, matching the base ceiling
- `git diff --check`


Sheraff:

PR #7849, opened 2026-07-18
issues

schiller-manuel:


Sheraff:
included in https://github.com/TanStack/router/pull/7805

PR #7861, opened 2026-07-20
fix(react-router): prevent MatchInner from throwing undefined when a redirected match's loadPromise is cleared

### LINKED ISSUE #7753: MatchInner can `throw undefined` when a redirected match's loadPromise was already cleared — escapes CatchBoundary and unmounts the whole app

Router

### Describe the bug

On initial page load, when an async `beforeLoad` takes a few hundred ms (in our case: an auth token refresh) and then throws `redirect()`, there is a race where `MatchInner` renders a match observed as `status === 'redirected'` **after** the redirect navigation has already settled. At that point the match's `loadPromise` has been cleared, so `MatchInner` executes `throw getMatchPromise(match, 'loadPromise')` — which evaluates to **`throw undefined`**.

`getMatchPromise` can return `undefined` by design of its lookup:

```ts
// packages/react-router/src/Match.tsx
const getMatchPromise = (match, key) => {
  return (
    router.getMatch(match.id)?._nonReactive[key] ?? match._nonReactive[key]
  )
}
```

…and `router-core` clears the promise once loading settles:

```ts
// packages/router-core/src/load-matches.ts
match._nonReactive.loadPromise?.resolve()
match._nonReactive.loadPromise = undefined
```

The thrown `undefined` is then not contained by `CatchBoundary`, because its render callback checks truthiness of the error:

```ts
// packages/react-router/src/CatchBoundary.tsx
children={({ error, reset }) => {
  if (error) {            // <- undefined is falsy
    return React.createElement(errorComponent, { error, reset })
  }
  return props.children   // <- children re-throw, boundary is considered failed
}}
```

`getDerivedStateFromError(undefined)` stores the falsy error, the boundary re-renders its children, they throw again, React treats the boundary as failed and escalates. With React 19 the error reaches the root, the whole tree is unmounted and `defaultOnUncaughtError` calls `reportError(undefined)` — the only console output is a stackless `Uncaught undefined`, and the user gets a permanent white screen (`#root` is empty).

### Steps to reproduce

Timing-dependent race, so hard to make deterministic, but the shape is:

1. SPA with `createRouter` (no SSR), React 19, `defaultPendingComponent` set.
2. Root-ish route `/` whose `beforeLoad` awaits something slow (~300–800 ms, e.g. a token refresh request) and then does `throw redirect({ to: '/somewhere' })`.
3. Cold-load the app at `/`.
4. Intermittently, the app white-screens with `Uncaught undefined` in the console and an empty `#root`. Reloading (when the slow work is no longer needed, e.g. token now fresh) loads fine.

We hit this consistently-intermittently in production; the window opens whenever the first render commits while the redirect is being processed.

### Expected behavior

Either the stale render suspends/retries (e.g. `getMatchPromise` falling back to a resolved promise instead of `undefined`), or at minimum the error boundary contains whatever was thrown — `CatchBoundaryImpl` tracking a `hasError` sentinel instead of relying on error truthiness — so a falsy thrown value can't unmount the entire app.

### Platform

- Package: @tanstack/react-router 1.170.17 (also present in current `main` sources)
- React 19, client-side only (no SSR)
- Browser: Chrome (any)

### Additional context

Happy to provide more details or test a patch. Workaround we're shipping: an app-level error boundary above `RouterProvider` that treats falsy thrown values as errors and remounts once.

gonzoblasco:
When an async `beforeLoad` throws `redirect()` and pending UI is enabled (`defaultPendingMs: 0`), a race can leave `MatchInner` rendering a match observed as `status === "redirected"` *after* the redirect navigation has already settled. At that point the match's `loadPromise` has been cleared (`undefined`), so `MatchInner` executes `throw getMatchPromise(match, "loadPromise")` — which evaluates to **`throw undefined`**.

The thrown `undefined` escapes `CatchBoundary` because its render callback checks truthiness of the error (`if (error)` — `undefined` is falsy). With React 19 the error reaches the root, the whole tree is unmounted, and the user gets a permanent white screen with `Uncaught undefined` in the console.

**The fix (two parts):**

1. **`MatchInner` (client + server branches):** guard the `throw` so a cleared `loadPromise` (`undefined`) is not thrown. The redirect has already been processed by the router at that point — React can safely render the component until the match store updates with the new route.

2. **`CatchBoundaryImpl`:** use a `hasError` sentinel instead of relying on error truthiness. This is defense in depth — if something else ever throws a falsy value, the boundary catches it instead of letting it unmount the app.

**Testing:**
- Added 2 tests that reproduce the race: a single async `beforeLoad` redirect and a chain of two async redirects, both with `defaultPendingMs: 0` to force pending publication mid-chain.
- All 919 existing tests pass (49 test files, 0 failures).

Fixes #7753

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error boundary handling so errors reset correctly when navigation state changes.
  * Prevented redirect and loading races from causing uncaught errors or unexpectedly unmounting the application.
  * Improved reliability for chained asynchronous redirects and pending UI transitions.

* **Tests**
  * Added regression coverage for asynchronous redirects, pending states, and multi-step navigation flows.

### DIFF
--- a/packages/react-router/src/CatchBoundary.tsx
+++ b/packages/react-router/src/CatchBoundary.tsx
@@ -46,12 +46,16 @@ class CatchBoundaryImpl extends React.Component<{
 
   static getDerivedStateFromProps(
     props: { getResetKey: () => string | number },
-    state: { resetKey?: string | number; error: Error | null },
+    state: {
+      resetKey?: string | number
+      error: Error | null
+      hasError: boolean
+    },
   ) {
     const resetKey = props.getResetKey()
 
-    if (state.error && state.resetKey !== resetKey) {
-      return { resetKey, error: null }
+    if (state.hasError && state.resetKey !== resetKey) {
+      return { resetKey, error: null, hasError: false }
     }
 
     return { resetKey }
--- a/packages/react-router/src/CatchBoundary.tsx
+++ b/packages/react-router/src/CatchBoundary.tsx
@@ -38,7 +38,11 @@ class CatchBoundaryImpl extends React.Component<{
   }) => React.ReactNode
   onCatch?: (error: Error, errorInfo: ErrorInfo) => void
 }> {
-  state = { error: null } as { error: Error | null; resetKey?: string | number }
+  state = { error: null, hasError: false } as {
+    error: Error | null
+    hasError: boolean
+    resetKey?: string | number
+  }
 
   static getDerivedStateFromProps(
     props: { getResetKey: () => string | number },
@@ -53,10 +57,10 @@ class CatchBoundaryImpl extends React.Component<{
     return { resetKey }
   }
   static getDerivedStateFromError(error: Error) {
-    return { error }
+    return { error, hasError: true }
   }
   reset() {
-    this.setState({ error: null })
+    this.setState({ error: null, hasError: false })
   }
   componentDidCatch(error: Error, errorInfo: ErrorInfo) {
     if (this.props.onCatch) {
@@ -65,7 +69,7 @@ class CatchBoundaryImpl extends React.Component<{
   }
   render() {
     return this.props.children({
-      error: this.state.error,
+      error: this.state.hasError ? this.state.error : null,
       reset: () => {
         this.reset()
       },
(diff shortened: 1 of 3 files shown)


schiller-manuel:
https://github.com/TanStack/router/pull/7805 will fix this properly

PR #7878, opened 2026-07-22
fix(start-client-core): skip console.log for AbortError in serverFnFetcher (#7873)

### LINKED ISSUE #7873: Start: serverFnFetcher unconditionally console.logs every server-function error, including benign cancellation AbortErrors

Start

### Describe the bug

`serverFnFetcher`'s `getResponse` logs *every* non-`Response` error thrown by a server-function call via `console.log`, then re-throws it. In `packages/start-client-core/src/client-rpc/serverFnFetcher.ts`:

  ```ts
  async function getResponse(fn) {
    let response
    try {
      response = await fn()
    } catch (error) {
      if (error instanceof Response) response = error
      else {
        console.log(error)   // ← logs every error, incl. benign cancellations
        throw error
      }
    }
    // ...
  }
  ```

  Two problems:

  1. **It's unconditional** — there is no `process.env.NODE_ENV` guard, so it also runs in production builds (bundlers don't strip `console.log` by default), spamming the browser console for real users.
  2. **It logs benign cancellations.** When a caller passes an `AbortSignal` to a server function and aborts it (e.g. a request cancelled on navigation/unmount), the RPC `fetch` rejects with `DOMException: signal is aborted without reason` (`AbortError`). `getResponse` `console.log`s it and re-throws. The
  re-thrown error is then handled by the caller (who deliberately aborted and no longer cares about the result), so the throw itself is fine — but the `console.log` leaves a stray, stack-trace-bearing console entry with no context.

  Because it's `console.log` (not `console.error`) on a *caught-and-rethrown* error, it can't be intercepted from application code: it never reaches a framework error hook or error boundary, and it isn't an unhandled promise rejection, so a `window` `unhandledrejection` listener doesn't catch it either.
  The only app-side workarounds are monkey-patching `console` or patching the package.

https://codesandbox.io/p/devbox/condescending-diffie-z59y9f

 1. Start Dev Server
 2. Open preview browser.
 3. Click the button
 4. examine browser dev console logs

### Expected behavior

Cancelling a server-function call should not log to the console. More generally, `getResponse` should not `console.log` errors it re-throws — the caller is responsible for handling and reporting them. At minimum the log should be removed; if diagnostic logging is intended, gate it behind
  `process.env.NODE_ENV !== 'production'` and skip cancellation errors (`AbortError` / `CancelledError`).

### Screenshots or Videos

_No response_

### Platform

- `@tanstack/react-start` 1.168.30 (`@tanstack/start-client-core` 1.170.13)
- React 19, Vite 8
- Browser: Chrome (reproduces in both dev and production builds)

### Additional context

_No response_

gonzoblasco:
Fixes #7873

`serverFnFetcher`'s `getResponse` was logging every non-Response error via `console.log`, including benign `AbortError`s from cancelled requests. This spammed the browser console in both dev and production builds.

Added a guard that skips the `console.log` for `DOMException` with `name === 'AbortError'` — the caller deliberately aborted and will handle the rejection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request cancellation handling so intentionally aborted requests no longer generate unnecessary error logs.
  * Other request errors continue to be reported as before.
  * This provides cleaner error reporting while preserving existing behavior for unexpected failures.


gonzoblasco:
Hey, just bumping this one - it's a small fix skipping a noisy console.log for AbortError. Would appreciate a maintainer review when you have a moment.

gonzoblasco:
Closing as superseded - current `main` no longer logs this error at all.

#8204 (`cbbfbe37`, merged 2026-09-16) removed the `console.log(error)` from `getResponse` entirely, so it now rethrows non-`Response` errors without logging. That fully covers this PR's intent, and the broader ask in #7873, which makes the `AbortError` guard redundant.

The branch conflicts with `main` for exactly that reason: the guarded line no longer exists. Rebased onto current `main`, the change would be empty, so there is nothing left to merge here.

PR #7893, opened 2026-07-24
fix(router-core): preserve encoded slashes in splat params

### LINKED ISSUE #7871: Splat params with an encoded `%2F` becomes indistinguishable from a path separator

Router

### Describe the bug

A splat param (`$` / `_splat`) spans multiple path segments, so the only way to consume it is to split it on `/`. But the router decodes the *whole* splat with `decodeURIComponent` before handing it over, which turns any `%2F` inside a segment into a literal `/`. By the time application code splits, the separator and the encoded slash are the same character and the original segmentation is unrecoverable.

Notably, the router is already careful about this elsewhere: `location.pathname` is decoded with `decodeURI` (via `decodePath`), which by design leaves reserved characters such as `%2F` encoded. So for the same URL, `location.pathname` preserves the distinction and `params._splat` destroys it.

## Reproduction

StackBlitz: 

```js
import {
  createRootRoute, createRoute, createRouter, createMemoryHistory,
} from "@tanstack/react-router";

const root = createRootRoute();
const named = createRoute({ getParentRoute: () => root, path: "/make/$make" });
const splat = createRoute({ getParentRoute: () => root, path: "/parts/$" });
const routeTree = root.addChildren([named, splat]);

async function inspect(url) {
  const router = createRouter({
    routeTree,
    history: createMemoryHistory({ initialEntries: [url] }),
  });
  await router.load();
  const match = router.state.matches.at(-1);
  console.log(url);
  console.log("  params           :", JSON.stringify(match.params));
  console.log("  location.pathname:", JSON.stringify(router.state.location.pathname));
  if (match.params._splat !== undefined) {
    console.log("  _splat.split('/'):", JSON.stringify(match.params._splat.split("/")));
  }
}

// "Ford/New Holland" is one value containing a slash, encoded as %2F.
await inspect("/make/Ford%2FNew%20Holland");
await inspect("/parts/make/Ford%2FNew%20Holland");
// For comparison, a genuinely two-segment path.
await inspect("/parts/make/Ford/New%20Holland");
```

Output:

```
/make/Ford%2FNew%20Holland
  params           : {"make":"Ford/New Holland"}
  location.pathname: "/make/Ford%2FNew Holland"

/parts/make/Ford%2FNew%20Holland
  params           : {"*":"make/Ford/New Holland","_splat":"make/Ford/New Holland"}
  location.pathname: "/parts/make/Ford%2FNew Holland"
  _splat.split('/'): ["make","Ford","New Holland"]

/parts/make/Ford/New%20Holland
  params           : {"*":"make/Ford/New Holland","_splat":"make/Ford/New Holland"}
  location.pathname: "/parts/make/Ford/New Holland"
  _splat.split('/'): ["make","Ford","New Holland"]
```

The last two URLs are different — one has an encoded slash inside a segment, the other has a real separator — and `location.pathname` still tells them apart. But they produce an identical `_splat`, so any consumer that splits on `/` cannot distinguish them.

https://stackblitz.com/github/rshaul/tanstack-repro

1. Go to https://stackblitz.com/github/rshaul/tanstack-repro
2. Inspect console

### Expected behavior

`_splat` should preserve the segmentation of the matched path, so that splitting it on `/` reproduces the segments the URL actually had

### Screenshots or Videos

_No response_

### Platform

- `@tanstack/react-router` 1.170.18
- `@tanstack/router-core` 1.171.15
- Node 24.15.0

### Additional context

_No response_

xianjianlf2:
Fixes #7871.

Splat params span multiple path segments, so decoding the whole splat with `decodeURIComponent` makes an encoded slash inside a segment indistinguishable from a real path separator.

This decodes splat params segment-by-segment and keeps `%2F` encoded inside each segment. Other encoded characters in the segment, like `%20`, `%23`, and `%3D`, are still decoded as before. The same helper is used for fuzzy `**` params.

Tests:
- `pnpm --dir packages/router-core test:unit --run tests/path.test.ts`
- `pnpm --dir packages/router-core test:eslint`
- `pnpm --dir packages/history build && pnpm --dir packages/router-core test:types:ts70`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed wildcard (`*`) and fuzzy catch-all (`**`) route parameter decoding so encoded slashes (`%2F`) stay within the same matched segment.
  * Improved handling of reserved characters inside wildcard and fuzzy catch-all matches.
  * Ensured encoded slashes are not treated the same as real path separators (`/`).
* **Tests**
  * Extended route matching tests to cover encoded-slash preservation and fuzzy catch-all behavior.

### DIFF
--- a/packages/router-core/src/new-process-route-tree.ts
+++ b/packages/router-core/src/new-process-route-tree.ts
@@ -860,6 +860,19 @@ type ParamExtractionState = {
   segment: number
 }
 
+function decodeSplatParam(value: string) {
+  return value
+    .split('/')
+    .map((part) =>
+      decodeURIComponent(
+        // Decode each path segment, but keep encoded slashes inside a segment
+        // distinct from the real slashes that separate splat segments.
+        part.replace(/%2F/gi, (match) => `%25${match.slice(1)}`),
+      ),
+    )
+    .join('/')
+}
+
 /**
  * This function is "resumable":
  * - the `leaf` input can contain `extract` and `rawParams` properties from a previous `extractParams` call
@@ -950,7 +963,7 @@ function extractParams<T extends RouteLike>(
         currentPathIndex + (n.prefix?.length ?? 0),
         path.length - (n.suffix?.length ?? 0),
       )
-      const splat = decodeURIComponent(value)
+      const splat = decodeSplatParam(value)
       // TODO: Deprecate *
       rawParams['*'] = splat
       rawParams._splat = splat
@@ -1313,7 +1326,7 @@ function getNodeMatch<T extends RouteLike>(
     }
     const splat = sliceIndex === path.length ? '/' : path.slice(sliceIndex)
     bestFuzzy.rawParams ??= Object.create(null)
-    bestFuzzy.rawParams!['**'] = decodeURIComponent(splat)
+    bestFuzzy.rawParams!['**'] = decodeSplatParam(splat)
     return bestFuzzy
   }
 
--- a/packages/router-core/src/new-process-route-tree.ts
+++ b/packages/router-core/src/new-process-route-tree.ts
@@ -860,6 +860,13 @@ type ParamExtractionState = {
   segment: number
 }
 
+/**
+ * Decodes a splat match while preserving encoded slashes within each segment.
+ *
+ * Splat values can span multiple URL segments. Decoding the whole value at once
+ * would make `%2F` indistinguishable from the literal `/` separators already in
+ * the matched path, so each segment is decoded independently.
+ */
 function decodeSplatParam(value: string) {
   return value
     .split('/')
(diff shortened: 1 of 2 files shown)


xianjianlf2:
Added a short docstring for the splat decoding helper to clarify why decoding happens segment-by-segment and why encoded slashes stay escaped.

Local check:
- `pnpm --dir packages/router-core test:unit --run tests/path.test.ts`

HarperZ9:
The core fix traces correctly — splitting on `/` and re-escaping `%2F` per segment keeps an inner encoded slash distinct from the real separators, so `Ford%2FNew Holland` survives.

One edge worth a test: a double-encoded literal `%252F` and a real encoded `%2F` both come out as `%2F` (the `/%2F/gi` replace doesn't match inside `%252F`, and `decodeURIComponent('%252F')` is also `%2F`), so the half-decoded `_splat` can't tell them apart. Might be the intended contract, but a `%252F` case would pin it down.

Also heads up — #8141 rewrites the same splat decode in `new-process-route-tree.ts` via a centralized `string-encoding` module, so the two will conflict. Worth coordinating so whichever lands keeps the `%2F`-preservation this one adds.

PR #7897, opened 2026-07-25
fix(router-core): stringify parent params for missing route templates

### LINKED ISSUE #7877: Link to a server-only route skips parent params.stringify on the client

Start and Router

### Describe the bug

When a parent route uses `params.parse` and `params.stringify`, a `<Link reloadDocument>` to a server-only child route can generate different hrefs during SSR and hydration.

The server applies the parent `params.stringify`, while the client does not.

### Reproduction

Create a parent route that maps the public `articles` segment to the internal `article` value:

```tsx
// routes/$postType/route.tsx
import { Outlet, createFileRoute } from '@tanstack/react-router'

export const Route = createFileRoute('/$postType')({
  params: {
    parse: ({ postType }) =>
      postType === 'articles'
        ? { postType: 'article' as const }
        : false,
    stringify: ({ postType }) => ({
      postType: postType === 'article' ? 'articles' : postType,
    }),
  },
  component: Outlet,
})
```

Render a link from its index route:

```tsx
// routes/$postType/index.tsx
import { Link, createFileRoute } from '@tanstack/react-router'

export const Route = createFileRoute('/$postType/')({
  component: IndexComponent,
})

function IndexComponent() {
  return (
    <Link
      to="/$postType/$postId/download"
      params={{
        postType: 'article',
        postId: '1',
      }}
      reloadDocument
    >
      Download
    </Link>
  )
}
```

Define the destination as a server-only route:

```ts
// routes/$postType/$postId/download.ts
import { createFileRoute } from '@tanstack/react-router'

  '/$postType/$postId/download',
)({
  server: {
    handlers: {
      GET: () => new Response('file'),
    },
  },
})
```

Then:

1. Start the application.
2. Visit `/articles`.
3. Check the hydration warning.
4. Click the download link.

### Actual behavior

SSR renders:

```txt
/articles/1/download
```

During hydration, the client generates:

```txt
/article/1/download
```

React reports an href hydration mismatch.

The DOM may continue to show `/articles/1/download`, but clicking the Link uses the client-generated `/article/1/download`, resulting in a 404.

The server-only route is available in the generated route types, so the Link passes TypeScript validation. However, it is absent from the client runtime route tree, so client-side URL generation does not apply the parent `params.stringify`.

This was discovered in the same application as #7731, but appears to be a separate issue involving server-only route pruning.

### Expected behavior

SSR and client-side URL generation should produce the same canonical href:

```txt
/articles/1/download
```

A type-safe Link to a server-only route should apply the same parent `params.stringify` chain on both sides.

If linking to server-only routes is intentionally unsupported, the generated route types should prevent this usage or the limitation should be documented.

### Current workaround

Adding an empty component keeps the destination in the client route tree:

```ts
  '/$postType/$postId/download',
)({
  component: () => null,
  server: {
    handlers: {
      GET: () => new Response('file'),
    },
  },
})
```

After adding the component, both SSR and the client generate `/articles/1/download`.

### Version

```txt
@tanstack/react-router: 1.170.18
@tanstack/react-start: 1.168.32
```

xianjianlf2:
Fixes #7877.

When Start prunes a server-only leaf route from the client route tree, a client-side `buildLocation()` call for that route template cannot find an exact `routesByPath` entry. The current fallback leaves `destRoutes` empty for route templates, so parent `params.stringify` hooks are skipped and the generated href can differ from SSR.

This keeps exact route-template matches unchanged, but when a template is missing it falls back to the longest known parent route branch. That gives parent routes a chance to canonicalize path params while still leaving the server-only leaf absent from the client route tree.

The regression test models the client route tree after a server-only child was pruned: only `/$postType` exists locally, while a link is built for `/$postType/$postId/download`. The parent stringify maps `article` back to the canonical `articles` segment.

Tests run:
- `PATH=/opt/homebrew/bin:$PATH pnpm --dir packages/history build`
- `PATH=/opt/homebrew/bin:$PATH pnpm --dir packages/router-core test:unit --run tests/build-location.test.ts -t "matched parent of a missing route template"`
- `PATH=/opt/homebrew/bin:$PATH pnpm --dir packages/router-core test:unit --run tests/build-location.test.ts`
- `PATH=/opt/homebrew/bin:$PATH pnpm --dir packages/router-core test:unit`
- `PATH=/opt/homebrew/bin:$PATH pnpm --dir packages/router-core test:types:ts70`
- `PATH=/opt/homebrew/bin:$PATH pnpm --dir packages/router-core test:eslint`
- `PATH=/opt/homebrew/bin:$PATH pnpm exec prettier --check packages/router-core/src/router.ts packages/router-core/tests/build-location.test.ts`

Note: `test:eslint` exits 0 with existing warnings in unrelated router-core files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved navigation to template-based destinations when a deeply nested route is not available in the client route tree.
  * Preserved parameter formatting and canonical URL generation using the nearest known parent route.
  * Ensured nested paths resolve correctly, including transformed URLs such as `/articles/1/download`.
  * Prevented incomplete route chains and incorrect destination URLs during navigation.


xianjianlf2:
Merged latest `origin/main` into `fix/server-route-link-stringify-7877` and resolved the router-core conflict against the current tuple-based matching code. The parent-route fallback for missing server-only leaf templates is preserved so params.stringify still canonicalizes links.

Validation:
- `pnpm install --frozen-lockfile`
- `pnpm exec prettier --check packages/router-core/src/router.ts packages/router-core/tests/build-location.test.ts`
- `pnpm --dir packages/history run build`
- `pnpm --dir packages/router-core exec vitest run tests/build-location.test.ts` (73 passed, no type errors)

Current head: `8a73eddc2e54d879403637b8b6c96234ac0b54d3`

schiller-manuel:
this is not supported for server only links

PR #7901, opened 2026-07-25
test(rsc): add dev-mode HMR e2e for RSC routes with a co-located server fn

sreetamdas:
### Description

Adds `e2e/react-start/rsc-hmr`: a dev-mode HMR e2e project for TanStack Start **RSC** routes, covering route files that co-locate a `createServerFn` with the route component.

Replaces #7621, which was a deliberate CI canary for a bug that has since been fixed upstream (vitejs/vite-plugin-react#1248, released in `@vitejs/plugin-rsc@0.5.30`). This is the regression suite @schiller-manuel asked for on that PR instead.

Built per his review there: a separate project rather than bolting onto `e2e/react-start/rsc`, following `e2e/react-start/hmr`'s CI hardening, multiple routes including one with a CSS module and one with no CSS, and both the vite and rsbuild adapters.

### Why a Start-level test, given the fix was upstream

Upstream's own regression test (`plugin-rsc`'s `examples/client-first`) exercises the plugin in isolation. It does not go through Start's `?tss-serverfn-split` transform, which is what puts a co-located route file into the rsc module graph in the first place, and it cannot cover the rsbuild adapter at all.

Worth being precise about the two lanes: `@vitejs/plugin-rsc` is Vite-only, so the **vite** lane is what guards against this specific bug regressing. The **rsbuild** lane goes through Rspack / `react-server-dom-rspack` and is adapter-parity coverage — valuable, but a different code path.

### Layout

`rsc-hmr` = `hmr`'s harness + `rsc`'s configs:

- `playwright.config.ts`, `tests/setup/global.{setup,teardown}.ts` — from `e2e/react-start/hmr`, so it inherits `waitForServer` + `preOptimizeDevServer` (cold compile and Vite's dep-optimize reload storm happen outside any test's timeout) and `createHmrFileEditor` with `normalizeSource`, so a run killed mid-edit self-heals instead of cascading.
- `vite.config.ts` / `rsbuild.config.ts` — from `e2e/react-start/rsc`, both with `rsc: { enabled: true }`.
- `playwrightModes`: `vite-ssr` and `rsbuild-ssr`. `workers: 1` and no shards, since the tests mutate fixture files on disk.
- Declares `mode: 'ssr'` and then runs a dev server, matching what `hmr` does — the nx playwright plugin's mode allowlist is `['ssr','spa','prerender','preview']` with no `dev`. Happy to wire this differently if you'd prefer.

### Fixtures

| Route | What it covers |
|---|---|
| `/co-located` | `createServerFn` + component in one file, no CSS — the original bug |
| `/co-located-css-module` | same, plus a CSS module import |
| `/separate-file` | control: server fn extracted to its own module (the documented workaround) |

Each test seeds client state (a counter plus a `window` sentinel) before editing, then asserts the new text renders **and** the state survived — so a full reload that happens to render the right text fails rather than passes.

The CSS cases are the ones I'd most want kept. The upstream guard decides whether to suppress client HMR by classifying importers as CSS vs non-CSS, so the interaction most at risk is a route with both kinds — in both directions: the component must still Fast Refresh, and a CSS-only edit must still hot-swap without a full reload. Both are asserted.

### Verification

Run locally against both adapters:

- `E2E_TOOLCHAIN=vite` — 4/4 pass
- `E2E_TOOLCHAIN=rsbuild` — 4/4 pass
- `tsc --noEmit` — clean

And, more usefully, the suite was checked against the **broken** plugin version. Pinning this project to `@vitejs/plugin-rsc@0.5.27` and re-running gives:

```
2 failed
  › Fast Refreshes a co-located route component and preserves state
  › Fast Refreshes when the route also imports a CSS module
2 passed
```

with the failure mode from the original report — `Received: "co-located-baseline"`, and only `(rsc) hmr update /src/routes/co-located.tsx?tss-serverfn-split` in the terminal with no client update. The CSS-only edit and the separate-file control stay green, since neither was ever affected. So the suite fails specifically on this bug rather than merely passing on a fixed version.

Note `0.5.27` is inside the `>=0.5.20` peer range that `@tanstack/react-start-rsc` currently declares, which is what #7900 addresses.

### Notes

- No changeset — `rsc-hmr` is `private: true` and unpublished.
- No changes needed to `nx.json`, `pnpm-workspace.yaml`, or any workflow: `e2e/react-start/*` is globbed as a workspace, and `scripts/nx/playwright-plugin.ts` synthesises targets from the `nx.metadata.playwrightModes` block in the project's own `package.json`.
- The extracted module in the control fixture is `-separateFile.tsx`, deliberately without a `.server.` suffix — import protection treats those as server-only and swaps in a client mock, which breaks the route instead of exercising the workaround. Flagging it because #7618's workaround wording suggests `*.server.ts`, which is misleading for a module a route imports directly.

## Summary by CodeRabbit

* **New Features**
  * Added a React Server Components HMR end-to-end test app with routes demonstrating server-rendered content, client state preservation, and CSS-module updates.
  * Enabled testing with both Vite and Rsbuild toolchains.
  * Added Playwright coverage verifying updates occur through HMR without full-page reloads.
* **Chores**
  * Added E2E configuration, strict TypeScript settings, global test setup and teardown, and ignore rules for development artifacts.


LadyBluenotes:

PR #7919, opened 2026-07-29
fix(start-plugin-core): fail fast when the Vite plugin runs on Vite older than v7

### LINKED ISSUE #7918: In spa mode _shell.html don generate on Vite v6

Start

### Describe the bug

When I use Vite v6 and @tanstack/react-start (v1.168.32) with spa mode, the `_shell.html` on build doesn't generate. In the terminal, I see the step to building the client environment and ssr but after this, no prerendered. In `dist` folder, we have `dist/client` without `_shell.html`.

https://github.com/apacholik-univio/tanstack-router-and-vite-6

Follow the instructions in the [TanStack Start | Build a Project from Scratch](https://tanstack.com/start/latest/docs/framework/react/build-from-scratch), but install Vite version 6:

```bash
pnpm add -D vite@v6 @vitejs/plugin-react
```

In `vite.config.ts` enable spa mode:
```
tanstackStart({
  spa: {
    enabled: true,
  }
})
```

Then run the build command.
The result is `dist/client` without `_shell.html`.

### Expected behavior

Independent of the Vite version (supported; the latest version from the 6.*.* was released 2 months ago), the `@tanstack/react-start/plugin/vite` plugin should work correct.

Alternatively, there should be a note in the documentation that it works starting with Vite v7.

It would also be helpful if the build process threw an exception indicating that it cannot build (?) the shell.

### Screenshots or Videos

_No response_

### Platform

- Start Version: v1.168.32
- OS: Windows
- Bundler: vite
- Bundler Version: 6.4.3

### Additional context

_No response_

Mohith26:
Fixes #7918, reported by @apacholik-univio, implementing the issue's requested fail-fast behavior.

Root cause: `_shell.html` (and all Start post-build steps: prerender, sitemap) run via the plugin-level `buildApp` hook. Vite 7 iterates `getSortedPlugins('buildApp')`, but Vite 6 only calls the config-level `builder.buildApp` and silently ignores plugin hooks, so on Vite 6 both environments build and post-build never runs. Vite 6 is outside the declared peer range (`vite: ">=7.0.0"`), but that peer is optional (for rsbuild users), so package managers never enforce it, which is how a Vite 6 app gets a silent broken build.

Making Vite 6 actually work would contradict the declared support range, so this adds an `assertSupportedViteVersion` guard called from the Start core plugin's `config` hook that throws a clear error naming the detected version, the steps that would be silently skipped, and the upgrade path.

Verified end to end with published packages: unpatched on Vite 6.4.3 reproduces the silent success with no `_shell.html`; patched on 6.4.3 fails fast with the actionable error; patched on Vite 7.3.6 and the in-repo e2e on 8.0.14 generate shells normally. Unit suite: 517 passed (11 new; the new test file fails without the fix). Changeset (patch, `@tanstack/start-plugin-core`), eslint/types/build targets all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * TanStack Start now fails fast with a clear compatibility error when used with Vite versions older than 7.
  * Prevents builds from appearing successful when post-build tasks such as prerendering or SPA shell generation are skipped.
  * Supported Vite versions continue to work as expected.

### DIFF
--- a/.changeset/fair-pugs-argue.md
+++ b/.changeset/fair-pugs-argue.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/start-plugin-core': patch
+---
+
+Fail fast with a clear error when the TanStack Start Vite plugin runs on Vite older than v7. Vite 6 and older never invoke the `buildApp` plugin hook, so builds appeared to succeed while post-build steps such as prerendering and SPA shell (`_shell.html`) generation were silently skipped.
--- a/packages/start-plugin-core/src/vite/plugin.ts
+++ b/packages/start-plugin-core/src/vite/plugin.ts
@@ -1,3 +1,4 @@
+import { version as viteVersion } from 'vite'
 import { crawlFrameworkPkgs } from 'vitefu'
 import {
   applyResolvedBaseAndOutput,
@@ -35,6 +36,7 @@ import {
 } from './output-directory'
 import { postServerBuild } from './post-server-build'
 import { serializationAdaptersPlugin } from './serialization-adapters-plugin'
+import { assertSupportedViteVersion } from './vite-version'
 import type {
   TanStackStartVitePluginCoreOptions,
   ViteRscForwardSsrResolverStrategy,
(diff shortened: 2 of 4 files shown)


schiller-manuel:
that's what version ranges are for. i don't see a need for manually verifying this on our end

PR #7925, opened 2026-07-30
fix(start-server-core): return 406 instead of 500 for non-HTML Accept headers

### LINKED ISSUE #7913: Start: page route returns 500 when Accept excludes HTML

Start

### Describe the bug

 `createStartHandler` refuses to render a page route when the request's `Accept`
  header does not allow HTML, and reports that refusal as `500`:

https://github.com/TanStack/router/blob/179d9b9b5eec50dc7833b4e2ef91b735d34bbc07/packages/start-server-core/src/createStartHandler.ts#L558-L561

  A failed content negotiation is a client-side condition, so this should be
  `406 Not Acceptable`. `500` says the server malfunctioned, and nothing did.

reproduction doesn't need any custom code

Against any Start app, on any page (non-server) route:
```sh
  curl -i -H 'Accept: application/json' https://example.com/
  curl -i -H 'Accept: image/webp'       https://example.com/
```
Both return 500 `{"error":"Only HTML requests are supported here"}`. Accept: */*, a browser Accept, and an absent header all return 200.

### Expected behavior

Should return `406 Not Acceptable` instead of `500 Server Error`.

### Screenshots or Videos

_No response_

### Platform

- @tanstack/react-start 1.168.32
- @tanstack/start-server-core 1.169.17 (current latest)

### Additional context

_No response_

okxint:
**Problem**

When a request's `Accept` header doesn't include HTML or `*/*`, `createStartHandler` correctly rejects it but returns `HTTP 500 Internal Server Error`. This is wrong — the problem is with the request, not the server.

**Fix**

Return `HTTP 406 Not Acceptable`, which is what the HTTP spec defines for failed content negotiation.

```diff
- { status: 500 },
+ { status: 406 },
```

Fixes #7913

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed code-split route loading when generated paths contain single quotes.
  * Requests that do not accept HTML now receive the appropriate `406 Not Acceptable` response instead of a server error.

### DIFF
--- a/packages/router-plugin/src/core/code-splitter/compilers.ts
+++ b/packages/router-plugin/src/core/code-splitter/compilers.ts
@@ -623,7 +623,7 @@ export function compileCodeSplitReferenceRoute(
                         ) {
                           programPath.unshiftContainer('body', [
                             template.statement(
-                              `const ${splitNodeMeta.localImporterIdent} = () => import('${splitUrl}')`,
+                              `const ${splitNodeMeta.localImporterIdent} = () => import('${splitUrl.replace(/'/g, "\\'")}')`,
                             )(),
                           ])
                         }
@@ -719,7 +719,7 @@ export function compileCodeSplitReferenceRoute(
                         ) {
                           programPath.unshiftContainer('body', [
                             template.statement(
-                              `const ${splitNodeMeta.localImporterIdent} = () => import('${splitUrl}')`,
+                              `const ${splitNodeMeta.localImporterIdent} = () => import('${splitUrl.replace(/'/g, "\\'")}')`,
                             )(),
                           ])
                         }
--- a/packages/start-server-core/src/createStartHandler.ts
+++ b/packages/start-server-core/src/createStartHandler.ts
@@ -557,7 +557,7 @@ export function createStartHandler<TRegister = Register>(
           return normalizeSsrResponse(
             Response.json(
               { error: 'Only HTML requests are supported here' },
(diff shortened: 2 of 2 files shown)


packages/router-plugin/src/core/code-splitter/compilers.ts
```diff
@@ -623,7 +623,7 @@ export function compileCodeSplitReferenceRoute(
                         ) {
                           programPath.unshiftContainer('body', [
                             template.statement(
-                              `const ${splitNodeMeta.localImporterIdent} = () => import('${splitUrl}')`,
```
schiller-manuel, line 626: this file should not be changed in this PR

schiller-manuel (changes requested):

packages/start-server-core/src/createStartHandler.ts
```diff
@@ -557,7 +557,7 @@ export function createStartHandler<TRegister = Register>(
           return normalizeSsrResponse(
             Response.json(
               { error: 'Only HTML requests are supported here' },
-              { status: 500 },
```
schiller-manuel, line 560: we should add an e2e test

EdwardSalkeld:
@okxint are you intending to fix this up to address review comments? if not, I can do so in a new pr

EdwardSalkeld:
Same change made by @schiller-manuel in https://github.com/TanStack/router/pull/8204 - no need for this

PR #7952, opened 2026-08-04
perf(react-router): bail out of Link re-renders when href and active state are unchanged

### LINKED ISSUE #7951: Link re-renders on every navigation even when its href and active state are unchanged
## Summary

`useLinkProps` subscribes every client-side `Link` to the location store with an identity selector, then derives `href` and `isActive` from the published location in downstream memos. The comparator can only ask "is this a different URL?", never "does this link care?" — so every `Link` on the page re-renders on every navigation, including the ones whose rendered output is byte-identical before and after.

On one page in our app (Mixcloud), **313 links persisted across a single navigation and exactly 1 changed its rendered output.** The other 312 re-rendered for nothing.

Fix and regression test in #7952.

## Not the same as #2011 / #2516

Worth separating up front, because it looks adjacent. [#2011](https://github.com/TanStack/router/discussions/2011) and the work that came out of it ([#2359](https://github.com/TanStack/router/pull/2359) → [#2516](https://github.com/TanStack/router/pull/2516)) reduced the **cost of each render**, via the `routesByPath` fast path in `buildLocation`. That landed and helps.

This is about the **number of renders**. The two are independent: after #2516 each `Link` render is cheaper, but every `Link` still renders on every navigation. In my measurements the remaining `buildLocation` time is already small — 1563 calls totalling 15.8ms — and the residue is the React render pass itself.

## Version

Reproduced on `@tanstack/react-router@1.170.17`. I diffed `packages/react-router/src/link.tsx` against the published `1.170.18` tarball and against `main` — **byte-identical in all three**, so this applies to the current release and to unreleased `main`.

## Mechanism

[`src/link.tsx:403-407`](https://github.com/TanStack/router/blob/main/packages/react-router/src/link.tsx):

```ts
const currentLocation = useStore(
  router.stores.location,
  (l) => l,                                  // publishes the whole location
  (prev, next) => prev.href === next.href,   // fires on any URL change
)
```

`currentLocation` then feeds two derivations, both of which list it as a dependency:

```ts
const next = React.useMemo(() => {
  const opts = { _fromLocation: currentLocation, ..._options }
  return router.buildLocation(opts as any)
}, [router, currentLocation, _options])
```

```ts
const isActive = React.useMemo(() => { /* compares currentLocation against next */ }, [
  /* …, */ currentLocation, externalLink, isHydrated, next.hash, next.pathname, next.search, router.basepath,
])
```

So both recompute on every navigation and the component re-renders regardless of whether either result differs.

The `_isServer` branch already avoids this deliberately — its comment says it computes active status "by reading from the location store directly" specifically to "avoid creating any router-state subscriptions". The client path has no equivalent, and `activeOptions` only tunes *how* activeness is compared, never *whether* the component subscribes.

## Measurements

Real app, ~578 anchors on the page, navigating between two sibling routes under the same layout. Only the leaf content and one nav tab change.

**How much of the work is wasted** — snapshotting each persisted anchor's `href` and `data-status` either side of the navigation:

| | |
| --- | --- |
| links persisting across the navigation | 313 |
| whose `href` or active status changed | **1** |
| whose derived output was identical | **312 (99.7%)** |

The one that changed was the tab being navigated to, gaining `data-status="active"`.

**What the wasted renders cost** — per-commit `selfBaseDuration`, attributed by walking fibers carrying React's `PerformedWork` flag:

| | mount | update |
| --- | --- | --- |
| whole page | 679.9ms | 200.6ms |
| TanStack `Link` | 97.6ms (759 mounts) | **24.1ms (364 renders)** |

~24ms of ~880ms on that navigation, or ~12% of update-render time. On a lighter navigation back to the smaller list it was 16.5ms of 65.5ms update time.

**Caveats so these aren't over-read:** development build, so absolute numbers are inflated relative to production — the proportions are the reliable part. Single machine, single page. And the mount column dominates the update column for this page, so the case for fixing this is that it is ~100% waste and scales with link count, not that it is the largest number on the page.

## Direction taken in the PR

Publish the location-derived values through the selector so the comparator can bail out, instead of publishing the location and deriving afterwards:

```ts
const { href, externalLink, isActive } = useStore(
  router.stores.location,
  selectLinkState,   // buildLocation + getHrefOption + isActive, memoized on router/_options/…
  compareLinkState,  // href === href && externalLink === externalLink && isActive === isActive
)
```

`buildLocation` still runs once per link per location change — what goes away is the React render and the host reconciliation under it.

One implementation note worth flagging, since it is why this isn't a one-liner: `Link` passes `_fromLocation: currentLocation` explicitly, and in `buildLocation` that is the head of a fallback chain (`dest._fromLocation || this.pendingBuiltLocation || this.latestLocation`, `router-core/src/router.ts:1857`). It is deliberately pinning resolution to the *subscribed* location rather than letting it fall through to `latestLocation`, which can differ mid-transition. The selector formulation preserves that, because it derives from the value being published — but a naive "subscribe to a boolean" would not, since `next` still needs the location for relative `to` resolution and param inheritance.

## Scope

**Only `packages/react-router` is addressed.** `packages/solid-router` and `packages/vue-router` have the same shape and would want the same treatment; I have not touched them, so please don't assume they're covered.

`packages/react-router/tests/link.bench.tsx` exists and I did not run it — it looks like the natural place for a perf guard if you want one.

matclayton:
Fixes #7951.

## Problem

`useLinkProps` subscribes every client-side `Link` to the location store with an identity selector:

```ts
const currentLocation = useStore(
  router.stores.location,
  (l) => l,
  (prev, next) => prev.href === next.href,
)
```

`href` and `isActive` are then derived in downstream memos that list `currentLocation` in their dependencies. So the comparator can only ask "is this a different URL?", never "does this link care?" — and every `Link` on the page re-renders on every navigation, including the ones whose rendered output is identical before and after.

On one page in our app (Mixcloud), **313 links persisted across a single navigation and exactly 1 changed its rendered output** — the tab gaining `data-status="active"`. The other 312 re-rendered for nothing. Full measurements are in #7951.

This is complementary to #2359 / #2516, which reduced the *cost* of each `Link` render via the `routesByPath` fast path. This reduces the *number* of renders. After #2516 each render is cheaper, but every link still renders on every navigation — the remaining `buildLocation` time measured at 15.8ms across 1563 calls, with the residue being the React render pass itself.

## Change

The location-derived values move into the selector, and `compareLinkState` compares the three resulting primitives, so a link whose `href` and active state are unaffected by a navigation bails out.

`buildLocation` still runs once per link per location change. What goes away is the React render and the host reconciliation beneath it.

Supporting details:

- The `isActive` and `externalLink` bodies move to module-level helpers unchanged, so the selector stays readable.
- `activeOptions` is spread into its four primitive fields in the selector's dependency list rather than depended on directly, because callers routinely pass an inline object literal — this matches what the previous `isActive` memo already did.
- `doPreload` no longer pre-supplies `_builtLocation`, because the built location is no longer held in render state. `preloadRoute` already falls back to `opts._builtLocation ?? this.buildLocation(opts)`, and `handleClick` has always let `router.navigate` build its own. Net cost is one extra `buildLocation` per hover.

The subscribed-location pinning is preserved deliberately. `Link` passes `_fromLocation: currentLocation`, and in `buildLocation` that is the head of `dest._fromLocation || this.pendingBuiltLocation || this.latestLocation` — pinning resolution to the subscribed snapshot rather than one that can differ mid-transition. The selector formulation keeps that, because it derives from the value being published. A naive "subscribe to a boolean" would not, since `next` still needs the location for relative `to` resolution and param inheritance.

## Tests

`tests/link.test.tsx` → `link re-render bail-out`. It renders two memoized components that call `useLinkProps`, navigates, and asserts that the link the navigation cannot affect does not re-render. The components are memoized so a re-render of the owning route component cannot be mistaken for the subscription firing, and the link options are module-stable for the same reason.

Verified in both directions by reverting the source change and keeping the test:

- without the change: `AssertionError: expected 4 to be 2` (one re-render, doubled by StrictMode)
- with the change: passes

Existing suite, against the unmodified baseline on the same commit:

| | baseline | this branch |
| --- | --- | --- |
| `test:unit` | 943 passed, 1 skipped | **944 passed**, 1 skipped |
| `test:types` | passes | passes |
| `test:eslint` | 0 errors, 97 warnings | 0 errors, 97 warnings |
| `test:build` | passes | passes |

## Scope

**Only `packages/react-router`.** `packages/solid-router` and `packages/vue-router` have the same shape and would want the same treatment — I have not touched them, so please don't assume they're covered. Happy to follow up on those if you'd like it done the same way.

`tests/link.bench.tsx` exists and I did not run it; it looks like the natural home for a perf guard here if you want one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link updates during navigation so active links refresh correctly.
  * Reduced unnecessary re-rendering for links whose destination and active state remain unchanged.
  * Improved handling of external URLs and blocked unsafe protocols.
  * Enhanced route preloading consistency by rebuilding destination details when needed.
  * Links now provide more reliable navigation behavior as the application’s location changes.


packages/react-router/src/link.tsx
```diff
@@ -563,13 +612,13 @@ export function useLinkProps<
...
-    router
-      .preloadRoute({ ..._options, _builtLocation: next } as any)
-      .catch((err) => {
-        console.warn(err)
-        console.warn(preloadWarning)
-      })
-  }, [router, _options, next])
+    // `preloadRoute` builds the location itself; it is no longer held in render
+    // state. Matches `handleClick`, which lets `router.navigate` build its own.
+    router.preloadRoute({ ..._options } as any).catch((err) => {
```
Sheraff: why not return the built location from `selectLinkState` and use it here? It's a cheap way to bypass a bunch of work (thouh admittedly "preload" work happens less frequently than "location update" re-renders)
(maybe it's not possible without causing a re-render on every location change, just asking in case it is. This might also be the only place in the repo where we use `_builtLocation` as a param for `preloadRoute`, so if we really don't use it anymore, we might be able to clean that code path up)

packages/react-router/src/link.tsx
```diff
@@ -31,6 +33,101 @@ import type {
   ValidateLinkOptionsArray,
 } from './typePrimitives'
 
+interface LinkState {
+  href: string | undefined
+  externalLink: string | undefined
+  isActive: boolean
+}
```
Sheraff: A tuple would minify better
```suggestion
type LinkState  = [
  href: string | undefined,
  externalLink: string | undefined,
  isActive: boolean,
]
```

packages/react-router/src/link.tsx
```diff
@@ -399,128 +496,80 @@ export function useLinkProps<
     ],
   )
 
+  const {
+    exact: activeExact,
+    explicitUndefined: activeExplicitUndefined,
+    includeHash: activeIncludeHash,
+    includeSearch: activeIncludeSearch,
+  } = activeOptions ?? {}
```
Sheraff: I think (needs to be checked) that we would save a few bytes by not destructuring here, and just

- using property access in the dependencies array
  ```ts
    activeOptions.exact,
    activeOptions.explicitUndefined,
    activeOptions.includeHash,
    activeOptions.includeSearch,
  ```
- and not re-building the object when calling `resolveIsActive`
  ```ts
    resolveIsActive(
      //...,
      activeOptions,
      //...,
    )
  ```

packages/react-router/src/link.tsx
matclayton: Good call, done in `2ec281e`. Measured on the unminified build it's 54 bytes off each of `dist/esm/link.js` and `dist/cjs/link.cjs`. Less after minify and gzip, but the direction holds. Kept the labelled-tuple form so the field names still document themselves.

packages/react-router/src/link.tsx
matclayton: Tried this: lint fails. Passing `activeOptions` straight through to `resolveIsActive` makes the body reference the whole object, so `react-hooks/exhaustive-deps` then wants it in the dependency array:

```
542:5  error  React Hook React.useCallback has a missing dependency: 'activeOptions'  react-hooks/exhaustive-deps
```

The destructure keeps the deps as primitives while the body still gets an object. Also worth noting the snippet needs `activeOptions?.exact` rather than `activeOptions.exact`, since `activeOptions` is optional and most links don't pass it, so the unguarded access throws on the common path.

Happy to switch if listing `activeOptions` itself is preferable. The cost is that callers passing an inline `activeOptions={{ exact: true }}` literal rebuild the selector on every render. That doesn't cause extra re-renders, since the comparator still returns the previous selection; it just re-runs `buildLocation` on renders that were happening anyway. Either trade works for me, just say which you'd rather have.

packages/react-router/src/link.tsx
matclayton: I don't think it can work, for the reason you suspected. Two ways to try it:

- **Include `next` in the compared state.** Its identity changes on every location change, so the comparator is always false and every link re-renders on every navigation. That removes the point of the PR.
- **Return it but leave it out of the comparator.** When the comparator reports equal, `useSyncExternalStoreWithSelector` hands back the previous selection, so `next` is stale. Harmless for an absolute `to`, wrong for relative `to` and inherited params: you'd preload a location built from the old one, silently.

On cleaning up the code path, not yet: `_builtLocation` is still passed to `preloadRoute` by `packages/solid-router/src/link.tsx:254` and `packages/vue-router/src/link.tsx:255`. It'd only become removable once those get the same treatment, which I've deliberately left alone here.

matclayton:
Pushed `bd4e5b7`, `2ec281e` and `8d04b0b`.

Both nitpicks are addressed: braces on the single-line `if` bodies, and `as any` dropped from `new URL(to)` since the guard already narrows `to` to `string`. The test now asserts the published state as well as the render counts; I confirmed that adds coverage by sabotaging the selector, where a constant tuple with a wrong href but correct active state now fails.

On CodSpeed: most regressions are Solid and Vue benchmarks, and this PR only touches `packages/react-router/src/link.tsx`, so they can't come from it. The memory numbers aren't plausible either (`mem server server-fn-churn (vue)`: 269.5 KB to 4,658.6 KB), and CodSpeed's banner says different runtime environments were compared.

The four React `client-nav` regressions are in scope, so I checked whether the selector runs `buildLocation` more often than the old `useMemo`. It halves it: 917 calls before, 459 after, same page and link counts. I can't rule out that those scenarios have few enough links that the bail-out has nothing to pay for while the selector still costs per render. Do you trust those numbers on this runner? If so I'll dig in; if they're known noisy I'd rather not chase them.

packages/react-router/src/link.tsx
Sheraff: it's ok to disable eslint locally if we can save some bytes (and we know that it is safe to do so)

packages/react-router/src/link.tsx
Sheraff: once we've confirmed this change is acceptable for React, I think it should also be applied to Solid and Vue to limit drift. And then if what you are saying is correct, we will be able to remove `_builtLocation` which saves some bytes again

packages/react-router/src/link.tsx
matclayton: Can re-check this in a bit, just vendoring this into our internal test suite to check if it finds any regressions.

Sheraff:
> On CodSpeed: most regressions are Solid and Vue benchmarks, and this PR only touches `packages/react-router/src/link.tsx`, so they can't come from it. The memory numbers aren't plausible either (`mem server server-fn-churn (vue)`: 269.5 KB to 4,658.6 KB), and CodSpeed's banner says different runtime environments were compared.

yes memory benchmarks should be ignored, i haven't managed to stabilize them yet.

> The four React `client-nav` regressions are in scope, so I checked whether the selector runs `buildLocation` more often than the old `useMemo`. It halves it: 917 calls before, 459 after, same page and link counts. I can't rule out that those scenarios have few enough links that the bail-out has nothing to pay for while the selector still costs per render. Do you trust those numbers on this runner? If so I'll dig in; if they're known noisy I'd rather not chase them.

Some CPU benchmarks are more stable than others, so to take with a grain of salt. `ssr request loop (solid)` is supposed to be one of the stable ones, and it appears here as a regression on code that was not touched so...

All the benchmarks can be run locally though, where you have more control over stability. That includes bundle size benchmarks and all the codspeed benchmarks (that just run on vitest locally).

There is also `e2e/react-start/flamegraph-bench` for benchmarks that really hammer the SSR server, but this PR is purely client-side, right?

packages/react-router/src/link.tsx
matclayton: Happy to have a go at those but it'll be much more AI driven and less me steering, as honestly I'm not a Vue/solid person, would you like them on this PR?

packages/react-router/src/link.tsx
matclayton: Done in `07081d4`. `activeOptions` now goes straight to `resolveIsActive`, with the dependency array listing the four fields and an `exhaustive-deps` disable above it.

Saves 269 bytes on each of `dist/esm/link.js` and `dist/cjs/link.cjs`, unminified.

Kept the `?.` on the dependency entries, since `activeOptions` is undefined on most links and the unguarded access throws on that path.

packages/react-router/src/link.tsx
matclayton: I looked at both. Two findings, and they both point away from the plan.

**The change does not port, because the problem is React specific.** What this PR removes is a React component re-render: `useLinkProps` re-runs and the whole `<a>` subtree reconciles on every navigation. Solid and Vue have no such thing. In Solid, `useLinkProps` runs once per link and `next` / `hrefOption` / `isActive` are `createMemo`s (`packages/solid-router/src/link.tsx:136-203`), so a location change recomputes those memos and updates only the bound attributes. Vue is the same shape with `computed` and a `computedProps` ref built in `setup()` (`packages/vue-router/src/link.tsx:226-244`, `427-449`). Neither re-runs the component body, so there is no re-render to bail out of.

Both also already gate on the href the same way this PR's base did, via the location store's equality option: `solid-router/src/link.tsx:128-131` and `vue-router/src/link.tsx:222-224`. Converting either to a tuple selector would do the same work at coarser granularity, and would lose the per-attribute update that fine-grained reactivity gives them. I would leave both alone.

**`_builtLocation` cannot be removed.** It is core redirect plumbing rather than a link optimisation, so it survives regardless of what the adapters do:

- `router-core/src/router.ts:2437` throws `redirect({ href, _builtLocation: nextLocation })` from core itself.
- `resolveRedirect` reads it at `router.ts:2804-2806` to skip rebuilding the location.
- `router.ts:2825` gates the dangerous-protocol check on its **absence**, so a redirect carrying an internally built location deliberately skips validation that an externally supplied `href` receives. Removing the field would change that security behaviour.

The only piece that becomes dead if all three adapters stop passing it is the `opts._builtLocation ??` fallback in `preloadRoute` (`router.ts:2891`), one line. The type, the redirect plumbing and the protocol gate all stay.

Worth adding that Solid and Vue pass an **already memoized** `next()` into `preloadRoute`, so for them the parameter is a genuine saving. Dropping it there would be a pessimisation, which is the opposite of the byte win. React is the odd one out here only because the selector cannot safely publish `next` (my earlier comment in this thread).

matclayton:
Ran `client-nav` locally as suggested, patched and base **interleaved**, three rounds per side, comparing medians. Interleaving mattered: run sequentially, the numbers moved by up to 10 points on scenarios where the code does provably identical work, so the first pass manufactured regressions that were really machine drift.

Two of the four React regressions (`head`, `control-flow`) come out as **improvements** locally. `rewrites` reproduced, and it was real.

### The rewrites regression was real, and is fixed

Links commonly pass inline `params={{ a: 'x' }}` / `search={{ _locale: 'fr' }}` literals. Those change identity on every parent render, which rebuilt `_options`, which changed the store selector's identity, which discarded `useSyncExternalStoreWithSelector`'s memoized selection. `buildLocation` then ran **twice** per navigation, once in the notification check and once in the render-phase selection, where the previous `useMemo` chain ran it once.

Measured on a replica of the `rewrites` scenario (six links, root subscribed to the pathname via `useLocation`, inline `params`/`search`), `buildLocation` calls per navigation:

| | base | this PR before the fix | this PR now |
| --- | --- | --- | --- |
| per navigation | 7 | 12 | 7 |

The arithmetic pins it: the one link with no inline props builds once, the five with inline props build twice. Fixed by keeping `_options` referentially stable while its contents are equal.

### Current numbers

Medians of three interleaved runs, hz, higher is better:

| scenario | base | this PR | delta |
| --- | --- | --- | --- |
| `client-links` | 62.3 | 121.1 | +94.2% |
| `client-side` | 252.4 | 324.2 | +28.4% |
| `client-route-tree-scale` | 186.4 | 218.7 | +17.3% |
| `client-async-pipeline` | 312.5 | 348.9 | +11.7% |
| `client-head` | 201.4 | 224.7 | +11.6% |
| `client-loaders` | 302.2 | 332.9 | +10.1% |
| `client-search-params` | 137.3 | 151.1 | +10.0% |
| `client-nested-params` | 165.2 | 175.9 | +6.5% |
| `client-preload` | 279.0 | 296.8 | +6.4% |
| `client-control-flow` | 152.1 | 160.5 | +5.5% |
| `client-rewrites` | 238.8 | 247.0 | +3.5% |
| `client-history` | 360.2 | 356.3 | -1.1% |
| `client-mount` | 614.5 | 583.7 | -5.0% |

### One regression left, and I have not explained it

`client-mount` is still about -5%. It reproduces with the first commit alone, so it comes from the core change rather than any of the refinements, but I have not isolated the mechanism: `buildLocation` counts are identical at mount (7 on both sides), and a standalone mount loop plus CPU profile showed parity, though that harness turned out to be timer bound and not sensitive enough to trust. Flagging it rather than claiming it is noise.

### The CodSpeed report above is stale

The workflow runs for the latest commits are sitting in `action_required`, so Benchmarks and Bundle Size have not re-run since `8d04b0b`. The CodSpeed comment still reflects the state before the `rewrites` fix. Whenever you next approve a run it should re-measure.

### Bundle size

Against base, from `benchmark:bundle-size`: +20 to +28 bytes raw and -7 to +27 bytes gzip across the eight React scenarios.

Also dropped a commit along the way. I had added an href memoization that cut `getHrefOption` from 5 calls per navigation to 0, which looked like the fix. It moved the `rewrites` benchmark by 0.05%, so it was reverted rather than carried for the sake of a tidy call-count.

matclayton:
Bundle cost is down to roughly nothing, pushed in `390fe99`.

Attribution first, since it turned out to be one hunk rather than the change as a whole. On `react-router.minimal`, against this branch's base:

| | gzip |
| --- | ---: |
| the re-render bail-out on its own | -25 B |
| plus the `_options` stabilisation | +52 B |
| net, what you measured | +27 B |

So the selector change pays for itself, and the whole cost was `useValueStable` and its call sites.

Two changes:

1. `activeOptions` is stabilised with the same helper, so the selector depends on one value rather than four destructured fields. The dependency array then matches what the body reads, so the `exhaustive-deps` disable is gone too.
2. Dropped the reference-equality guard inside `useValueStable`. `deepEqual` already returns early on `a === b` (`router-core/src/utils.ts:371`), so the guard only avoided a function call that returns immediately.

Across the eight React scenarios, gzip goes from +10..+44 to **-24..+11**, and raw bytes are uniformly negative, -37 to -45 on every scenario.

Two things I tried and rejected:

- `replaceEqualDeep` is already in the bundle and is shorter to call, but it allocates a copy on every non-identical call, which is the common case here (inline `params` / `search` literals). That is a per-link-per-render path, so not worth the bytes.
- Keeping the `!==` guard costs up to 7 B gzip per scenario. I only dropped it after confirming the work it avoids is a function call that returns immediately.

944 unit tests, 6 TS versions and eslint all still clean, and the `rewrites` scenario is still ahead of base (+2.5%, four interleaved rounds).

packages/react-router/src/link.tsx
```diff
@@ -563,13 +603,13 @@ export function useLinkProps<
...
-    router
-      .preloadRoute({ ..._options, _builtLocation: next } as any)
-      .catch((err) => {
-        console.warn(err)
-        console.warn(preloadWarning)
-      })
-  }, [router, _options, next])
+    // `preloadRoute` builds the location itself; it is no longer held in render
+    // state. Matches `handleClick`, which lets `router.navigate` build its own.
+    router.preloadRoute({ ..._options } as any).catch((err) => {
```
Sheraff: I think this object isn't mutated by `preloadRoute` so we probably don't need the shallow clone
```suggestion
    router.preloadRoute(_options as any).catch((err) => {
```

matclayton: Confirmed, applied in `0dbecc4`.

Worth saying why I checked rather than just taking it: `_options` is now the stabilised props object, shared across renders and held in a ref, so a write in there would corrupt the caller's props and poison the `useValueStable` baseline.

Traced the path, and nothing writes to it:

- `preloadRoute` reads `opts._builtLocation`, then hands the object to `buildLocation(opts)` and `matchRoutes(next, { dest: opts })`.
- `build()` only reads `dest` fields. The `Object.assign` calls in there target locally created `Object.create(null)` objects.
- `applySearchMiddleware` and `buildMiddlewareChain` only read `dest.search`.
- Search middlewares never receive `dest`. The signature is `({ search, next })`, and `functionalUpdate(dest.search, currentSearch)` calls the user function with `currentSearch`.

The only `opts` mutations in router-core are `redirect.ts:124` and `133`, and those act on a redirect's own options, which arrive as `{ ...err.options }`.

The stronger argument is that the spread was never defensive in the first place: it existed to add `_builtLocation`, so removing that left it vestigial. That holds regardless of whether the core internals stay read-only.

gzip is unchanged, raw is -5 B and brotli -35 B. The actual gain is one fewer object allocation per hover.

`handleClick` keeps its spread, since it augments with `replace` / `resetScroll` / `viewTransition`.

matclayton:
We’ve been running this patch in various forms for a day, and so far nothing has shown up in bugs except the rerender wins

Sheraff:
> We’ve been running this patch in various forms for a day, and so far nothing has shown up in bugs except the rerender wins

can you push something (empty commit) to the branch to revive the CI after github's outage? The benchmark action doesn't even appear here anymore

PR #7990, opened 2026-08-06
perf(react-router): stop persisted matches re-rendering on every navigation

### LINKED ISSUE #7989: Every persisted route match re-renders on every navigation
# Every persisted route match re-renders on every navigation

## Summary

In `@tanstack/react-router`, navigating from one route to a sibling re-renders
**every mounted `Match` and `MatchInner`** — including the matches that stayed,
whose rendered output is byte-for-byte identical before and after. Each of those
re-renders walks the whole `MatchView` → `Suspense` → `CatchBoundary` →
`CatchNotFound` → `MatchInner` chain.

On a deep layout tree (app shell → section → list → detail) that is 3–4 wasted
component subtree renders per navigation, on the hot path of every link click.
Measured in a production app with a deep layout tree; reproduced here in a
minimal test.

There are two independent causes. Both are present on `main`; one of them is
present on the released 1.170.x line as well.

## Cause 1 — `loadedAt` is observed by every match (released 1.170.x line)

`Match` subscribes to `router.stores.loadedAt` for one reason: to feed
`CatchBoundary`'s `getResetKey`, so a route error clears on the next navigation.

```tsx
const resetKey = useStore(router.stores.loadedAt, (loadedAt) => loadedAt)
...
<MatchView router={router} matchId={matchId} resetKey={resetKey} matchState={matchState} />
```

`router-core` writes `loadedAt` on every navigation, so this makes **every
mounted match** re-render on every navigation — even routes that have no
`errorComponent` at all and can therefore never use the reset key.

On `main` this specific subscription is gone (the reset key is the match
identity now), so cause 1 does not apply there.

## Cause 2 — `MatchInner` / `Match` select the whole match by identity (both lines)

`matchRoutesInternal` / `buildMatches` re-mints every *staying* match on every
navigation:

```ts
match = {
  ...existingMatch,
  cause,
  search: ...,
  _strictSearch: strictMatchSearch,   // fresh object
  searchError,
}
```

and the load pass then bumps `fetchCount` and swaps `abortController` (1.170.x)
or flips `cause` `'enter' → 'stay'` (main). So the match store *always*
republishes with a new object identity, and the subscription

```tsx
const match = useStore(matchStore, (value) => value)   // no equality fn
```

always reports a change — even though none of the changed fields is rendered.

The counting probe reports exactly which keys changed for a staying match on a
sibling navigation:

```
"changedKeys": [
  "_strictSearch(identity-only)",
  "context(identity-only)",
  "abortController(identity-only)",
  "fetchCount(value)"          // 1.170.17
]
"changedKeys": [
  "_strictSearch(identity-only)",
  "context(identity-only)",
  "abortController(identity-only)",
  "cause(value)"               // main
]
```

`identity-only` means the value is structurally equal to the previous one and
only the object identity changed. Nothing in that list reaches the DOM through
`Match`/`MatchInner`.

On main the same identity subscription lives in `Match` itself
(`useStore(matchStore, (value) => value)`), and it feeds `MatchView` and
`MatchInner` the whole match object as a prop, so `React.memo` on `MatchInner`
can never bail out.

## Minimal reproduction

A four-level tree — root (which re-renders per navigation, as a real app shell
does) → `/section` (with `validateSearch` + loader) → `/section/list` (loader,
inline `<Link search={{...}}>` object literals) → leaf `a`/`b`. A `basepath` is
configured so path rewriting is exercised. `React.memo` is patched to count
renders of each `*Impl` component, and every match store is subscribed to for
publish counting.

The probe added in the accompanying PR
(`packages/react-router/tests/match-rerender-probe.test.tsx`) is exactly that,
and it prints `PROBE_SIBLING_NAV` / `PROBE_SEARCH_NAV` / `PROBE_TWO_NAVS`
reports.

## Measured counts (before the fix)

Four matches are mounted (`__root__`, `/section`, `/section/list`, leaf). Three
of them stay across a sibling navigation.

| Scenario | branch | `MatchImpl` renders | `MatchInnerImpl` renders | Should be |
| --- | --- | --- | --- | --- |
| sibling `/section/list/a → /b` | 1.170.17 | 4 | 4 | 1, 1 |
| sibling `/section/list/a → /b` | main | 4 | 4 | 1, 1 |
| search-only `?tab=all → ?tab=mine` | 1.170.17 | 4 | 4 | 0, 0 |
| search-only `?tab=all → ?tab=mine` | main | 4 | 4 | 0, 0 |
| two navigations back and forth | 1.170.17 | 8 | 8 | 2, 2 |
| two navigations back and forth | main | 8 | 8 | 2, 2 |

Supporting counts on the sibling navigation (identical before and after any
render-side fix, since `router-core` is untouched):

- staying match store publishes: **3**
- of those, publishes whose *whole match* is structurally equal to the previous
  one: **0** (`fetchCount`/`cause` really did change — the object is not deeply
  equal, it is just that nothing rendered depends on it)
- `router.stores.loadedAt` fires: **1** per navigation (1.170.x)

The search-only case is the clearest: no match enters or leaves, `search`
reaches route components through `useSearch`, and yet all four `Match` **and**
all four `MatchInner` re-render.

## Why it is not just `React.memo`'s job

`Match` and `MatchInner` are already `React.memo`'d. The re-render is not driven
by a parent prop change, it is driven by their own store subscriptions
reporting a change. `memo` cannot help; the subscription has to select less.

## Note for anyone measuring this

`deepEqual` in this repo defaults to `ignoreUndefined: true`, so it will call
two matches equal when one has an extra `undefined`-valued key. The probe
deliberately uses its own strict structural comparison for reporting, so
"identity-only" means identity-only and nothing else.

matclayton:
# perf(react-router): stop persisted matches re-rendering on every navigation

Fixes the churn described in the accompanying issue (#7989): every mounted
`Match` and `MatchInner` re-rendered on every navigation, including matches that
stayed and rendered identical output.

Closes #7989.

Branch: `match-rerender-main` (3 commits, off `abf9b81b1f`).

## What changed

`packages/react-router/src/Match.tsx` only.

**1. `Match` selects the fields its subtree renders, instead of the whole match
by identity.**

```diff
-const match = useStore(matchStore, (value) => value)
-return <MatchView router={router} match={match!} />
+const selection = useStore(
+  matchStore,
+  (match) => selectMatchFields(router, routeId, match),
+  matchSelectionEqual,
+)
+return <MatchView router={router} routeId={routeId} selection={selection} />
```

The selection is a labelled tuple of `[matchId, ssr, status, error, remountKey,
lazy]`. `loaderDeps`, `_strictParams` and `_strictSearch` only reach the output
through `remountDeps`, so the remount key is computed inside the selector and
compared as a string — which also retires the `MatchInner` `key` `useMemo`,
which never held anyway (a fresh `_strictSearch` object invalidated it every
navigation). `MatchInner` now receives primitives, so its `React.memo` can
actually bail out.

**2. The per-navigation identity that resets `CatchBoundary` is observed in a
wrapper that only mounts for routes that have an `errorComponent`.**

```tsx
function ResettableCatchBoundary({ routeId, ...props }) {
  const router = useRouter()
  const resetKey = (isServer ?? router.isServer)
    ? router.stores.byRoute.get(routeId)!.get()
    : useStore(router.stores.getMatchStore(routeId), (match) => match)

  return <CatchBoundary getResetKey={() => resetKey} {...props} />
}
```

`getResetKey={() => match}` was the whole reason `Match` needed the match
identity. Moving that read down means only routes that can actually show an
error pay for it, and because `props.children` is the element `MatchView`
already created, the subtree below the wrapper bails out even when the wrapper
itself re-renders.

**3. `route._lazy` is part of the selection.** A lazy route's options are
`Object.assign`ed onto the route object in place, and a re-offered pending match
is the only signal that the components this subtree renders have just been
replaced. Without it,
`tests/issue-4467-lazy-route-pending.test.tsx › a lazy pending component is
offered while the eager loader is still pending` fails (it keeps showing
`Loading default` instead of `Loading lazy page`). `_lazy` is `@internal`, so it
is stripped from the published declarations and is read through a narrow local
structural type.

## Red / green counts

Counted by patching `React.memo` to wrap every `*Impl` component with a counter
(`packages/react-router/tests/match-rerender-probe.test.tsx`, added in the first
commit as reporting-only and asserted in the third). Four matches mounted, three
of them staying across a sibling navigation.

### `main` (this PR)

| Scenario | `MatchImpl` before | after | `MatchInnerImpl` before | after |
| --- | --- | --- | --- | --- |
| sibling `/section/list/a → /b` | 4 | **1** | 4 | **1** |
| search-only `?tab=all → ?tab=mine` | 4 | **0** | 4 | **0** |
| two navigations back and forth | 8 | **2** | 8 | **2** |

`OutletImpl` is unchanged (1 / 0 / 2) — the fix does not suppress renders that
are genuinely required.

Store-side counts, unchanged by the fix because `router-core` is untouched:
3 staying-match publishes per sibling navigation, 0 of which are structurally
equal to the previous value.

### Released `1.170.17` line, for reference

The same shape of fix applies there, and it splits into two independently
measurable hunks. `Match` on 1.170.x also subscribes to `router.stores.loadedAt`
purely for `getResetKey`.

| Scenario | baseline | + `loadedAt` hunk only | + both hunks |
| --- | --- | --- | --- |
| sibling — `MatchImpl` / `MatchInnerImpl` | 4 / 4 | **1** / 4 | **1** / **1** |
| search-only — `MatchImpl` / `MatchInnerImpl` | 4 / 4 | **0** / 4 | **0** / **0** |
| two navs — `MatchImpl` / `MatchInnerImpl` | 8 / 8 | **2** / 8 | **2** / **2** |

## Unit suites

`packages/react-router`, `vitest run` (typecheck enabled), full suite:

| Branch | Baseline | Patched |
| --- | --- | --- |
| `main` | 74 files, 993 passed / 3 failed / 1 skipped (997) | 74 files, **996 passed** / 1 skipped (997) |
| `1.170.17` | 45 files, 914 passed / 3 failed / 1 skipped (918) | 45 files, **917 passed** / 1 skipped (918) |

The 3 baseline failures are the probe's own count assertions, i.e. the red side
of this change. No other test changes state, and `Type Errors  no errors` on
both sides of both branches.

## Error-boundary reset semantics

This is the behaviour the change most risks, since it moves what feeds
`getResetKey`. Two tests cover it, and both pass **before and after** on both
branches (so they are guards, not new behaviour being introduced):

1. *a layout that throws during render resets its boundary on the next
   navigation* — the throwing layout **stays mounted** across the navigation, so
   a resetKey change is the only thing that can clear the boundary.
2. *error thrown by a staying layout after the last commit still resets on the
   next navigation* — the timing edge: the component renders fine, then a later
   `setState` makes it throw **after** the last store/`loadedAt` tick. The next
   navigation must still reset.

The repo's existing `errorComponent.test.tsx`,
`disableGlobalCatchBoundary.test.tsx` and `not-found.test.tsx` also pass
unchanged.

### One sizing note

An app that sets a router-level `defaultErrorComponent` resolves an
`errorComponent` for *every* route, so `ResettableCatchBoundary` mounts once per
match rather than only for routes that declare their own error component. That is
bounded: the residual cost is the wrapper + `CatchBoundary` pair, and the subtree
below still bails on element identity (`props.children` is the element
`MatchView` already created), so the per-match render savings this PR delivers are
unaffected. If you want the residual per-match subscription gone even in the
default-error-component case, the direction would be `CatchBoundary` reading its
reset signal only while it is actually errored — that is deliberately out of scope
here and would be a follow-up rather than something to fold into this change.

## Benchmarks

Strictly **interleaved A/B**, never A×n then B×n — sequential batching on this
box invents 10–15% swings. Each round: swap the built `@tanstack/react-router`
`dist` for the other side, rebuild the benchmark app, run the bench. Medians
across rounds; `vitest bench` `time: 10_000`, `warmupIterations: 100`.

### `benchmarks/client-nav` (react), 8 interleaved rounds per side

| Metric | baseline median | patched median | Δ | baseline range | patched range | paired wins |
| --- | --- | --- | --- | --- | --- | --- |
| throughput (hz, higher better) | 57.26 | 58.94 | **+2.9%** | 53.73–59.54 | 57.50–60.20 | 6/8 |
| mean (ms, lower better) | 17.47 | 16.97 | **−2.9%** | 16.80–18.61 | 16.61–17.39 | 6/8 |
| min (ms, lower better) | 14.17 | 14.06 | −0.8% | 13.98–14.57 | 13.95–14.18 | 5/8 |

**Honest reading:** the median moves the right way and the patched side is
visibly tighter, but the harness cannot cleanly resolve an effect this size. The
round-to-round spread of the *baseline against itself* is ~10%, wider than the
~3% median delta, and 2 of 8 paired rounds went the other way (sign test on 6/8
is not significant). Machine: 4 vCPU, load average 0.7–0.9 from unrelated
background work, per-run `rme` 0.8–1.2%. Treat CodSpeed on CI as the deciding
measurement; this is not a claim of a 3% win.

### `benchmarks/ssr` (react), 4 interleaved rounds per side

| Metric | baseline median | patched median | Δ |
| --- | --- | --- | --- |
| throughput (hz) | 43.20 | 43.44 | +0.6% |
| mean (ms) | 23.15 | 23.02 | −0.6% |
| min (ms) | 20.02 | 19.93 | −0.5% |

**No measurable change**, which is expected: the fix targets re-render churn
across *client* navigations, and SSR renders each match once. Reported to show
there is no regression, not a win. (`benchmarks/memory` was not run.)

Never traded runtime for bytes: nothing in this change adds work to a render
path in order to shrink output.

## Bundle size

Bundle size is a review gate here, so: gzip first (`gzip -9`), then raw, then
brotli (quality 11), per emitted file. Measured on `packages/react-router/dist`
from `vite build`. The build of the unmodified tag is **byte-identical to the
published 1.170.17 tarball across all 299 dist files**, so these deltas are
attributable to the source change and nothing else.

Attribution is per hunk, on the `1.170.17` line where the change splits cleanly
into the `loadedAt` hunk and the field-selection hunk (the selector-only column
is `full − loadedAt`):

| File | Metric | baseline | `loadedAt` hunk | selector hunk | **both** |
| --- | --- | --- | --- | --- | --- |
| `esm/Match.js` | **gzip** | 2878 | +24 | +71 | **+95** |
| `esm/Match.js` | raw | 12912 | +138 | −141 | **−3** |
| `esm/Match.js` | brotli | 2554 | +25 | +75 | **+100** |
| `cjs/Match.cjs` | **gzip** | 3011 | +25 | +69 | **+94** |
| `cjs/Match.cjs` | raw | 14909 | +210 | −246 | **−36** |
| `cjs/Match.cjs` | brotli | 2662 | +36 | +66 | **+102** |
| whole `dist/esm` JS | **gzip** | 41020 | +24 | +71 | **+95** |
| whole `dist/cjs` JS | **gzip** | 46177 | +25 | +69 | **+94** |

`esm/index.js`, `esm/index.dev.js` and `cjs/index.cjs` are byte-identical in
every variant — the change is contained to the `Match` chunk.

So: **+95 B gzip** on the esm payload (+0.23% of 41.0 kB), and raw output
actually *shrinks*, because retiring the `key` `useMemo` and the two dead
`isServer` branches removes more code than the wrapper adds.

### Tuples over objects

The `MatchInner`/`Match` selection publishes a **labelled tuple**, not an
object. Measured both, same logic:

| File | Metric | object selection | labelled tuple | tuple saves |
| --- | --- | --- | --- | --- |
| `esm/Match.js` | gzip | 3001 | 2973 | **28 B** |
| `esm/Match.js` | raw | 13073 | 12909 | 164 B |
| `cjs/Match.cjs` | gzip | 3130 | 3105 | **25 B** |
| `cjs/Match.cjs` | raw | 15037 | 14873 | 164 B |

### One build-output note worth knowing

The production build **keeps `/** … *\/` JSDoc blocks** in `dist` and drops `//`
line comments. Writing the `ResettableCatchBoundary` rationale as a JSDoc block
cost **+335 B gzip** instead of +95 B — the comment was 70% of the change's
gzip footprint. Internal, non-exported helpers here therefore document
themselves with `//`.

## Dead code removed

- `MatchView`'s `resetKey` prop and its type (1.170.x) — the boundary owns it now.
- `MatchInner`'s `key` `useMemo` and its six-entry dependency array (both
  branches) — the key is computed in the store selector.
- 1.170.x only, two unreachable server branches inside `MatchInner`'s
  **client** path: the `if (isServer ?? router.isServer)` guard around
  `minPendingPromise`, and the whole server `status === 'error'` branch.
  `MatchInner` has its own `if (isServer ?? router.isServer) { … return }`
  block far above that returns before either is reachable. (On `main` the
  server error branch *is* live — `MatchInner` has no separate server block
  there — so it is kept.)

No other parameter or local became unused; `eslint` is clean on the changed
files.

## Solid

**The same pattern does not apply, and no fix is needed.** Not implemented.

`packages/solid-router/src/Match.tsx` has no component-level re-render to
suppress. `currentMatch` is a `createMemo` over the same per-route store, and
every consumer — `Show`, `Switch`, `Dynamic`, the `getResetKey` passed to
`CatchBoundary` — is its own fine-grained reactive computation. A fresh match
identity re-runs only the computations that read it, and only patches DOM where
the derived value actually changed. There is no `MatchView`/`MatchInner`
subtree that gets invalidated wholesale, so there is nothing for a field
selection to save.

Two details worth noting for reviewers:

- Solid's `getResetKey={currentMatch}` is the direct analogue of this PR's
  `ResettableCatchBoundary`, and it already costs nothing extra: it is a signal
  read inside the boundary, not a re-render of every match.
- Solid deliberately depends on the whole match identity in exactly one place,
  and its comment names the reason: *"Lazy route option mutations become
  observable with the next client match publication."* That is precisely the
  dependency the React port had to re-add as the `route._lazy` selection slot —
  independent confirmation that it is a real render input rather than a
  workaround.

## How to review

1. Read `selectMatchFields` and ask whether anything `MatchView`/`MatchInner`
   renders is missing from the tuple. That is the whole correctness argument.
2. Read `ResettableCatchBoundary` and confirm the two error-reset tests in the
   probe file are the ones you would have asked for — especially the second, the
   "error tripped after the last store tick" edge.
3. Everything else is mechanical: props threaded as primitives instead of a
   match object.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary component re-renders during navigation, improving routing performance.
  * Preserved correct behavior for pending, not-found, error, and server-rendered routes.
  * Improved error-boundary reset behavior so it occurs only when appropriate.
  * Maintained correct component remounting during route transitions.

* **Tests**
  * Added coverage for navigation rendering, route updates, repeated navigations, route-store changes, and error-boundary resets.

### DIFF
--- a/packages/react-router/src/Match.tsx
+++ b/packages/react-router/src/Match.tsx
@@ -12,6 +12,7 @@ import { SafeFragment } from './SafeFragment'
 import { renderRouteNotFound } from './renderRouteNotFound'
 import { ScrollRestoration } from './scroll-restoration'
 import { ClientOnly } from './ClientOnly'
+import type { ErrorRouteComponent } from './route'
 import type {
   AnyRoute,
   AnyRouteMatch,
@@ -37,6 +38,72 @@ const outletMatchSelectionEqual = (
   b: OutletMatchSelection,
 ) => a[0] === b[0] && a[1] === b[1]
 
+type MatchSelection = [
+  matchId: string | undefined,
+  ssr: boolean | 'data-only' | undefined,
+  status: AnyRouteMatch['status'] | undefined,
+  error: unknown,
+  remountKey: string | undefined,
+  lazy: LazyRouteState,
+]
+
+// `_lazy` is marked `@internal`, so it is stripped from the published
+// declarations router-core's consumers compile against.
+type LazyRouteState = Promise<void> | true | undefined
+
+const matchSelectionEqual = (a: MatchSelection, b: MatchSelection) =>
+  a[0] === b[0] &&
+  a[1] === b[1] &&
+  a[2] === b[2] &&
+  a[3] === b[3] &&
+  a[4] === b[4] &&
+  a[5] === b[5]
+
+const emptyMatchSelection: MatchSelection = [
+  undefined,
+  undefined,
+  undefined,
+  undefined,
+  undefined,
+  undefined,
+]
+
+// `buildMatches` re-mints every staying match on every navigation (a fresh
+// object with a fresh `_strictSearch`, `context` and `abortController`, and an
+// updated `cause`), so a match store always republishes with a new identity.
+// Selecting only the fields this subtree renders lets a staying route bail out.
+// `loaderDeps`/`_strictParams`/`_strictSearch` reach the output solely through
+// the remount key, so the key is computed here and compared as a string.
+// `route._lazy` is selected too: a lazy route's options are assigned onto the
+// route in place, and a re-offered pending match is the only signal that the
+// components this subtree renders have just been replaced.
+function selectMatchFields(
+  router: ReturnType<typeof useRouter>,
+  routeId: string,
+  match: AnyRouteMatch | undefined,
+): MatchSelection {
+  if (!match) return emptyMatchSelection
+
+  const route = router.routesById[routeId] as AnyRoute
+  const remountFn =
+    route.options.remountDeps ?? router.options.defaultRemountDeps
+  const remountDeps = remountFn?.({
+    routeId,
+    loaderDeps: match.loaderDeps,
+    params: match._strictParams,
+    search: match._strictSearch,
+  })
+
+  return [
+    match.id,
+    match.ssr,
+    match.status,
+    match.error,
+    remountDeps ? JSON.stringify(remountDeps) : undefined,
+    (route as { _lazy?: LazyRouteState })._lazy,
+  ]
+}
+
 export const Match = React.memo(function MatchImpl({
   routeId,
 }: {
@@ -46,23 +113,61 @@ export const Match = React.memo(function MatchImpl({
 
   if (isServer ?? router.isServer) {
     const match = router.stores.byRoute.get(routeId)!.get()!
-    return <MatchView router={router} match={match} />
+    return (
+      <MatchView
+        router={router}
+        routeId={routeId}
+        selection={selectMatchFields(router, routeId, match)}
+      />
+    )
   }
 
   const matchStore = router.stores.getMatchStore(routeId)
   // eslint-disable-next-line react-hooks/rules-of-hooks
-  const match = useStore(matchStore, (value) => value)
-  return <MatchView router={router} match={match!} />
+  const selection = useStore(
+    matchStore,
+    (match) => selectMatchFields(router, routeId, match),
+    matchSelectionEqual,
+  )
+  return <MatchView router={router} routeId={routeId} selection={selection} />
 })
+
+// `CatchBoundary` resets its error whenever `getResetKey()` changes, and the
+// match identity is what changes per navigation. Observing it here rather than
+// in `Match` keeps that per-navigation update off every mounted route: only
+// routes that actually have an `errorComponent` pay for it, and because
+// `props.children` is the element `MatchView` already created, the subtree
+// below bails out.
+function ResettableCatchBoundary({
+  routeId,
+  ...props
+}: {
+  routeId: string
+  children: React.ReactNode
+  errorComponent?: ErrorRouteComponent
+  onCatch?: (error: Error, errorInfo: React.ErrorInfo) => void
+}) {
+  const router = useRouter()
+  const resetKey =
+    (isServer ?? router.isServer)
+      ? router.stores.byRoute.get(routeId)!.get()
+      : // eslint-disable-next-line react-hooks/rules-of-hooks -- condition is static
+        useStore(router.stores.getMatchStore(routeId), (match) => match)
+
+  return <CatchBoundary getResetKey={() => resetKey} {...props} />
+}
 
 function MatchView({
   router,
-  match,
+  routeId,
+  selection,
 }: {
   router: ReturnType<typeof useRouter>
-  match: AnyRouteMatch
+  routeId: string
+  selection: MatchSelection
 }) {
-  const route: AnyRoute = router.routesById[match.routeId]
+  const [matchId, ssr, status, error, remountKey] = selection
+  const route: AnyRoute = router.routesById[routeId]
 
   const pendingElement = renderPending(router, route)
 
@@ -77,7 +182,7 @@ function MatchView({
       router.options.notFoundRoute?.options.component)
     : route.options.notFoundComponent
 
-  const resolvedNoSsr = match.ssr === false || match.ssr === 'data-only'
+  const resolvedNoSsr = ssr === false || ssr === 'data-only'
   const ResolvedSuspenseBoundary =
     (route.options.wrapInSuspense ??
     pendingElement ??
@@ -86,7 +191,7 @@ function MatchView({
       : SafeFragment
 
   const ResolvedCatchBoundary = routeErrorComponent
-    ? CatchBoundary
+    ? ResettableCatchBoundary
     : SafeFragment
 
   const ResolvedNotFoundBoundary = routeNotFoundComponent
@@ -98,28 +203,28 @@ function MatchView({
     : SafeFragment
   return (
     <ShellComponent>
-      <matchContext.Provider value={match.routeId}>
+      <matchContext.Provider value={routeId}>
         <ResolvedSuspenseBoundary fallback={pendingElement}>
           <ResolvedCatchBoundary
-            getResetKey={() => match}
+            routeId={routeId}
             errorComponent={routeErrorComponent as any}
             onCatch={(error, errorInfo) => {
               // Forward not found errors (we don't want to show the error component for these)
               if (isNotFound(error)) {
-                error.routeId ??= match.routeId
+                error.routeId ??= routeId
                 throw error
               }
               if (process.env.NODE_ENV !== 'production') {
-                console.warn(`Warning: Error in route match: ${match.id}`)
+                console.warn(`Warning: Error in route match: ${matchId}`)
               }
               routeOnCatch?.(error, errorInfo)
             }}
           >
             <ResolvedNotFoundBoundary
               fallback={(error) => {
-                error.routeId ??= match.routeId
+                error.routeId ??= routeId
 
-                if (error.routeId !== match.routeId) {
+                if (error.routeId !== routeId) {
                   throw error
                 }
 
@@ -131,10 +236,20 @@ function MatchView({
             >
               {resolvedNoSsr ? (
                 <ClientOnly fallback={pendingElement}>
-                  <MatchInner match={match} />
+                  <MatchInner
+                    routeId={routeId}
+                    status={status}
+                    error={error}
+                    remountKey={remountKey}
+                  />
                 </ClientOnly>
               ) : (
-                <MatchInner match={match} />
+                <MatchInner
+                  routeId={routeId}
+                  status={status}
+                  error={error}
+                  remountKey={remountKey}
+                />
               )}
             </ResolvedNotFoundBoundary>
           </ResolvedCatchBoundary>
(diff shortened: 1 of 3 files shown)


packages/react-router/src/Match.tsx
```diff
@@ -37,6 +38,72 @@ const outletMatchSelectionEqual = (
...
+
+// `buildMatches` re-mints every staying match on every navigation (a fresh
+// object with a fresh `_strictSearch`, `context` and `abortController`, and an
+// updated `cause`), so a match store always republishes with a new identity.
+// Selecting only the fields this subtree renders lets a staying route bail out.
+// `loaderDeps`/`_strictParams`/`_strictSearch` reach the output solely through
+// the remount key, so the key is computed here and compared as a string.
+// `route._lazy` is selected too: a lazy route's options are assigned onto the
+// route in place, and a re-offered pending match is the only signal that the
+// components this subtree renders have just been replaced.
```
Sheraff, line 79: wouldn't something else also change at the same time when the components are replaced? like the `status` for example?
This feels quite hacky, and "bolted on top", i'm not super convinced

packages/react-router/src/Match.tsx
```diff
@@ -37,6 +38,72 @@ const outletMatchSelectionEqual = (
...
+    loaderDeps: match.loaderDeps,
+    params: match._strictParams,
+    search: match._strictSearch,
+  })
+
+  return [
+    match.id,
+    match.ssr,
+    match.status,
+    match.error,
```
Sheraff, line 101: why not just return the `match`? since we're already doing a custom equality function, we can restrict the equality check to those specific keys without having to then pass them all as individual props to the MatchInner

packages/react-router/src/Match.tsx
```diff
@@ -37,6 +38,72 @@ const outletMatchSelectionEqual = (
...
+  a[5] === b[5]
+
+const emptyMatchSelection: MatchSelection = [
+  undefined,
+  undefined,
+  undefined,
+  undefined,
+  undefined,
+  undefined,
+]
```
Sheraff, line 69: wouldn't just `[]` do the same thing?

Sheraff (commented):
If i understand correctly, we're doing all of this so that `<Comp />` does not re-render? Or just so that `MatchView` and `MatchInner` don't re-render?

Because I think that 
- `Comp` does not re-render (it is memoized on very narrow and stable dependencies)
- `MatchView` and `MatchInner` are probably pretty cheap to re-render, aren't they?
  - `MatchInner` is 2 `useMemo` and 3 `if`
  - `MatchView` seems fairly trivial as well, no logic, just some JSX

Do we get noticeable performance benefits from this change?

matclayton:
Thanks for looking at it, I'm also not convinced right now, unlike the Link PR. I'm still profiling it internally and seeing if its a meaningful speed up. We have some slowness in this area, but its not clear to me this is actually solving the main issue, it helps, but it might be negligible. I'll look in over it the next day or so, I thought I'd marked it as draft for now, my mistake!

PR #8015, opened 2026-08-08
fix(react-router): useMatchRoute w/ React Compiler

### LINKED ISSUE #4499: useMatchRoute not updating when using React Compiler

Sheraff:
fixes https://github.com/TanStack/router/issues/4499

## Bug Fixes

* Improved `useMatchRoute` behavior during navigation and server-side rendering.
* Preserved callback stability across unchanged renders and batched state updates.
* Ensured route matching responds correctly to location and router status changes.

## Tests

* Added React Compiler end-to-end coverage for route matching.
* Verified navigation between routes, nested routing, and matched-route display.
* Added regression coverage for callback identity and state updates.

### DIFF
--- a/docs/router/api/router/useMatchRouteHook.md
+++ b/docs/router/api/router/useMatchRouteHook.md
@@ -3,7 +3,9 @@ id: useMatchRouteHook
 title: useMatchRoute hook
 ---
 
-The `useMatchRoute` hook is a hook that returns a `matchRoute` function that can be used to match a route against either the current or pending location.
+The `useMatchRoute` hook returns a `matchRoute` function that can be used to match a route against either the current or pending location. The hook subscribes the component to changes in the router state used for matching, making it useful when the match result affects what the component renders.
+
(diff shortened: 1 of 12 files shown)


kamalbennani:
Thank you @Sheraff for taking over my PR :bow: appreciate it

Sheraff:
> Thank you @Sheraff for taking over my PR 🙇 appreciate it

I tried to credit you as a co-author of the commit, but it looks like your aircall github account (`kamalbennani-aircall`) was not the correct one to use, sorry

kamalbennani:
That's fine; what matters is that the fix has landed!

PR #8026, opened 2026-08-10
fix(router-core): stream post-</body> render output instead of quarantining it

ryansolid:
# fix(router-core): stream post-`</body>` render output instead of quarantining it (out-of-order SSR renderers)

## Problem

`transformStreamWithRouter` assumes the renderer holds `<body>` open until the render
stream completes — true for React's `renderToPipeableStream`, where `</body></html>`
are the final bytes. Solid's `renderToStream` has the opposite shape (unchanged since
Solid 1.x): it flushes a **complete document** (`…</body></html>`) as the shell, then
keeps streaming out-of-order boundary fragments (`<template>` + `$df()` swap scripts)
*after* `</body>`. Browsers reparent trailing content into `<body>`; that is the
protocol.

The transform treats everything from the first `</body>` onward as "tail"
(`MergeState.HoldingTail` → `appendTail`) and holds it until render **and**
serialization finish, so its `$_TSR` script injections land before `</body>`. Under
Solid's shape that quarantines every boundary fragment until the end of the stream.
Two measured consequences (TanStack Start, Solid 2.0 line, production build; fixture:
three queries under separate boundaries settling at 25 ms / ~717 ms / ~1500 ms):

**1. No progressive paint — every boundary is gated on the slowest query on the page.**

Server byte stream, before the fix (real-Chrome UA):

| t (ms) | chunk | content |
|---|---|---|
| 46.5 | 0 | Shell with all three fallbacks, `$_TSR.router=` blob |
| 62.8 | 1 | `shell` query resolution script (data on the wire at ~63 ms) |
| 739.4 | 2 | `renderOnly` query script |
| 1536.8 | 3 | `slow` query script |
| 1538.0 | 5 | **Held tail: `</body></html>` + all three boundary `<template>`s + swap scripts** |

Client side, the `shell` boundary's data was in the query cache at ~92 ms, but its
markup arrived at ~1538 ms — a ~1.45 s window where the fallback stayed visible with
the data already hydrated. Each boundary's content is delayed to the end of the
*entire* stream.

**2. Hard failure above 64 KB.** The tail buffer is capped (`MAX_TAIL_CHARS`,
64 KB). A streamed boundary rendering ~100 KB of HTML throws
`SSR stream tail exceeded maximum buffer` server-side, the connection dies
mid-response (`net::ERR_INCOMPLETE_CHUNKED_ENCODING`), and the user is left with a
permanently stuck loading fallback — the corresponding query hydrated as a pending
promise that never resolves, so the client cannot recover. Any content-heavy streamed
route hits this.

## What the transform must guarantee (and why it buffered)

The tail-holding exists for exactly one reason: injected router HTML (the `$_TSR`
state blob and streamed seroval script chunks) must land **inside `<body>`**, i.e.
before the document terminator. With React's shape that is nearly free — the held
tail is just the final `</body></html>`. The bug is that the hold is keyed on *first
sight of `</body>`* rather than on *the terminator itself*, so a renderer that keeps
producing after the terminator gets its entire remaining output quarantined.

The transform's other invariant — injected HTML only enters the stream at safe
closing-tag boundaries, never splitting app bytes mid-tag — is independent of the
tail hold and is preserved (see below).

## Fix

Hold **only the document terminator**, and keep streaming everything after it:

- On seeing `</body>`, the transform captures `</body>[whitespace]</html>`
  (case-insensitive, incremental across chunk splits, whitespace bounded at 256
  chars) into `pendingTail` and emits it last, exactly as before.
- Everything after the captured terminator re-enters the normal scan-and-inject
  pipeline (new `MergeState.PostBody`): fragments flush immediately, and router
  script injections interleave with them. Scripts landing after `</body>` on the
  wire is precisely Solid's own protocol — the parser reparents them into `<body>`,
  and the relocated terminator means the *final document* is better-formed than the
  renderer's raw output, not worse.
- If the bytes after `</body>` don't look like `[whitespace]</html>` (unknown
  renderer shape), the transform degrades gracefully: it holds only `</body>` and
  streams the rest. No path buffers unbounded app output anymore.
- Injection safety in pass-through: injected HTML is written immediately only when
  the scanner is at a safe boundary (no partial trailing tag held in `leftover`);
  otherwise it queues and flushes at the next closing-tag boundary, same as before
  `</body>`. This matters because the bounded-`leftover` overflow path can flush
  output that ends mid-element (e.g. inside `<script>` text), where an immediate
  injection would corrupt the stream.

### Ordering guarantees, before vs. after

- `$_TSR` scripts still always precede the emitted terminator (held tail flushed
  last at `tryFinish`, after any remaining router HTML). Unchanged.
- Data-script-before-dependent-markup for streamed queries: previously an accident
  of the quarantine (scripts flushed immediately while all fragments were held).
  Now it holds by arrival order — the seroval chunk for a settled query is injected
  at the settle event, at or before the moment the renderer flushes that boundary's
  fragment, and the e2e capture confirms each data script precedes its fragment on
  the wire. Note the client is also order-tolerant here: fragments swap fallback
  DOM without needing the data, and hydrated pending promises resolve whenever the
  script executes.

### The 64 KB cap

`MAX_TAIL_CHARS` and the `SSR stream tail exceeded maximum buffer` failure are
removed, not raised: the held tail is now structurally bounded by the terminator
matcher (≤ `</body>` + 256 chars whitespace + `</html>`). Remaining buffers, audited:

| buffer | bound | on overflow |
|---|---|---|
| `pendingTail` | ~270 chars by construction | n/a (matcher gives up and streams through) |
| `leftover` (split-tag window) | 2 KB rolling | older bytes flushed (pre-existing) |
| `pendingRouterHtml` | 16 MB | hard error (pre-existing; router's own scripts, not app output) |
| `pendingWrites` (downstream backpressure) | 16 MB | hard error (pre-existing; consumer stalled) |

App render output can no longer terminate the response.

## Cross-framework safety

- **React shape is byte-identical.** When `</body></html>` are the final render
  bytes, nothing follows the terminator: the capture completes, pass-through never
  carries app bytes, late serialization scripts still flush before the held
  terminator. A new test asserts exact output equality for a React-shaped sequence,
  and it passes unchanged against the pre-fix transform (verified by stashing the
  fix and re-running: the 4 Solid-shape tests fail on the old code, the React
  byte-identity test passes).
- Suites: `@tanstack/router-core` unit suite fully green (105 files, 1575 passed,
  3 pre-existing expected-fails); `@tanstack/react-router` unit suite fully green;
  `@tanstack/solid-router` unit suite fully green.
- One existing test rewritten deliberately: `tail overflow errors and runs cleanup`
  asserted the old kill-the-connection behavior for >64 KB post-`</body>` content.
  It is now `large post-</body> content streams through instead of erroring`, which
  asserts the same input survives, with the rationale in a comment.

## Tests added (`packages/router-core/tests/transformStreamWithRouter.test.ts`)

New describe block `out-of-order (post-</body>) streaming`, with the measured
timeline as the repro description:

- Solid-shaped fixture: complete document, then fragments — asserts each fragment is
  released **before the stream ends** (mid-stream assertion via a continuous
  collector), that an injected data script between fragments flushes immediately and
  precedes the next fragment, and that the document still ends with exactly one
  `</body></html>`.
- A >64 KB fragment (the measured hard-failure case) streams through and the
  connection survives to a clean close.
- Terminator split across chunks (`</body>` … `</ht` … `ml>`) and uppercase
  `</BODY>\n</HTML>` variants — captured and emitted last, byte-exact output asserted.
- React-shaped fixture asserting byte-identical output (see above).
- Injection while a partial fragment tag is pending queues until the next safe
  boundary (the stream-corruption guard).

## End-to-end verification (instrumented Start app, Solid 2.0 line)

Same fixture and probes as the original investigation; the fix was applied to the
installed `router-core` dist as a minimal surgical patch (identical logic) so the
measured delta is exactly this change. Baseline re-measured on the same machine
immediately before.

**Server byte stream (fragment arrival):**

| boundary (query settle) | before | after |
|---|---|---|
| `shell` (25 ms) | 1588 ms (held tail) | **83 ms** |
| `renderOnly` (~717 ms) | 1588 ms (held tail) | **768 ms** |
| `slow` (~1500 ms) | 1588 ms (held tail) | **1561 ms** |
| `</body></html>` | start of held tail | final 14-byte chunk |

Each data script (chunks 1, 5, 9) precedes its boundary's fragment on the wire.

**Client timeline (headless Chrome, real-Chrome UA):**

| event | before | after |
|---|---|---|
| `shell` data in cache | ~80 ms | ~70 ms |
| `shell` boundary content rendered | 1511.8 ms | **77.5 ms** |
| `renderOnly` boundary content rendered | 1514.3 ms | **710.6 ms** |
| `slow` boundary content rendered | 1515.1 ms | 1508.8 ms |
| first counter click handled | 460 ms | 451 ms |

The ~1.45 s data-in-cache→content-visible window for the fast boundary is gone; each
boundary now paints at its own data-arrival time instead of the slowest query's.
Client `queryFn` executions remain **0** (all DOM tokens are the SSR tokens), console
clean apart from a favicon 404.

**64 KB case (`/big`, ~100 KB boundary):** before — server threw
`SSR stream tail exceeded maximum buffer`, connection died at ~98 KB read
(`ERR_INCOMPLETE_CHUNKED_ENCODING`), fallback stuck forever. After — response
completes cleanly: 394,564 bytes, all 1505 `<li>` rows, terminates with
`</body></html>`, zero server errors.

## Notes for reviewers

- `findHtmlBoundary`'s `</body>`-detection is unchanged; in `PostBody` a literal
  `</body>` in content is treated as an ordinary closing-tag boundary rather than a
  second terminator.
- The fast path (`reserveStreamFastPath`) is untouched.
- A widening cast was needed in the read loop (`state as MergeState`) because the
  state transitions moved into helper closures and TypeScript's control-flow
  narrowing otherwise pins `state` to its initial value.

## Summary by CodeRabbit

* **New Features**
  * Improved server-rendered page streaming for faster, more reliable delivery.
  * Supports large and split document terminators, including varied letter casing.
  * Preserves correct ordering for dynamically rendered page fragments.
  * Handles content that appears after the document body closes.

* **Bug Fixes**
  * Prevented streaming interruptions when output exceeds previous buffering limits.
  * Improved compatibility with React-generated output.
  * Prevented malformed or incomplete page endings from disrupting streamed content.

### DIFF
--- a/packages/router-core/src/ssr/transformStreamWithRouter.ts
+++ b/packages/router-core/src/ssr/transformStreamWithRouter.ts
@@ -45,18 +45,29 @@ const MIN_CLOSING_TAG_LENGTH = 4
 const DEFAULT_SERIALIZATION_TIMEOUT_MS = 60000
 const DEFAULT_LIFETIME_TIMEOUT_MS = DEFAULT_SERIALIZATION_TIMEOUT_MS * 2
 const MAX_LEFTOVER_CHARS = 2048
-const MAX_TAIL_CHARS = 64 * 1024
 const MAX_ROUTER_HTML_CHARS = 16 * 1024 * 1024
 const MAX_PENDING_WRITE_CHARS = 16 * 1024 * 1024
 
-// Merge lifecycle: body bytes can stream, router HTML must precede tail,
-// terminal states own close/error/cleanup exactly once.
+const BODY_CLOSE_LENGTH = '</body>'.length
+const HTML_CLOSE = '</html>'
+// Whitespace tolerated between </body> and </html> while capturing the
+// document terminator. Past this the shape is treated as unknown and bytes
+// stream through instead of buffering.
+const MAX_TERMINATOR_WHITESPACE = 256
+
+// Merge lifecycle: body bytes can stream, router HTML must precede the held
+// document terminator, terminal states own close/error/cleanup exactly once.
 const MergeState = {
   ReadingBody: 0,
+  // Saw </body>; capturing the document terminator (</body>[ws]</html>).
   HoldingTail: 1,
-  AppDone: 2,
-  Draining: 3,
-  Done: 4,
+  // Terminator captured (or shape unknown): app bytes stream through again
+  // (out-of-order fragments from renderers like Solid that emit a complete
+  // document up front), and the held terminator is emitted last.
+  PostBody: 2,
+  AppDone: 3,
+  Draining: 4,
+  Done: 5,
 } as const
 
 type MergeState = (typeof MergeState)[keyof typeof MergeState]
@@ -124,6 +135,40 @@ function findHtmlBoundary(str: string): number {
   }
 
   return lastClosingTagEnd
+}
+
+/**
+ * Incrementally match the document terminator `</body>[whitespace]</html>`
+ * (case-insensitive) at the start of `tail`, which is known to begin with
+ * `</body>`. Returns the end index of the terminator when fully matched,
+ * -1 while more bytes could still complete it, and -2 when the shape cannot
+ * match (unknown renderer shape — caller should stop buffering).
+ */
+function matchDocumentTerminator(tail: string): number {
+  let i = BODY_CLOSE_LENGTH
+  let whitespace = 0
+  while (i < tail.length) {
+    const code = tail.charCodeAt(i)
+    if (
+      code === 32 || // space
+      code === 9 || // \t
+      code === 10 || // \n
+      code === 13 || // \r
+      code === 12 // \f
+    ) {
+      if (++whitespace > MAX_TERMINATOR_WHITESPACE) return -2
+      i++
+      continue
+    }
+    break
+  }
+  for (let j = 0; j < HTML_CLOSE.length; j++) {
+    const at = i + j
+    if (at >= tail.length) return -1
+    // `| 32` lowercases ASCII letters and is a no-op for `<`, `/`, `>`.
+    if ((tail.charCodeAt(at) | 32) !== HTML_CLOSE.charCodeAt(j)) return -2
+  }
+  return i + HTML_CLOSE.length
 }
 
 /**
@@ -580,7 +625,8 @@ function makeMainStream(
   // between-chunk text buffer; keep bounded to avoid unbounded memory
   let leftover = ''
 
-  // captured bytes from </body> onward; must stay behind router scripts.
+  // Held document terminator (</body>[ws]</html>, bounded by construction);
+  // emitted last so injected router scripts always land inside <body>.
   let pendingTail = ''
 
   let streamBarrierLifted = false
@@ -639,7 +685,14 @@ function makeMainStream(
       cleanup(err)
       return
     }
-    if (state === MergeState.HoldingTail) {
+    // Past </body> the injected HTML is self-contained script/markup and can
+    // be written immediately — but only when the scanner sits at a safe
+    // boundary (no partial app tag pending in `leftover`), so an injection
+    // can never split an app tag mid-bytes.
+    if (
+      state === MergeState.HoldingTail ||
+      (state === MergeState.PostBody && !leftover)
+    ) {
       flushPendingRouterHtml()
       writeChunk(html)
     } else {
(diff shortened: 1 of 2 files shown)


Sheraff:
FYI the "Bundle Size", "Labeler", and "PR / Preview" are expected to fail on external forks. "PR / Test" isn't. It looked like an `nx` infra issue so I started a 2nd run.

Also I'm still trying to stabilize the codspeed benchmarks. So you can fully ignore the memory benchmarks altogether, and take the CPU benchmarks ("Simulation") with a grain of salt.

TylerRick:
Independently verified this against a deterministic Solid 2 reproduction of the hard-failure mode
(TanStack Start, `@tanstack/solid-start@2.0.0-rc.3` / `solid-js@2.0.0-rc.3`,
`@tanstack/router-core@1.171.22`): one route whose query is deliberately not primed by the loader,
so the boundary is always pending at shell flush and ~460 KB of rows stream after `</body>`.

On stock 1.171.22, every browser-UA request dies mid-stream — ~5.8 KB delivered, connection closed
with data outstanding, `SSR stream tail exceeded maximum buffer` in the server log — while
`isbot`-matched UAs (plain `curl`) always succeed via the all-ready path, which made the failure
look environment-dependent in our app when it was UA-dependent.

With this PR's pkg.pr.new build of `@tanstack/router-core`: 5/5 browser-UA requests complete at
the full ~467 KB matching the bot-UA byte counts, no tail-buffer errors, all `$_TSR` scripts land
inside `<body>`, the deferred boundary content arrives, and the document terminates cleanly with
`</body></html>`.

We currently carry a `pnpm patch` raising `MAX_TAIL_CHARS` to 8 MiB in production to keep Solid
streaming SSR alive; this PR makes that patch unnecessary (and also removes the progressive-paint
gating the cap never addressed). Happy to share the repro if useful.

PR #8038, opened 2026-08-11
fix(router-core): keep useLocation on the rendered route's location

### LINKED ISSUE #8037: useLocation returns pathname of route being navigated to

Router

### Describe the bug

In our application we have a BackButton component that is rendered on multiple routes. The BackButton component uses `useLocation` and other values to determine its own visibility. We observed that the BackButton gets hidden when the application is about to navigate to a new route even though it should stay visible as long as the route it's rendered on is still visible.

https://github.com/ulrichstark/tanstack-repro-useLocation-returns-incorrect-pathname

1. Setup

```sh
git clone https://github.com/ulrichstark/tanstack-repro-useLocation-returns-incorrect-pathname
cd tanstack-repro-useLocation-returns-incorrect-pathname
npm i
npx vite
````

2. Open browser console
3. Click on "Posts"
4. BackButton component logs "/posts" into console
5. Click on "Home"
6. BackButton component logs "/" into console

### Expected behavior

BackButton component should not log "/" into console because it's not rendered on that route.

ulrichstark:
This change is AI generated.

useLocation read router.stores.location, which is written when a navigation starts rather than when its matches are published, so a component on the route being left re-rendered once with the destination's pathname before unmounting.

Add stores.presentedLocation, published atomically with the match lane inside setMatches, and read it from useLocation in react, solid, and vue. router.state.location is unchanged and still reports the requested location while it loads.

Closes #8037

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected `useLocation` behavior across React, Solid, and Vue routers so it reflects the location associated with currently rendered routes.
  - Prevented components from observing a destination pathname before the corresponding route is presented.
  - Preserved access to the requested location through `router.state.location` during loading.

- **Documentation**
  - Clarified when `useLocation` updates and how to observe requested locations during navigation.

- **Tests**
  - Added regression coverage for stale pathname reporting during navigation.

### DIFF
--- a/.changeset/use-location-presented-lane.md
+++ b/.changeset/use-location-presented-lane.md
@@ -1,0 +1,14 @@
+---
+'@tanstack/router-core': patch
+'@tanstack/react-router': patch
+'@tanstack/solid-router': patch
+'@tanstack/vue-router': patch
+---
+
+`useLocation` now returns the location that produced the matches currently
+being rendered instead of the location the router has parsed but not yet
+loaded. A component on the route being left no longer re-renders with the
+destination's pathname before it unmounts. The presented location is
+published on the same lane as the matches it describes, so the two can never
+disagree. `router.state.location` is unchanged and still reports the
+requested location while it loads.
--- a/docs/router/api/router/useLocationHook.md
+++ b/docs/router/api/router/useLocationHook.md
@@ -5,6 +5,8 @@ title: useLocation hook
 
 The `useLocation` method is a hook that returns the current [`location`](./ParsedLocationType.md) object. This hook is useful for when you want to perform some side effect whenever the current location changes.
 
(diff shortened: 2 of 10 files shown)


Sheraff:
i actually kinda like this proposal. But this would be a breaking change, some people actually rely on `useLocation` tracking the URL and not the matches. We might have to delay this for the next major

ulrichstark:
Since the original issue is fixed for me, we can do anything you like with this PR. Do you want me to close it or turn it into a draft PR?

Sheraff:
> Since the original issue is fixed for me, we can do anything you like with this PR. Do you want me to close it or turn it into a draft PR?

I'll keep it on the side for v2

- replace the current `useLocation` with this proposed version (basically a `location` that has the exact same lifecycle as the matches, so it's always "what is rendered")
- and if we really still need it: `usePendingLocation` would be the current implementation, tracking the URL instead of what is rendered
- and we drop `resolvedLocation` unless we find someone who can justify the need for it

PR #8051, opened 2026-08-12
ci: changesets v3 & split release workflow

Sheraff:
split release workflow = better security
https://e18e.dev/docs/publishing.html#standard-workflow

changesets v3 enables it
https://changesets.dev/guide/migration

Closes https://github.com/TanStack/router/pull/7692 (a previous attempts, w/ pre-release changesets packages)

## Summary by CodeRabbit

* **Release Process**
  * Improved automated release handling with clearer steps for versioning, packaging, publishing, and creating release announcements.
  * Added safer, more controlled release execution across supported branches and release channels.
  * Improved build artifact handling and job-level safeguards for more reliable releases.

* **Maintenance**
  * Updated release tooling and configuration to support the latest Changesets format.
  * Streamlined version updates by removing an unnecessary formatting step.


.github/workflows/release.yml
```diff
@@ -2,63 +2,155 @@ name: Release
 
 on:
   push:
-    branches: [main, '*-pre', '*-maint']
+    # we do not support '*-maint' branches at this time
```
schiller-manuel, line 5: what is the `-maint` suffix?

.github/workflows/release.yml
```diff
@@ -2,63 +2,155 @@ name: Release
...
+    branches: [main, '*-pre']
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: false
 
 env:
-  NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
```
schiller-manuel, line 12: don't we need this? why did we have it?

cc @beaussan

package.json
```diff
@@ -44,7 +44,7 @@
     "format": "prettier --experimental-cli --ignore-unknown '**/*' --write",
     "changeset": "changeset",
     "changeset:publish": "changeset publish",
-    "changeset:version": "changeset version && node scripts/update-example-deps.mjs && pnpm install --lockfile-only --ignore-scripts --no-frozen-lockfile && pnpm format",
```
schiller-manuel, line 47: why did we format here? are package.json misformatted maybe afterwards?

.github/workflows/release.yml
Sheraff: it's down there in the `pack` job now (for the `build` step). Limiting the permissions/tokens to only where they are required

package.json
Sheraff: i'm not entirely sure *what* needed to get formatted before, but now `changeset` uses whichever formatter it finds in the repo (prettier for us) so we should be able to remove it

.github/workflows/release.yml
Sheraff: i think it was meant for "maintenance" branches, but i'm not entirely sure

.github/workflows/release.yml
```diff
@@ -2,63 +2,155 @@ name: Release
...
+      contents: read
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
-          fetch-depth: 0
-          persist-credentials: true # changesets/action pushes Release PR commits
+          persist-credentials: false
       - name: Setup Tools
         uses: TanStack/config/.github/setup@e4b48f16568324f76f467aa4c2aac2f05db632c3 # main
```
Sheraff: no need for `TanStack/config/.github/setup` if we're not using node, pnpm, or installed dependencies in this job

.github/workflows/release.yml
```diff
@@ -2,63 +2,155 @@ name: Release
...
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - name: Setup Tools
+        uses: TanStack/config/.github/setup@e4b48f16568324f76f467aa4c2aac2f05db632c3 # main
```
Sheraff: no need for `TanStack/config/.github/setup` if we're not using node, pnpm, or installed dependencies in this job

PR #8055, opened 2026-08-12
fix(react-router): avoid Suspense above root documents

### LINKED ISSUE #8053: Root-route `pendingComponent` wraps `<html>` in an unhydratable Suspense since 1.170.19, discarding the whole SSR document

Router

### Describe the bug

Since **1.170.19**, setting `pendingComponent` on the **root route** wraps the whole document in a `<Suspense>` boundary that React cannot hydrate. React recovers by client-rendering the boundary — which is the entire document — so the SSR HTML is painted and then thrown away on every page load. The page still *looks* correct, which is why this is easy to miss.

`45c4ad8d6` ("fix(router-core): lane match loader rewrite", #7805) removed the root-route guard from `MatchView`:

```diff
   const ResolvedSuspenseBoundary =
-    // If we're on the root route, allow forcefully wrapping in suspense
-    (!route.isRoot || route.options.wrapInSuspense || resolvedNoSsr) &&
     (route.options.wrapInSuspense ??
-      PendingComponent ??
+    pendingElement ??
     ((route.options.errorComponent as any)?.preload || resolvedNoSsr))
       ? React.Suspense
       : SafeFragment
```

The removed conjunct is what kept a *mere* `pendingComponent` from wrapping the root — previously the root only got a boundary if it opted in explicitly via `wrapInSuspense`, or if SSR was off. That distinction looks deliberate given the comment, and dropping it changes behaviour for any app whose root route renders `<html>` from `component` (which is the shape used by `examples/react/start-bare` itself).

**Why React can't hydrate it.** When a Suspense boundary contains the `<html>`/`<head>`/`<body>` singletons, React's Fizz hoists them into the document preamble and writes the boundary's start marker *inside* `<body>`, followed by preamble-contribution comments:

```
<body><!--$--><!--html--><!--head--><!--body--><div>…
```

On the client the Suspense fiber sits *above* `<html>`, so when React reaches it the hydration cursor is still up in the preamble — it lands on a `<script>` in `<head>`. `canHydrateHydrationBoundary` only scans forward over *siblings*, so it can never reach the marker down in `<body>`; it returns `null` and `throwOnHydrationMismatch` fires on the Suspense fiber. React's own diff shows exactly this:

```
<MatchImpl routeId="__root__">
  <MatchView match={{id:"__root__/", …}}>
    <SafeFragment>
      <Suspense fallback={<TSRPendingComponent>}>
-       <script type="module" src="/@id/virtual:tanstack-start-dev-client-entry">
```

React then calls `clearHydrationBoundary`, which walks from the marker to the matching `<!--/$-->` removing everything — including releasing the `<html>`/`<head>`/`<body>` singletons via those preamble-contribution comments — and re-renders it all client-side.

Not React-version-specific: I checked the preamble markers are emitted identically by `react-dom` 19.2.6, 19.2.8 and a 19.3 canary, so this affects React 19.2 stable, not just canaries.

Your own `examples/react/start-bare`, plus one line — no separate project needed. I ran it standalone (its deps are all published versions, so `npm install` works outside the monorepo) on `@tanstack/react-router@1.170.27` / `@tanstack/react-start@1.168.44` / `react-dom@19.2.8` / `vite@8.2.1`:

```diff
 // examples/react/start-bare/src/routes/__root.tsx
 export const Route = createRootRoute({
   head: () => ({
     links: [{ rel: 'stylesheet', href: appCss }],
   }),
   component: RootComponent,
+  pendingComponent: () => null,
 })
```

Happy to publish this as a standalone git repo if you'd rather have a clone-and-run link — it's literally the above, so I didn't want to add a redundant repo to the pile.

1. `cd examples/react/start-bare && pnpm dev` (or `npm install && npx vite dev` standalone).
2. Open the page with the browser console open — **no error**, and `view-source` shows `<body><div>…`.
3. Add `pendingComponent: () => null` to the root route as above.
4. Reload. The console now shows an uncaught **"Hydration failed because the server rendered HTML didn't match the client"**, and `view-source` shows `<body><!--$--><!--html--><!--head--><!--body-->…`.

To see the blast radius rather than just the error, count how much of the document React discards. Paste this before load (DevTools → Sources → snippet, or a Playwright `addInitScript`):

```js
let removed = 0
new MutationObserver(rs => {
  for (const r of rs) if (r.target === document.body) removed += r.removedNodes.length
}).observe(document, { childList: true, subtree: true })
setTimeout(() => console.log('body child nodes removed:', removed), 4000)
```

- Unmodified: **2** (unrelated script cleanup).
- With `pendingComponent`: **13**, in `clearHydrationBoundary` order — `$`, `html`, `head`, `body`, `div`, `$`, `main`, `/$`, `div`, `script`, `/$`. That's the whole document.

### Expected behavior

A `pendingComponent` on the root route shouldn't put a Suspense boundary above `<html>`. Either restore the `!route.isRoot` guard so the root needs an explicit `wrapInSuspense` opt-in, or — if the root genuinely should be wrappable now — keep the boundary below the document singletons so React can hydrate it.

Worth noting the current behaviour has no good workaround from userland other than "don't set `pendingComponent` on the root route", which is what we've done. The guard is still absent in 1.170.27, so upgrading doesn't help.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.170.27 / 1.168.44 (also reproduced on 1.170.23 / 1.168.40; **not** present on 1.170.18)
- OS: macOS 15 (Darwin 24.4.0, arm64)
- Browser: Chromium (HeadlessChrome via Playwright)
- Browser Version: 148.0.7778.96
- Bundler: vite
- Bundler Version: 8.2.1

### Additional context

I bisected across every published release from 1.170.17 to 1.170.27 by checking the compiled `ResolvedSuspenseBoundary` expression in `dist/esm/Match.js`: the guard is present in 1.170.17 and 1.170.18, and absent from 1.170.19 onwards.

For anyone hitting this before it's fixed: the symptom is silent. The page renders correctly and, depending on your setup, the error may only appear as an uncaught `pageerror` rather than a `console.error`, so it can hide behind an `onRecoverableError` handler that doesn't log. The reliable tell is the SSR HTML — if the first bytes after `<body>` are `<!--$--><!--html--><!--head--><!--body-->`, your whole document is being re-rendered on the client.

Sheraff:
Fixes #8053

## Summary

- avoid implicitly wrapping document-owning root routes in Suspense
- keep root Suspense behavior for roots with `shellComponent`, explicit `wrapInSuspense`, and selective SSR (`ssr: false` or `ssr: data-only`)
- retain the existing SSR document while a hydrated document-owning root is pending instead of replacing `<html>` with its pending UI
- preserve normal pending UI behavior for pure client-rendered roots and nested routes

## Regression origin: #7805

Before #7805, `MatchView` deliberately prevented a root route from getting an implicit route-level Suspense boundary merely because it defined a `pendingComponent`:

```tsx
(!route.isRoot || route.options.wrapInSuspense || resolvedNoSsr) &&
  (route.options.wrapInSuspense ??
    PendingComponent ??
    ((route.options.errorComponent as any)?.preload || resolvedNoSsr))
```

For a normal SSR root, `route.isRoot` was true, `wrapInSuspense` was not enabled, and `resolvedNoSsr` was false. The first condition therefore selected `SafeFragment`, leaving a document-owning root shaped like this during hydration:

```tsx
<RootComponent>
  <html>
    <head />
    <body>...</body>
  </html>
</RootComponent>
```

This was an intentional exception, documented by the adjacent comment about only allowing the root to be forcefully wrapped. The separate boundary in `Matches.tsx` was also disabled during SSR and hydration, while the first child route could still suspend safely through the boundary rendered by `Outlet` inside the root document.

The `Match.tsx` rewrite in #7805 removed the root-route condition and reduced the selection to:

```tsx
route.options.wrapInSuspense ??
  pendingElement ??
  ((route.options.errorComponent as any)?.preload || resolvedNoSsr)
```

Changing `PendingComponent` to `pendingElement` was not itself the problem: both are truthy when a pending component exists. The regression came from dropping the `!route.isRoot || ...` gate. After that change, adding even `pendingComponent: () => null` implicitly produced this tree:

```tsx
<Suspense fallback={pendingElement}>
  <RootComponent>
    <html>
      <head />
      <body>...</body>
    </html>
  </RootComponent>
</Suspense>
```

React treats `<html>`, `<head>`, and `<body>` as document singletons and hoists them into the server-rendered document preamble. The Suspense markers are emitted lower in `<body>`, but the client fiber says that Suspense is above `<html>`. React cannot associate that marker with the client boundary, reports a hydration mismatch, clears the boundary, and client-renders the entire SSR document. The page can still look correct afterward, which makes the lost SSR DOM identity easy to miss.

The lane-loader architecture did not inherently require removing this rule; the guard was lost during the broad renderer simplification. The current fix restores the document-ownership invariant while accounting for the newer architecture:

- nested routes can still receive implicit boundaries
- a root with `shellComponent` is safe because the shell owns `<html>/<head>/<body>` and the boundary is placed inside it
- explicit `wrapInSuspense` remains a force opt-in
- selective-SSR roots retain the boundary required for client rendering
- an SSR root that directly owns the document is not implicitly wrapped by `pendingComponent`

The fix also handles the modern lane renderer's pending branch. When an unwrappable hydrated root becomes pending, it keeps rendering its retained SSR content rather than replacing `<html>` with pending UI. Pure CSR roots continue to render their pending component normally. Before #7805, the older loading/presentation flow did not expose this direct pending-root replacement path, so restoring only the old boundary expression would not cover all behavior under the new architecture.

## Reproduction

The hydration issue is now reproduced directly in `e2e/react-start/dev-ssr-styles` using React 19.2.3; upgrading to React 19.2.8 is not required.

The Playwright regression:

- captures an SSR DOM node during HTML parsing, before hydration begins
- installs a `MutationObserver` before application code runs
- waits for client hydration to complete
- asserts that the captured SSR node was neither removed nor replaced
- collects browser console and page errors related to hydration

This uses DOM removal and node identity as the behavioral oracle rather than React internal stream comment markers.

### Main versus this fix

The production test was also run in an isolated worktree based on `origin/main` at `38485038c5`, with only the E2E fixture commit cherry-picked and all workspace artifacts rebuilt from that checkout.

- `main`: fails with `captured: true`, `removed: true`, and `replaced: true`
- this branch: passes with the same test and command; the SSR node remains attached and retains its identity

The destructive hydration recovery reproduces in both Vite development and a built production application, so the issue is not dev-only. In development React also reports the hydration mismatch; in production the DOM identity check directly detects the destructive recovery.

## Coverage

- the browser E2E is the authoritative #8053 regression and runs in the existing dev matrix plus a focused built-production run
- existing React Router unit tests cover unwrapped document-owning roots, retention of root content while pending, and Suspense inside an explicit root shell
- the earlier streaming assertion against `<!--html--><!--head--><!--body-->` was removed because it was an indirect, React-internal proxy and did not perform hydration

## Tests

- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/react-router:test:unit --outputStyle=stream --skipRemoteCache` (1,010 passed, 1 skipped)
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/react-router:test:types --outputStyle=stream --skipRemoteCache` (TypeScript 5.6 through 7.0)
- `CI=1 NX_DAEMON=false pnpm nx run @tanstack/react-router:test:eslint --outputStyle=stream --skipRemoteCache` (0 errors)
- `CI=1 pnpm run test:e2e` in `e2e/react-start/dev-ssr-styles` (six dev configurations plus production)
- `CI=1 pnpm run test:e2e:prod` in `e2e/react-start/dev-ssr-styles`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SSR and hydration behavior for root routes that render document content.
  - Preserved server-rendered content during hydration and navigation instead of replacing it with pending UI prematurely.
  - Ensured pending states appear only when appropriate, including after the configured minimum delay.

- **Tests**
  - Added end-to-end coverage confirming document content remains intact through hydration and loading transitions.


e2e/react-start/dev-ssr-styles/src/routes/__root.tsx
```diff
@@ -13,16 +14,27 @@ export const Route = createRootRoute({
       { name: 'viewport', content: 'width=device-width, initial-scale=1' },
     ],
   }),
+  pendingComponent: () => null,
   component: RootComponent,
 })
 
 function RootComponent() {
+  useEffect(() => {
+    document.documentElement.dataset.hydrated = 'true'
```
schiller-manuel: why are we using this particular e2e test as a regression test? this doesnt have to do anything with dev styles, right?

maybe just add a separate e2e project?

e2e/react-start/dev-ssr-styles/tests/issue-8053-root-document-hydration.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+            (removedNode instanceof Element && removedNode.contains(ssrNode))
+          ) {
+            state.removed = true
+          }
+        }
+      }
+    }).observe(document, { childList: true, subtree: true })
+  })
+
+  const hydrationErrors: Array<string> = []
```
schiller-manuel: do we really need to manually check errors? doesnt our custom test fixture from @tanstack/router-e2e-utils handle that already?

packages/react-router/tests/issue-7986-retained-pending.test.tsx
```diff
@@ -480,7 +480,7 @@ test('a global not-found destination does not retain the mounted root success',
   })
 
   expect(await screen.findByTestId('pending')).toBeVisible()
-  expect(screen.getByTestId('content')).not.toBeVisible()
+  expect(screen.queryByTestId('content')).not.toBeInTheDocument()
```
schiller-manuel: why was this changed? same question for the other similar changes down below

Issue #3455, opened 2025-02-16
`<StrictMode/>` breaks `beforeLoad` context injection on parent matches when using `<Navigate/>`


JortRoelofs:
### Which project does this relate to?

Router

### Describe the bug

For context: I'm using OIDC for client side authentication, having users land at a url `oidc/callback` where a `<Navigate>` component is rendered once the user is authenticated, navigating them to the protected route they previously tried to access. Now obviously, the `useEffect` hook in the component is run twice when running in `StrictMode`, which means that the `navigate` function is called twice. The protected route is located under a `/_protected` virtual route where the auth data is injected into the route contexts in a `beforeLoad` function. Child routes that are protected can then access the data from the context. However, the loading of matches is cancelled due to the second `navigate` call and for some reason this does not properly merge the contexts on the second run, causing the parent context value to be omitted for children. For some reason, the value injected in the parent `beforeLoad` function can be accessed in the direct child `beforeLoad` function, but not for nested children and other ways to access the router context. Considering this does work when `StrictMode` is disabled and the contexts should be merged according to the typing, this seems to be unintended behaviour.

### Your Example Website or App

https://stackblitz.com/edit/tanstack-router-xs8ubhn9?file=src%2Froutes%2F_protected%2Fauthenticated.tsx

### Steps to Reproduce the Bug or Issue

1. Run the example with the router in `StrictMode`.
2. Click on the 'trigger' button, causing a `Navigate` component to be rendered.
3. For the pending parent route, the `beforeLoad` function is called, and the `value` attribute is injected into the context.
4. The second call to `navigate` inside the `Navigate`'s `useEffect` causes the match loading to be cancelled and restarted, seemingly before the `loader` function is called.
5. Both the parent's and child's `beforeLoad` functions are called, but the parent's context is not injected properly into the child's context for the `loader` function.
6. Now, the parent's context attributes are missing from the context.
7. For reference, when commenting out `<React.StrictMode>` and re-running, the parent attribute is injected into the child's context.

### Expected behavior

As prescribed above, I expect a parent's route context to be injected into a child's context, even when `StrictMode` is enabled.

### Platform

- OS: Windows 10
- Browser: Chrome, Firefox
- Version: 1.105.0

### Additional context

In my personal project, the `beforeLoad` function is asynchronous. The example uses a synchronous function for simplicity to show that it's not deterministic, but it should work properly for asynchronous functions as well.

PR #8066, opened 2026-08-14
fix(react-router): stop Navigate re-issuing its navigation on every render

### LINKED ISSUE #8060: `<Navigate>` re-issues its navigation on every render, causing an unbounded loop

Router

### Describe the bug

`<Navigate>` re-issues its navigation on every render.

Its only guard is an identity check on the JSX props object:

```js
const previousPropsRef = React.useRef(null)
useLayoutEffect(() => {
  if (previousPropsRef.current !== props) {
    navigate(props)
    previousPropsRef.current = props
  }
}, [router, props, navigate])
```

React allocates a fresh props object on every render, so `previousPropsRef.current !== props` is always true and the guard has never prevented anything.

That only becomes observable when the component rendering `<Navigate>` re-renders. The reproducer covers the three realistic ways that happens:

| Case | Renders of the component holding `<Navigate>` | Destination `beforeLoad` runs | Loop |
| --- | --- | --- | --- |
| 1. subscribes to router state | 26 | n/a | yes |
| 2. external store emits during a pending navigation | 32 | 25 | yes |
| 3. `search` as an updater function | 26 | n/a | yes |

**Case 1 is self-sustaining and needs nothing else.** Issuing the navigation changes router state, which re-renders a component subscribed to that state, which re-issues the navigation. No external input, and the destination has no `beforeLoad` at all. The cost is unbounded render churn: with the reproducer's stop raised to 100000, it reaches 100001 renders in under half a second.

**Case 2 is the one that hurts in production.** An unrelated subscription re-renders the component while the navigation is pending, and each re-issue supersedes the in-flight navigation, so the destination's `beforeLoad` is restarted over and over and the navigation never settles.

**Case 3** shows that comparing the props by value rather than by identity would not be sufficient either: `search` and `params` accept updater functions, which are usually declared inline and so are a fresh value on every render too.

Worth noting: `packages/solid-router` and `packages/vue-router` both run the navigation in `onMount` / `onMounted`, so they issue it once and never re-issue. React's `Navigate` is the only adapter that re-issues, and the presence of this guard suggests the once-only semantics were intended here too.

I mention that because #6672 is effectively this same defect in solid-router and was closed with "throw a `redirect` in `beforeLoad` instead". I don't think that answer applies here: this isn't a user-authored effect, it's library code carrying an explicit guard against re-issuing that is a no-op.

https://github.com/kamalbennani/router/tree/repro/navigate-loop

1. Clone https://github.com/kamalbennani/router/tree/repro/navigate-loop (branch `repro/navigate-loop`)
2. `npm install`
3. `npm run dev`
4. Open the dev server URL and click **1. redirect component subscribes to router state**

The counters on the page report **26 renders** of the component holding `<Navigate>` and **loop detected: true**. Cases 2 and 3 are linked from the same page.

The reproducer stops rendering `<Navigate>` after 25 renders so the tab stays usable. Raise `MAX_RENDERS` in `src/main.jsx` to see that none of the cases terminate on their own.

### Expected behavior

I expected `<Navigate>` to issue its navigation once, but it re-issues on every render, and when the component re-renders as a consequence of that navigation the result is an unbounded loop.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.170.27 (also verified on 1.131.7, 1.136.17, 1.136.18)
- OS: macOS
- Browser: Chrome
- Browser Version: 149
- Bundler: vite
- Bundler Version: 5.4

### Additional context

**This is not a regression.** I checked, because #5905 moved this effect from `React.useEffect` to `useLayoutEffect` and that looked like a plausible culprit. It isn't:

| Version | Renders (case 1) | Loop |
| --- | --- | --- |
| 1.131.7 | 26 | yes |
| 1.136.17 (last before #5905) | 26 | yes |
| 1.136.18 (first with #5905) | 26 | yes |
| 1.170.27 | 26 | yes |

1.136.17 and 1.136.18 straddle #5905 exactly and behave identically. I also flipped the single line back to `React.useEffect` on current `main`, rebuilt, and the loop was unchanged. The defect predates 1.131.7, and any fix should keep `useLayoutEffect` so #5905's flicker fix stands.

We hit case 2 in production: a redirect component holding an Apollo subscription, pointed at a route with an async `beforeLoad`. The page sat on a permanent loading state and issued 4511 requests before the tab died. Nothing in our 4500-test suite caught it, since it only manifests in a browser against a real async guard.

I have a fix and an e2e fixture ready if you're open to a PR. It guards on the resolved destination rather than the props object, keeping `useLayoutEffect`. Happy to adjust the approach if you'd prefer a different direction, for example making React's `Navigate` fire once on mount to match the Solid and Vue adapters.

kamalbennani:
Fixes #8060.

## Problem

`Navigate` guards its navigation on an identity check against the props object:

```js
const previousPropsRef = React.useRef(null)
useLayoutEffect(() => {
  if (previousPropsRef.current !== props) {
    navigate(props)
    previousPropsRef.current = props
  }
}, [router, props, navigate])
```

React allocates a fresh props object on every render, so this never holds and the navigation is re-issued on every render.

That guard came from #3465 ("only navigate once in StrictMode", fixing #3455). It works for StrictMode because both invocations of the double-invoked effect close over the same props object. Across a genuine re-render it cannot hold, so the general case was never covered.

The consequence is an unbounded loop. Subscribing to router state is sufficient on its own: issuing the navigation changes router state, which re-renders the component, which re-issues the navigation. No external input and no async destination required - with the fixture's stop raised to 100000 it reaches 100001 renders in under half a second.

An async destination makes it worse rather than being a precondition: it stays pending across the re-renders, so each re-issue supersedes the previous one and restarts its `beforeLoad`. That is the shape that produces unbounded requests. We hit it in production - a redirect component holding a data-fetching subscription, 4511 requests before the tab died.

**This is not a regression.** I checked #5905 (which moved this to `useLayoutEffect`) because it looked like the culprit. It isn't - 1.136.17 and 1.136.18 straddle it and behave identically, and flipping the line back to `React.useEffect` on `main` changes nothing. The defect predates 1.131.7, and this PR keeps `useLayoutEffect` so #5905's flicker fix stands.

## Fix

Guard on the resolved destination instead of the props object.

A value comparison of the props is not sufficient: `search` and `params` accept updater functions, which are usually declared inline and so are a fresh value on every render too. Resolving the location collapses those to a concrete href. `Link` already builds the location on every render, so this is a cost the router is used to paying.

**This does not stop `Navigate` from navigating on updates.** A `Navigate` whose destination genuinely changes still re-navigates; only re-issuing the identical resolved destination is suppressed.

The effect closes over the committed render's props rather than reading a ref written during render, so a render React discards cannot influence which options get used.

## Tests

New e2e fixture `e2e/react-router/navigate-component`, covering the three ways the redirect component gets re-rendered:

| Case | Destination | What it shows |
| --- | --- | --- |
| subscribes to router state | sync | self-sustaining, needs nothing else |
| external store emits during a pending navigation | async | re-issues restart the destination guard |
| `search` as an updater function | sync | value-comparing props would not fix this |

All three fail on `main` and pass with this change.

`Navigate` now resolves its destination during render, which also runs on the server where the navigation effect does not, so `packages/react-router/tests/navigate-component.test.tsx` adds a server-render check for the plain and updater-function forms.

Verification: 1012 react-router unit tests pass (1010 before, +2 added), 0 lint errors, no type errors, and the `basic` and `basic-file-based` e2e suites pass.

## Note on the lockfile

The `pnpm-lock.yaml` diff contains peer-resolution churn beyond the new fixture's own entry. That is pre-existing: I confirmed an empty workspace package produces a 2-line diff, but any package with dependencies triggers a full peer re-resolution, because the committed lockfile is stale relative to pnpm 11.9.0. Happy to drop the fixture into an existing e2e app instead if you'd rather avoid it.

## Note on the other adapters

`solid-router` and `vue-router` run this in `onMount` / `onMounted` and issue the navigation exactly once. React's `Navigate` is the only adapter that re-issues. If you'd prefer React match them by firing once on mount, I'm glad to redo it that way - it's a larger behavior change, so I went with the smaller one that makes the existing guard work.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `<Navigate>` repeatedly triggering navigation when components re-render.
  - Navigation now completes once when search parameters are generated dynamically or components run in Strict Mode.

- **Tests**
  - Added coverage for router-state, external-store, server-rendering, and Strict Mode scenarios.
  - Added end-to-end checks to detect navigation loops, verify destination rendering, and confirm expected navigation requests.

### DIFF
--- a/.changeset/olive-jokes-hunt.md
+++ b/.changeset/olive-jokes-hunt.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-router': patch
+---
+
+Stop `Navigate` from re-issuing its navigation on every render. The guard compared the props object by identity, which is fresh on every render, so a component that re-rendered while the destination was still pending superseded its own in-flight navigation and never committed it. The guard now compares the resolved destination.
--- a/e2e/react-router/navigate-component/index.html
+++ b/e2e/react-router/navigate-component/index.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Navigate component test</title>
+  </head>
+  <body>
+    <div id="app"></div>
+    <script type="module" src="/src/main.tsx"></script>
+  </body>
+</html>
--- a/e2e/react-router/navigate-component/package.json
+++ b/e2e/react-router/navigate-component/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "tanstack-router-e2e-navigate-component",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "dev": "vite --port 3000",
+    "dev:e2e": "vite",
+    "build": "vite build && tsc --noEmit",
+    "preview": "vite preview",
+    "start": "vite",
+    "test:e2e": "rm -rf port*.txt; playwright test --project=chromium"
+  },
+  "dependencies": {
+    "@tanstack/react-router": "workspace:^",
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0"
+  },
+  "devDependencies": {
+    "@playwright/test": "^1.61.0",
+    "@tanstack/router-e2e-utils": "workspace:^",
+    "@types/react": "^19.0.8",
+    "@types/react-dom": "^19.0.3",
+    "@vitejs/plugin-react": "^6.0.1",
+    "vite": "^8.0.14"
+  }
+}
--- a/e2e/react-router/navigate-component/playwright.config.ts
+++ b/e2e/react-router/navigate-component/playwright.config.ts
@@ -1,0 +1,25 @@
+import { defineConfig, devices } from '@playwright/test'
+import { getTestServerPort } from '@tanstack/router-e2e-utils'
+import packageJson from './package.json' with { type: 'json' }
+
+const PORT = await getTestServerPort(packageJson.name)
+const baseURL = `http://localhost:${PORT}`
+
+export default defineConfig({
+  testDir: './tests',
+  workers: 1,
+  reporter: [['line']],
+  use: { baseURL },
+  webServer: {
+    command: `VITE_NODE_ENV="test" VITE_SERVER_PORT=${PORT} pnpm dev:e2e --port ${PORT}`,
+    url: baseURL,
+    reuseExistingServer: !process.env.CI,
+    stdout: 'pipe',
+  },
+  projects: [
+    {
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'] },
+    },
+  ],
+})
--- a/e2e/react-router/navigate-component/src/main.tsx
+++ b/e2e/react-router/navigate-component/src/main.tsx
@@ -16,15 +16,20 @@ import {
  * render, so an identity check never holds and the navigation is re-issued on
  * every render.
  *
- * That is only observable when the component rendering `<Navigate>` re-renders
- * while the navigation is still pending. Then each re-render supersedes the
- * in-flight navigation before its `beforeLoad` can settle, the navigation never
- * commits, and the component keeps re-rendering: the app is stuck on a loading
- * state while requests pile up.
+ * That is only observable when the component rendering `<Navigate>` re-renders,
+ * which the three routes below are the realistic ways of causing.
  *
- * The two routes below are the realistic ways a component gets re-rendered
- * during a pending navigation. `MAX_RENDERS` bounds the loop so a regressed
- * build fails the assertions instead of exhausting the browser tab.
+ * Subscribing to router state is enough on its own: issuing the navigation
+ * changes router state, which re-renders the component, which re-issues the
+ * navigation. No external input and no async destination are needed.
+ *
+ * An async destination makes it worse rather than being a precondition. It
+ * stays pending across the re-renders, so each re-issue supersedes the previous
+ * one and restarts its `beforeLoad`, turning the render loop into unbounded
+ * requests.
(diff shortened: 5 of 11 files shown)


MILLERMARRU (commented):
I traced the guard logic in useNavigate.tsx and the bug is real and exactly as described: `previousPropsRef.current !== props` inside a `useLayoutEffect` that lists `props` in its own dependency array can never suppress a re-issue, since React gives JSX props a fresh object identity on every render regardless of whether any prop value changed. Guarding on the resolved `href` from `router.buildLocation` instead is the right fix, and the reasoning for why a plain value-comparison of props wouldn't be enough is solid, `search`/`params` accept inline updater functions, which are also fresh every render. The e2e fixture covering three separate re-render triggers (router-state subscription, external store, updater-function search) is convincing, and I appreciate the explicit note ruling out PR #5905 as the source of the regression by testing both sides of that commit directly.

Worth flagging that #8064 looks like a competing fix for the same underlying issue (#8060), I compared the two and this one is more complete (e2e coverage plus the #5905 check), so I'd lean toward this one over #8064.

kamalbennani:
Hi @Sheraff, I'd love to get your thoughts on the issue & the provided fix
:ty:

Sheraff:
> Hi @Sheraff, I'd love to get your thoughts on the issue & the provided fix :ty:

i agree w/ the issue. Not sure about the fix.
- `buildLocation(...)` is a somewhat expensive operation, it might be nice to avoid it *in render* if possible
- i think `<Navigate />` is meant as a "navigate on mount" and this PR makes it *reactive*

maybe we should just use an empty dependency array? `useLayoutEffect(..., [])`

PR #8081, opened 2026-08-16
feat(solid-router): delegate lazyRouteComponent to Solid 2's lazy()

brenelz:
## Problem

`lazyRouteComponent` predates Solid 2's `lazy()` and hand-rolls what it now
provides: a load state machine over `Dynamic`, with documented hydration
caveats (a conditional `createMemo` in the render path) and none of the
SSR asset story — route chunks rendered on the server got no per-module
client asset resolution through Solid's client-assets manifest, relying
solely on TanStack's route-keyed manifest.

## Mechanism

`lazyRouteComponent` now delegates to solid-js `lazy()` (~60 lines, most
of them comments):

- The importer wrapper selects the configured export and forwards the
  module's `$$moduleUrl` (appended by `@solidjs/vite-plugin`'s SSR
  transform), so server-side `lazy()` resolves the route chunk's client
  assets — stylesheet links, modulepreload hints, and hydration gating —
  through the same manifest seam as any other `lazy()` component.
  Positional hydration remains the fallback. Duplicate emission against
  TanStack's route-keyed manifest is absorbed by the renderer's resource
  dedupe (by identity/URL).
- The module-not-found reload-once dance (stale deploy → one
  sessionStorage-gated refresh) is unchanged, living in the importer
  wrapper's rejection handler.
- `preload()` keeps its contract: a memoized `Promise<void>` that never
  rejects, cleared on failure so a later preload retries, one import per
  server.

## Interim solid-js patch (to be upstreamed)

Solid's `lazy()` caches its module promise permanently — rejections
included — which would seal in a failed chunk download for the page's
lifetime and make TanStack's download-retry contract (error boundary →
retry → re-import) impossible without wrapping `lazy()` in a retry state
machine. This PR carries a pnpm patch for solid-js 2.0.0-rc.0 (same play
as the earlier vite-plugin-solid `isEntry` patch, removed once upstream
shipped) that:

- clears the cached module promise on rejection (client and server), so
  the next preload or render retries the import — matching how browsers
  treat failed dynamic imports;
- fixes an unhandled-rejection leak (`lazy()`'s internal `.then` had no
  rejection handler);
- on the server, stops a transient import failure from poisoning every
  subsequent SSR request (the shared rejected promise previously threw
  for the rest of the process lifetime).

Retry semantics improve with it: a failed hover-preload is retried at
render (previously the recorded error was thrown without re-attempting),
and the error-boundary retry flow is unchanged when the render-time
attempt also fails. `tests/component-preload-retry.test.tsx` pins the new
shape.

Note: until the fix ships in a solid-js release, the retry behavior only
holds inside this repo (the patch applies to the workspace, not to
published consumers) — one more reason to upstream promptly.

## Verification

All on `solid-router-v2-pre` (solid-js `2.0.0-rc.0` + patch,
`@solidjs/vite-plugin` `3.0.0-next.28`):

- `@tanstack/solid-router` unit (client + server modes): pass, including
  the updated `component-preload-retry` suite — except the 4 pre-existing
  branch failures unrelated to this change (`remountDeps` ×3 from the
  Solid 2 `onMount` removal; `issue-7986` ×1), which fail identically
  without these commits.
- Type tests across ts56–ts70: pass. eslint: 0 errors.
- e2e (`basic-file-based-code-splitting`, `solid-start/basic`,
  `start-manifest`): running, results to follow in a comment.


brenelz:
## e2e verification + two hydration fixes (747a786ba0)

The initial e2e sweep surfaced two real defects, both now fixed:

**1. Orphaned SSR DOM on client navigation (20 `basic` e2e failures).** Solid's `lazy()` preloaded-module hydration shortcut (`_$HY.modules[key]` → `() => cached.default`) assumes the registered chunk's default export is the component. For `lazyRouteComponent`'s named-export wrapper the chunk default is `undefined`, so the boundary silently rendered nothing, never claimed the SSR'd DOM, and navigation couldn't dispose it — old route content stacked up under new routes. The interim solid-js patch now gates the shortcut on a bundler-injected `moduleUrl` (the transform's marker that the default-export convention holds); un-transformed wrappers fall through to their own import.

**2. Per-route CSS isolation broken (4 `start-manifest` e2e failures).** Forwarding the route module's `$$moduleUrl` registered the whole *routes chunk* through the renderer's asset channel — and since route chunks bundle many routes, direct SSR of `/r2` emitted the union of every route's CSS. Route chunks now opt out explicitly (`$$moduleUrl: null`, a patch-added semantic that also suppresses the unresolvable-module warning): route assets stay on TanStack's route-keyed manifest, which is per-route precise, and hydration stays synchronous because `preload()` warms `lazy()`'s component cache before the router renders. User-level `lazy()` components inside routes remain fully registered.

Both patch hunks are upstream candidates alongside the rejection-retry fix. The longer-term convergence (making route chunks safely registrable) is per-component default-export split chunks in the router plugin — follow-up territory.

**Results:** solid-start `basic` **80 passed** (was 60/20) · `start-manifest` **10 passed** including the previously failing lazy→static shared-CSS navigation · unit/type/lint green apart from the four pre-existing failures fixed on the #8080 branch.

brenelz:
## Restacked on #8080 — route chunks now opt in to the client-assets manifest (cffa310f65)

Retargeted this PR onto the useHead branch, which changes the asset story for the better:

- **The earlier `$$moduleUrl: null` opt-out is gone.** The 4 stylesheet failures that motivated it were *duplicate links* — Solid's asset channel and TanStack's in-tree manifest links both emitting the same URL — not wrong granularity: split route modules (`?tsr-split=component`) have per-module client-manifest entries with exactly their own CSS. With #8080's head registry underneath, its URL-keyed resource dedupe reconciles the two sources, so `lazyRouteComponent` now forwards the real module id and route chunks get per-route CSS + modulepreload hints through Solid's channel.
- **The solid-js patch shrinks to exactly the two upstream-tracked fixes**: rejection retry (solidjs/solid#2999 — already fixed on `next` in 3c68ab21) and the hydration-shortcut moduleUrl gate (solidjs/solid#3011). When both ship in an rc, the patch deletes cleanly.

**Verification on the stack:** unit 59 files / 865 tests fully green (first configuration where everything passes, including the CatchBoundary fixes from #8080); solid-start `basic` e2e 80 passed; `start-manifest` 12 passed including the modulepreload-hint and hydration asset-map tests.

brenelz:
## Patch retired — now on vanilla solid-js 2.0.0-rc.1

rc.1 (released today) ships both halves of what the pnpm patch carried: the `lazy()` rejection-retry fix (solidjs/solid#2999) and the `{ export }` option with its dev-loud hydration guard (solidjs/solid#3011). The updated bundler transform arrives via `babel-preset-solid` 2.0.0-rc.1 through `@solidjs/vite-plugin` 3.0.0-next.30.

This PR now contains **no patches** — `patches/` and `patchedDependencies` are gone, solid pins are bumped monorepo-wide to the rc.1 line, and `lazyRouteComponent` needed zero changes (the backport had already put it in its final shape against the real API).

**Verification on vanilla rc.1:** unit 59 files / 865 tests green · solid-start `basic` e2e 80 passed · `start-manifest` 12 passed.

birkskyum (approved):

PR #8085, opened 2026-08-16
perf(router-core): encode and decode search params without URLSearchParams

anonrig:
## Summary

Split out of the combined hot-path PR so bundle size and performance can be measured independently.

`encode` / `decode` no longer construct a `URLSearchParams` on every call. They still match `application/x-www-form-urlencoded` behavior, including reserved characters, spaces, lone surrogates, and malformed percent escapes. `undefined` values are omitted. Duplicate keys still decode as arrays.

Sibling PRs from the original combined change:

- #8086 — `cleanPath` regex skip
- #8089 — `parseHref` intern / simple-path sanitize skip
- #8090 — `interpolatePath` client fast path
- #8091 — SSR match-id one-pass dehydrate

## Test plan

- [x] `packages/router-core/tests/qss.test.ts`
- [x] `packages/router-core/tests/searchParams.test.ts` (existing coverage)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query-string encoding and decoding compatibility with standard URL form behavior.
  * Spaces are encoded as `+`, while reserved characters and lone surrogate values are handled safely.
  * Undefined values are omitted, malformed escapes are tolerated, and repeated keys are preserved as arrays.
  * Query strings with optional leading `?` and empty segments are now handled consistently.

* **Tests**
  * Expanded coverage for special characters, malformed input, repeated encoding, and object mutations.

### DIFF
--- a/packages/router-core/src/qss.ts
+++ b/packages/router-core/src/qss.ts
@@ -3,13 +3,160 @@
  * Copyright (c) Luke Edwards luke.edwards05@gmail.com, MIT License
  * https://github.com/lukeed/qss/blob/master/license.md
  *
- * This reimplementation uses modern browser APIs
- * (namely URLSearchParams) and TypeScript while still
- * maintaining the original functionality and interface.
+ * This reimplementation matches `URLSearchParams` encode/decode
+ * (application/x-www-form-urlencoded) without constructing one on
+ * every call.
  *
  * Update: this implementation has also been mangled to
  * fit exactly our use-case (single value per key in encoding).
  */
+
+function percentEncodeByte(code: number): string {
+  return '%' + (code + 256).toString(16).toUpperCase().slice(1)
+}
+
+function hexValue(code: number): number {
+  if (code >= 48 && code <= 57) {
+    return code - 48
+  }
+  if (code >= 65 && code <= 70) {
+    return code - 55
+  }
+  if (code >= 97 && code <= 102) {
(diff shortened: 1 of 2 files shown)


schiller-manuel:
please split this up into separate PRs. looking at the diff i sense quite the increase in bundle size, so we need to track each individual change's implication.  
we also need benchmarks for the affected functions to understand how much improvement each change brings.
only then can we balance between bundle size vs performance, possibly ending up with a different implementation on client vs. server.

anonrig:
@schiller-manuel i cannot split because of this repositories multiple pull request per user limit.

schiller-manuel:
i wasn't aware of this limit, however it is set to 5 but you only have 3 PRs open?
Screenshot 2026-08-16 at 20 45 07

will whitelist you so this limit shouldnt affect you anymore

anonrig:
@schiller-manuel i'll split this pr to multiple now. thanks.

anonrig:
@schiller-manuel split this into separate PRs so each change can be measured on its own (bundle size vs performance):

- #8085 — `qss` encode/decode without `URLSearchParams` (this PR, now qss-only)
- #8086 — `cleanPath` regex skip
- #8089 — `parseHref` intern / simple-path sanitize skip
- #8090 — `interpolatePath` client fast path
- #8091 — SSR match-id one-pass dehydrate

PR #8093, opened 2026-08-17
feat(react-router): allow pendingComponent: false to opt out of the default pending fallback

### LINKED ISSUE #7773: Add a way to opt out of the default pending component
## Problem

Routes that render their own loading state have no way to opt out of the router's default pending component.

`pendingComponent` currently falls back to the configured default whenever a route is pending. Setting the option to `null` or `undefined` does not disable that fallback.

## Proposed solution

Add an explicit opt-out value:

```tsx
export const Route = createFileRoute('/example')({
  pendingComponent: 'never',
})
```
This should prevent the default pending component from rendering while preserving the route's pending behavior.

Opening this issue because there was [PR #2658](https://github.com/TanStack/router/pull/2658) addressing it but it is outdated now.

nishant-batra:
### Description

Fixes #7773.

Adds a way to opt a route out of the router-wide default pending fallback entirely, using the same convention `errorComponent` already ships with:

```ts
// packages/react-router/src/route.tsx
errorComponent?: false | null | undefined | ErrorRouteComponent
pendingComponent?: false | null | undefined | RouteComponent // <- this PR
```

### Why this is a one-line, logic-free change

Both `pendingComponent` and `errorComponent` are read through the same `?? default` pattern, and `??` only falls through on `null`/`undefined` — `false` is not nullish, so it survives.

- `Match.tsx#renderPending` (render): `route?.options.pendingComponent ?? router.options.defaultPendingComponent`, then `if (!PendingComponent) return null`.
- `load-client.ts` (pending-status timing calc): same `??`, then `if (!component ...) return`.

`false ?? default` stays `false`, and the subsequent truthy checks already treat falsy as "nothing to render" — same as `errorComponent`'s `routeErrorComponent ? CatchBoundary : SafeFragment`. So the runtime already does the right thing with `false`; it's only blocked by the type today. No changes needed in either file, confirmed by tracing every reference to `.pendingComponent` in the React path.

### Test plan

Added `packages/react-router/tests/issue-7773-pending-component-opt-out.test.tsx`:
- A route with `pendingComponent: false` never shows the router-wide `defaultPendingComponent`, even with `defaultPendingMs: 0`, and still renders its real content once the loader resolves.
- A contrast case: a sibling route *without* the opt-out still shows the default normally, confirming the type widening doesn't disturb default inheritance.

Ran together with the existing pending/error-boundary suites — 64/64 passing, no regressions.

- [x] `test:types` (tsc on `src`) — clean
- [x] `eslint` on changed files — clean
- [x] `prettier --check` — clean
- [x] `vitest run` on the new + related pending/error test files — 64/64 pass

### Status

Opening as a **draft** — proposed this approach on #7773 and haven't gotten maintainer sign-off yet, since it's technically an API surface change per CONTRIBUTING.md. Marking draft rather than requesting review until that lands.

### AI assistance

Used AI to help trace the `??`/fallback-chain call sites and draft the diff and test. Reviewed and verified the control-flow claims myself and can walk through the reasoning in review.

## Summary by CodeRabbit

* **New Features**
  * Routes can now opt out of displaying the router’s default pending/loading component by setting `pendingComponent` to `false`.
  * Routes without this setting continue to use the shared pending fallback.

* **Documentation**
  * Updated route option documentation to describe the new opt-out behavior.

### DIFF
--- a/packages/react-router/tests/issue-7773-pending-component-opt-out.test.tsx
+++ b/packages/react-router/tests/issue-7773-pending-component-opt-out.test.tsx
@@ -1,0 +1,96 @@
+import { act, cleanup, render, screen } from '@testing-library/react'
+import { afterEach, expect, test, vi } from 'vitest'
+import {
+  Outlet,
+  RouterProvider,
+  createControlledPromise,
+  createMemoryHistory,
+  createRootRoute,
+  createRoute,
+  createRouter,
+} from '../src'
+
+afterEach(() => {
+  cleanup()
+  vi.useRealTimers()
+})
+
+// https://github.com/TanStack/router/issues/7773
+test('pendingComponent: false suppresses the router-wide default pending fallback', async () => {
+  vi.useFakeTimers()
+
+  const loaderGate = createControlledPromise<string>()
+  const rootRoute = createRootRoute({ component: Outlet })
+  const optedOutRoute = createRoute({
+    getParentRoute: () => rootRoute,
+    path: '/opted-out',
+    pendingComponent: false,
+    loader: () => loaderGate,
+    component: () => (
+      <div data-testid="content">{optedOutRoute.useLoaderData()}</div>
+    ),
+  })
+  const router = createRouter({
+    routeTree: rootRoute.addChildren([optedOutRoute]),
+    history: createMemoryHistory({ initialEntries: ['/opted-out'] }),
+    defaultPendingComponent: () => <div data-testid="pending">Pending</div>,
+    defaultPendingMs: 0,
+    defaultPendingMinMs: 100,
+  })
+
+  render(<RouterProvider router={router} />)
+
+  await act(async () => {
+    await vi.advanceTimersByTimeAsync(0)
+  })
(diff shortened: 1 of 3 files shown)


packages/react-router/tests/issue-7773-pending-component-opt-out.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+    history: createMemoryHistory({ initialEntries: ['/opted-out'] }),
+    defaultPendingComponent: () => <div data-testid="pending">Pending</div>,
+    defaultPendingMs: 0,
+    defaultPendingMinMs: 100,
+  })
+
+  render(<RouterProvider router={router} />)
+
+  await act(async () => {
+    await vi.advanceTimersByTimeAsync(0)
```
schiller-manuel: advancing by 0? does that do anything? same below

nishant-batra: You're right, advancing by 0ms doesn't do anything, that was missed before. Thanks for flagging it, fixed now in 5c44c9af90.

PR #8105, opened 2026-08-18
fix(solid-router): defer adapter hydration payloads

### LINKED ISSUE #8104: Solid Start 2.0.0-rc.0: custom serialization adapters emitted into Solid’s SSR stream run before hydrate() installs the adapter map — uncaught `$_TSR.t is undefined`

Start

### Describe the bug

With a custom serialization adapter registered in `start.ts` and a route whose loader primes a query through `@tanstack/solid-router-ssr-query`, every page load throws:

```
TypeError: Cannot read properties of undefined (reading 'get')
```

The dehydrated query state reaches Solid's own SSR resource stream, not just the router's pipeline, so the adapter expression is emitted twice:

- inside Solid's stream script (`_$HY.r["0"] = …`) as a bare `$_TSR.t.get("Money")(1234)` — no `$_TSR.p(...)` wrapper, and ahead of the router's `$tsr-stream-barrier` script that defines `self.$_TSR`. Solid's stream scripts run at parse time, so this one throws (or reports `$_TSR is not defined` when it parses before the router's bootstrap);
- inside the router's barrier script, correctly wrapped in `$_TSR.p(...)` and buffered until `hydrate()` installs the adapter map.

A loader that simply returns the adapter-typed value does NOT reproduce it — that path is serialized by the router's pipeline alone. It takes the ssr-query dehydration path, which is the ordinary shape for an app reading session or loader data through the query cache.

https://github.com/TylerRick/tanstack-solid-ssr-query-adapter-repro

1. `pnpm install`
2. `pnpm dev`
3. Open http://localhost:5597 with the console visible — one uncaught `TypeError` per load.
4. Optional confirmation in the served HTML: the same `$_TSR.t.get("Money")(…)` expression appears once unbuffered inside `_$HY.r[...]`, and once inside `$_TSR.p(...)` in the barrier script.

### Expected behavior

Adapter expressions emitted into Solid's SSR stream are buffered the way the router's own chunks are — or the adapter map is installed before any stream script can run — so a page that streams adapter-typed data through ssr-query loads without an uncaught error.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: `@tanstack/solid-start`, `@tanstack/solid-router`, `@tanstack/solid-router-ssr-query` 2.0.0-rc.0; `@tanstack/solid-query` 6.0.0-rc.0; `@tanstack/router-core` 1.171.16
- solid-js / `@solidjs/web`: 2.0.0-rc.0
- OS: Linux
- Browser: Chromium 141
- Bundler: vite 8.2.1 (vite-plugin-solid 3.0.0-next.27)

### Additional context

Ours carries a `Temporal.PlainDate` on the session, so every page load for a user with that field set threw. Workaround we ship: an inline head script, rendered first, that installs `$_TSR` as an accessor with a passthrough `t` map, so nothing throws whichever script parses first and `hydrate()`'s real map still wins.

Related: #5737 (closed, different cause), #7706 (adapter deserialization broken by calling a server fn from the client entry).

Fnine59:
Fixes #8104.

Solid's SSR serializer can emit hydration data before the router bootstrap and before `hydrate()` installs the custom adapter map. Adapter expressions in those eager scripts therefore read `$_TSR.t` too early.

This change routes a Solid hydration payload through the router's existing buffer once a custom adapter participates in serialization. If the payload arrives before the bootstrap, it creates only an early buffer that the real bootstrap adopts. The complete payload and all later serializer chunks stay in order, so `hydrate()` installs the adapter map and runs the real `fromSerializable` function before Solid hydration consumes the value; there is no identity fallback.

Regression coverage exercises both missing-`$_TSR` and missing-`.t` timing, payloads that arrive after initialization, preservation of the pre-bootstrap queue, restoration of a `Money` instance and its method, and the unchanged default serializer path when no adapters are configured.

## Reproduction

- [Minimal reproduction repository](https://github.com/TylerRick/tanstack-solid-ssr-query-adapter-repro)


packages/router-core/src/ssr/tsrScript.ts
```diff
@@ -16,5 +16,5 @@ self.$_TSR = {
   p(script) {
     !this.initialized ? this.buffer.push(script) : script()
   },
-  buffer: [],
+  buffer: self.$_TSR?.buffer || [],
```
brenelz, line 19: we can't really make router-core changes on this branch

brenelz (changes requested):

Issue #8115, opened 2026-08-19
useRouteContext() renders with beforeLoad-provided context stripped during pending presentation


antur84:
### Which project does this relate to?

Router

### Describe the bug

Since the lane loader rewrite (1.170.29+), a client-side navigation can render an already-mounted route with a context that's missing every key its `beforeLoad` returns.

`contextualize` (router-core `dist/esm/load-client.js`) updates the in-flight lane match in two steps: it assigns `match.context = { ...parentContext, ...routeContext }` *before* awaiting the route's `beforeLoad`, and merges the result back in only after the await. Meanwhile `offerPending` fires at `pendingMs` and presents shallow clones of those same lane matches. If the timer lands inside the await, the presented root match still has `status: "success"` (it sits above the pending boundary), so the root component re-renders from it — and `useRouteContext()` is suddenly missing everything `beforeLoad` provides.

It only needs three ordinary ingredients: a pending component on the navigation target (or a `defaultPendingComponent`), a `pendingMs` shorter than an async root `beforeLoad` (we use `defaultPendingMs: 150`; on mobile devices a 150ms+ `beforeLoad` is routine), and a component that reads that context during render. With `router.invalidate()` involved it's worse — invalidated matches present with zero delay (`match.invalid ? 0 : pendingMs`), so even a fast `beforeLoad` window gets rendered.

### Your Example Website or App

https://github.com/antur84/tanstack-router-pending-context-loss-repro

`pnpm install && pnpm build && pnpm start`, open the app with the console showing, click either button. One attempt per page load (a retained destination never goes pending), and note the repro disables `defaultPreload: 'intent'` — hover preloading moves the navigation outside the window, which is also why this skews hard to touch devices in production. The `old-versions-clean` branch is the identical app pinned to 1.170.15 / 1.168.25: same clicks, no context loss.

### Expected behavior

A match being reloaded keeps its previous (complete) context visible until the new `beforeLoad` result is in — the rendered UI should never observe the half-built context.

### Actual behavior

The root component re-renders with the stripped context, crashes into the root `errorComponent` for a moment, and the navigation then completes. Consumers that cache the value are broken past the flash — in our app the locale from root `beforeLoad` seeds the i18n runtime, so this crashed ~500 users/hour to the error screen after we upgraded (reverting stopped it).

### Suggested fix

[...]

PR #8116, opened 2026-08-19
fix(router-core): keep a reloading match's previous beforeLoad context until it settles

antur84:
Closes the window described in #8115 structurally. `contextualize` assigned `{ ...parentContext, ...routeContext }` to the live lane match before awaiting `beforeLoad`, so anything observing the match mid-load (the pending presentation before #8084's gating; `matches` inside a beforeLoad; future callers) saw a context stripped of every beforeLoad-provided key.

The match is now seeded from the committed same-id match during the window, keeps the exact fresh merge on the no-beforeLoad path, and resets to it on every non-success settle — preserving the "failure clears the previous generation" semantics pinned by `preload-beforeload-reuse.test.ts`.

The new test observes the window directly from inside a reloading `beforeLoad` and fails without this change (`[undefined]` vs `['en']`). Standalone repro from the issue: https://github.com/antur84/tanstack-router-pending-context-loss-repro.

Ran locally: router-core (1590), react-router (1027), and `test:unit` across start-client-core, start-server-core, react-start-client/server, both ssr-query packages, and solid-router — all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved previously loaded route context while refreshed route data is still loading.
  * Maintained route context during validation failures, unsuccessful loads, and handled loading errors.
  * Prevented temporary loss of context for routes without a loading hook.

* **Tests**
  * Added regression coverage for context retention during asynchronous route reloads.

### DIFF
--- a/packages/react-router/tests/issue-8115-beforeload-context-window.test.tsx
+++ b/packages/react-router/tests/issue-8115-beforeload-context-window.test.tsx
@@ -1,0 +1,51 @@
+import * as React from 'react'
+import { act, cleanup, render, screen } from '@testing-library/react'
+import { afterEach, expect, test, vi } from 'vitest'
+import {
+  RouterProvider,
+  createMemoryHistory,
+  createRootRoute,
+  createRouter,
+} from '../src'
+
+afterEach(() => {
+  cleanup()
+  vi.useRealTimers()
+})
+
+// https://github.com/TanStack/router/issues/8115
+test('#8115: a reloading match keeps its previous beforeLoad context until the new result lands', async () => {
+  vi.useFakeTimers()
+
+  const observed: Array<unknown> = []
+  let runs = 0
+  const rootRoute = createRootRoute({
+    beforeLoad: async ({ matches }) => {
+      runs++
+      if (runs > 1) {
+        observed.push((matches[0] as any).context?.locale)
+      }
+      await new Promise((resolve) => setTimeout(resolve, 100))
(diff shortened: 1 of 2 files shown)


antur84:
Field data supporting this fix: after upgrading to 1.170.30 / router-core 1.171.25 (which includes #8084), the context loss from #8115 recurred in our production app — at roughly 0.5% of the original rate (~3 events/hour vs ~500 users/hour), but through a different vector than the one #8084 gates.

Every captured stack shows a **synchronous** React flush (`flushSyncWorkAcrossRoots_impl` → `performSyncWorkOnRoot`) rendering the root route with the stripped context — never the `pendingMs` presentation from the original report. The sequence we infer:

1. A same-id reload runs `contextualize`, which mutates the **committed (live) match object's** `.context` to `{ ...parentContext, ...routeContext }` before awaiting `beforeLoad` (load-client.ts — the window this PR closes).
2. Right after that write and before the await, the router touches its store (`setFetching(...'beforeLoad')` / `status = 'pending'`).
3. A `useSyncExternalStore` subscriber flushes synchronously in the same task, and that render reads the live match's half-built context via `useRouteContext()`.

#8084's presentation gating can't help here because the observed object is the committed match itself, not a presented lane clone. Consistent with a task-scheduling race on a microtask-scale window, all seven production hits were mobile or low-end devices (Android WebView, Chrome Mobile), and it isn't reproducible with a plain navigation.

PR #8131, opened 2026-08-20
perf(ssr-query): Batch streamed query dehydration

scttcper:
Queries that finish during SSR currently dehydrate and enqueue one at a time, walking the full query cache for every query. This batches queries that finish in the same Query scheduler turn. Later queries still stream separately, pending work flushes when rendering finishes, and cleanup drops anything left over.

Measured with a production TanStack Start SSR route rendering separate `useSuspenseQuery` children with 20 records each. Before and after requests were interleaved and consumed the full response.

| Query completion pattern | Chunks | Response size | Mean latency |
|---|---:|---:|---:|
| 10 timers in one turn | 10 -> 1 | -4.6% | -1.6% |
| 10 timers staggered 1 ms apart | 10 -> 4 | -3.0% | neutral |
| 50 callbacks in one turn | 50 -> 1 | -6.5% | -8.1% |

Tests cover batching across scheduler turns, refetch deduplication, custom dehydration filters, render completion, and request cleanup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server-side query streaming by batching completed queries into fewer updates.
  * Prevented pending queries from being streamed after request cleanup or stream closure.
  * Ensured queued query results are flushed before the stream closes.

### DIFF
--- a/packages/router-ssr-query-core/src/index.ts
+++ b/packages/router-ssr-query-core/src/index.ts
@@ -199,6 +199,8 @@ export function setupCoreRouterSsrQueryIntegration<TRouter extends AnyRouter>({
       }
       if (!pendingQueryHashes) {
         pendingQueryHashes = new Set()
+        // QueryCache listeners run inside notifyManager.batch. Scheduling the
+        // flush collects queries completed during the same scheduler turn.
         notifyManager.schedule(() => {
           try {
             flushPendingQueries()
--- a/packages/router-ssr-query-core/src/index.ts
+++ b/packages/router-ssr-query-core/src/index.ts
@@ -1,4 +1,5 @@
 import {
+  notifyManager,
   dehydrate as queryDehydrate,
   hydrate as queryHydrate,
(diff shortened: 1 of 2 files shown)


scttcper:
The CodSpeed `-5.31%` report appears to be a baseline artifact. None of the flagged benchmarks import `router-ssr-query-core`; the base Solid memory job crashed inside CodSpeed, and #8130 against the same base reported the exact same `211.1 ms -> 227.2 ms` client-nav result. The report is also still on the first commit, while benchmark runs for the current head need maintainer approval.

Base failure: https://github.com/TanStack/router/actions/runs/32280657076/job/96158388236

Same result on #8130: https://github.com/TanStack/router/pull/8130#issuecomment-5356303083

schiller-manuel:
ignore the codspeed benchmarks please, they are unfortunately not stable

schiller-manuel:
thanks for the initial idea, i took it further in https://github.com/TanStack/router/pull/8175

PR #8134, opened 2026-08-21
fix(start-server-core): return 499 when the client disconnects mid-request

### LINKED ISSUE #7991: Start: a client disconnect is reported as an unhandled 500, and an app cannot opt out

Start

### Describe the bug

When a client goes away mid-request — a reload, a closed tab, an aborted `fetch` — Start rejects the request with `signal.reason`, and h3 classifies that rejection as an unhandled server error: a `console.error` plus a 500, once per abandoned request.

The cancellation propagation itself is correct and welcome (it landed for us in `@tanstack/start-server-core@1.169.19`; before that, disconnects were silently ignored). What's missing is the classification: a client hanging up is a normal outcome, but Start has no vocabulary to distinguish it from a genuine failure, and no seam for the app to supply one.

This matters most for routes that hold long-lived upstream connections. We proxy ElectricSQL shape requests; a single authenticated page holds ~10 long-polls, so every reload emits 10 unhandled 500s. At that rate the log noise buries real errors.

The reproduction contains two routes:

- `/api/naive` — a long-lived handler that ignores `request.signal`.
- `/api/careful` — the same handler written to do everything right: it observes `request.signal`, tears down its work, and resolves a clean `204`. It never rejects.

**Both produce an identical unhandled 500**, which is the core of the report — the app-side fix does not exist.

https://github.com/jwaltz/start-client-disconnect-repro

1. Clone the reproduction repo
2. `bun install`
3. `bun run build`
4. `bun run serve` (terminal 1)
5. `bun run reproduce` (terminal 2) — starts each request and aborts it after 250 ms

### Expected behavior

A request abandoned by its own client should not be logged as an unhandled server error. Ideally either:

- Start treats a rejection that is exactly the aborted request's `signal.reason` as a terminal non-error, or
- Start threads a config through to `toResponse` (h3 already supports `silent` and `onError`) so an app can classify it.

Genuine upstream failures must stay visible either way — including the race where a real error is thrown and the client disconnects immediately afterward.

### Screenshots or Videos

Server output — one block per disconnect, including the handler that returned a clean 204:

```
error: The connection was closed.
      cause: DOMException { code: 20, name: "AbortError", message: "The connection was closed." },
     status: 500,
  unhandled: true,

      at prepareResponse (.../h3-v2.mjs:181:37)
      at toResponse (.../h3-v2.mjs:150:19)
```

### Platform

- `@tanstack/react-start`: 1.168.38
- `@tanstack/start-server-core`: 1.169.21
- `@tanstack/react-router`: 1.170.21
- `h3-v2`: 2.0.1-rc.20
- `srvx`: 0.11.22
- `vite`: 8.2.1, `nitro`: 3.0.260610-beta (preset `bun`)
- Bun: 1.4.0-canary.1

Also observed under the Node dev adapter, where the abort originates in `srvx/dist/adapters/node.mjs` on early response close — so this is not Bun-specific.

### Additional context

Three seams we checked, all closed (paths in `@tanstack/start-server-core@1.169.21`):

1. **Returning a Response from the handler doesn't help.** `executeMiddleware` re-checks `signal.aborted` *after* awaiting the handler and throws `signal.reason` over the returned Response (`dist/esm/createStartHandler.js:199-207`). That's what `/api/careful` demonstrates.
2. **A custom `server.ts` wrapper is too late.** `createStartHandler` returns `requestHandler(startRequestResolver)` internally (`dist/esm/createStartHandler.js:419`), so by the time app code sees the value, h3 has already logged and converted it.
3. **h3's escape hatches are unreachable.** `toResponse(val, event, config = {})` logs when `error.unhandled && !config.silent`, and marks any non-`HTTPError` as `unhandled = true`. Start calls `toResponse(value, h3Event)` with no config (`dist/esm/request-response.js`), so neither `silent` nor `onError` can be supplied.

The only mechanism we found was patching `request-response.js` to inspect the rejection before it reaches h3. That works, but pins the app to one exact transitive version of `start-server-core` — a routine dependency bump silently un-applies it — so we've reverted it and are living with the noise.

Possibly related: #6069 (uncatchable "connection was closed" DOMException flooding logs), #7873 (`serverFnFetcher` logging benign cancellation AbortErrors), #7748 (client abort mid-SSR crashing `vite dev` via the stream watchdog), #7929 (aborted SSR render returning 200). They look like the same missing concept viewed from four places: no shared notion of "this stopped because the client left."

naoya7076:
Fixes #7991.

## Problem

`startRequestResolver` in `createStartHandler` has no `catch` around its `try/finally`. When a client disconnects, the abort is re-thrown internally as `signal.reason` so that in-flight work unwinds and gets disposed. That rejection then escapes the request boundary. h3 marks it as unhandled, logs it with `console.error`, and responds 500.

The app has no way to opt out:

- Start calls `toResponse(value, h3Event)` without a config, so h3's `silent` / `onError` options never apply.
- A handler that observes the signal and returns a response gets discarded, because `executeMiddleware` re-throws `signal.reason` at the end when the signal is aborted.

More background in my comment on the issue: https://github.com/TanStack/router/issues/7991#issuecomment-5342562494

## Change

Catch the abort at the outermost request boundary and return `499 Client Closed Request`:

```ts
} catch (err) {
  if (request.signal.aborted && err === request.signal.reason) {
    return new Response(null, { status: 499, statusText: 'Client Closed Request' })
  }
  throw err
}
```

- The internal `throw signal.reason` calls are untouched. They are still needed to drive cleanup. Only the final classification changes.
- The check compares identity with `request.signal.reason` instead of matching `err.name === 'AbortError'`, so unrelated `AbortError`s thrown by app code still propagate as real errors.
- This is option 1 from https://github.com/TanStack/router/issues/7991#issuecomment-3493614284. The fix sits at the shared boundary, so it covers SSR, server routes, and server functions at once.
- A patch changeset for `@tanstack/start-server-core` is included.

### Why 499

Per [RFC 9110 §15.6](https://www.rfc-editor.org/rfc/rfc9110#section-15.6), 5xx means the server itself failed. A disconnect is the client withdrawing the request, so 500 misclassifies it. There is no standard code for this case because the response never reaches the client; it only matters for logs and middleware. nginx's non-standard [499 (`NGX_HTTP_CLIENT_CLOSED_REQUEST`)](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_request.h) became the de facto convention for it.

The adjacent layers already handle it this way: [h3's `proxy()` returns 499](https://github.com/h3js/h3/blob/main/src/utils/proxy.ts) when `event.req.signal.aborted` (since 2.0.1-rc.23), and srvx's node adapter suppresses `console.error` for aborted requests. Start is currently the only layer in the stack that turns a disconnect into an unhandled 500.

## Behavioral change

Disconnects that previously produced a 500 and an error log now produce a 499 and no log. If you alert on 5xx rates, expect fewer entries. The client never sees either response, so nothing changes on the wire.

## Tests

13 existing assertions in the `createStartHandler request cancellation` suite expected `status === 500` after `requestController.abort(...)`. They now expect 499. That is the whole extent of the behavior change. The side-effect assertions in those tests (cleanup runs once, streams disposed with the abort reason, render never called) are unchanged. The one 500 assertion that is not abort-related (middleware throwing a real error) is also unchanged.

Three new tests assert that a disconnect returns `499 Client Closed Request` on each request path:

- SSR render
- server route handler
- server function call

To run locally:

```sh
pnpm nx run @tanstack/start-server-core:test:unit -- tests/createStartHandler.test.ts
```

`test:unit` (33 passed), `test:types`, and `test:eslint` pass on the package.

## Verification

I applied this diff with `pnpm patch` to `@tanstack/start-server-core@1.169.26` in a production Start app (server routes proxying Connect RPC to a backend), and compared the same screens with and without it:

| | `AbortError` logs | unhandled 500s |
|---|---|---|
| without patch | 6 | 6 |
| with patch | 0 | 0 |

## Related PRs

- #6539 returns 499 for aborts inside the React SSR stream. Same conclusion, narrower scope. No file overlap.
- #7930 makes aborted SSR throw instead of returning a 200 shell. Complementary: once it throws, this boundary classifies it. No file overlap.
- No open PR touches this file or addresses #7991.

If you prefer a different status code, or want to settle the policy discussion in #7991 first, I'm happy to adjust.

## Summary by CodeRabbit

* **Bug Fixes**
  * Client disconnects during server-side rendering, route handling, or server function calls now return HTTP 499 “Client Closed Request” instead of an internal server error.
  * Prevents unnecessary rendering after a client connection has been closed.
* **Release**
  * Includes a patch release for the server core package.

### DIFF
--- a/.changeset/client-disconnect-499.md
+++ b/.changeset/client-disconnect-499.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/start-server-core': patch
+---
+
+Return `499 Client Closed Request` instead of an unhandled 500 when the client disconnects mid-request. Applies to SSR, server routes, and server functions.
--- a/packages/start-server-core/src/createStartHandler.ts
+++ b/packages/start-server-core/src/createStartHandler.ts
@@ -779,6 +779,15 @@ export function createStartHandler<TRegister = Register>(
       request.signal.throwIfAborted()
       responseOwnsCleanup = response.serverSsrCleanup === 'stream'
       return response.response
+    } catch (err) {
+      // classify a client disconnect as 499 instead of an unhandled 500
+      if (request.signal.aborted && err === request.signal.reason) {
+        return new Response(null, {
+          status: 499,
+          statusText: 'Client Closed Request',
+        })
+      }
+      throw err
     } finally {
       if (router?.serverSsr && !responseOwnsCleanup) {
         // Clean up router SSR state if it was set up but won't be cleaned up by the callback
--- a/packages/start-server-core/tests/createStartHandler.test.ts
+++ b/packages/start-server-core/tests/createStartHandler.test.ts
@@ -406,7 +406,7 @@ describe('createStartHandler request cancellation', () => {
       const cancellation = new Error('request disconnected')
       requestController.abort(cancellation)
 
-      expect((await response).status).toBe(500)
+      expect((await response).status).toBe(499)
       expect(routeSignal?.aborted).toBe(true)
       expect(routeSignal?.reason).toBe(cancellation)
       expect(render).not.toHaveBeenCalled()
@@ -463,7 +463,7 @@ describe('createStartHandler request cancellation', () => {
     await renderStarted
     requestController.abort(new Error('request disconnected'))
 
-    expect((await response).status).toBe(500)
+    expect((await response).status).toBe(499)
     expect(cleanupCalls).toBe(1)
     expect(router.serverSsr).toBeUndefined()
 
@@ -503,7 +503,7 @@ describe('createStartHandler request cancellation', () => {
     const cancellation = new Error('request disconnected')
     requestController.abort(cancellation)
 
-    expect((await response).status).toBe(500)
+    expect((await response).status).toBe(499)
     resolveRender(new Response(new ReadableStream({ cancel })))
     await vi.waitFor(() => {
       expect(cancel).toHaveBeenCalledTimes(1)
@@ -542,7 +542,7 @@ describe('createStartHandler request cancellation', () => {
     const cancellation = new Error('request disconnected')
     requestController.abort(cancellation)
 
-    expect((await response).status).toBe(500)
+    expect((await response).status).toBe(499)
     resolveMiddleware(new Response(new ReadableStream({ cancel })))
     await vi.waitFor(() => {
       expect(cancel).toHaveBeenCalledTimes(1)
@@ -594,7 +594,7 @@ describe('createStartHandler request cancellation', () => {
 
       await renderStarted
       requestController.abort(new Error('request disconnected'))
-      expect((await response).status).toBe(500)
+      expect((await response).status).toBe(499)
 
       resolveRender({
         response: new Response('stream'),
@@ -657,7 +657,7 @@ describe('createStartHandler request cancellation', () => {
       await middlewareStarted
       requestController.abort(new Error('request disconnected'))
 
-      expect((await response).status).toBe(500)
+      expect((await response).status).toBe(499)
       await vi.waitFor(() => {
         expect(consoleError).toHaveBeenCalledWith(cleanupError)
       })
@@ -731,7 +731,7 @@ describe('createStartHandler request cancellation', () => {
     await middlewareStarted
     requestController.abort(new Error('request disconnected'))
 
-    expect((await response).status).toBe(500)
+    expect((await response).status).toBe(499)
     expect(render).not.toHaveBeenCalled()
 
     releaseMiddleware({
@@ -779,7 +779,7 @@ describe('createStartHandler request cancellation', () => {
     await innerStarted
     requestController.abort(new Error('request disconnected'))
 
-    expect((await response).status).toBe(500)
+    expect((await response).status).toBe(499)
     expect(outerFinally).toHaveBeenCalledOnce()
     expect(render).not.toHaveBeenCalled()
   })
@@ -812,7 +812,7 @@ describe('createStartHandler request cancellation', () => {
     await innerStarted
     requestController.abort(new Error('request disconnected'))
 
-    expect((await response).status).toBe(500)
+    expect((await response).status).toBe(499)
     expect(render).not.toHaveBeenCalled()
   })
 
@@ -838,7 +838,7 @@ describe('createStartHandler request cancellation', () => {
       {},
     )
 
-    expect(response.status).toBe(500)
+    expect(response.status).toBe(499)
     expect(render).not.toHaveBeenCalled()
   })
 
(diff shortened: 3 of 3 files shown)


radist2s:
Thanks for putting this together. We've been hitting this exact issue repeatedly in Playwright E2E: routine client disconnects produce a wall of `AbortError` / unhandled 500 logs even though the suite succeeds. Classifying the request's own abort reason as 499 at the Start boundary looks like the right fix and would make CI logs much more useful. We'd really appreciate seeing this merged.

Sheraff:
499 actually seems reasonable 
- h3 https://github.com/h3js/h3/blob/1892ee9cae06533c7db72a5213bceda61cc1d58d/src/utils/proxy.ts#L362-L374
- nginx https://github.com/nginx/nginx/blob/master/src/http/ngx_http_request.h#L133-L143
- asp.net https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.statuscodes.status499clientclosedrequest?view=aspnetcore-10.0

naoya7076:
Thanks @Sheraff — good to have the 499 precedent confirmed.

We're carrying this as a `pnpm patch` in production; without it, unhandled 500s fill our logs on every navigation that cancels in-flight calls. @radist2s hits the same in Playwright runs.

Rebased onto main. Two things:

- If the direction looks right, I'd appreciate a review. It's one catch block in `createStartHandler.ts`; the rest is tests.
- The test workflows still haven't run here — only CodeRabbit and Socket have reported. Could someone kick them off? `test:unit` (33 passed), `test:types` and `test:eslint` pass locally.

frenzzy:
Another production data point, and a temporary package for anyone else waiting on this.

We run TanStack Start (1.168.x) behind a GCP load balancer on our own h2c server (srvx `NodeRequest`). Over 24h, 202 of 203 `AbortError: This operation was aborted` unhandled-500 prints matched an LB `client_disconnected_before_any_response` entry for the same request, down to the millisecond. So every one of them was a browser leaving, not a server fault. Each one printed h3's ~20-line `console.error` at ERROR and counted against our error budget.

I checked this PR against a real build: an h2c client that sends `NGHTTP2_CANCEL` mid-request to a server route. On 1.169.28 it prints `status: 500, unhandled: true`. With this change, nothing reaches h3 (499). The identity check against `request.signal.reason` is what makes it safe for us: an upstream `AbortSignal.timeout()` inside a handler is also an `AbortError`, and it still surfaces as a real error.

Until this lands, I've published the commit cherry-picked onto the exact `1.169.28` that `@tanstack/react-start@1.168.46` pins (all 121 package tests pass; only `dist/esm/createStartHandler.js` differs):

- Package: [@frenzzy/start-server-core](https://www.npmjs.com/package/@frenzzy/start-server-core)
- Version: `1.169.28-client-disconnect-499.0`
- Tag: `client-disconnect-499`

```json
"overrides": {
  "@tanstack/start-server-core": "npm:@frenzzy/start-server-core@1.169.28-client-disconnect-499.0"
}
```

It's a drop-in replacement until an official release includes this fix. Thanks @naoya7076 for the fix. I'd love to see it merged.

radist2s:
I’m really glad this problem isn’t just bothering me, but the bloke next to me as well. Let’s sort this mess out already.

PR #8136, opened 2026-08-21
chore(start): remove redundant splitChunks config from React plugin

chenjiahan:
TanStack Start already configures Rsbuild to split async chunks. This prevents `@rsbuild/plugin-react` from adding its chunk groups, so `splitChunks: false` is redundant.

This PR removes the option from benchmarks, e2e fixtures, and docs without changing build output.

## Summary by CodeRabbit

* **Improvements**
  * React applications now use the default code-splitting behavior, enabling split chunks for improved bundle optimization.
  * Updated React examples, documentation, and end-to-end scenarios to reflect the default configuration.


chenjiahan:
These CI failures look like permission issues, the workflows couldn’t post comments or labels due to the fork’s read-only token.

PR #8145, opened 2026-08-21
perf(router-core): evaluate head/scripts across matches in parallel

Sheraff:
## Summary
- `projectLane` (load-server.ts + load-client.ts) awaited each match's head/scripts(/headers) sequentially — total latency was the **sum** of all resolutions instead of the max.
- Now invokes all callbacks in the rendered prefix synchronously (route order), then awaits/applies results one match at a time in route order with the original per-match error/abort/break logic.

## Behavioral equivalence (strict)
Each preserved ordering has a dedicated test in `tests/head-scripts-parallel.test.ts` (10 tests):
- results applied in route order even when later matches resolve first
- early-break semantics (`ssr:false` / error / `_notFound`) unchanged
- deterministic route-order error surfacing; sync-throw handling; silent abort break
- zero unhandled rejections (catch handlers attached to every promise, incl. discarded ones)

**Caught during verification:** a first implementation wrapped results in an extra `.then` layer; that single extra microtask tick flipped a real scheduling race so `runBackground`'s publish landed *after* `router.load()` resolved — caught by the existing `public-hydration-contract.test.ts`. Fixed by awaiting the raw attempt promise (identical await-chain depth to original).

## Performance
N=6 heads × 60ms: sequential 362ms → projectLane **62ms** (~max, not sum).

## Verification
test:unit ✅ (107 files / 1616 tests) · eslint ✅ · types ✅

## Residual risks (documented in RESULT-perf-task4.md)
1. head fns for the whole prefix are invoked up-front: a fn reading *another* match's already-applied head result would see it unapplied (loaders are guaranteed complete first; contract tests pass).
2. Under abort, later prefix callbacks may be invoked before the abort is observed.

### DIFF
--- a/RESULT-perf-task4.md
+++ b/RESULT-perf-task4.md
@@ -1,0 +1,82 @@
+# RESULT — perf/task4-parallel-head-scripts
+
+## Change
+
+`projectLane` evaluated each match's head/scripts(/headers) callbacks
+sequentially — a `Promise.all` **per match** inside a sequential loop — so
+total latency was the **sum** of all callback resolutions. Both implementations
+now:
+
+1. Walk the rendered prefix first (unchanged break conditions), invoking every
+   qualifying match's head/scripts(/headers) callbacks synchronously **in route
+   order** (identical invocation order to before).
+2. Await and apply the results strictly **one match at a time in route order**,
+   reusing the exact original per-match apply/error/abort/break logic.
+
+Files:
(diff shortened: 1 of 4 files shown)


Sheraff:
Superseded by #8151, which contains only the route asset parallelization. The independent byte-reduction changes were split into #8150. Both target main.

Issue #7331, opened 2026-05-03
Solid Router route-scoped accessors can throw during navigation teardown


s3bba:
### Which project does this relate to?

Router

### Describe the bug

In Solid 2.0 and TanStack Solid Router, route scoped accessors such as `Route.useParams()` and `Route.useSearch()` can throw after navigation has moved away from the route that created the accessor.

The test creates a route component that calls `postRoute.useParams()`, schedules a delayed read of that accessor, navigates away, and then lets the delayed read run. On current code, the delayed read throws:

```text
Invariant failed: Could not find an active match from "/posts/$postId"
```

Obviosly you would never write such code for the actual applications, but the async reading of accessors can be triggered in createEffect() calls.

Reduced test code snippet:

```tsx
component: function PostComponent() {
  const params = postRoute.useParams()

  // Simulated async work that reads params accessor
  delayedRead = delayedReadGate.then(() => {
    try {
      params()
    } catch (err) {
      delayedError = err
    }
  })

  return <h1>Post {params().postId}</h1>
}
```

```tsx
await router.navigate({ to: '/other' })
// Trigger async work
releaseDelayedRead()
await delayedRead

if (delayedError) {
  throw new Error(
    `Route-scoped useParams threw after navigation away: ${
      delayedError instanceof Error ? delayedError.message : String(delayedError)
    }`,
  )
}
```

The current behavior demonstrates that a route-created accessor can become throwing solely because active matches no longer contain that route id.

### Complete minimal reproducer

https://github.com/TanStack/router/pull/7330

### Steps to Reproduce the Bug

Run the added test in the pull request

```
CI=1 NX_DAEMON=false pnpm nx run @tanstack/solid-router:test:unit --outputStyle=stream --skipRemoteCache -- tests/useMatch.test.tsx
```

```
@tanstack/solid-router: ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
@tanstack/solid-router:  FAIL   @tanstack/solid-router  tests/useMatch.test.tsx > useMatch > route-scoped useParams should remain readable from async work created by the route during navigation away
@tanstack/solid-router: Error: Route-scoped useParams threw after navigation away: Invariant failed: Could not find an active match from "/posts/$postId"
@tanstack/solid-router:  ❯ tests/useMatch.test.tsx:170:13
@tanstack/solid-router:     168|
@tanstack/solid-router:     169|     if (delayedError) {
[...]

PR #8169, opened 2026-08-26
fix: preserve shouldThrow in route-scoped hooks

### LINKED ISSUE #3482: shouldThrow option for useParams not working

### LINKED ISSUE #8168: Route-scoped useMatch ignores shouldThrow

Router

### Describe the bug

Current route: `/`
Target route: `/orgs/$orgId`

Generic:

```ts
useMatch({
  from: '/orgs/$orgId',
  shouldThrow: false,
})
```

→ `undefined`

Route-scoped:

```ts
orgDetailRoute.useMatch({
  shouldThrow: false,
})
```

→ invariant error

Route-scoped `useMatch` drops `shouldThrow` while delegating to the standalone `useMatch` hook.

https://github.com/mizukendesu/route-usematch-verify

1. Clone the reproducer and run `pnpm install`
2. Run `pnpm dev:before`
3. Open `/orgs/acme` — both APIs return a match
4. Navigate to Home — generic `useMatch` returns `undefined`, route-scoped `useMatch` throws

### Expected behavior

`undefined` when the target route is inactive and `shouldThrow` is false.

### Platform

- Router / Start Version: 1.170.32
- OS: macOS
- Browser: Chrome
- Bundler: vite
- Bundler Version: 6.2.0

### Additional context

Published `@tanstack/react-router@1.170.32`. The sidebar stays mounted on Home, so reading the child `/orgs/$orgId` match from that shell is optional.

mizukendesu:
## 🎯 Changes

Route-scoped `useMatch`, `useSearch`, and `useParams` wrappers were dropping `shouldThrow` when delegating to the standalone hooks. As a result, `shouldThrow: false` fell back to the default `true` behavior and threw when the target route was inactive.

This PR fixes that across **React, Solid, and Vue** by:

- forwarding all hook options via spread (`{ ...opts, from: this.id }`) in `RouteApi`, `Route`, `RootRoute`, and `LazyRoute`
- adding `TThrow` to route-scoped hook types so `shouldThrow: false` correctly returns `undefined` instead of being a type error
- adding runtime and type regression coverage for each framework

### Spread vs explicit forwarding

Compared corrected explicit forwarding against spread on `*.minimal` scenarios:

| Scenario | explicit (gzip) | spread (gzip) |
| --- | ---: | ---: |
| react-router.minimal | 85862 | 85865 |
| solid-router.minimal | 34005 | 33982 |
| vue-router.minimal | 50647 | 50627 |

Differences are negligible. Spread was chosen to avoid future option-forwarding omissions.

Fixes #8168
Fixes #3482

Related to #3642 — Overlaps with #3642, which previously explored `shouldThrow` support for route-scoped `useSearch` / `useParams`.

Related to #7331

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Route-scoped `useMatch`, `useSearch`, and `useParams` hooks now forward all supplied options.
  * Using `shouldThrow: false` correctly returns optional results, including `undefined` for inactive routes.
  * Type definitions now accurately reflect optional results across React, Solid, and Vue routers.

* **Bug Fixes**
  * Improved consistency for route, route API, and lazy route hooks across supported frameworks.

* **Tests**
  * Added coverage for optional return values and inactive-route behavior.

### DIFF
--- a/.changeset/route-scoped-use-match-should-throw.md
+++ b/.changeset/route-scoped-use-match-should-throw.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/react-router": patch
+---
+
+Fix route-scoped useMatch APIs to forward the shouldThrow option.
--- a/.changeset/route-scoped-use-match-should-throw.md
+++ b/.changeset/route-scoped-use-match-should-throw.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/react-router": patch
+'@tanstack/react-router': patch
 ---
 
 Fix route-scoped useMatch APIs to forward the shouldThrow option.
--- a/.changeset/route-scoped-use-match-should-throw.md
+++ b/.changeset/route-scoped-use-match-should-throw.md
@@ -1,5 +1,7 @@
 ---
 '@tanstack/react-router': patch
+'@tanstack/solid-router': patch
+'@tanstack/vue-router': patch
 ---
 
-Fix route-scoped useMatch APIs to forward the shouldThrow option.
+Fix route-scoped `useMatch`, `useSearch`, and `useParams` APIs to forward the `shouldThrow` option and preserve optional return types when `shouldThrow: false`.
--- a/packages/react-router/src/fileRoute.ts
+++ b/packages/react-router/src/fileRoute.ts
@@ -221,6 +221,7 @@ export class LazyRoute<TRoute extends AnyRoute> {
     return useMatch({
       select: opts?.select,
       from: this.options.id,
+      shouldThrow: opts?.shouldThrow,
       structuralSharing: opts?.structuralSharing,
     } as any) as any
   }
--- a/packages/react-router/src/fileRoute.ts
+++ b/packages/react-router/src/fileRoute.ts
@@ -218,12 +218,7 @@ export class LazyRoute<TRoute extends AnyRoute> {
   }
 
   useMatch: UseMatchRoute<TRoute['id']> = (opts) => {
-    return useMatch({
-      select: opts?.select,
-      from: this.options.id,
-      shouldThrow: opts?.shouldThrow,
-      structuralSharing: opts?.structuralSharing,
-    } as any) as any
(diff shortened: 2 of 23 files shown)


Sheraff (commented):
thank you for this PR
- could you do the same for `useSearch` and `useParams`?
- can you also take a look and see if Solid and Vue have the same issue (and if so, handle it there too)?
- since we're forwarding all options, would it reduce the bundle size to spread instead of explicitly listing all keys? `useMatch({ ...opts, from: this.id })`

mizukendesu:
@Sheraff 

Thanks for the review — all three points are addressed in the latest push:

1. **React `useSearch` / `useParams`** — now forward `shouldThrow` (+ `TThrow` on route-scoped types)
2. **Solid / Vue** — same bug in all three route-scoped hooks; fixed across all four wrappers
3. **Spread vs explicit** — compared on `*.minimal`; gzip diffs were negligible, so I went with spread to avoid future forwarding omissions (details + table in the PR description)

Added per-framework runtime/type regression tests. Let me know if you'd rather land explicit forwarding anyway.

PR #8181, opened 2026-08-27
fix: return AsyncRouteComponent from lazyRouteComponent

anonrig:
## Summary

`lazyRouteComponent` currently returns:

```ts
T[TKey] extends (props: infer TProps) => any
  ? AsyncRouteComponent<TProps>
  : never
```

tsgolint (and similar checkers that cannot instantiate that conditional) resolve the whole function as `error`. Call sites then fail type-aware lint (`no-unsafe-assignment`, `no-unsafe-call`) even though `tsc` is fine.

This keeps `T` / `TKey` named-export inference, moves the props conditional into a default `TProps` type parameter, and always returns `AsyncRouteComponent<TProps>`.

Applies to `@tanstack/react-router`, `@tanstack/solid-router`, and `@tanstack/vue-router`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type resolution for `lazyRouteComponent` across React, Solid, and Vue routers.
  * Prevented valid default and named component exports from being incorrectly typed as unavailable.
  * Preserved accurate component prop inference for lazy-loaded routes.

* **Documentation**
  * Added release metadata for the router package updates.

### DIFF
--- a/.changeset/lazy-route-component-return-type.md
+++ b/.changeset/lazy-route-component-return-type.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/react-router': patch
+'@tanstack/solid-router': patch
+'@tanstack/vue-router': patch
+---
+
+Return `AsyncRouteComponent` from `lazyRouteComponent` instead of a `never` conditional
+
+The previous `T[K] extends (props: infer P) => any ? AsyncRouteComponent<P> : never` return type is resolved as `error` by tsgolint (and similar checkers that cannot instantiate that conditional). Named-export inference is unchanged via a default `TProps` type parameter.
--- a/packages/react-router/src/lazyRouteComponent.tsx
+++ b/packages/react-router/src/lazyRouteComponent.tsx
@@ -16,12 +16,11 @@ import type { AsyncRouteComponent } from './route'
 export function lazyRouteComponent<
   T extends Record<string, any>,
   TKey extends keyof T = 'default',
+  TProps = T[TKey] extends (props: infer P) => any ? P : unknown,
 >(
   importer: () => Promise<T>,
   exportName?: TKey,
-): T[TKey] extends (props: infer TProps) => any
-  ? AsyncRouteComponent<TProps>
-  : never {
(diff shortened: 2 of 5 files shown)


rbalicki2 (approved):

MILLERMARRU (commented):
Checked the actual type change, not just the changeset description. This moves the `T[TKey] extends (props: infer P) => any ? AsyncRouteComponent<P> : never` conditional out of the return-type position into a default type parameter (`TProps = T[TKey] extends (props: infer P) => any ? P : unknown`), and the return type is now unconditionally `AsyncRouteComponent<TProps>`. That's exactly the kind of conditional tsgolint's narrower type-checker can't instantiate in return position (per the PR description), and moving it to a defaulted parameter sidesteps that without changing what `tsc` itself infers, confirmed by the new `lazyRouteComponent.test-d.tsx` (`expectTypeOf(Comp).toEqualTypeOf<AsyncRouteComponent<{ id: string }>>()` for both default and named export cases).

One behavioral question worth asking explicitly rather than assuming: before this change, a non-function `T[TKey]` resolved the whole return type to `never`. Now it resolves to `AsyncRouteComponent<unknown>` instead. In practice I don't think this weakens anything, since `never` is a bottom type and was already assignable to any downstream position (a route's `component` field, JSX, etc.) without erroring either way, so the `never` branch was never doing real guard duty at call sites, only in isolated contexts that check the type directly (like the new test file does with `.not.toEqualTypeOf<never>()`). Still, worth having a maintainer confirm nothing internal to the router (route matching, `component` prop typing) branches on `AsyncRouteComponent<T> extends never` or similar, since that's the one place this technically isn't type-identical to before.

Applied consistently across all three adapters (react/solid/vue-router) with the same signature shape, which is the right scope, this bug affects all three the same way since they share the same conditional-return pattern.

schiller-manuel:
so tsgolint cannot handle proper TS syntax?
is there an issue for that at tsgolint? sounds like they should fix it instead.

also, i think we have many more conditional return types.

> and similar checkers

which ones?

PR #8187, opened 2026-08-28
fix(react-start-client): render visible() through GenericHydrate

EunusHosen:
## 🎯 Changes

`visible()` in `@tanstack/react-start-client` is the only hydration strategy that carries its own renderer (`VisibleHydrate`) instead of composing the `start-client-core` strategy with `GenericHydrate` the way `idle()`, `interaction()`, `media()` and `condition()` do. That renderer's `HydrationBoundary` placed a `useEffect` after a conditional `use()`:

```ts
if (!g.r) reactUse(g.p)   // skipped on the replay, once the gate has opened
React.useEffect(() => { o?.() }, [o])
```

When a component suspends on `use()` and React later **replays** it (`renderWithHooksAgain`, rerender dispatcher), React only switches back to the mount dispatcher from inside `use()` itself (`useThenable`). With the call skipped on the replay, the effect registers against the update path of a fiber that never mounted and React throws `Update hook called on initial render` — its own source annotates that branch with *"reached when the component suspends, resumes, then renders an additional hook"*. During hydration that surfaces as `#422` (recoverable hydration error): React client-renders the boundary, so the deferred chunk is fetched immediately and nothing is deferred.

This PR makes `visible()` what its siblings already are:

```ts
withHydrationRenderer(coreVisible(options), GenericHydrate)
```

`GenericHydrate`'s `HydrationGate` has no hooks after `use()` (its `onHydrated` effect lives in a child), so the replay is safe. As a consequence `visible()` also gains the marker attributes, delegated intent handling, procedural `prefetch` and preserved fallback HTML the other strategies have. The `VisibleHydrate` export of the internal `hydration/visible` module goes away; it was never re-exported from the package entry.

**Reproduction (browser):** a `<Hydrate when={visible({ rootMargin: '400px' })}>` boundary whose marker starts beyond `viewport height + rootMargin`, then scrolled into view. Dev console: `Update hook called on initial render…` from `HydrationBoundary` (`hydration/visible.js`), wrapped in *"There was an error while hydrating but React was able to recover…"*; the split chunk is requested at once. With this change: no errors, the chunk is requested only when the marker enters the band, and the marker's `data-ts-hydrate-when` attribute is cleared after hydration. Verified in a TanStack Start app on React 19.2.8, in both `vite dev` and a production build.

**Tests:** adds a `Hydrate` test for `visible()` — dormant with the marker off screen, hydrates once the (stubbed) `IntersectionObserver` fires, `onRecoverableError` never called, marker attribute cleared, `onHydrated` once. Note for reviewers: the replay path is not reachable in the jsdom harness (React unwinds and retries there rather than replaying, with or without `startTransition`), so the crash itself was verified in the browser as above; the test pins the shared-renderer behaviour and the `visible`-specific wiring.

Ran locally in `packages/react-start-client`: `vitest` (28/28), `eslint` (0 errors; the 5 warnings pre-exist), `tsc`, `vite build`, `publint --strict` and `attw`.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved visible hydration reliability when content enters the viewport.
  * Prevented hydration failures that could trigger unnecessary client rendering and immediate deferred loading.

* **New Features**
  * Visible hydration now uses the same rendering behavior as other hydration strategies.
  * Off-screen content remains server-rendered and hydrates in place when it becomes visible.

* **Tests**
  * Added coverage confirming successful viewport-triggered hydration without recoverable errors.


schiller-manuel:
please create an issue instead of PRs

PR #8192, opened 2026-08-29
feat(solid): named single-flight source + loadFlightTarget, the cache-agnostic trigger

ryansolid:
**Requires solid-js / `@solidjs/web` 2.0.0-rc.6 (released) — the branch bumps the Start packages and e2e apps to rc.6 and floors the peer range there. Stacked on #8193 (merge that first): the retired `setupRouterSsrQueryIntegration` breaks on rc.6, so the e2e apps must be on the userland glue before this lands.**

## Summary

Solid's single-flight channel is multi-source: a mutation response carries a keyed envelope of per-cache slices, each routed to the consumer subscribed under its source id. Previously Start claimed the single unnamed slot on both halves — so any other cache wanting mutation-response data (e.g. solid-query, whose provider subscribes under `"sq"` in [TanStack/query#11326](https://github.com/TanStack/query/pull/11326)) displaced the router's flight data: whichever library registered last silently won.

This PR moves the router's flight data onto its own source id (`"tsr"`):

- **Server:** `collectSolidStartFlightData` registers additively via `registerFlightDataSource(SOLID_START_FLIGHT_SOURCE, ...)` instead of claiming the per-handler unnamed slot. A user-supplied `collectFlightData` option now keeps the unnamed slot to itself — its data folds *alongside* the router's rather than replacing it.
- **Client:** the flight consumer subscribes under the same id and receives exactly the router's slice of the keyed envelope.

Net effect: a Start app using solid-query gets router loader/match refresh **and** query-cache hydration from one mutation round trip, with no coordination between the libraries.

## `loadFlightTarget`: the trigger as a public primitive

The factoring behind multi-source flight is: **the router owns the trigger** (running the matched routes' data functions for the URL the client will show after the mutation), **each cache owns its slice** (what to extract from the loaded state). This PR makes the trigger public — `loadFlightTarget` in `@tanstack/solid-router/ssr/server`:

```ts
import { registerFlightDataSource } from '@solidjs/web/server-functions/server'
import { loadFlightTarget } from '@tanstack/solid-router/ssr/server'
import { FLIGHT_DATA_SOURCE, dehydrateSettled } from '@tanstack/solid-query'

registerFlightDataSource(FLIGHT_DATA_SOURCE, (event, outcome) => {
  if (!outcome.targetUrl) return undefined
  const queryClient = createQueryClient()
  return loadFlightTarget({
    router: createAppRouter(queryClient),
    event,
    outcome,
    collect: async () => {
      const state = await dehydrateSettled(queryClient)
      return state.queries.length > 0 ? state : undefined
    },
  })
})
```

It derives the flight request (the outcome's pre-digested `targetUrl`, the mutation's cookie effects already folded in), points the router at the target through a fresh memory history, runs `router.load()` inside the flight request-event scope, and hands the loaded router to `collect`. Errors are contained per Solid Router's own collector convention — flight data is an optimization, so a failure omits the slice rather than surfacing as a mutation error (core also contains per-source: [solidjs/solid@ec523607](https://github.com/solidjs/solid/commit/ec523607)).

Start's `tsr` collector consumes the same primitive, keeping only its bespoke halves: start-context acquisition, its serialized-redirect target resolution, and match-state extraction.

Known divergence from Solid Router's collector, deliberate for now: `outcome.revalidateKeys` scoping isn't honored — it comes from Solid's `X-Revalidate` convention, which TanStack mutations don't produce.

## Pre-landing checklist (rc.6 shipped — all done)

- [x] Delete the feature detection in `solid-rpc-flight-client.ts` (`getFlightDataSourceIds` probe) — subscribes directly under `SOLID_START_FLIGHT_SOURCE`.
- [x] Delete the legacy fallback in `server-functions-handler.ts` (the unnamed `collectFlightData`-slot assignment) — registers via `registerFlightDataSource` unconditionally.
- [x] Bump the `@solidjs/web` peer floor — floored at rc.6 rather than rc.5, since rc.5's settle-walk regression breaks the solid-query pairing this protocol exists for.
- [x] Bump Start packages and solid-start e2e apps to solid rc.6 (manifests + lockfile; the branch previously had a manifest/lockfile mismatch).

## Compatibility

- The named-source protocol requires `@solidjs/web` 2.0.0-rc.6+ on both halves; client and server resolve the same install, so the halves cannot disagree.
- The existing `isSolidStartFlightData` shape guard stays — redundant-but-harmless under named sources.

## Verification

- `nx build @tanstack/solid-router` and `nx build @tanstack/solid-start` green, `publint --strict` + `attw` pass.
- The composed collector runs end-to-end in solidjs/templates' `fullstack-tanstack` (solidjs/templates#287), which uses exactly the `loadFlightTarget` + `dehydrateSettled` composition above.
- The single-flight e2e (`e2e/solid-start/server-functions`) runs on rc.6 with the named path active (`X-Single-Flight: tsr`): 29 passed, 1 skipped, locally — on top of #8193's userland query glue.


ryansolid:
Pushed two additions: `refactor(solid-start): require named flight sources, drop the pre-rc.5 fallback` (the runtime now assumes solid-js 2.0.0-rc.5+, where two-argument `subscribeFlightData` and `registerFlightDataSource` exist) and the e2e expectation update for the named `X-Single-Flight: tsr` header. This PR is the base of a stack: #8213 (native SSR match transfer) and #8214 (Solid-owned transport/payload) build on it.

brenelz (approved):

brenelz:
Benchmarks are failing with

```
 TypeError: registerFlightDataSource is not a function
     ❯ dist/server/server.js:6938:25
        6936|  endpoint: "/_serverFn/"
        6937| });
        6938| registerFlightDataSource("tsr", collectSolidStartFlightData);
           |                         ^
        6939| var solidNoJSHandler = createNoJSHandler();
        6940| async function handleSolidServerFunctionRequest(request, options = {})…
        ````

brenelz:
And some e2e with

> [MISSING_EXPORT] "registerFlightDataSource" is not exported by "../../../node_modules/.pnpm/@solidjs+web@2.0.0-rc.4_solid-js@2.0.0-rc.4/node_modules/@solidjs/web/server-functions/dist/server.js".
   ╭─[ node_modules/.nitro/vite/services/ssr/index.js:8:114 ]

PR #8193, opened 2026-08-29
refactor(solid): retire solid-router-ssr-query — native channels carry the Router + Query pairing

ryansolid:
**Draft — for discussion.** Works against published packages today (`@tanstack/solid-query@6.0.0-rc.1`, `solid-js@2.0.0-rc.4`); all three converted e2e suites pass.

## Summary

On the Solid v2 line, `@tanstack/solid-router-ssr-query` bundles three unrelated things, and each now has a better home:

1. **SSR transport (the bulk of the package — `router-ssr-query-core`'s dehydrate + query stream + client hydrate).** Native to solid-query v6: `QueryClientProvider` serializes the request's cache into Solid's hydration registry (content-addressed by query hash, promise-valued at fetch-dispatch so pending queries stream, covering never-mounted loader prefetches) and primes the client cache from it. **Running the package alongside it ships every query payload twice** — which is what the e2e apps were doing: once through the router's dehydration stream, once through the hydration registry.
2. **`wrapQueryClient`.** A JSX expression — the router's own `Wrap` option:

```tsx
Wrap: (props) => (
  <QueryClientProvider client={queryClient}>{props.children}</QueryClientProvider>
)
```

3. **`handleRedirects`.** Runtime navigation glue, not an SSR concern: cache-driven fetches (mount fetches, background refetches, mutations) run outside the router, so a `queryFn`/`mutationFn` throwing `redirect()` needs handing to `router.navigate`. That's a small userland composition of public APIs on both sides — `isRedirect`/`resolveRedirect` from the router, `config.onError` on the query/mutation caches (see `routeCacheRedirects` in the converted apps). Loader-driven redirects were never the package's: they propagate through the loader and the router handles them on both sides.

This PR converts the three Solid Start e2e apps (`basic-solid-query`, `server-functions`, `server-routes`) to the composition, removes the dependency, and marks the package deprecated for the v2 line in its README.

## Verification

- `basic-solid-query`: 6/6 — SSR hydration, nested layouts, suspense transitions, query-preserving navigation, all on the native transport only.
- `server-functions`: 29/29 — **including `redirect-test` (mount-time query redirect via the `onError` glue) and `redirect-test-ssr` (render-time query redirect resolving through the stream handler)**, demonstrating `handleRedirects` is fully replaced.
- `server-routes`: 2/2.

## Notes

- The same composition is the reference pattern in solidjs/templates' `fullstack-tanstack` (bare vite + TanStack Router + Query, no Start — solidjs/templates#287), which is the point: none of it is Start-specific.
- Complementary to [#8192](https://github.com/TanStack/router/pull/8192) (named single-flight sources + `loadFlightTarget`): this PR is the SSR/read side going native, that one is the mutation/write side. Together they remove the need for any `*-ssr-query` package on Solid.
- `docs/router/integrations/query.md`'s Solid tab still shows the old setup — happy to rewrite it once there's agreement on direction.
- React/Vue packages untouched; the `router-ssr-query-core` transport remains correct for frameworks without a native serialization channel.

### DIFF
--- a/.changeset/retire-solid-router-ssr-query.md
+++ b/.changeset/retire-solid-router-ssr-query.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/solid-router-ssr-query': patch
+---
+
+Deprecated: Solid Query's `QueryClientProvider` now carries the Router + Query SSR pairing natively (registry hydration, named single-flight sources, redirect handling via userland glue), so this integration package is no longer needed. See the README for the migration.
--- a/e2e/solid-start/basic-solid-query/package.json
+++ b/e2e/solid-start/basic-solid-query/package.json
@@ -17,7 +17,6 @@
     "@tanstack/solid-query-devtools": "^6.0.0-rc.1",
     "@tanstack/solid-router": "workspace:^",
     "@tanstack/solid-router-devtools": "workspace:^",
-    "@tanstack/solid-router-ssr-query": "workspace:^",
     "@tanstack/solid-start": "workspace:^",
     "redaxios": "^0.5.1",
     "solid-js": "^2.0.0-rc.4",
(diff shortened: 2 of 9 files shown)


brenelz:
I'm on board with this. Would be one less package to maintain.

> Running the package alongside it ships every query payload twice 

That doesn't seem ideal

birkskyum (approved):

brenelz (approved):

PR #8213, opened 2026-09-02
feat(solid-router): native SSR match transfer, hydration-claiming boot, provider-owned dispatch

ryansolid:
## Summary

Phase 1 of the Solid-native SSR plan (RFC included in the diff: `RFC-solid-native-ssr.md`). Replaces the React-inherited SSR bootstrapping in the Solid adapter with Solid's own hydration machinery:

- **Registry match transfer**: the server serializes router match state (`loaderData`, status, `__beforeLoadContext`) into Solid's hydration registry under `tsr:` keys — the same channel `solid-query` uses — instead of relying on the `$_TSR` script for match data. Deferred `loaderData` fields stream natively through seroval promise serialization.
- **Hydration-claiming boot**: the client primes the router from the registry in the `Router` constructor, before `hydrate()` runs. Matches are committed outside the hydration render's owned scope, so downstream apps no longer need boot-time `router.load()` workarounds (e.g. the `bootLoad` pattern) or prefetch pausing.
- **Provider-owned server dispatch**: `RouterProvider` kicks off `router.load()` on the server and parks the render on it via an async memo, preserving blocking loader semantics through Solid's async SSR (`renderToStream`) with no explicit `await router.load()` in entry files.

Includes a protocol-less external-SSR repro harness (`packages/solid-router/repro-external-ssr/`) that round-trips server render → hydration in jsdom and asserts loader single-run, node claiming, deferred streaming order, and interactivity.

Stacked on #8192 (named single-flight sources) — review the last 4 commits.

## Test plan

- [x] `@tanstack/solid-router` unit + server-mode suites
- [x] External-SSR repro harness (streamed chunks, deferred loaderData, zero loader re-runs after hydrate)
- [x] Solid Start e2e suites (basic, server-functions, basic-solid-query among others)


Sheraff:
just a small FYI: we try to use `isServer ?? router.isServer` instead of just `isServer` so that in `development` mode, `router.isServer` can define the value (in production, `isServer` is always defined `true` or `false`, so this gets dead-code eliminated anyway)

This helps people who instantiate a router in their unit tests

brenelz (approved):

PR #8215, opened 2026-09-02
fix(router-utils): clone cached nodes to release Babel's traversal cache

sverrejoh:
## 🎯 Changes

`@babel/traverse` keys its `NodePath` and `Scope` caches on node identity using a
WeakMap. The Start compiler keeps a reference to the initializer node in its module
cache for the life of the dev server, so that entry is never collected, and neither
is the whole file's traversal graph hanging off it.

Cache a cloned node instead, to prevent the original staying alive as a WeakMap key.
The only readers look at the node type and `callee`/`object`/`property`/`name`, so a
clone is equivalent for them. Build output was byte-identical across every run.

This gives 11.6% less peak dev-server memory in our app.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compiler memory management by preventing cached module information from retaining source-file traversal data.
  * Preserved module initializer information using detached copies, reducing the risk of unnecessary memory retention during compilation.

### DIFF
--- a/.changeset/start-compiler-module-info-ast-retention.md
+++ b/.changeset/start-compiler-module-info-ast-retention.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-utils': patch
+---
+
+Stop the Start compiler's module cache from retaining parsed ASTs. `extractModuleInfoFromAst` now stores a detached clone of each binding's initializer, so a cached module no longer keeps the `@babel/traverse` `NodePath` and `Scope` graph of the file it came from reachable.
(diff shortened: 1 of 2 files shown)


schiller-manuel:
cloning is compute intensive AFAIK. this needs proper benchmarking

sverrejoh:
@schiller-manuel You're right, cloning is slower. I have this alternative implementation #8296, which creates a descriptor and caches that instead. This is an order of magnitude cheaper than cloning.

Issue #7218, opened 2026-04-17
instanceof RunnableDevEnvironment fails with vite-plus alias


ultra2207:
### Which project does this relate to?

Router

### Describe the bug

IF you do vp dev the error occurs. It says Cannot get /. Note that using npm run dev or bun run dev which runs vite under the hood works fine.

The fix in https://github.com/TanStack/router/issues/6982#issuecomment-4269260919 fixes the issue.

### Your Example Website or App

https://github.com/ultra2207/tanstack_test

### Steps to Reproduce the Bug or Issue

1. git clone into the reproducer
2. Install Vite-plus
3. Run vp dev and open the url

### Expected behavior

it will show "Cannot GET /"

### Platform

When using Vite-plus cli.

davyleotte-eng:
Muito bom

schiller-manuel:
please provide a reproducer that can be cloned directly without having to execute any more installation. also try different package managers to find out if it depends on that

ultra2207:
you can clone what i linked above directly, but yeah it is an issue with vite+ specifically. Using npm or bun to start thr vite server does fine. However since the fix for the issue is in tanstack router code and not in vite-plus code i opened the issue here.

Reproduction simply involves git cloning the above, installing vite-plus, cd into it and then do vp run.

schiller-manuel:
why dont you install vite-plus then in the repo? it might matter which package manager you use for that step.

ultra2207:
the regular vp install using irm https://vite.plus/ps1 | iex installs globally, not in the repo. Also just putting it in the repo wont work as its different for windows/mac/linux. Not to mention adding to path which the command does automatically.

ultra2207:
yo

roll:
It seems to be we are running into the same, a mostly empty Tanstack Start app does work with `vite dev` and 404 on any route with `vp dev` (pnpm managed)

kevinkod:
Corroborating data point from a different setup (npm workspaces, not the OP's global vite-plus install), in case it helps move this out of "information needed".

**Environment**
- npm workspaces monorepo (not pnpm/bun), `vite` aliased to vite-plus via root `overrides`: `"vite": "npm:@voidzero-dev/vite-plus-core@latest"`
- `@tanstack/react-start` SSR app, `vp dev`
[...]

PR #8235, opened 2026-09-04
fix(start-plugin-core): install the dev server middleware when vite is duplicated, and warn when it is skipped

### LINKED ISSUE #7614: TanStack Start dev server returns "Cannot GET /" on Vite 8 (SSR middleware silently skipped, then hard-throws on opt-in)

Router

## Summary

On Vite 8 (including but not limited to **Vite+**, the new Vite-based toolchain by VoidZero), the TanStack Start dev server plugin fails to register its SSR middleware. The result is the request falls through to the underlying Connect/Express default and the browser sees a plain `404 Not Found` with body `Cannot GET /`. The failure is silent — no log line, no error overlay.

If the user opts in via `tanstackStart({ vite: { installDevServerMiddleware: true } })` to bypass the silent skip, the next line throws:

```
Error: cannot install vite dev server middleware for TanStack Start since the SSR environment is not a RunnableDevEnvironment
```

…even though the SSR environment **is** in fact a `RunnableDevEnvironment`. Two independent bugs in `@tanstack/start-plugin-core/dist/esm/vite/dev-server-plugin/plugin.js` are responsible. This issue proposes a code-level fix in the plugin (not a `node_modules` patch).

## Environment

| Package | Version |
|---|---|
| `@tanstack/react-start` | `1.168.25` |
| `@tanstack/start-plugin-core` | `1.171.17` |
| `@tanstack/router-core` | `1.171.13` |
| `@voidzero-dev/vite-plus-core` (aliased `vite`) | `0.1.24` |
| Vite (bundled inside vite-plus-core) | `8.0.14` |
| Node | `>=20` |
| Package manager | pnpm (workspace) |

The trigger is Vite 8 (the rest is incidental). Any project that ends up on Vite 8 with a monorepo / dual-vite-resolution layout will hit this. The user-reported case is Vite+, but the underlying cause is generic to Vite 8.

## Reproduction

Minimal monorepo layout:

```
apps/web/
  package.json          // "vite": "^8.0.0" in devDeps, no extra config
  vite.config.ts        // plugins: [tanstackStart(), react()]
  src/
    router.tsx          // exports getRouter() with createRouter + routeTree
    routes/
      __root.tsx        // createRootRouteWithContext
      index.tsx         // createFileRoute("/")
src/                    // no other entry files
```

```ts
// apps/web/vite.config.ts
import { defineConfig } from "vite-plus"
import { tanstackStart } from "@tanstack/react-start/plugin/vite"
import viteReact from "@vitejs/plugin-react"

export default defineConfig({
  plugins: [tanstackStart(), viteReact()],
})
```

```bash
cd apps/web
vp run dev      # or: vite dev
# ➜  Local:   http://localhost:3000/
```

```bash
curl -i http://localhost:3000/
# HTTP/1.1 404 Not Found
# Content-Type: text/html; charset=utf-8
# Content-Length: 139
#
# <!DOCTYPE html>
# <html lang="en"><head><meta charset="utf-8"><title>Error</title></head>
# <body><pre>Cannot GET /</pre></body></html>
```

No error in the dev server log. No Vite error overlay. The TanStack Start server entry is never invoked.

## Expected vs Actual

| | Expected | Actual |
|---|---|---|
| `curl http://localhost:3000/` | `200 OK` with SSR-rendered HTML of `routes/index.tsx` | `404 Not Found`, body `Cannot GET /` |
| Server log | `tanstack-start: dev server listening` (or similar) — middleware registered | No TanStack Start log line; only Vite's default request log |

## Root cause

Both bugs live in **`@tanstack/start-plugin-core/dist/esm/vite/dev-server-plugin/plugin.js`** (the `configureServer` post-hook that registers the catch-all SSR middleware).

### Bug 1 — Overly broad early-return condition

```js
// lines 55–59 (current)
if (installMiddleware == void 0) {
  if (viteDevServer.config.server.middlewareMode) return;
  if (!isRunnableDevEnvironment(serverEnv) || "dispatchFetch" in serverEnv) return;
}
```

The intent of `"dispatchFetch" in serverEnv` was to opt out for the new `FetchableDevEnvironment`, which talks to the server via `dispatchFetch` instead of `runner.import`. But in Vite 8, **`RunnableDevEnvironment` also has a `dispatchFetch` method** — it inherits it from the `DevEnvironment` base to satisfy the `Environment` interface contract. The `RunnableDevEnvironment` class declaration in `vite/dist/node/index.d.ts`:

```ts
declare class RunnableDevEnvironment extends DevEnvironment {
  // ...
  dispatchFetch(request: Request): Promise<Response>;
}
```

So `"dispatchFetch" in serverEnv` is `true` for *every* dev environment, and the post-hook unconditionally returns. The middleware is never registered.

Verified by adding a `configureServer` debug hook in the user's `vite.config.ts`:

```
=== DEBUG ENV ===
typeof ssr: object
constructor.name: RunnableDevEnvironment
has dispatchFetch: false     ← ← ← Runnable, no dispatchFetch on the INSTANCE
has runner: true
server.middlewareMode: false
=== END DEBUG ===
```

(The `"dispatchFetch" in serverEnv` check was matching the *property on the class* at plugin load time, not the instance the dev server actually created.)

### Bug 2 — `instanceof` across ESM module boundaries

```js
// line 65 (current)
if (!isRunnableDevEnvironment(serverEnv)) throw new Error(
  "cannot install vite dev server middleware for TanStack Start since the SSR environment is not a RunnableDevEnvironment"
);
```

`isRunnableDevEnvironment(env)` is implemented as `env instanceof RunnableDevEnvironment`, where `RunnableDevEnvironment` is imported by the plugin from `"vite"`. In a monorepo (or any setup where vite is aliased / re-exported under multiple module specifiers), Node's ESM module cache can end up with **two distinct `RunnableDevEnvironment` class identities**:

- The one **Vite+**'s bundled Vite uses to construct the SSR env (`new RunnableDevEnvironment(...)` in `vite-plus-core/dist/vite/node/chunks/node.js`).
- The one **`@tanstack/start-plugin-core`** imported at module load time, bound inside the closure of `isRunnableDevEnvironment`.

The two classes share the same source but live in different module-graph entries. `env instanceof RunnableDevEnvironment` is therefore `false` even though `env.constructor.name === "RunnableDevEnvironment"`.

This is the failure that becomes visible only after the user sets `installDevServerMiddleware: true` to skip Bug 1's early-return:

```
Error: cannot install vite dev server middleware for TanStack Start since
the SSR environment is not a RunnableDevEnvironment
```

Verified the diagnosis end-to-end by registering a custom `configureServer` post-hook that does the **same** `serverRunner.import("virtual:tanstack-start-server-entry")` call as the plugin would, with **no `instanceof` check** — the page renders correctly (`200 OK`, body contains the expected SSR markup). This isolates the failure to the plugin's gating logic, not to the underlying import / SSR pipeline.

## Suggested fix

In `@tanstack/start-plugin-core/dist/esm/vite/dev-server-plugin/plugin.js`, replace the two identity-based guards with a **feature check** (the middleware only ever uses `serverEnv.runner.import`, so check exactly that). This avoids both the `instanceof` cross-module problem and the dispatchFetch false positive:

```diff
--- a/src/vite/dev-server-plugin/plugin.ts
+++ b/src/vite/dev-server-plugin/plugin.ts
@@
+// True iff `env` exposes the runner.import API this middleware relies on.
+// Feature check (not instanceof) to survive ESM module-graph boundaries
+// in monorepos and to avoid false negatives on Vite 8 where
+// RunnableDevEnvironment also exposes a `dispatchFetch` method.
+function supportsRunnerImport(env: unknown): env is { runner: { import: (id: string) => Promise<unknown> } } {
+  return (
+    typeof env === "object" &&
+    env !== null &&
+    "runner" in env &&
+    typeof (env as any).runner === "object" &&
+    (env as any).runner !== null &&
+    typeof (env as any).runner.import === "function"
+  );
+}
+
 return [{
   name: "tanstack-start-core:dev-server",
   /* ... */
   configureServer(viteDevServer) {
     /* ... */
     return () => {
       const serverEnv = viteDevServer.environments[VITE_ENVIRONMENT_NAMES.server];
       if (!serverEnv) throw new Error(`Server environment ${VITE_ENVIRONMENT_NAMES.server} not found`);
       const clientEnv = viteDevServer.environments[VITE_ENVIRONMENT_NAMES.client];
       const installMiddleware = installDevServerMiddleware;
       if (installMiddleware === false) return;
       if (installMiddleware == void 0) {
         if (viteDevServer.config.server.middlewareMode) return;
-        if (!isRunnableDevEnvironment(serverEnv) || "dispatchFetch" in serverEnv) return;
+        if (!supportsRunnerImport(serverEnv)) return;
       }
-      if (!isRunnableDevEnvironment(serverEnv)) throw new Error("cannot install vite dev server middleware for TanStack Start since the SSR environment is not a RunnableDevEnvironment");
+      if (!supportsRunnerImport(serverEnv)) throw new Error(
+        "cannot install vite dev server middleware for TanStack Start: the SSR environment does not expose runner.import(). " +
+        "This middleware requires a RunnableDevEnvironment (or equivalent)."
+      );
       viteDevServer.middlewares.use(async (req, res) => { /* unchanged */ });
     };
   }
 }];
```

The `import { isRunnableDevEnvironment } from "vite"` line can be removed (it's only used in the two checks above). Same for the `import { ... } from "vite"` if it was only there for that symbol.

### Why this is the right shape

- **Feature check, not type check.** The middleware only ever calls `serverEnv.runner.import(ENTRY_POINTS.server)`. The only thing we need to know is "can I call `.import` on it?". That's a property check, not an identity check.
- **Cross-module safe.** A feature check is robust to class identity mismatches across ESM module-graph boundaries (the root cause of Bug 2).
- **Vite 8 safe.** It does not look at `dispatchFetch` at all, so the `RunnableDevEnvironment`-now-inherits-`dispatchFetch` behavior in Vite 8 (Bug 1) is irrelevant.
- **Forward compatible.** If Vite later adds a third dev-env flavor that also exposes `runner.import`, this code keeps working.

### Suggested follow-up

Once the check is fixed, the `installDevServerMiddleware` option loses most of its value (it was only ever a workaround for this exact bug). Two options:

1. **Deprecate it** in a minor release, document the migration, remove in the next major.
2. **Keep it** as a "force" override that skips the feature check entirely (useful for custom SSR transports that do their own thing outside the Vite dev server). This is essentially the current behavior — just fix the bug behind it.

I'd vote for (1) + (2) combined: keep the option but redefine its semantics as "skip the check", deprecate the old behavior, document in the changelog.

## Workaround (until fixed)

Pin Vite to v7. Vite 7's `RunnableDevEnvironment` does not expose `dispatchFetch` (the method was added in Vite 8), so Bug 1's check is false and the middleware is installed normally. Bug 2 may still bite in monorepos on Vite 7 — switching to a single Vite resolution (one Vite in the lockfile, no alias) is enough to dodge it.

For projects that *must* stay on Vite 8 / Vite+, the only currently-available workaround is patching `node_modules/@tanstack/start-plugin-core/dist/esm/vite/dev-server-plugin/plugin.js`. I would **not** recommend that for production.

## Related

- `pnpm` override setup: `vite: npm:@voidzero-dev/vite-plus-core@latest` + `overrides: { vite: "catalog:" }` to ensure a single Vite resolution across the workspace. This is the recommended configuration for Vite+ monorepos and is what surfaces the cross-module `instanceof` issue.
- `@tanstack/router-core` v1.171.x ships a `tanstack-router/skills/router-core/SKILL.md` family. A skill for "TanStack Start dev server plugin" covering the environment-type contract would be a good add.

## Checklist

- [x] Reproduction steps
- [x] Environment versions
- [x] Root cause with `file:line` references
- [x] Source-level fix (no node_modules patching)
- [x] Workaround for users who can't wait
- [x] Forward-compatibility note

null

null

### Expected behavior

null

### Screenshots or Videos

_No response_

### Platform

- Bundler: [e.g. vite]

### Additional context

_No response_

audi0417:
## 🎯 Changes

Fixes #7614. Also the root cause of #7218, which has been on `information needed` since April for want of a reproducer that does not require installing Vite+ - the reproduction below needs only two plain npm copies of vite.

The Start dev server can stop server rendering with no diagnostic at all: the browser gets `Cannot GET /`, and the Start plugin logs nothing. Two separate things combine to produce that.

`isRunnableDevEnvironment()` is a bare `instanceof RunnableDevEnvironment` against the copy of vite that imports it. When the dev server is created by a *different* copy of vite - Vite+, or a workspace that ends up with a duplicated vite install - the check returns `false` for an environment that is structurally runnable. `devServerPlugin` then takes the "another plugin must already serve SSR" branch and returns without installing its middleware, so the request falls through to Vite's default handler and 404s.

Opting out of that branch with `vite: { installDevServerMiddleware: true }` does not help, because the next check throws `cannot install vite dev server middleware for TanStack Start since the SSR environment is not a RunnableDevEnvironment` - for an environment that is one.

This PR does two things:

- Adds `isRunnableEnvironment()`, which falls back to a structural `'runner' in environment` check when the `instanceof` helper says no. That is the same reasoning the neighbouring `dispatchFetch` check already uses to avoid `isFetchableDevEnvironment` for nitro. `runner` is a prototype getter on `RunnableDevEnvironment`, so `in` resolves it without instantiating the module runner.
- Splits the combined skip condition, so "nitro already serves SSR" (silent, intended) stays distinct from "nothing here can server render" (now a warning naming `installDevServerMiddleware`). The second case previously produced no output at all, which is what makes this failure so hard to diagnose from the outside.

### Reproduction

`examples/react/start-basic` on Vite 8.2.2 against the published `@tanstack/start-plugin-core@1.171.39`, with a second vite copy (8.0.14, installed as an npm alias) creating the dev server so that the server and the plugin resolve different copies. No Vite+ involved, and `middlewareMode` is `false` throughout, so this is not the intentional middleware-mode skip. This matches the finding in #7218 that the `@tanstack/*` version is not the variable.

| Setup | `GET /` before | `GET /` after |
| --- | --- | --- |
| One vite copy, with nitro | 200, 3597 bytes | 200, 3597 bytes |
| One vite copy, without nitro | 200, 3597 bytes | 200, 3597 bytes |
| Two vite copies | **404 `Cannot GET /`**, no plugin output | 200, 3597 bytes |
| Two vite copies + `installDevServerMiddleware: true` | **throws on startup** | 200, 3597 bytes |

The guard itself, on the same environment object:

```
vite (copy that owns the server) isRunnableDevEnvironment: true
vite (copy the plugin resolves)  isRunnableDevEnvironment: false
typeof ssr.runner: object   dispatchFetch present: false   ctor: RunnableDevEnvironment
```

### What this does not cover

I reproduced the duplicated-vite condition with two npm copies of vite rather than by running Vite+ itself, so this is verified against the mechanism - `instanceof` across module instances - rather than against the `vp dev` CLI specifically. Happy to confirm against Vite+ before this lands if you would rather see that.

The new warning fires for any SSR environment that is neither runnable nor exposes `dispatchFetch`. A plugin that serves SSR some other way would now see a warning where it previously saw nothing; `installDevServerMiddleware: false` silences it, which the message says. Say the word if you would rather that case stayed quiet.

### Tests

`packages/start-plugin-core/tests/vite/dev-server-plugin.test.ts` covers the four branches: a runnable environment from another vite instance installs the middleware and does not throw when opted into, an environment that cannot render warns instead of skipping silently, and an environment that another plugin already serves stays silent. The first three fail on `main`.

Ran `eslint`, `tsc`, and `vitest` for `@tanstack/start-plugin-core` (32 files, 516 tests, no type errors). The 5 pre-existing eslint errors in `handleCreateServerFn.ts` and `manifestBuilder.ts` are present on a clean tree and untouched here.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved development server middleware installation when multiple Vite copies are present.
  - Added support for detecting compatible server-rendering environments across Vite versions or instances.
  - Added clearer warnings when no server-renderable environment is available.
  - Avoided unnecessary warnings when another plugin already handles request processing.

- **Tests**
  - Added coverage for middleware installation, opt-in behavior, warnings, and compatibility with alternate Vite environments.

### DIFF
--- a/.changeset/dev-server-middleware-duplicated-vite.md
+++ b/.changeset/dev-server-middleware-duplicated-vite.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/start-plugin-core': patch
+---
+
+Detect a runnable SSR environment structurally so the dev server middleware is still installed when the plugin and the dev server resolve different copies of vite, and warn instead of silently skipping the middleware when no environment can server render. Previously such a setup fell through to Vite's default handler and served `Cannot GET /` with no diagnostic, and opting in via `installDevServerMiddleware: true` threw `the SSR environment is not a RunnableDevEnvironment` for an environment that was in fact runnable.
--- a/packages/start-plugin-core/src/vite/dev-server-plugin/plugin.ts
+++ b/packages/start-plugin-core/src/vite/dev-server-plugin/plugin.ts
@@ -6,7 +6,12 @@ import {
   collectDevStyles,
   normalizeCssModuleCacheKey,
 } from './dev-styles'
-import type { Connect, DevEnvironment, PluginOption } from 'vite'
+import type {
+  Connect,
+  DevEnvironment,
+  PluginOption,
+  RunnableDevEnvironment,
+} from 'vite'
 import type { GetConfigFn } from '../../types'
 
 export function devServerPlugin({
@@ -144,16 +149,20 @@ export function devServerPlugin({
             }
 
             // do not install middleware if SSR env in case another plugin already did
-            if (
-              !isRunnableDevEnvironment(serverEnv) ||
-              // do not check via `isFetchableDevEnvironment` since nitro does implement the `FetchableDevEnvironment` interface but not via inheritance (which this helper checks)
-              'dispatchFetch' in serverEnv
-            ) {
+            // do not check via `isFetchableDevEnvironment` since nitro does implement the `FetchableDevEnvironment` interface but not via inheritance (which this helper checks)
+            if ('dispatchFetch' in serverEnv) {
+              return
+            }
+
+            if (!isRunnableEnvironment(serverEnv)) {
+              viteDevServer.config.logger.warn(
+                `[tanstack-start] the "${VITE_ENVIRONMENT_NAMES.server}" environment is neither runnable nor fetchable, so the dev server middleware was not installed and requests will not be server rendered. If another plugin serves them, set \`installDevServerMiddleware: false\` to silence this warning.`,
+              )
               return
             }
           }
 
-          if (!isRunnableDevEnvironment(serverEnv)) {
+          if (!isRunnableEnvironment(serverEnv)) {
             throw new Error(
               'cannot install vite dev server middleware for TanStack Start since the SSR environment is not a RunnableDevEnvironment',
             )
@@ -253,6 +262,22 @@ export function devServerPlugin({
       },
     },
   ]
+}
+
+/**
+ * `isRunnableDevEnvironment` is an `instanceof` check against the copy of vite this
+ * package resolves. When the dev server is created by a different copy of vite - Vite+,
+ * or a workspace that ends up with a duplicated vite install - that check is false even
+ * though the environment is a runnable one, so fall back to a structural check.
+ * This is the same reason the `dispatchFetch` check above does not go through
+ * `isFetchableDevEnvironment`.
+ * @param environment
+ * @returns
+ */
+function isRunnableEnvironment(
+  environment: DevEnvironment,
+): environment is RunnableDevEnvironment {
+  return isRunnableDevEnvironment(environment) || 'runner' in environment
 }
 
 /**
--- a/packages/start-plugin-core/tests/vite/dev-server-plugin.test.ts
+++ b/packages/start-plugin-core/tests/vite/dev-server-plugin.test.ts
@@ -1,0 +1,85 @@
+import { describe, expect, test, vi } from 'vitest'
+import { devServerPlugin } from '../../src/vite/dev-server-plugin/plugin'
+import { VITE_ENVIRONMENT_NAMES } from '../../src/constants'
+import type { GetConfigFn } from '../../src/types'
+
+/**
+ * A runnable SSR environment created by a *different* copy of vite than the one
+ * this plugin imports, as happens with Vite+ or a duplicated vite install.
+ * It is structurally a RunnableDevEnvironment, but `instanceof` against our copy
+ * of vite is false, so `isRunnableDevEnvironment()` rejects it.
+ */
+class ForeignRunnableDevEnvironment {
+  get runner() {
+    return { import: () => Promise.resolve({}) }
+  }
+}
+
+function runDevServerPlugin({
+  serverEnv,
+  installDevServerMiddleware,
+}: {
+  serverEnv: unknown
+  installDevServerMiddleware?: boolean
+}) {
+  const warn = vi.fn()
+  const use = vi.fn()
+  const viteDevServer = {
+    config: {
+      server: { middlewareMode: false },
+      experimental: { bundledDev: false },
+      logger: { warn },
+    },
+    environments: { [VITE_ENVIRONMENT_NAMES.server]: serverEnv },
+    middlewares: { use },
+    watcher: { add: vi.fn() },
+  }
+
+  const [plugin] = devServerPlugin({
+    getConfig: (() => ({})) as unknown as GetConfigFn,
+    devSsrStylesEnabled: false,
+    installDevServerMiddleware,
+  }) as Array<any>
+
+  const postHook = plugin.configureServer(viteDevServer as any)
+  postHook?.()
+
+  return { warn, use }
+}
+
+describe('dev server plugin middleware installation', () => {
+  test('installs the middleware for a runnable SSR environment from another vite instance', () => {
+    const { use, warn } = runDevServerPlugin({
+      serverEnv: new ForeignRunnableDevEnvironment(),
+    })
+
+    expect(use).toHaveBeenCalledTimes(1)
+    expect(warn).not.toHaveBeenCalled()
+  })
+
+  test('does not throw when opting in with a runnable environment from another vite instance', () => {
+    expect(() =>
+      runDevServerPlugin({
+        serverEnv: new ForeignRunnableDevEnvironment(),
+        installDevServerMiddleware: true,
+      }),
+    ).not.toThrow()
+  })
+
+  test('warns instead of silently skipping when the SSR environment cannot render', () => {
+    const { use, warn } = runDevServerPlugin({ serverEnv: {} })
+
+    expect(use).not.toHaveBeenCalled()
+    expect(warn).toHaveBeenCalledTimes(1)
+    expect(warn.mock.calls[0]?.[0]).toContain('installDevServerMiddleware')
+  })
+
+  test('stays silent when another plugin already serves the SSR environment', () => {
+    const { use, warn } = runDevServerPlugin({
+      serverEnv: { dispatchFetch: () => Promise.resolve(new Response()) },
+    })
+
+    expect(use).not.toHaveBeenCalled()
+    expect(warn).not.toHaveBeenCalled()
+  })
+})


Sheraff:
shouldn't the fix be on the `vite` side if `isRunnableDevEnvironment` can fail like this?

audi0417:
You are right that the defect is in `isRunnableDevEnvironment`. It is a bare `env instanceof RunnableDevEnvironment` in both versions I checked (8.0.14 and 8.2.2), so it answers "did my copy of vite construct this" rather than "is this environment runnable". A brand check upstream would fix it properly for everyone, and I could not find an existing vite issue for it - happy to open one and link it here.

I put the fallback here because this function already made the same call once. Two lines up, the `dispatchFetch` check is written as a property test, with a comment noting that nitro implements `FetchableDevEnvironment` structurally rather than by inheritance, "which this helper checks". So this follows an existing precedent in the file rather than introducing a new one. The other consideration is the peer range: this package supports `vite >=7.0.0`, so an upstream fix only reaches a project once it is on whatever version carries it.

The warning is separable from that question. Today an SSR environment that cannot render produces a 404 with no log line at all, which is the part that makes this hard to diagnose from the outside, and that holds whether or not the guard is fixed upstream.

Happy to go whichever way you prefer: keep the fallback as an interim measure, or narrow this to just the warning and let the guard be fixed in vite.

Sheraff:
please speak in human to me, my brain is exhausted from AI-speak

audi0417:
Sorry about that. English isn't my first language, so I used ChatGPT to help organize my thoughts. I didn't mean to make the comment sound so AI-ish.

I agree that the issue should ultimately be fixed in Vite. I added the fallback here so TanStack Start can still work with duplicated Vite versions before that fix lands.

If you'd prefer, I can remove the fallback, keep the warning, and open a Vite issue for the `instanceof` problem.

Sheraff:
> Sorry about that. English isn't my first language, so I used ChatGPT to help organize my thoughts. I didn't mean to make the comment sound so AI-ish.
> 
> I agree that the issue should ultimately be fixed in Vite. I added the fallback here so TanStack Start can still work with duplicated Vite versions before that fix lands.
> 
> If you'd prefer, I can remove the fallback, keep the warning, and open a Vite issue for the `instanceof` problem.

No worries, it's not my first language either, but now i understand you!
opening an issue on Vite is probably a good idea yes.
I'll ask the maintainers their opinion for what we should do in the meantime

Sheraff:
In what situations would someone have multiple versions of Vite at the same time?

zOadT:
FYI: https://github.com/voidzero-dev/vite-plus/issues/1391 should be the corresponding issue in the vite-plus repo. I am not sure if this should be solved by vite-plus or vite.

audi0417:
The concrete case I ran into is Vite+. `vite` is aliased to `@voidzero-dev/vite-plus-core`, while `vite-plus` also depends on it directly. So Node can end up loading what is essentially the same package through two different paths.

In this case, it can happen even with the same version — you still get two module instances, which is enough to break `instanceof`.

#7218  and #7614  look related too.

PS: @zOadT thanks for the link. https://github.com/voidzero-dev/vite-plus/issues/1391 looks like the same issue from the Vite+ side.

schiller-manuel:
fixed by https://github.com/voidzero-dev/vite-plus/issues/1391

PR #8242, opened 2026-09-04
fix(react-router,solid-router,vue-router): key the stale-chunk reload guard on the importer, not the error message

### LINKED ISSUE #3327: Auto-reload on failed lazy route loading can only work once per session on Safari

SethThoburn:
## 🎯 Changes

Fixes #3327

Improve reload behavior on lazy chunk loading error.

When the module for a lazy chunk 404's, typically caused by a deploy since the
last page load, `lazyRouteComponent` reloads the module to pick up the new
build. To avoid a reload-loop, we store the error message in `sessionStorage`
and don't reload if we've already "used up" the reload for that error.

This approach works very well on Chrome and Firefox, where the error is unique
to the missing module, meaning if another deploy happens during the same
session, we get a new error and can have another reload.

On Safari however, that error message is very plan, just "Importing a module
script failed." So any chunk loading error causes future deploys to throw
without a reload for that session because the errors are not unique per failed
module.

The solution I chose is to clear the error on a successful recovery. This still
prevents the reload loop, because if we don't recover successfully, the key is
still there and we won't reload. But it means that in Safari, the error is
cleared, and we can reload on the next chunk error in the same session.

My original strategy was to key not by the error message but by the import name,
but this has an edge case that could show up with certain bundlers. Namely, if
import names are deterministic, an import could 404, trigger a reload, then 404
again on a future deploy in the same session. The `lazyRouteComponent` would
view that as a reload loop and throw the 404. In my research, this could happen
with certain configurations of webpack.

I verified my approach by first reproducing the error in Safari, and then I
tried with the fix. Also tested in Chrome to ensure the approach works there.

## ✅ Checklist

- [x] I have followed the steps in the [[Contributing guide](https://github.com/TanStack/router/blob/main/CONTRIBUTING.md)](https://github.com/TanStack/router/blob/main/CONTRIBUTING.md).

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a [[changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery from stale lazy-loaded route chunks after deployments across React, Solid, and Vue routers.
  * Prevented repeated reload loops when browsers report identical module-loading errors, including Safari.
  * Successfully recovered modules now reset reload protection, allowing future stale-chunk recovery during the same session.

* **Tests**
  * Added end-to-end coverage for stale chunk recovery in Chromium and Safari-compatible WebKit browsers.
  * Added coverage for sequential deployments and successful recovery state reset.


packages/router-core/src/utils.ts
```diff
@@ -504,6 +504,59 @@ export function isModuleNotFoundError(error: any): boolean {
...
+  let importerSource = ''
+  try {
+    importerSource = Function.prototype.toString.call(importer)
+  } catch {
+    // Reading the source is best-effort; fall back to the message below.
+  }
+
+  return `tanstack_router_reload:${hashReloadKeyInput(
+    importerSource || error.message,
+  )}`
```
Sheraff: why do we need a try/catch and a fallback here? in which cases could this fail?

packages/router-core/src/utils.ts
```diff
@@ -504,6 +504,59 @@ export function isModuleNotFoundError(error: any): boolean {
...
+  error: { message: string },
+): string {
+  let importerSource = ''
+  try {
+    importerSource = Function.prototype.toString.call(importer)
+  } catch {
+    // Reading the source is best-effort; fall back to the message below.
+  }
+
+  return `tanstack_router_reload:${hashReloadKeyInput(
```
Sheraff: do we really need to hash here? what kinds of strings are we expecting to get from the importer stringification? it all goes into the temporary session storage in the end, so we can hash for perf if the strings are huge, but otherwise we shouldn't worry, right?

SethThoburn: Yeah, I think you are right. Will remove it.

packages/router-core/src/utils.ts
SethThoburn: Planning to remove this. Actually reworking the PR a bit which is why I moved it to draft.

SethThoburn:
Moving this PR to draft while I make some changes. Basically, there are circumstances where the filename (used as the key) would be the same between builds (if using webpack rather than vite) causing this same error to show up in  safari and also in Chrome et al. 

To avoid this regression and fix that edge case in Safari, I'm going to clear the key from session storage after a successful reload.

packages/router-core/src/utils.ts
SethThoburn: Removed.

SethThoburn:
PR updated!

Sheraff:
This is a pretty expensive change at +225B gzip for the react.minimal scenario, could we get something smaller? (the bundle size benchmark and skill are both in the codebase, agents are pretty good at lowering that). We're always aiming for 0 or negative of course, but any reduction would be appreciated.

SethThoburn:
@Sheraff, pushed an updated commit that benchmarks at +13 for react-router.minimal. It looks like before it was +44, not +225, but I think it's less bad now regardless.

PR #8246, opened 2026-09-05
fix(start): return 404 for unknown server function ids

### LINKED ISSUE #7363: TanStack Start server functions should return 404 when requested with an invalid ID, currently 500

Start

### Describe the bug

Server functions live at the `/_serverFn/$id` route, and they work well. If the build changes, the IDs change - this is good.

However, when a server function is requested at an old ID that no longer exists, the framework returns a 500 and you'll see a bunch of associated logging for the stack trace.

This is causing me, and presumably others, issues when deploying to production because these 500s trigger alerts etc. when they are in fact erroneous. Any other instance of a missing resource would be treated as a 404.

The cause of the issues in production comes from caching. Some of this we can control as website publishers, but some of it we can't. For example, I am currently being repeatedly hit by BingBot during its routine indexing operations (which I want) but it caches and follows server function routes in ways that don't seem to always abide by headers or TTLs. Each time it hits an old server function route that it has discovered and cached from a previous load, I'll get a 500 error.

I've personally worked around this by:
- adding more specific caching headers to the root, route, asset and serverFn routes
- adding more specific robot.txt instructions to disallow serverFn routes
- adding middleware to suppress the "Server function info not found" 500 errors and turn them into 404s

These three things are things that I think the framework should do for you.

https://github.com/neverstew/server-fn-500s

1. Create an empty project - I used the basic template from the docs
2. Build it with `npm run build`
3. Preview with `npx vite preview`
4. `curl -i http://localhost:3000/_serverFn/test`

```
neverstew@mattbook ~/d/a/start-basic> curl -i http://localhost:3000/_serverFn/test
HTTP/1.1 500 Internal Server Error
Vary: Origin
content-type: application/json;charset=UTF-8
content-length: 53
Date: Thu, 07 May 2026 21:14:00 GMT
Connection: keep-alive
Keep-Alive: timeout=5

{"status":500,"unhandled":true,"message":"HTTPError"}⏎
```

```
neverstew@mattbook ~/d/a/start-basic> npx vite preview

 ╭────────── [Build Info] ────────────╮
 │                                    │
 │  - Build Directory: .output        │
 │  - Date: 07/05/2026, 22:13:44      │
 │  - Nitro Version: 3.0.260311-beta  │
 │  - Nitro Preset: node-server       │
 │                                    │
 ╰────────────────────────────────────╯

  ➜  Local:   http://localhost:3000/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
Error: Server function info not found for test
    at getServerFnById (file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:144:27)
    ... 8 lines matching cause stack trace ...
    at async Object.fetch (file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:1586:10) {
  cause: Error: Server function info not found for test
      at getServerFnById (file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:144:27)
      at handleServerAction (file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:587:23)
      at file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:1411:15
      at AsyncLocalStorage.run (node:internal/async_local_storage/async_context_frame:63:14)
      at runWithStartContext (file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:199:22)
      at serverFnHandler (file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:1404:13)
      at next (file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:1261:19)
      at executeMiddleware (file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:1279:9)
      at startRequestResolver (file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:1417:42)
      at async Object.fetch (file:///Users/neverstew/dev/abitat/start-basic/.output/server/_ssr/ssr.mjs:1586:10),
  status: 500,
  statusText: undefined,
  headers: undefined,
  data: undefined,
  body: undefined,
  unhandled: true
}
```

### Expected behavior

A nicely handled response with something informative and a 404 status.
No stack trace in the logs.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.167.65
- OS: macOS, Linux, (presumably Windows - untested)
- Browser: N/A
- Bundler: vite
- Bundler Version: 8.0.0

### Additional context

_No response_

ousamabenyounes:
## 🎯 Changes

Fixes #7363.

Server function ids are minted per build. After a redeploy, caches, crawlers and tabs that have not reloaded keep calling ids the new build does not know about. Those requests were coming back as an unhandled `500` with a full stack trace in the server logs, which fires production alerts for what is really a missing resource.

The unknown-id case was also the one resolver failure `handleServerAction` could not see: it resolved the server function *outside* its `try`, so the rejection escaped the handler entirely.

- `generateServerFnResolverModule` still throws a regular `Error` for an id that is absent from the manifest, but sets a flag on it (`SERVER_FN_NOT_FOUND`, exported from `start-server-core/src/constants.ts` and interpolated into the emitted module, so both packages share one definition).
- `handleServerAction` resolves the function inside a `try`. A flagged error becomes a `404 text/plain` response and is logged outside production; every other resolution failure — a module that fails to import, a missing export — is rethrown and keeps its current behaviour.

Deliberately kept a real `Error` rather than a `notFound()`-shaped value: `createSsrRpc`, the RSC `loadServerAction` and `ServerFunctionSerializationAdapter` all call the resolver with no `catch`, so a not-found-shaped throw would have made an infrastructure error render the app's `notFoundComponent` and would have dropped the message and stack on those paths. Only the request handler acts on the flag.

The `404` body omits the requested id so the response never echoes back a caller-supplied path segment; the id is logged server-side outside production.

Scope notes:
- The dev-only Vite resolver is unchanged. Dev ids are base64url-encoded `{file, export}` payloads rather than manifest lookups, so the cross-deployment skew this fixes cannot happen there, and a loud dev error is more useful than a quiet `404`.
- The issue also suggests cache headers and `robots.txt` guidance for `/_serverFn` routes. Those are separate config/API decisions and are not part of this PR.

## Test verification (RED → GREEN)

New tests: 4 in `start-plugin-core` (the emitted resolver is written to a temp file and imported, so the generated JS is really executed) and 5 in `start-server-core`.

**RED** — production sources reverted to `main`, only the new tests applied:

```
--- start-server-core tests/server-functions-handler.test.ts ---
 × answers 404 when the id is not in this build
 × logs the unresolved id outside production
 × stays quiet in production, where stale ids are routine traffic
 Test Files  1 failed (1)
      Tests  3 failed | 2 passed (5)

--- start-plugin-core tests/server-fn-resolver-module.test.ts ---
 × flags an id that is absent from the manifest so the handler can answer 404
 × leaves a resolution failure that is not a missing id unflagged
 Test Files  1 failed (1)
      Tests  2 failed | 2 passed (4)
```

The tests that pass in both columns are the guards: a resolvable id still dispatches, and a genuine resolver failure still propagates instead of being masked as a `404`.

**GREEN** — full unit suites on the branch (`main` baseline was 118 and 512):

```
--- start-server-core ---
 Test Files  8 passed (8)
      Tests  123 passed (123)

--- start-plugin-core ---
 Test Files  32 passed (32)
      Tests  517 passed (517)
 Type Errors  no errors
```

**End-to-end, against a real production build** (`e2e/react-start/server-functions`, `vite build` + `srvx --prod`, `NODE_ENV=production`). Same-origin headers because a bare `curl` is rejected by the CSRF middleware with `403` before reaching the handler:

```
### WITHOUT FIX — main
$ curl -si .../_serverFn/test -H 'Origin: ...' -H 'x-tsr-serverFn: true'
HTTP/1.1 500 Internal Server Error
content-type: application/json;charset=UTF-8

{"status":500,"unhandled":true,"message":"HTTPError"}

server stderr:
Error: Server function info not found for test
    at getServerFnById (.../dist/server/server.js:4459:27)
  cause: Error: Server function info not found for test
      at getServerFnById (.../dist/server/server.js:4459:27)

### WITH FIX
HTTP/1.1 404
content-type: text/plain; charset=utf-8

Server function not found

server stderr: (nothing)
```

Client side, this also means the call now rejects with `Error: Server function not found` instead of resolving with the serialized `500` body.

### Full local suite

Replayed what `pnpm run test:pr` covers for the affected packages, on the upstream base and again on this branch — identical results, so no regression:

```
                      main (baseline)      this branch
format                PASS                 PASS
test:eslint           PASS                 PASS
test:types            PASS                 PASS   (ts56 - ts70)
test:unit             PASS                 PASS
build                 PASS                 PASS
test:build            PASS                 PASS   (publint --strict + attw)
```

Also green: the `tanstack-react-start-e2e-server-functions` e2e project (53 vite + 53 rsbuild + 3 rsbuild-cache passed) — the rsbuild-cache specs are the ones asserting the missing-id diagnostic stays absent from dev-server logs while the registry is intact, which is why the log was kept rather than dropped.

Bundle size: **0 byte gzip delta across all 18 benchmark scenarios**, as expected for a server-only change.

The full `test:e2e` matrix (~130 example/e2e apps, distributed over Nx Cloud agents in CI) was not reproducible on one machine; only the server-functions project above was run locally.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests for unavailable or outdated server-function IDs now return a clear `404` response.
  * Client calls to unavailable server functions now reject with an error instead of returning a serialized server error response.
  * Other server-function resolution failures continue to behave as before.

* **Documentation**
  * Added guidance for handling server-function version mismatches, including prompting users to reload.

### DIFF
--- a/.changeset/stale-server-fn-id-not-found.md
+++ b/.changeset/stale-server-fn-id-not-found.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/start-plugin-core': patch
+'@tanstack/start-server-core': patch
+---
+
+Answer `404` instead of an unhandled `500` when a server function is requested with an id that is not in the build's manifest. Every build mints new server function ids, so caches, crawlers and tabs that have not reloaded keep calling ids from a previous deployment, and each one was surfacing as an unhandled server error with a full stack trace. The generated resolver now flags that specific case, the request handler answers `404` and logs the unresolved id outside production, and every other resolution failure keeps its current behaviour. On the client, a call to a stale id now rejects with an error instead of resolving with the serialized `500` body.
--- a/docs/start/framework/react/guide/server-functions.md
+++ b/docs/start/framework/react/guide/server-functions.md
@@ -421,6 +421,8 @@ Server functions are addressed by a generated, stable function ID under the hood
 By default, IDs are SHA256 hashes of the same seed to keep bundles compact and avoid leaking file paths.
 If two server functions end up with the same ID (including when using a custom generator), the system de-duplicates by appending an incrementing suffix like `_1`, `_2`, etc.
 
+Because the IDs belong to a specific build, a deployment can still receive requests for IDs it no longer knows about — from a cached response, a crawler, or a tab that has not reloaded since the previous release. Those requests are answered with a `404`, and the client-side call rejects with an error rather than resolving. If your app should survive a version skew instead of surfacing the error, catch it where the server function is called and prompt a reload.
+
 Customization:
 
 You can customize function ID generation for the production build by providing a `generateFunctionId` function when configuring the TanStack Start build tool plugin.
--- a/packages/start-plugin-core/src/start-compiler/server-fn-resolver-module.ts
+++ b/packages/start-plugin-core/src/start-compiler/server-fn-resolver-module.ts
@@ -1,3 +1,4 @@
+import { SERVER_FN_NOT_FOUND } from '@tanstack/start-server-core/constants'
 import type { ServerFn } from './types'
 
 interface ResolverManifestEntry {
@@ -50,7 +51,13 @@ function getResolverBody(): string {
 export async function getServerFnById(id, access) {
   const serverFnInfo = manifest[id]
   if (!serverFnInfo) {
-    throw new Error('Server function info not found for ' + id)
+    // Every build mints new ids, so a cache, a crawler or a tab that has not
+    // reloaded keeps requesting ids from a previous deployment. Flagged so the
+    // request handler can answer 404; still a regular Error so the callers that
+    // are not serving an HTTP request keep the message and the stack.
+    const error = new Error('Server function info not found for ' + id)
+    error[${JSON.stringify(SERVER_FN_NOT_FOUND)}] = true
+    throw error
   }
 __CLIENT_REFERENCED_CHECK__
   const fnModule = serverFnInfo.module ?? (await serverFnInfo.importer())
--- a/packages/start-plugin-core/tests/server-fn-resolver-module.test.ts
+++ b/packages/start-plugin-core/tests/server-fn-resolver-module.test.ts
@@ -1,0 +1,133 @@
+import { mkdtemp, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { pathToFileURL } from 'node:url'
+import { afterEach, describe, expect, test } from 'vitest'
+import {
+  SERVER_FN_NOT_FOUND,
+  isServerFnNotFound,
+} from '@tanstack/start-server-core/constants'
+import { generateServerFnResolverModule } from '../src/start-compiler/server-fn-resolver-module'
+
+const KNOWN_SERVER_FN_ID = 'known-server-fn-id'
+const STALE_SERVER_FN_ID = 'stale-server-fn-id'
+
+const createdDirs: Array<string> = []
+
+afterEach(async () => {
+  await Promise.all(
+    createdDirs
+      .splice(0)
+      .map((dir) => rm(dir, { force: true, recursive: true })),
+  )
+})
+
+/**
+ * The resolver is emitted as source text, so the only faithful way to observe
+ * its runtime behaviour is to write it out and import it as a real module.
+ */
+async function loadGeneratedResolver(
+  opts: { extractedFilename?: string } = {},
+) {
+  const source = generateServerFnResolverModule({
+    serverFnsById: {
+      [KNOWN_SERVER_FN_ID]: {
+        functionName: 'knownServerFn',
+        extractedFilename: opts.extractedFilename ?? './known-server-fn.mjs',
+      } as never,
+    },
+    includeClientReferencedCheck: false,
+  })
+
+  const dir = await mkdtemp(join(tmpdir(), 'tsr-server-fn-resolver-'))
+  createdDirs.push(dir)
+
+  await writeFile(
+    join(dir, 'known-server-fn.mjs'),
+    'export const knownServerFn = () => "ok"\n',
+    'utf8',
+  )
+  const modulePath = join(dir, 'resolver.mjs')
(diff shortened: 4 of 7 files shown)


ousamabenyounes:
Addressed the review comment in d81f330.

`isServerFnNotFound` accepted an inherited `tssServerFnNotFound`, so an unrelated resolver failure whose prototype happened to carry the marker would have been answered with `404` instead of surfacing. The doc comment already specified an own property, so the check now matches it via `Object.prototype.hasOwnProperty.call`.

Added a regression test (`does not treat an inherited marker as a missing id`) that throws an `Error` subclass carrying the marker on its prototype and asserts the handler rethrows it. Confirmed it fails without the guard and passes with it.

Verified locally: `test:eslint`, `test:types`, `test:unit`, `build` and `test:build` for the affected packages, all green and matching the `main` baseline.

ousamabenyounes:
Rebased onto `main` and resolved the merge conflicts. The upstream frame-protocol refactor had moved `TSS_CONTENT_TYPE_FRAMED_VERSIONED` to `@tanstack/start-client-core`, so I kept `main`'s imports and only re-added `isServerFnNotFound`. The two `*.test.ts` files were add/add conflicts; I merged the resolver suites and moved the handler not-found cases into their own `server-functions-handler-not-found.test.ts` to keep the module mocks isolated.

Verified locally: on unmodified `main` source the not-found tests are RED (3 fail — the id still surfaces as 500); with the fix they are GREEN. `nx test:unit` passes for `@tanstack/start-server-core` (247) and `@tanstack/start-plugin-core` (resolver suite 9), and `test:types` + `test:eslint` pass for both. CI on the pushed SHA is still running — not claiming it green.

Nothing left open on my side; awaiting maintainer review.

PR #8251, opened 2026-09-05
perf(router): use URL.canParse for absolute URL checks

Sheraff:
## 🎯 Changes

Use `URL.canParse` for checks that only need to know whether a string is an absolute URL. This avoids constructing a URL object on success and throwing on relative paths or malformed URLs.

React, Solid, Vue, navigation, and redirects share an internal `isAbsoluteUrl` helper with a constructor fallback for older browsers. The Start and Nitro build plugins use `URL.canParse` directly because their minimum Node versions support it. Existing protocol checks and internal-link fast paths remain in place.

React and Solid links use `!isSafeInternal(to) && isAbsoluteUrl(to)`: slash/dot paths skip parsing, and other destinations go directly to URL validation without separate type or colon checks. The internal helper accepts optional `undefined` and returns a boolean.

Local Node 25.8.1 validation benchmark results, median of seven samples of 100,000 checks after warmup, in ns/check:

| Input | Constructor + catch | Native check with fallback detection | Speedup |
| --- | ---: | ---: | ---: |
| Absolute URL | 217 | 114 | 1.9× |
| Relative path | 3,920 | 40 | 97.5× |
| Malformed URL | 3,988 | 82 | 48.4× |
| Mixed | 2,899 | 55 | 52.4× |

These are validation microbenchmarks, not whole-router speedups. React and Solid already skip parsing most internal links.

Validation:

- `pnpm test:unit` (30 affected projects), including existing framework link/redirect coverage.
- `pnpm test:eslint` (33 affected projects) and `pnpm test:types` (37 affected projects).
- React basic example browser smoke test with native support enabled and disabled.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of absolute URLs across links, navigation, redirects, and application configuration.
  - Preserved support for older browsers without `URL.canParse` through a compatibility fallback.
  - Maintained existing external-link handling, including dangerous-protocol protection.
  - Applied consistent absolute-URL handling across React, Solid, Vue, and core routing experiences.
  - Improved handling of optional or undefined navigation destinations without changing expected routing behavior.


packages/router-core/tests/isAbsoluteUrl.test.ts
Sheraff, line 1: remove this test file

packages/react-router/src/link.tsx
```diff
@@ -298,16 +295,15 @@ export function useLinkProps<
 
       // Only attempt URL parsing when it looks like an absolute URL.
       if (typeof to === 'string' && to.indexOf(':') > -1) {
-        try {
-          new URL(to)
+        if (isAbsoluteUrl(to)) {
```
Sheraff: we don't need nested `if`s here, `isAbsoluteUrl` can just be part of the condition

packages/react-router/src/link.tsx
```diff
@@ -226,11 +226,10 @@ export function useLinkProps<
     if (
       typeof to === 'string' &&
       !safeInternal &&
-      // Quick checks to avoid `new URL` in common internal-like cases
+      // Quick checks to avoid URL parsing in common internal-like cases
       to.indexOf(':') > -1
     ) {
-      try {
-        new URL(to)
```
Sheraff, line 233: we don't need nested `if`s here, `isAbsoluteUrl` can just be part of the condition

packages/react-router/src/link.tsx
```diff
@@ -80,8 +81,7 @@ function resolveExternalLink(
   if (typeof to !== 'string' || to.indexOf(':') === -1) {
     return undefined
   }
-  try {
-    new URL(to)
+  if (isAbsoluteUrl(to)) {
```
Sheraff: we can re-arrange this into a single `if` block here

packages/solid-router/src/link.tsx
```diff
@@ -178,8 +179,7 @@ export function useLinkProps<
     const safeInternal = isSafeInternal(to)
     if (safeInternal) return undefined
     if (typeof to !== 'string' || to.indexOf(':') === -1) return undefined
-    try {
-      new URL(to as any)
+    if (isAbsoluteUrl(to)) {
```
Sheraff: we can re-arrange this into a single `if` block

packages/start-plugin-core/tests/planning.test.ts
Sheraff, line 1: is this testing something useful? or are we just basically testing the built-in implementation of `URL.canParse`?

scripts/benchmarks/url-validation.mjs
Sheraff, line 1: remove this, useful for the PR, but not to keep in the repo

schiller-manuel (approved):

PR #8252, opened 2026-09-05
perf(router): share Link pathname interpolation

schiller-manuel:
## 🎯 Changes

Reuse pathname interpolation across Links with a bounded router-scoped cache, collect cache metadata during the interpolation pass, and simplify React Link path handling to meet the bundle budgets.

This is PR 1 of a stacked series.

## ✅ Checklist

- [x] I have tested code changes locally with the relevant test commands.

## 🚀 Release Impact

- [x] This change affects published code, and I have generated changesets.
- [ ] This change is docs/CI/dev-only.


packages/router-core/src/router.ts
```diff
@@ -1879,6 +1898,71 @@ export class RouterCore<
...
+                params,
+                decoder,
+                undefined,
+                undefined,
+                this.isServer,
+              )
+            : interpolatePathname(path, params, decoder))
+        )
+      }
+      key = keys.length === 1 ? value : key + value.length + ':' + value
```
Sheraff: could we use a separator that is less likely to collide with actual values? (like `\0`)

schiller-manuel: good idea

packages/router-core/src/router.ts
```diff
@@ -1879,6 +1898,71 @@ export class RouterCore<
...
+            )
+          : interpolatePathname(path, params, decoder, undefined, keys)
+      plan = [keys, createSieveCache<string, string>(128), decoder]
+      this.pathCache.set(path, plan)
+    }
+    const [keys, paths] = plan
+    let key = ''
+    for (const name of keys) {
+      const value = params[name]
+      if (typeof value !== 'string') {
```
Sheraff: does this mean that omitted optional params can never be cached? this would mean that a website where `/{-$lang}/foo` uses a default locale for the most commonly used language then no path would ever be cached?

schiller-manuel: need to add a test for this

packages/router-core/src/router.ts
schiller-manuel: not necessary, encoding makes sure this does not collide

packages/router-core/src/router.ts
schiller-manuel: fixed

PR #8259, opened 2026-09-05
perf(router-core): reduce navigation promise chains

Sheraff:
## 🎯 Changes

Reduce Promise overhead in navigation while preserving cancellation and loader readiness:

- Synchronous client `beforeLoad` results skip the cancellable Promise wrapper and its abort listener. Keep the `await` checkpoint so a queued replacement navigation can cancel stale work before its loader starts.
- Combine component-chunk error handling and loader readiness into one async task. Return reportable chunk failures directly; otherwise wait for loader data before publishing readiness. Remove the empty transaction-completion continuation.
- Resolve static server SSR policies synchronously. Functional policies still assimilate their results into a native Promise; synchronous throws become rejected Promises so queued request cancellation wins before context/error hooks run.

Controller allocation and public types are unchanged. Controllers still own shared loader flights, cached data, request cancellation, and deferred cleanup. Promise detection assumes ordinary Promises, including foreign-realm Promises; custom stateful `then` getters are not an additional compatibility requirement.

### Measurements

Local allocation measurements compare production baseline `07b3bc971d` with `b8e8fd5`. Each lane contains a root plus eight nested routes. Identical temporary harnesses counted `async_hooks` Promise resources around one operation, outside timed loops. Absolute counts include harness awaits; the deltas are the useful result.

| Scenario | Before | After | Fewer Promise resources |
| --- | ---: | ---: | ---: |
| Client, no hooks/chunks | 147 | 119 | 28 |
| Client, eight synchronous beforeLoad hooks | 187 | 135 | 52 |
| Client, eight Promise beforeLoad hooks | 187 | 159 | 28 |
| Client, six sync and two Promise hooks | 187 | 141 | 46 |
| Client, asynchronous component preloads | 187 | 159 | 28 |
| Client, blocking loaders and component preloads | 275 | 247 | 28 |
| Server, static SSR policies | 68 | 50 | 18 |
| Server, synchronous functional SSR policies | 84 | 74 | 10 |
| Server, Promise functional SSR policies | 84 | 74 | 10 |
| Server, mixed SSR policies | 72 | 56 | 16 |

The chunk/completion changes save three Promise resources per match plus one per navigation. Each synchronous `beforeLoad` additionally saves three Promise resources and an abort listener registration/removal. Returning `undefined` has the same additional saving as returning a context object: 159 → 135 in the eight-hook comparison against the chunk/completion changes alone.

**Bundle size before the final chunk-error cleanup (`b8e8fd5`):** React Router minimal was **85,821 → 85,828 gzip bytes (+7 B)**. Both local measurement and [CI's 18-scenario comparison](https://github.com/TanStack/router/pull/8259#issuecomment-5555454557) show **+7–16 gzip bytes** overall, with unchanged JS file counts. The chunk/completion changes together save 6 bytes; the synchronous-hook guard adds 13 bytes to that composition. Independently, chunk consolidation and completion cleanup each saved 5 bytes; gzip effects are not additive.

**Latest chunk-error cleanup:** A fresh comparison against HEAD `bbd9d9ad` (after merging main) measures **−10 to +5 gzip bytes** across 18 scenarios: 11 shrink, two are unchanged, and five grow. React Router minimal is **85,828 → 85,829 B (+1 B)**; its initial gzip decreases by 3 B. Raw JS decreases by 16–18 B, and JS file counts are unchanged. These are incremental deltas for the cleanup, separate from the original PR comparison above.

**CPU simulation:** All six client-navigation/SSR jobs for React, Solid, and Vue passed on `b8e8fd5`. [CodSpeed reports](https://github.com/TanStack/router/pull/8259#issuecomment-5555551020) no CPU simulation regressions and one significant improvement: Solid loader navigation **202.8 → 196.7 ms** (about 3% less time). The report flags runtime-environment differences, so the magnitude remains provisional. **CodSpeed memory results are excluded from the performance assessment.**

Local timing runs had substantial outliers, including unchanged control paths, so they are diagnostic rather than evidence of a general navigation speedup. The allocation measurements and CI CPU simulation results are the basis for this tradeoff. Temporary investigation benchmarks and the results document are omitted from the final diff; regression tests remain.

<details>
<summary>Other ideas investigated</summary>

- Removing the client `beforeLoad` await entirely lets queued replacement navigation start a stale loader. Retain the scheduling checkpoint while skipping the synchronous wrapper. Server `beforeLoad` already awaits directly, so it has no corresponding wrapper to remove.
- Using the shared `isPromise` helper added 35 gzip bytes to the chunk/completion version; the inline guard adds 13 bytes instead.
- Skipping `Promise.all` for synchronous head/scripts added 18 gzip bytes and was left out to limit bundle growth.
- Moving abort-listener cleanup into existing settlement callbacks removed one Promise per wait but added 2 gzip bytes and slowed the local 80-wait benchmark by 3.1% for values, 6.6% for fulfilled Promises, and 5.8% for rejected Promises (RME 0.12–0.42%). It was left out despite the allocation reduction.
- Fresh server matching allocates controllers that lane cloning subsequently replaces. Eliminating that duplication requires distinguishing private fresh matches from reused or exposed controllers; blindly reusing them is unsafe for repeated loads. That ownership change is outside this patch.

</details>

### Validation

[The full affected CI unit/E2E, type, lint, and build checks passed on `b8e8fd5`](https://github.com/TanStack/router/actions/runs/33998482855/job/101393069847). Those CI results predate the final chunk-error cleanup; the new head will run CI again. Local validation of the corrected cleanup passed 1,712 core unit tests (plus four existing expected failures), TypeScript 5.6–7 checks, ESLint with 26 existing warnings, and 33 Chromium redirect tests. The chunk/readiness changes also passed 27 React pending/presentation tests and three React Start SSR/hydration tests.

Retained regression coverage includes synchronous context inheritance, native/foreign Promises, replacement navigation, cancellation before Promise settlement and late rejection, chunk failures and control flow, blocking readiness, SSR inheritance, and queued request cancellation.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Performance**
  * Reduced unnecessary Promise allocations during client navigation and server-side rendering.

* **Bug Fixes**
  * Improved cancellation handling when navigation is superseded during asynchronous route loading.
  * Preserved correct redirect, not-found, and chunk-failure behavior during reentrant navigation.
  * Improved static SSR policy inheritance across nested routes and ensured request cancellation takes precedence over SSR outcomes.

* **Documentation**
  * Clarified cancellation and loading behavior for synchronous and asynchronous route hooks.


packages/router-core/src/load-server.ts
```diff
@@ -232,7 +239,9 @@ async function contextualize(
     const match = lane.matches[index]!
     const route = getRoute(router, match)
     try {
-      match.ssr = await resolveSsr(router, lane, index)
+      const ssr = resolveSsr(router, lane, index)
+      // Functional policies are assimilated into a native Promise above.
+      match.ssr = ssr instanceof Promise ? await ssr : ssr
```
schiller-manuel, line 244: why not use `isPromise`?

Sheraff: if we *know* it's going to be a genuine `Promise` then this check is cheaper than `isPromise()`
(and also `isPromise` is currently unused, so there isn't even an upside to "just use the same thing as elsewhere")

PR #8261, opened 2026-09-06
perf(start-plugin-core): skip unused manifest build data

LadyBluenotes:
## 🎯 Changes

Start's Vite and Rsbuild client-build normalizers populate two indexes that the manifest never reads, and collect CSS text even when `server.build.inlineCss` is disabled.

Remove those indexes and collect CSS content only when the resolved inline-CSS setting enables it. Preserve route stylesheet metadata and enabled inline-CSS behavior. Add regressions for disabled capture, enabled capture, and the build-hook configuration wiring in both bundlers.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Performance**
  * Reduces build processing by skipping CSS content capture when inline CSS is disabled.
  * Avoids generating unused build index data.

* **Bug Fixes**
  * Preserves stylesheet links while including inline CSS content only when configured.
  * Prevents inline CSS capture during development builds.
  * Applies inline CSS settings consistently across Vite and Rsbuild builds.

* **Tests**
  * Added coverage for inline CSS behavior across Vite and Rsbuild builds.

### DIFF
--- a/packages/start-plugin-core/tests/rsbuild/normalized-client-build.test.ts
+++ b/packages/start-plugin-core/tests/rsbuild/normalized-client-build.test.ts
@@ -63,9 +63,14 @@ describe('normalizeRspackClientBuild', () => {
     expect(manifest.inlineCss).toBeUndefined()
   })
 
-  test.each([false, true])(
-    'captures inline CSS according to the resolved config (%s)',
-    (enabled) => {
+  test.each([
+    { action: 'build', enabled: false, captureCss: false },
+    { action: 'build', enabled: true, captureCss: true },
+    { action: 'dev', enabled: false, captureCss: false },
+    { action: 'dev', enabled: true, captureCss: false },
+  ])(
+    'captures inline CSS according to the action and resolved config ($action, $enabled)',
+    ({ action, enabled, captureCss }) => {
       const readCss = vi.fn(() =>
         Buffer.from('.card{background:url(./dot.svg)}'),
       )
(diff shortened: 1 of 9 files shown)


packages/start-plugin-core/tests/rsbuild/normalized-client-build.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      clientBuild,
+      routeTreeRoutes: {
+        __root__: {},
+        '/posts': { filePath: '/routes/posts.tsx' },
+      },
+      basePath: '/assets',
+    })
+
+    expect(readCss).not.toHaveBeenCalled()
+    expect(getAssets).not.toHaveBeenCalled()
```
schiller-manuel: Please keep these tests focused on functional behavior rather than enforcing the optimization. Whether CSS assets are read or an internal map is allocated isn't part of the functional contract I want these tests to cover.

Remove the CSS-read/enumeration call assertions here and in the other new Rsbuild/Vite cases, along with the internal-map allocation assertion. Keep the expectations for correct stylesheet links, inline CSS content when enabled, and development behavior. If existing tests already cover those outcomes, drop the redundant new cases.

schiller-manuel (changes requested):
Please revise the added tests to cover functional behavior rather than enforce the optimization, as described in the inline comment.

schiller-manuel (approved):

PR #8270, opened 2026-09-06
fix(solid): avoid repeated hydration scans and script execution

LadyBluenotes:
## 🎯 Changes

Solid hydration boundaries retain their marker element through a ref, avoiding a document-wide marker scan for each boundary. External and inline script creation are mutually exclusive, preventing a script with both `src` and string children from being inserted twice.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented external scripts from executing twice when they include both a source URL and inline content.
  * Improved hydration reliability by ensuring each hydration region uses its corresponding marker.
  * Ensured externally loaded scripts are properly removed when their associated content is unmounted.
* **Tests**
  * Added coverage for script rendering and cleanup across multiple inline-content scenarios.
  * Verified that hydration callbacks receive the correct marker for each hydration region.


packages/solid-router/tests/Scripts.test.tsx
```diff
@@ -54,6 +54,33 @@ afterEach(() => {
...
+        history: createMemoryHistory(),
+      })
+      await router.load()
+
+      const { unmount } = render(() => <RouterProvider router={router} />)
+      const scripts = document.querySelectorAll(
+        'script[src="/external-script.js"]',
+      )
+      expect(scripts).toHaveLength(1)
+      expect(scripts[0]?.textContent).toBe('')
```
brenelz, line 76: is this correct?

`expect(scripts[0]?.textContent).toBe('')`

brenelz (approved):

PR #8276, opened 2026-09-06
perf(router-plugin): avoid redundant compiler work

LadyBluenotes:
## 🎯 Changes

Route splitting parsed the same reference source three times and generated source maps that the Webpack/Rspack adapters discarded. Reuse one AST for grouping detection, shared-binding analysis and reference compilation, and skip map generation when those adapters receive no input map.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added source map generation for router code-split output when supported by the build environment.
  * Preserved incoming source maps for Webpack and Rspack builds while avoiding maps that would be discarded.
  * Improved code-split compilation consistency by reusing parsed route information.
  * Maintained identical generated code and source map results when route information is reused during compilation.

### DIFF
--- a/packages/router-plugin/src/core/code-splitter/compilers.ts
+++ b/packages/router-plugin/src/core/code-splitter/compilers.ts
@@ -148,13 +148,14 @@ const allCreateRouteFns = [
  * AND at least one non-split property. Only locally-declared module-level
  * bindings are candidates (not imports — bundlers dedupe those).
  */
-export function computeSharedBindings(opts: {
-  code: string
-  filename?: string
-  codeSplitGroupings: CodeSplitGroupings
-}): Set<string> {
-  const ast = parseAst(opts)
-
+export function computeSharedBindings(
+  opts: {
+    code: string
+    filename?: string
+    codeSplitGroupings: CodeSplitGroupings
+  },
+  ast = parseAst(opts),
+): Set<string> {
   // Early bailout: collect all module-level locally-declared binding names.
   // This is a cheap loop over program.body (no traversal). If the file has
   // no local bindings (aside from `Route`), nothing can be shared — skip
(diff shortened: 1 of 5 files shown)


packages/router-plugin/src/core/code-splitter/compilers.ts
```diff
@@ -1392,11 +1395,12 @@ export function compileCodeSplitSharedRoute(
  * This function should read get the options from by searching for the key `codeSplitGroupings`
  * on createFileRoute and return it's values if it exists, else return undefined
  */
-export function detectCodeSplitGroupingsFromRoute(opts: ParseAstOptions): {
+export function detectCodeSplitGroupingsFromRoute(
+  opts: ParseAstOptions,
+  ast = parseAst(opts),
```
brenelz, line 1400: so it only parses the ast if something isn't passed in?

brenelz (approved):
Looks good to me but I'd probably have @schiller-manuel take a look first

schiller-manuel:
i need to take a much closer look, but do we mutate the AST that is reused here? if yes, thats probably not correct then.

LadyBluenotes:
> i need to take a much closer look, but do we mutate the AST that is reused here? if yes, thats probably not correct then.

Yeah, reference compilation mutates it, but that runs last. Grouping detection and shared-binding analysis only read it, and the AST isn’t reused after compilation. Virtual/shared modules each parse their own AST

packages/router-plugin/src/core/code-splitter/compilers.ts
LadyBluenotes: Yeah if an AST is passed in, it uses that otherwise it parses a new one

Issue #8237, opened 2026-09-04
Server functions: requests without the x-tsr-serverFn header return unhandled 500s instead of a 4xx


puncsky:
## Summary

`handleServerAction` returns **unhandled 500s** (instead of 4xx responses) for two request shapes on the `/_serverFn/<id>` endpoint, reproducible on a fresh app with the latest packages (`@tanstack/react-start` **1.168.49**, `@tanstack/react-router` **1.170.32**, vite 7). Bots hit this daily in production on our site: one server-function endpoint answered **7,230×200**, **6,263×403** (CSRF, working as intended) and **6×500** in a 24-hour window — the 500s are crawlers replaying captured `/_serverFn/…?payload=…` URLs with browser headers minus the one custom header.

## Reproduction

A minimal app (5 source files, no extra deps):

```ts
// src/serverFn.ts
import { createServerFn } from "@tanstack/react-start";
export const getPost = createServerFn({ method: "GET" }).handler(
  async () => ({ title: "hello from the server function" }),
);
```

```tsx
// src/routes/index.tsx
import { createFileRoute } from "@tanstack/react-router";
import { getPost } from "../serverFn";
export const Route = createFileRoute("/")({
  loader: async () => ({ post: await getPost() }),
  component: () => <h1>repro</h1>,
});
```

plus `src/routes/__root.tsx` (`createRootRoute` with `<Outlet />`), `src/router.tsx` exporting `createRouter()`/`getRouter()` over `routeTree.gen`, and a vite config of `tanstackStart() + react()`. After `vite build`, serve `dist/server/server.js` through any `fetch(Request)` host (we used a 20-line `node:http` adapter).

### Shape 1 — request without `x-tsr-serverFn` (the main one)

```
curl -H 'Sec-Fetch-Site: same-origin' \
     'http://localhost:3000/_serverFn/<id>?payload=<valid seroval envelope>'   # → 500
curl -H 'x-tsr-serverFn: true' -H 'Sec-Fetch-Site: same-origin' \
     'http://localhost:3000/_serverFn/<id>?payload=<same envelope>'           # → 200
```

The envelope is exactly what the client proxy sends: `JSON.stringify(seroval.toJSON({data: {...}}))`, URL-encoded as `?payload=…`. **The same valid payload returns 200 with the header and 500 without it.** A no-payload request without the header also 500s.

Server-side stderr shows only the masked form — the cause repeats the message, so nothing names the defect:

```
Error: Internal Server Error
[...]

PR #8285, opened 2026-09-07
fix(start-server-core): return 400 for invalid JSON server function payloads

devtechedge:
## 🎯 Changes

A server function request with malformed JSON currently 500s. GET `JSON.parse(payloadParam)` and POST `request.json()` throw `SyntaxError`, the handler logs `Server Fn Error!`, and the client sees an unhandled 500. That's a bad request, not a server fault — crawlers replaying `/_serverFn/...` URLs with garbage payloads show up in production monitoring as 500s.

`handleServerAction` already returns a thrown `Response` from its outer catch, so both parse sites now throw `new Response('Invalid server function payload', { status: 400 })` on `SyntaxError` and rethrow anything else. seroval/`fromJSON` still 500s; this only covers invalid JSON.

This is the malformed-JSON case from #8237. The missing `x-tsr-serverFn` path is already in #8220, so this PR does not close that issue.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid JSON in server function GET payloads and request bodies now returns a clear `400` response instead of an unhandled `500` error.
  - Server functions are no longer invoked when their request payload contains malformed JSON.

- **Tests**
  - Added coverage for malformed JSON in both GET and POST server function requests.

### DIFF
--- a/.changeset/clear-json-payload-400.md
+++ b/.changeset/clear-json-payload-400.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/start-server-core': patch
+---
+
+Return 400 for invalid JSON on server function GET payloads and JSON request bodies.
--- a/packages/start-server-core/src/server-functions-handler.ts
+++ b/packages/start-server-core/src/server-functions-handler.ts
@@ -139,9 +139,21 @@ export const handleServerAction = async ({
             throw new Error('Payload too large')
           }
           // If there's a payload, we should try to parse it
-          const payload: any = payloadParam
-            ? parsePayload(JSON.parse(payloadParam))
-            : {}
+          let payload: any = {}
+          if (payloadParam) {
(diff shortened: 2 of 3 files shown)


devtechedge:
Hello - soft nudge on this PR when bandwidth allows.

Malformed JSON on a server-function request was bubbling as a `500`; this returns a `400` `Response` for `SyntaxError` on both the GET and POST parse paths so crawler junk stops looking like a server fault.

It only covers the invalid-JSON case called out in #8237, with a focused test and changeset in `start-server-core`, and leaves `seroval` / `fromJSON` behavior alone.

I am happy to adjust the error body, expand coverage, or rebase if `main` has moved.

Whenever you get a chance is fine - the branch is still green and ready on my side.

PR #8292, opened 2026-09-08
fix(react-router,router-core): hydrate Link active state against the server location

### LINKED ISSUE #8233: Link hydrates with wrong active state when a navigation starts before the route content has hydrated

Router

### Describe the bug

`<Link>` hydrates with the wrong active state (`data-status="active"`, `aria-current="page"`, `active` class) when a navigation starts after the shell has hydrated but before the route content's Suspense boundary has. React logs a hydration mismatch for every `<Link>` in the not-yet-hydrated content that points at the pending destination.

Chain (verified against `@tanstack/react-router` 1.170.32, `@tanstack/router-core` 1.171.27, `@tanstack/react-start` 1.168.49):

1. Start hydrates inside `startTransition(() => hydrateRoot(…))` and the root `<Outlet>` wraps child routes in `<Suspense>` (`Match.tsx`). The shell commits first; route content hydrates later in its own task.
2. A click on an already-hydrated shell `<Link>` runs `router.navigate()`. `loadClientRoute` sets `stores.location` synchronously (`load-client.ts`, inside the `status.set("pending")` batch); `resolvedLocation` moves only after the new matches commit. While the loader is in flight the old page stays mounted with the new location in the store.
3. The route content's boundary hydrates. `useLinkProps` derives the active state from `useStore(router.stores.location, …)`, and `@tanstack/react-store`'s `useStore`/`useSelector` passes the same live `source.get()` to `useSyncExternalStoreWithSelector` as both `getSnapshot` **and** `getServerSnapshot` (`packages/react-store/src/useSelector.ts`). React calls `getServerSnapshot` during hydration, gets the post-click location, and renders links to the pending destination as active — against HTML the server rendered as inactive.

React's `getServerSnapshot` contract is "what the server rendered with"; using the live getter is only safe while nothing mutates the store between SSR and hydration. Deferred boundary hydration plus a user click is exactly that gap.

Sibling of #8198 (same window, `useMatch` invariant variant). This one does not throw and is reachable in any Start app with nav links in the root layout.

https://github.com/k3dom/tanstack-link-hydration-repro

`pnpm install && pnpm repro` starts `vite dev`, loads `/` in Chromium with 6x CPU throttling (DevTools' "mid-tier mobile" preset), clicks the header `<Link to="/about">` as soon as the shell has hydrated, and counts React's hydration mismatch errors. Routes: `__root.tsx` has a header `<Link to="/about">` above `<Outlet />`; `index.tsx` has 500 `<Link to="/about">`s in the body; `about.tsx` has a 3s loader.

1. Open `/` (dev server, or a production build with CPU throttling — the window is the gap between shell hydration and route content hydration).
2. As soon as the header is interactive, click the header `<Link to="/about">`. `/about` has a slow loader, so the navigation stays pending.
3. Console:

   ```
   A tree hydrated but some attributes of the server rendered HTML didn't match the client properties.
     <Home>
       <main>
         <p>
           <a
   +         className="active"
   -         className={null}
             href="/about"
   +         data-status="active"
   -         data-status={null}
   +         aria-current="page"
   -         aria-current={null}
   ```

   Reported for every `<Link to="/about">` inside the route content; the header link itself was already hydrated and is fine.

Measured by the script (fresh clone, 6x throttle, reproduces on every run):

```
shell hydrated at:            702ms
header link clicked at:       1078ms
route content hydrated at:    1832ms
hydration mismatch errors:    1
```

Same run without the click: 0 errors. At 1x the window is ~90ms and Playwright's click lands after the content has hydrated, so it does not reproduce — it needs a slow device or a heavy page.

### Expected behavior

Route content hydrates against the location the server rendered with, so the output matches the HTML; React then re-renders with the live location and flips the active state. That is what the `getServerSnapshot` / `getSnapshot` split exists for.

Either would close the window:

- `useStore` accepts a separate server snapshot, and the router supplies the location captured at hydration time for `stores.location` until hydration completes.
- `useLinkProps` resolves the active state against the hydration-time location while hydrating. `useHydrated()` is already in scope there (used only for `includeHash`).

There is no user-land opt-out: `useLinkProps` always emits `data-status`/`aria-current` when active regardless of `activeProps`, and no `activeOptions` flag selects `resolvedLocation` (which `router.matchRoute()` already uses while pending).

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: `@tanstack/react-router` 1.170.32, `@tanstack/router-core` 1.171.27, `@tanstack/react-start` 1.168.49
- React: 19.2.8 / react-dom 19.2.8
- OS: Linux
- Browser: Chromium 149 (Playwright-driven)
- Bundler: Vite 8.2.2

### Additional context

Impact is mostly dev-time: React 19 only logs attribute mismatches in development and never patches them. In production the consequence is a stale DOM attribute on a reused `<a>` fibre when source and destination share a route.

shunn2:
## 🎯 Changes

Fixes [#8233.](https://github.com/TanStack/router/issues/8233)

`useLinkProps` derives its `href` and active state from `useStore(router.stores.location, …)`. `@tanstack/react-store` passes the same live `source.get()` as both `getSnapshot` and `getServerSnapshot`, so React reads the *current* location while a boundary is hydrating. A navigation that starts after the shell hydrates but before the route content does publishes its destination to `stores.location` synchronously (`loadClientRoute`), so every `<Link>` in the not-yet-hydrated content that points at the pending destination hydrates as active against HTML the server rendered as inactive.

- `router-core`: `hydrate()` records the location it claims the SSR payload with on `router._hydrationLocation` (internal field, no public API change).
- `react-router`: `selectLinkState` resolves against that location until `useHydrated()` flips, then re-derives against the live location. This also fixes `href` for relative links, which the server builds from the SSR location while the client was building it from the pending one.

Active state is deliberately not suppressed while hydrating: the server renders `data-status="active"` for the current page's links, so suppressing it would create the mismatch in the opposite direction. The tests assert both directions.

`resolvedLocation` is not a usable source here. It moves to the destination once the navigation commits, and it stays `undefined` until the initial client load commits when a route sets `ssr: false`. Both cases have a test that fails with `resolvedLocation` and passes with the captured hydration location.

### Behaviour

Unchanged when no navigation is in flight, because both locations are then equal, and unchanged in CSR, where `_hydrationLocation` is `undefined` and `useHydrated()` is already `true` on the first render. Applications that hydrate without calling `hydrate()` keep today's behaviour.

## ✅ Checklist

- `pnpm nx run-many --target=test:unit --projects=@tanstack/react-router,@tanstack/router-core` — `router-core` 121 files / 1809 passed / 4 expected fail, `react-router` 90 files / 1126 passed / 1 skipped, no type errors
- `pnpm nx run-many --target=test:eslint --projects=@tanstack/react-router,@tanstack/router-core` — 0 errors (the pre-existing 26 and 100 warnings are unchanged)
- `pnpm nx run-many --target=test:types --projects=@tanstack/react-router,@tanstack/router-core` — passes on TS 5.6, 5.7, 5.8, 5.9, 6.0 and 7.0

## 🚀 Release Impact

## Bug Fixes

* Fixed `<Link>` hydration so active states, URLs, `data-status`, `aria-current`, and active classes match the server-rendered location.
* Prevented hydration mismatches when navigation begins before content inside a Suspense boundary finishes hydrating.
* Links now transition to the current client location after hydration completes.
* Improved redirect and URL handling, including safer treatment of external, malformed, and potentially dangerous destinations.
* Prevented duplicate `beforeunload` prompts during document navigation.

### DIFF
--- a/packages/react-router/src/link.tsx
+++ b/packages/react-router/src/link.tsx
@@ -423,7 +423,7 @@ export function useLinkProps<
   // the comparator only sees the location, not whether this link's output moved.
   // eslint-disable-next-line react-hooks/rules-of-hooks
   const selectLinkState = React.useCallback(
-    (location: ParsedLocation): LinkState => {
+    (liveLocation: ParsedLocation): LinkState => {
       const directExternalLink = resolveExternalLink(
         to,
         router.protocolAllowlist,
@@ -432,6 +432,14 @@ export function useLinkProps<
         return [directExternalLink ?? undefined]
       }
 
+      // React runs this selector as `getServerSnapshot` while a boundary is
+      // hydrating, so it must reproduce the server output; `stores.location`
+      // already holds the destination while the loader is in flight. The
+      // `isHydrated` flip re-runs the selection against the live location.
+      const location =
+        isHydrated || !router._hydrationLocation
+          ? liveLocation
+          : router._hydrationLocation
       const next = router.buildLocation({
         _fromLocation: location,
         ..._options,
--- a/.changeset/quiet-links-hydrate.md
+++ b/.changeset/quiet-links-hydrate.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/router-core': patch
+'@tanstack/react-router': patch
+---
+
+Fix `<Link>` hydrating with the wrong active state (`data-status`, `aria-current`, active class) when a navigation starts before a Suspense boundary containing the link has hydrated. `hydrate()` now records the location the server rendered with, and links resolve their `href` and active state against it while hydrating, then follow the live location once hydrated.
--- a/.changeset/quiet-links-hydrate.md
+++ b/.changeset/quiet-links-hydrate.md
@@ -1,6 +1,8 @@
 ---
 '@tanstack/router-core': patch
 '@tanstack/react-router': patch
+'@tanstack/solid-router': patch
+'@tanstack/vue-router': patch
 ---
 
-Fix `<Link>` hydrating with the wrong active state (`data-status`, `aria-current`, active class) when a navigation starts before a Suspense boundary containing the link has hydrated. `hydrate()` now records the location the server rendered with, and links resolve their `href` and active state against it while hydrating, then follow the live location once hydrated.
+Fix `<Link>` hydrating with the wrong active state (`data-status`, `aria-current`, active class) when a navigation starts before the component containing the link has hydrated. `hydrate()` now records the location the server rendered with, and links resolve their `href` and active state against it while hydrating, then follow the live location once hydrated. Affects React, Solid and Vue: none of the three rectifies the mismatched attributes, so the stale state persisted on the element.
--- a/packages/react-router/tests/issue-8233-link-hydration-active-state.test.tsx
+++ b/packages/react-router/tests/issue-8233-link-hydration-active-state.test.tsx
@@ -1,0 +1,305 @@
+import * as React from 'react'
+import { act } from '@testing-library/react'
+import { hydrateRoot } from 'react-dom/client'
+import { renderToString } from 'react-dom/server'
+import { afterEach, expect, test, vi } from 'vitest'
+import { createMemoryHistory } from '@tanstack/history'
+import { dehydrateSsrMatchId } from '../../router-core/src/ssr/ssr-match-id'
+import { hydrate } from '../src/ssr/client'
+import {
+  Link,
+  Outlet,
+  RouterProvider,
+  createControlledPromise,
+  createRootRoute,
+  createRoute,
+  createRouter,
+} from '../src'
+import type { MockInstance } from 'vitest'
+import type { AnyRouter, RouteComponent } from '../src'
+
+const testCleanups: Array<() => void | Promise<void>> = []
+
+afterEach(async () => {
+  while (testCleanups.length) {
+    await testCleanups.pop()!()
+  }
+  vi.restoreAllMocks()
+  window.$_TSR = undefined
+  document.body.innerHTML = ''
+})
+
+function Home() {
+  return (
+    <main>
+      <Link to="/about" data-testid="content-link">
+        about
+      </Link>
+    </main>
+  )
+}
+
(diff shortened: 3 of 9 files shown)


packages/react-router/src/link.tsx
```diff
@@ -497,7 +497,17 @@ export function useLinkProps<
...
+      // React runs this selector as `getServerSnapshot` while a boundary is
+      // hydrating, so it has to reproduce the server output. A navigation that
+      // started before this boundary hydrated has already published its
+      // destination to `stores.location`, so resolve against the location the
+      // server rendered with until hydration completes. The `isHydrated` flip
+      // re-runs the selection against the live location right after.
+      const location =
+        isHydrated || !router._hydrationLocation
+          ? liveLocation
+          : router._hydrationLocation
```
Sheraff, line 442: Can this be simplified to `router._hydrationLocation ?? liveLocation` if we unset the `_hydrationLocation` at the end of hydration? Would that result in a smaller bundle size?

Sheraff:
Does this issue reproduce in Solid and Vue too? In which case the fix should be applied to all 3

packages/react-router/src/link.tsx
shunn2: I think the unset can't be scoped correctly. A boundary can still be hydrating while other components mount fresh client-side(a navigation commits before a root-layout Suspense boundary hydrates), and those fresh mounts need the live location.

size
react-router/dist/esm/link.js 2,602 → 2,645 (+43)
router-core/dist/esm/load-client.js 7,620 → 7,630 (+10)

shunn2:
> Does this issue reproduce in Solid and Vue too? In which case the fix should be applied to all 3

Reproduces in all three. I'll fix Vue and Solid too.

Vue is the same as React: `[Vue warn]: Hydration class mismatch on .../about`, and the DOM after mount is byte-identical to the server HTML. Solid reads the live location while hydrating as well

shunn2:
@Sheraff hi Is there anything else I should do or any feedback you can give me?

PR #8314, opened 2026-09-09
fix(router-core): omit defaulted search params for retain-then-strip Links

### LINKED ISSUE #8309: Links don't fully follow `retainSearchParams` and `stripSearchParams`

Router

### Describe the bug

When combining `retainSearchParams` and `stripSearchParams`, it seems their effect is inconsistent.

Example:
```ts
const SearchSchema = z.object({
  myParam: z.string().default('foo'),
});

  component: RootComponent,
  validateSearch: SearchSchema,
  search: {
    middlewares: [
      retainSearchParams(true),
      stripSearchParams({ myParam: 'foo' }),
    ],
  },
});
```
The goal is to keep search params while navigating (e.g. retain global filter states) but not clutter the URLs as long as the default values are still active.

Now `const { myParam } = Route.useSearch();` yields `"foo"`, but the parameter is not part of the URL in the browser - so far so good.

But unfortunately `Link`s don't quite follow this logic. They still point to `<BASE_URL>?myParam=foo`. Consequently a link does not match its own URL and is not show as being active.

Links seem to behave correctly if I use either `retainSearchParams` OR `stripSearchParams`, but not with both.

https://stackblitz.com/edit/github-oubutstq?file=src%2Froutes%2F__root.tsx

1. Use both `retainSearchParams` and `stripSearchParams` in `search.middlewares`.
2. Provide a search schema with a property that has a default value.
3. Strip that property if it has the default value.
4. Place a link and add a style when the link is active.
5. See that the link include the property with default value and is not shown as active.

### Expected behavior

I expected the link to not include the property as long as it has the given default value. And that it is shown as active if the URL matches.

### Screenshots or Videos

_No response_

### Platform

- Router / Start Version: 1.170.33
- OS: Linux
- Browser: Brave
- Browser Version: 1.94.121
- Bundler: vite
- Bundler Version: 8.0.14

### Additional context

_No response_

devtechedge:
## Changes

Fixes #8309.

When `retainSearchParams` is listed before `stripSearchParams`, Links that do not pass a `search` prop were still building hrefs like `/?myParam=foo` even though the browser URL correctly omitted the default. That mismatch also prevented the Link from showing as active.

Root cause: `buildLocation` without a `search` option ends the middleware chain with `{}`. Object-form `stripSearchParams` only recorded removals for keys still present on that result, so an outer `retainSearchParams` put the validated defaults from `fromSearch` back into the href.

This patch marks default-valued keys as removed when they were present on the incoming search but dropped by `next()`, matching how array-form strip already behaves. Non-default retained params are unchanged.

Regressions cover:
- `buildLocation` with retain-then-strip and no `search` option
- retaining non-default values in that same order
- Link href omission + `data-status="active"`

## Checklist

## Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed link generation so default search parameters are omitted when search-parameter retention and stripping are used together.
  - Corrected active-link matching when default search values are validated but not included in the URL.
  - Preserved non-default search values during navigation.

- **Tests**
  - Added coverage for generated links, active states, and navigation behavior involving default search parameters.

### DIFF
--- a/.changeset/retain-strip-link-defaults.md
+++ b/.changeset/retain-strip-link-defaults.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/router-core': patch
+---
+
+Fix `retainSearchParams` before `stripSearchParams` so Link hrefs omit default search params instead of keeping them and breaking active matching.
--- a/packages/react-router/tests/link.test.tsx
+++ b/packages/react-router/tests/link.test.tsx
@@ -6293,6 +6293,65 @@ describe('search middleware', () => {
     expect(postsLink).toHaveAttribute('data-status', 'active')
   })
 
+  test('retainSearchParams before stripSearchParams omits defaults from Link href and stays active', async () => {
+    // Regression for https://github.com/TanStack/router/issues/8309
+    const defaults = { myParam: 'foo' }
+    const rootRoute = createRootRoute({
+      validateSearch: z.object({
+        myParam: z.string().default('foo'),
+      }),
+      search: {
+        middlewares: [retainSearchParams(true), stripSearchParams(defaults)],
+      },
+      component: () => {
+        const { myParam } = rootRoute.useSearch()
+        return (
+          <>
+            <div data-testid="search-value">{myParam}</div>
+            <Link data-testid="home-link" to="/">
+              Home
+            </Link>
+            <Link data-testid="about-link" to="/about">
+              About
+            </Link>
+            <Outlet />
+          </>
+        )
+      },
+    })
+    const indexRoute = createRoute({
+      getParentRoute: () => rootRoute,
+      path: '/',
+      component: () => <h1>Index</h1>,
+    })
+    const aboutRoute = createRoute({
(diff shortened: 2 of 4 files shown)


devtechedge:
Hello - gentle check-in on this router-core Link search fix whenever review time opens up.

When `retainSearchParams` is listed before `stripSearchParams`, Links without a `search` prop were still building hrefs like `/?myParam=foo` even though the browser URL correctly omitted the default, which also broke the active state (`Fixes #8309`).

Root cause: `buildLocation` without a `search` option ended the middleware chain with `{}`, so object-form `stripSearchParams` did not record removals that an outer `retainSearchParams` later put back from `fromSearch`.

This marks default-valued keys as removed when they were present on the incoming search but dropped by `next()`, with regressions for `buildLocation`, retained non-defaults, and Link href / `data-status="active"` behavior, plus a changeset.

Happy to rebase onto `main`, expand cases, or adjust the middleware notes if that helps - no rush from my side.

PR #8327, opened 2026-09-09
perf(router-core): streamline route initialization

schiller-manuel:
## 🎯 Changes

Reuse parsed route segments, simplify route-owned interpolation, stabilize path decoders, and initialize route trees directly with explicit indexes.

This is PR 12 of a stacked series and is based on `optimize-link-vue-match`.

## ✅ Checklist

- [x] I have tested code changes locally with the relevant test commands.

## 🚀 Release Impact

- [x] This change affects published code, and I have generated changesets.
- [ ] This change is docs/CI/dev-only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path parameter interpolation and encoding, including wildcard and optional parameters.
  * Preserved original parameter names and fallback route behavior across route-tree rebuilds.
  * Devtools now hides navigation actions when required path parameters are missing.

* **Performance**
  * Improved route-tree construction and path generation by reusing parsed route information.

* **Documentation**
  * Clarified that allowed path parameter characters must be configured when creating a router and cannot be changed afterward.


packages/router-core/src/new-process-route-tree.ts
```diff
@@ -647,11 +601,6 @@ export function findRouteMatch<
   return result
 }
 
-/** Trim trailing slashes (except preserving root '/'). */
-export function trimPathRight(path: string) {
```
schiller-manuel, line 651: uses the imported one instead that already existed

packages/router-core/tests/original-param-names.test.ts
```diff
@@ -0,0 +1,139 @@
...
+      { id: '/first', path: '$id/detail', fullPath: '/$id/detail' },
+      {
+        id: '/alias',
+        path: '$name',
+        fullPath: '/$name',
+        children: [
+          {
+            id: '/alias/detail',
+            path: 'detail',
+            fullPath: '/$name/detail',
```
Sheraff, line 41: why are these route ids not related to their paths? (e.g. id `/first` and path `/$id`)
Is that supposed to be valid?

packages/router-core/tests/path-decoder.bench.ts
```diff
@@ -0,0 +1,63 @@
+import { bench, describe, expect } from 'vitest'
+import { createMemoryHistory } from '@tanstack/history'
+import { BaseRootRoute, BaseRoute } from '../src'
+import { createTestRouter } from './routerTestUtils'
+
+if (process.env.TSR_LINK_PERF === '1') {
```
Sheraff, line 6: why is this behind a flag? there is not risk of executing it w/ the tests since it's a `.bench.ts` file, right?

schiller-manuel: it allows running these many tests only if wanted, so you can run the other benchmarks but not those. but I guess we could just use standard name based filtering instead?

packages/router-core/tests/route-interpolation.test.ts
```diff
@@ -0,0 +1,261 @@
...
+test.each([
+  { path: '/static', params: {}, missing: false },
+  { path: '/$id', params: {}, missing: true },
+  { path: '/$id', params: { id: undefined }, missing: false },
+  { path: '/$id', params: { id: '' }, missing: false },
+  { path: '/{-$id}', params: {}, missing: false },
+  { path: '/pre{-$id}suffix', params: { id: null }, missing: false },
+  { path: '/files/$', params: {}, missing: true },
+  { path: '/files/$', params: { _splat: '' }, missing: true },
+  { path: '/files/{$}.txt', params: { _splat: 0 }, missing: true },
```
Sheraff, line 23: are we sure this should count as missing? `_splat: 0` is provided. It's probably an edge case, but it doesn't seem impossible that some people might use numbers as params

packages/router-core/tests/path-decoder.bench.ts
Sheraff: yeah since we don't run benchmarks in CI, we never really run "all benchmarks" right? (just talking about `vitest bench` here, not codspeed obviously)

i personally would rather run them w/ name filtering, than have to remember that `TSR_LINK_PERF` is a thing and that sometimes i have to provide it

PR #8340, opened 2026-09-10
fix(start-server-core): correctly iterate response headers

Sheraff:
## 🎯 Changes

`setResponseHeaders(new Headers(...))` currently copies no headers because `Object.entries()` does not iterate a `Headers` object. Iterate its entries directly, replacing existing response values on the first occurrence and appending subsequent `Set-Cookie` values. Since `Headers` iteration groups entries by name, remembering the previous name avoids allocating a lookup object and handles headers named `constructor` or `__proto__` correctly. Passing the live object returned by `getResponseHeaders()` back into `setResponseHeaders()` returns immediately, preserving every cookie without mutating the object during iteration.

Credit to @rhyek (Carlos González) for the original fix and tests in #5537. This PR carries forward all five original tests, strengthens the cookie assertions, and adds coverage for header-name collisions, repeated cookie replacement, identical cookies, combined ordinary headers, preservation of unrelated headers, and passing the live response headers back to the setter.

Validation:

- Server-core: 181 unit tests passed; ESLint and TypeScript 5.6–7.0 checks passed.
- React Start production SSR: all 37 redirect tests passed, including multiple `Set-Cookie` preservation.
- Full bundle benchmark: all 18 client scenarios have unchanged raw, gzip, Brotli, and initial-load sizes against the original PR implementation on the same base.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed response header handling so existing values are replaced correctly.
  - Preserved multiple values for headers such as `Set-Cookie`.
  - Improved handling of headers with special names and unrelated existing headers.
  - Ensured response headers remain available and accurate when updated or retrieved.

- **Tests**
  - Added coverage for setting, replacing, combining, and retrieving response headers and cookies.

### DIFF
--- a/packages/start-server-core/src/request-response.ts
+++ b/packages/start-server-core/src/request-response.ts
@@ -218,6 +218,9 @@ export function setResponseHeaders(
   headers: TypedHeaders<ResponseHeaderMap>,
 ): void {
   const event = getH3Event()
+  if (event.res.headers === headers) {
+    return
+  }
   let previousName = ''
   // Headers iteration groups entries by name, including separate Set-Cookie values.
   for (const [name, value] of headers) {
--- a/packages/start-server-core/src/request-response.ts
+++ b/packages/start-server-core/src/request-response.ts
@@ -218,8 +218,11 @@ export function setResponseHeaders(
   headers: TypedHeaders<ResponseHeaderMap>,
 ): void {
   const event = getH3Event()
-  for (const [name, value] of Object.entries(headers)) {
-    event.res.headers.set(name, value)
+  let previousName = ''
+  // Headers iteration groups entries by name, including separate Set-Cookie values.
+  for (const [name, value] of headers) {
+    event.res.headers[name === previousName ? 'append' : 'set'](name, value)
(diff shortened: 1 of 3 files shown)


packages/start-server-core/src/request-response.ts
```diff
@@ -218,8 +218,11 @@ export function setResponseHeaders(
...
 ): void {
   const event = getH3Event()
-  for (const [name, value] of Object.entries(headers)) {
-    event.res.headers.set(name, value)
+  let previousName = ''
+  // Headers iteration groups entries by name, including separate Set-Cookie values.
+  for (const [name, value] of headers) {
+    event.res.headers[name === previousName ? 'append' : 'set'](name, value)
+    previousName = name
   }
```
Sheraff, line 229: Maybe we should early return if both header objects are the same object
```
if (event.res.headers === headers) return
```
this would avoid issues where callers pass the live object returned by `getResponseHeaders()` back into `setResponseHeaders()`

PR #8353, opened 2026-09-11
perf(start-client-core): avoid unused middleware array

LadyBluenotes:
## 🎯 Changes

Use an indexed loop instead of `map` when registering server-function middleware. This avoids the unused result array and the iteration callback. Composition order, factory expansion, and sparse-array handling stay unchanged. The loop captures the initial array length, so entries appended during registration are not visited. Request execution is untouched.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server-function middleware registration for consistent ordering and reliable handling of nested, frozen, empty, and sparse middleware lists.
  * Middleware added during registration is no longer unexpectedly included.

* **Tests**
  * Added coverage for middleware concatenation, input preservation, and edge-case array behavior.

### DIFF
--- a/.changeset/start-middleware-registration.md
+++ b/.changeset/start-middleware-registration.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/start-client-core': patch
+---
+
+Avoid an unused array when registering server-function middleware.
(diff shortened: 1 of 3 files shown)


Sheraff:
would a for loop be even better than using array methods?

LadyBluenotes:
> would a for loop be even better than using array methods?

yeah youre right. ill update

PR #8382, opened 2026-09-11
perf(router-core): stop structurally sharing search and state in buildLocation

schiller-manuel:
Stacked on #8380.

## 🎯 Changes

`buildLocation` ran `nullReplaceEqualDeep` over the built `search` and `replaceEqualDeep` over the built `state` against the current location on every build. That work only affected object identity, and nothing reads that identity from a *built* location:

- Links select `href` and `isActive` from the built location; active state is computed by value and ignores history state.
- `commitLocation` already compares state by value (`deepEqual` over `_getUserHistoryState`), so pre-shared children were at most a shortcut there.
- The identity consumers actually rely on (selecting `location.state.user` or a nested search value without rerendering) comes from `parseLocation`, which stabilizes the committed location against the previous one. That sharing is untouched, as are the `matchRoutes` sites for search, params and loaderDeps. Both helpers stay imported and used.

On the server `replaceEqualDeep` was already a passthrough, so client and server builds now produce the same shapes.

### Removed lines (`packages/router-core/src/router.ts`, inside `buildLocation`'s `build`)

```ts
// Structural sharing only affects identity, so it does not make the
// location depend on the current one.
nextSearch = nullReplaceEqualDeep(lightweight[2 /* search */], nextSearch)
```

```ts
// Identity-only, as above.
nextState = replaceEqualDeep(currentLocation.state, nextState)
```

`nextSearch` became a `const`; `nextState` is now a single `const` conditional (`EMPTY_RECORD` without `dest.state`, `current().state` for `true`, `dest.state(current().state)` for a function, otherwise `dest.state`). The `usedCurrent` tracking is unchanged: only reads through `current()` / `currentMatch()` mark a build as location-dependent, so literal search/state keep hitting the per-options cache from #8370. `lightweight[2]` is still returned for `fromSearch()`.

### Behavior changes

1. **No build-time identity sharing.** A literal `search` or `state` is returned as the caller's object; `buildLocation(...).search.nested` is no longer reference-equal to the current location's nested value. After the navigation commits, `router.state.location.search.nested` / `.state.nested` *do* keep the previous reference (via `parseLocation`).
2. **Key order.** Because `replaceEqualDeep` returned the *current* search when contents were equal, `buildLocation` used to re-serialize the current key order. At `?a=1&b=2`, `buildLocation({ search: { b: 2, a: 1 } })` produced `?a=1&b=2`, so navigating there was a same-location no-op. It now produces `?b=2&a=1` and pushes a new history entry.
3. **No caller mutation.** The caller's `state` object flows straight to `commitLocation`; that path already copies before adding `__hashScrollIntoViewOptions`, and `@tanstack/history` copies before adding `__TSR_key`/`key`/index. Pinned by tests (plain and frozen state objects).
4. A destination without `search` yields the frozen `EMPTY_RECORD` rather than a fresh or shared object. Nothing downstream mutates the built search (all consumers copy); the full unit suites pass.

### Tests

- Removed `explicit state structurally shares unchanged nested values` (asserted build-time sharing).
- Added `describe('buildLocation - no structural sharing with the current location')`:
  - built `state` is the caller's object; after `navigate`, `location.state.user` keeps the previous reference
  - same for `search` with a nested `filter` subtree
  - `navigate` does not mutate a caller-supplied `state` (keys unchanged; a frozen state commits without throwing)
  - key order at the `buildLocation` level (`searchStr`/`href` follow the requested order)
  - key order at the navigation level (same order → `history.length` unchanged; different order → new entry with `?b=2&a=1`)
- 4 of the 5 new tests fail against the previous implementation; the mutation test is a regression guard.
- No other test in the repo asserted build-time sharing or depended on the old key-order behavior (`utils.test.ts` line 951 tests `replaceEqualDeep` itself and is untouched).

### Measurements

`pnpm benchmark:bundle-size:run --scenario react-router.minimal`: gzip **86026 → 86004 (−22)**, initial −23, raw −41, brotli +65.

### Verification

- `@tanstack/router-core:test:unit` 133 files / 3303 passed
- `@tanstack/react-router:test:unit` 91 files / 1174 passed
- `@tanstack/solid-router:test:unit` 68 + 7 files / 942 + 51 passed
- `@tanstack/vue-router:test:unit` 71 files / 968 passed
- `@tanstack/router-core:test:types`, `@tanstack/router-core:test:eslint` (0 errors), `git diff --check`, prettier

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - `buildLocation` no longer shares `search` and `state` objects with the current location before navigation is committed.
  - Navigation and location-building operations no longer mutate caller-provided state, including frozen objects.
  - Search key order is preserved, and equal searches with different key orders create distinct history entries.

- **Tests**
  - Added coverage for state isolation, nested reference stabilization after navigation, search ordering, and caller-state immutability.


Sheraff:
@codex if it is not shared in buildLocation, where is it shared now to ensure reference stability?

Sheraff (approved):

PR #8390, opened 2026-09-12
perf(router): keep the location cache client-only and slim Link SSR

schiller-manuel:
Stacked on #8385.

### Why

- The per-options `buildLocation` cache (`staticLocations`, from the location-reuse layers) only ever hits on the client: server renders never repeat an options object. Yet every server build still created the `WeakMap`, looked it up per `buildLocation`, and Solid/Vue server Links (which pass a fresh `{ _fromLocation, ...options }` per render) stored an entry per render.
- Rendering a React `Link` on the server copied its props four times: `Link` split off `_asChild` with an object rest, `useLinkProps` split the router options off with another object rest, the result was assembled with spreads, and `Link` copied once more to drop `type`/`disabled`. Profiling showed the object rest alone was ~85% of the hook's self time: V8 checks every key of an object rest against the whole 35-entry exclusion list (~470 ns per call vs ~95 ns for a key-set copy).

### What changed

1. **router-core: `staticLocations` is client-only.** The field is now `WeakMap | undefined`; its creation in `update()` and `setRoutes()`, the read at the top of `buildLocation`, and the write at its end are all guarded by `!(isServer ?? this.isServer)`. `isServer` from `@tanstack/router-core/isServer` is a per-bundle constant (`false` in browser builds, `true` in server builds), so bundlers fold the guards: server bundles contain no cache at all, client bundles keep the cache logic unchanged.
2. **react-router: slimmer `Link` server render, no extra client copies.** `useLinkProps` is now a thin wrapper over `useLinkPropsFor(options, ref, host?)`. `Link` calls it with its host (`'a'` or the `createLink` component), so the hook omits `disabled` for the anchor host and `type` for both hosts, and `Link` passes the result straight to `createElement` (no copies). The server branch is a separate `getServerLinkProps` referenced only inside `if (isServer ?? router.isServer)`, so it and its `LINK_OPTION_KEYS` set are dropped from client bundles. It reads the few options it needs directly, splits element props with the key set, fills the props object in place with the same precedence and attribute order as the previous spreads, and no longer calls `useForwardedRef` (that hook now sits after the server return; a server render never re-renders and the returned `ref` is the forwarded ref). The client path keeps its original rest destructure (plus `_asChild`/`type` extraction and `host !== 'a'` around `disabled`).
3. **router-core `encodePathLikeUrl`** tests with one merged character class `/[\s\u0080-\uFFFF]/` instead of the alternation `/\s|[^\u0000-\u007F]/`. Verified identical `test()` results for every UTF-16 code unit (0x0000–0xFFFF) plus surrogate-pair samples; the replacement regex is unchanged. About 10% cheaper.
4. **react-router: one state-prop precedence rule for every Link** (second commit, prompted by the review question on the `blockedLink` branch). `activeProps`/`inactiveProps` used to override `ref` and the event handlers on every link *except* a blocked one (destination with a disallowed scheme), where the forwarded ref and the router's handlers won. That exception was a leftover of the hardening change's original "router wins" ordering, kept alive by a test when state props were later allowed to override element props. It bought no safety (`href`, `disabled`, `target` are assigned after the state props in both orders, which is what keeps a blocked destination inert), it made the same `inactiveProps` attach their `ref`/`onClick` on one inactive link and silently drop them on another, and it differed from Solid's Link. Blocked links now follow the same rule as every other link; the server loses its if/else, the client loses two conditional spreads and the `blockedLink` flag. The blocked-link test pins the uniform rule together with the routing attributes it already checked. Noted in the changeset.

### Measurements (local, macOS arm64 / Node 24.8.0 — not CI)

Paired-runner and Start SSR numbers for the first commit are the stack author's local measurements:

- Link SSR paired runner (`@benchmarks/react-link-performance:test:perf:stable --mode ssr`, ABBA blocks, 4 fresh processes per case) against a dist snapshot of the previous stack top: all 13 cases "faster", CPU −19.5% .. −40.3% (shared-params −39.4%, unique-params −40.3%, param-updaters −36.4%, location-updaters −23.8%, relative −23.6%, middleware −19.5%, numeric-params −37.5%, optional-params −38.0%, splats −31.2%, encoding −21.1%, masks −29.4%, rewrites −20.7%, active −32.9%).
- Link client paired runner (`--mode client`): all 13 cases "faster", CPU −4.6% .. −9.2% (from the removed `Link` copies).
- Start SSR request loop (`benchmarks/ssr` react baseline app, interleaved builds, 4+4 slots, rme < 1.2%): 3.120 → 2.890 ms per 10-request loop (−7.4%), 320.5 → 346.1 hz (+8.0%).

Second commit (uniform precedence), paired runner with 3 repeats against the first commit:

- Link client: shared-params CPU −11.6% [−15.1, −7.9], encoding −13.2% [−20.2, −5.7] (two fewer spreads per render).
- Link SSR: within noise.

Bundle sizes measured in this PR's worktree with the official runner (`benchmark:bundle-size:run`, cache-free rebuilds on both sides):

- `react-router.minimal` gzip: **86012 → 86019 (+7)**; raw −31, brotli −55. Per-hunk attribution: the router-core hunks alone are **−12** gzip (the guards fold away and the regex is shorter); the `link.tsx` changes cost the remainder (the `_asChild`/`type` rest exclusions, the `type` re-add for the public hook and the `host !== 'a'` guards are not fully paid back by the removed `Link` copies). The second commit is −2 gzip on its own.
- Link SSR benchmark bundle (`benchmarks/client-nav/link-performance/dist/ssr/app.js`): **197832 → 189850 bytes**, because the client-only Link code no longer ships to the server.

### Bundle-content assertions (verified after `TSR_LINK_PERF=1 pnpm nx run-many --target=build:ssr,build:client --projects=@benchmarks/react-link-performance`)

- `dist/ssr/app.js`: **0** `staticLocations` references (was 4), **contains** `getServerLinkProps` and the `LINK_OPTION_KEYS` set, **0** `useForwardedRef` references.
- `dist/client/app.js`: **4** `staticLocations` references (unchanged), **0** `getServerLinkProps`, **0** `"preloadIntentProximity"` string literal (the key set is gone; the two remaining `preloadIntentProximity` occurrences are the client rest destructure, as before).

### Tests run

- `@tanstack/router-core:test:unit` — 133 files, **3303 passed | 4 expected fail**
- `@tanstack/react-router:test:unit` — 91 files, **1174 passed | 1 skipped** (re-run after the second commit)
- `@tanstack/solid-router:test:unit` — 942 passed | 1 skipped (+ 51 in the second project)
- `@tanstack/vue-router:test:unit` — 968 passed | 1 skipped
- `test:types` + `test:eslint` for `@tanstack/router-core` and `@tanstack/react-router` — types clean, eslint 0 errors (only pre-existing warnings, none in `link.tsx`, none on changed lines of `router.ts`/`utils.ts`)
- `git diff --check` and prettier on all changed files — clean

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link behavior during server rendering by avoiding unnecessary location caching and prop handling.
  * Standardized active and inactive link properties across regular, custom, and blocked links.
  * Ensured router-controlled attributes such as `href`, `disabled`, and `target` are applied consistently.
  * Prevented unsupported attributes from being forwarded to standard anchor elements.


packages/react-router/src/link.tsx
```diff
@@ -546,6 +553,130 @@ function resolveStateProps(
...
+    resolveStateProps(isActive, activeProps, inactiveProps, className, style)
+
+  // State props may override `ref`, but not on a blocked link (assigned first).
+  if (blockedLink) {
+    Object.assign(props, resolvedStateProps)
+    props.ref = forwardedRef
+  } else {
+    props.ref = forwardedRef
+    Object.assign(props, resolvedStateProps)
+  }
```
Sheraff: why does this distinction matter?

packages/react-router/src/link.tsx
```diff
@@ -515,8 +482,48 @@ export function useLinkProps<
 
 const STATIC_EMPTY_OBJECT = {}
 const STATIC_ACTIVE_OBJECT = { className: 'active' }
-const STATIC_DISABLED_PROPS = { role: 'link', 'aria-disabled': true }
-const STATIC_ACTIVE_PROPS = { 'data-status': 'active', 'aria-current': 'page' }
+// Options consumed by the router and never forwarded to the element.
+const LINK_OPTION_KEYS = /* @__PURE__ */ new Set([
```
Sheraff, line 484: i'm assuming this was verified to only exist in the server build?

schiller-manuel: yes. not leaking into the client since only referenced in the server branch from getServerLinkProps

packages/react-router/src/link.tsx
schiller-manuel: simplified

Sheraff (approved):

PR #8404, opened 2026-09-12
perf(react-router): fewer hooks per Link

schiller-manuel:
Stacked on #8401.

## 🎯 Changes

A `Link` render ran about 25 hook slots. React allocates a hook object per slot per render, plus a deps array for every `useMemo`/`useCallback`/`useEffect`, so the slot count sets both the re-render cost and the retained fiber memory of a mounted `Link`. Five of those slots did not need to exist:

1. The three `useValueStable` refs for `search`, `params` and `activeOptions` become one `useStableValues` ref. Each entry is replaced only when its own contents change, so the returned identities behave exactly as before (including the `ignoreUndefined: false` semantics that distinguish `{}` from `{ category: undefined }`).
2. `_options` and its `dest` copy come out of one `useMemo` instead of two.
3. `doPreload` becomes a module function `preloadLink(router, options)`; the `useCallback` existed only to give the effects a stable identity.
4. The IntersectionObserver effect and the render-preload effect merge into one preload effect whose cleanup also cancels a pending intent timer, exactly as the observer hook's cleanup did. `useIntersectionObserver` in `utils.ts` had no other consumer in `@tanstack/react-router` and is removed (the `solid-router`/`vue-router` packages keep their own separate copies).

`hasRenderFetched` stays: without it a link with an updater-function `search` would preload on every render.

Public API is unchanged. Touches only `packages/react-router/src/link.tsx` and `packages/react-router/src/utils.ts` (+60/−97).

### Measurements (local, macOS arm64, Node 24.8.0, 200 mounted Links, vs. the previous commit)

- Retained heap per mounted app: 2949 → 2782 KB (−6%, about −0.8 KB per Link)
- Mount (router creation + first render) median: 2.05 → 1.90 ms
- Link client paired runner (4 repeats): shared-params CPU −3.0% [−5.3, −0.6] faster; active −3.3% [−8.4, 2.1] inconclusive
- `react-router.minimal` gzip: 85984 → 85968 (−16 bytes)

### Bundle-content assertions (verified locally)

Built with `TSR_LINK_PERF=1 pnpm nx run-many --target=build:ssr,build:client --projects=@benchmarks/react-link-performance --skipNxCache`:

- `benchmarks/client-nav/link-performance/dist/ssr/app.js`: contains **no** `staticLocations`, **no** `useIntersectionObserver`, and **does** contain `getServerLinkProps` ✅
- `benchmarks/client-nav/link-performance/dist/client/app.js`: **contains** `staticLocations`, **no** `getServerLinkProps`, **no** `useIntersectionObserver` ✅

### Tests run (all with `--skipNxCache`)

- `@tanstack/react-router:test:unit` — 91 files, **1174 passed, 1 skipped**, no type errors
- `@tanstack/react-router:test:types` — passes on TS 5.6, 5.7, 5.8, 5.9, 6.0 and 7.0
- `@tanstack/react-router:test:eslint` — **0 errors**, 100 pre-existing warnings (none in `link.tsx` or `utils.ts`)
- `git diff --check HEAD~1` — clean
- `pnpm exec prettier --check packages/react-router/src/link.tsx packages/react-router/src/utils.ts` — clean

## ✅ Checklist

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md). — The existing changeset `.changeset/brisk-links-serve.md` from the lower layers of this stack already declares a `@tanstack/react-router` patch and covers this change; no new changeset is added.

## Summary by CodeRabbit

* **Performance**
  * Improved link preloading behavior for render, viewport, delayed, and touch interactions.
  * Improved stability when link destinations, search values, parameters, and active options change.


packages/react-router/src/link.tsx
```diff
@@ -347,26 +350,38 @@ function useLinkPropsFor<
...
+    ) {
+      observer = new IntersectionObserver(
+        (entries) => enqueuePreload(entries.pop()),
+        { rootMargin: '100px' },
+      )
+      observer.observe(innerRef.current)
+    }
+    return () => {
+      observer?.disconnect()
+      enqueuePreload()
```
Sheraff, line 382: why do we `enqueuePreload` on cleanup?

Sheraff: My bad, it is also the cleanup function
```ts
     if (!e) {
        cancelPreload(innerRef)
        return
      }
```

Sheraff (approved):

PR #8405, opened 2026-09-12
perf(react-router): memoize Link and forward refs without an effect

schiller-manuel:
Stacked on #8404.

## 🎯 Changes

When a layout re-renders on navigation, for instance because it reads `useLocation`, every Link below it re-renders too. Profiling 200 mounted Links showed that re-render dominated by React's per-hook update work, the element creation and the anchor's prop diff, not by the Link's own logic.

A Link's output depends only on its props, the router context and the location store, all of which React tracks for memoized components, so `Link` is now wrapped in `React.memo`. Router options (`ROUTER_OPTION_KEYS`) are compared by value, because destinations are usually inline object literals; element props are compared by reference, since they may hold arbitrary, even cyclic, user data.

The memo adds one fiber per Link. To pay for it, `useForwardedRef` is removed: its `useImperativeHandle` was the only layout effect on a Link and ran for every Link although most never receive a ref. The link keeps a plain `useRef`; when a ref is forwarded, one callback fills both. That callback is memoized on the forwarded ref, so a consumer's ref is notified on mount, when it changes and on unmount, exactly as with `useImperativeHandle`'s `[]` deps. The hook had no other consumer and is deleted from `utils.ts` (it was never re-exported from the package index). `enqueuePreload` no longer doubles as the cleanup path: the preload effect cancels the intent timer directly.

A new test, `tests/link-memo.test.tsx`, pins that a parent re-render with equal props (new inline objects) does not re-render the Link, and that a changed destination does. It fails without the memo with `expected 2 to be 1`. It also pins that element props are compared by reference only (a new JSX child re-renders, a cyclic element prop does not throw), that a stable callback ref is notified once per element rather than per Link render, and that a cleanup returned by a callback ref runs on unmount.

Public API unchanged.

### Measurements (local: macOS arm64, Node 24.8.0, against the previous commit)

- Link client paired runner, 200 Links under a re-rendering layout (3 repeats): shared-params CPU **-54%**, masks **-56%**, rewrites **-60%** (memo hits). Links whose props can never compare equal (updater functions, inline `activeProps` functions): param-updaters +2.5% [-5.2, 10.9], location-updaters -1.0%, active -2.0%, all inconclusive (the extra fiber and the removed layout effect cancel out).
- Retained heap per mounted app 2782 -> 2738 KB; mount unchanged within noise.
- `react-router.minimal` gzip 85968 -> 86004 (+36); the branch stays 8 bytes under the stack base (86012).
- Ref mechanism in isolation (200 components, production React 19.2.3): mount is ~5% cheaper without the `useImperativeHandle` layout effect; the 200-Link benchmark app mounts 4.97 -> 4.69 ms and retains 23 KB less. Memoizing the merged callback instead of recreating it per render costs +2.5% [1.4, 3.7] on `param-updaters` and +1.4% [0.4, 2.5] on `relative` (every Link re-renders there); the other workloads are inconclusive, and against `useImperativeHandle` all 13 are equal or better.

### Bundle-content assertions (verified on this branch)

Built with `TSR_LINK_PERF=1 pnpm nx run-many --target=build:ssr,build:client --projects=@benchmarks/react-link-performance`:

- `benchmarks/client-nav/link-performance/dist/ssr/app.js`: contains **no** `staticLocations`, **no** `useForwardedRef`, and **does** contain `getServerLinkProps`.
- `benchmarks/client-nav/link-performance/dist/client/app.js`: **contains** `staticLocations` and `areLinkPropsEqual`, and **no** `getServerLinkProps`, **no** `useForwardedRef`.

### Tests run

- `@tanstack/react-router:test:unit`: 92 files, 1178 passed, 1 skipped
- `@tanstack/react-router:test:types`: pass (TS 5.6, 5.7, 5.8, 5.9, 6.0, 7.0)
- `@tanstack/react-router:test:eslint`: 0 errors (100 pre-existing warnings, none in the touched files)
- `git diff --check HEAD~1`: clean
- `pnpm exec prettier --check` on `src/link.tsx`, `src/utils.ts`, `tests/link-memo.test.tsx`: clean

## ✅ Checklist

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md). The existing changeset `.changeset/brisk-links-serve.md` from the lower layers of this stack already covers `@tanstack/react-router` as a patch, so no new changeset is added.

## Summary by CodeRabbit

- **Performance**
  - Improved link rendering efficiency by avoiding unnecessary updates when link properties remain unchanged.
  - Link destinations and active styling continue to update correctly when relevant route parameters change.

- **Bug Fixes**
  - Improved handling of link targets, styles, classes, and preload visibility events for more consistent navigation behavior.
  - Improved forwarded ref handling, including reliable cleanup when links are replaced or unmounted.


packages/react-router/src/link.tsx
```diff
@@ -379,12 +370,21 @@ function useLinkPropsFor<
...
+  // A forwarded ref is filled alongside the link's own ref.
+  props.ref = forwardedRef
+    ? (element: Element | null) => {
+        innerRef.current = element
+        if (typeof forwardedRef === 'function') {
+          return forwardedRef(element)
+        }
+        forwardedRef.current = element
+      }
+    : innerRef
```
Sheraff: Is this really better than the native `useImperativeHandle`?

packages/react-router/src/link.tsx
```diff
@@ -821,9 +822,38 @@ export const Link: LinkComponent<'a'> = React.forwardRef<Element, any>(
...
+  next: Record<string, unknown>,
+): boolean {
+  let extraKeys = 0
+  for (const key in next) {
+    extraKeys++
+    if (
+      prev[key] !== next[key] &&
+      !(
+        ROUTER_OPTION_KEYS.has(key) &&
+        deepEqual(prev[key], next[key], { ignoreUndefined: false })
```
Sheraff: this will deep compare react `children`, right? isn't there a  chance that this might be very expensive and almost always return that they are not equal?

packages/react-router/src/link.tsx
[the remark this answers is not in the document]
schiller-manuel: Good catch, this was a real regression against `useImperativeHandle`. With the inline closure every Link render produced a new ref, so React detached and re-attached it: a consumer's callback ref was called with `null` and the element again on every location-driven re-render (3 calls after one navigation instead of 1), and an object ref was cleared and refilled inside the commit.

Fixed in c63f63d: the merged callback is memoized with `useCallback` on `forwardedRef`, which restores the `useImperativeHandle(ref, …, [])` contract (notified on mount, when the consumer's ref changes, on unmount). Test added in `link-memo.test.tsx`; it fails on the previous commit with `expected 1 call, got 3`.

I did not take the cleanup wrapper that clears `innerRef.current`. Nothing reads `innerRef.current` after a detach without a re-attach: the viewport observer reads it in an effect that only runs while the element is attached, the intent timeout map is keyed by the ref object rather than its value, and React re-attaches the same memoized callback before any effect can run again. A consumer's cleanup is still passed through to React unchanged. The wrapper measured +19 gzip on `react-router.minimal` for no observable behavior; the memoization alone is neutral (−4 together with the comparator tweak from the other thread).

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

schiller-manuel: Fair question, so I measured it in isolation (production React 19.2.3, jsdom, macOS arm64, Node 24.8, medians).

**Isolated mechanism**, 200 minimal `forwardRef` components with the same hooks except the ref one:

| variant | mount | update | callback-ref calls per update (200 comps) |
|---|---|---|---|
| `useImperativeHandle(ref, () => inner.current, [])` (main) | 618 µs | 196 µs | 0 |
| inline merged callback (this PR before c63f63d) | 580 µs (−6%) | 188 µs | **400** |
| `useCallback`-memoized merged callback (now) | 588 µs (−5%) | 193 µs | 0 |

**Real 200-Link benchmark app** (`@benchmarks/react-link-performance`, `shared-params`, 60 runs): mount 4.97 ms with `useImperativeHandle` → 4.69 ms memoized (−5.7%), retained heap −23 KB per mounted app (≈115 B per Link). On the navigation workloads the paired runner can't tell them apart (12/13 inconclusive, `rewrites` −4.6% [−9.0, −0.2]).

So per Link it's ~1 µs at mount and a few allocations per update. Small, but it is the better mechanism for structural reasons:

1. `useImperativeHandle` is a **layout effect**. Every Link fiber got `Update | LayoutStatic` at mount and was visited in the synchronous layout phase before paint, even for the ~99% of Links that never receive a ref (the effect is registered unconditionally; for a null ref it runs and does nothing). It was the only layout effect on a Link; now there is none. In dev StrictMode it was also double-invoked.
2. Per render it does `deps.concat([ref])` (two arrays), `imperativeHandleEffect.bind(null, create, ref)`, the `create` closure and an effect node, on both fibers of the pair — that is the retained heap above. `useCallback` is one hook slot with a one-element deps array.
3. It is a misuse of the API: `useImperativeHandle` exists to expose a *custom* handle. Forwarding the DOM node is what a ref does natively; merging two refs is the textbook callback-ref case.
4. It is slightly more correct. `useImperativeHandle(…, [])` snapshots `innerRef.current` once; if a `createLink` host swaps its root element without the Link re-rendering, the consumer's ref kept pointing at the old node. A callback ref on the host element follows the actual element.

CodeRabbit was right that the inline version I first pushed lost the `[]`-deps semantics (the 400 calls above). That's fixed in c63f63d with `useCallback` on `forwardedRef`, which is exactly the `[ref]` dep `useImperativeHandle` used internally. Cost of that hook vs. the inline version: +2.5% [1.4, 3.7] on `param-updaters` and +1.4% [0.4, 2.5] on `relative`, the two workloads where every Link re-renders per navigation; the other 11 are inconclusive. Against `useImperativeHandle` it is equal or better everywhere.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

packages/react-router/src/link.tsx
schiller-manuel: No — `children` never reaches `deepEqual`. The condition was `prev[key] !== next[key] && !(ROUTER_OPTION_KEYS.has(key) && deepEqual(...))`, so for any key that isn't a router option (`children`, `onClick`, `style`, `className`, `data-*`, …) the inner `&&` short-circuits before `deepEqual` and a changed reference is simply "not equal" → re-render. That's the standard `React.memo` behavior: a JSX child is a new element each parent render and the Link re-renders; a string child compares equal by value.

Only `ROUTER_OPTION_KEYS` fall through to `deepEqual`: `to`, `params`, `search`, `hash`, `state`, `mask`, `activeOptions`, `activeProps`, … — the router's own inputs, which the Link already deep-compares on every render in `useStableValues` (and which `buildLocation` used to `replaceEqualDeep` on main). Scalars among them exit `deepEqual` on the first two checks and functions (`search={(prev) => …}`) return false immediately, so the recursion only happens for the inline `params={{ id }}` / `search={{ page }}` literals it exists for.

Why value comparison at all: those literals are new objects on every parent render, so a plain shallow `memo` would never hit for the majority of real Links — the −54% / −56% / −60% in the description are all memo hits on inline destinations. A one-level shallow compare would be almost the same code path (`deepEqual` on a flat object of primitives *is* a shallow compare) but would miss nested `search` values, which is exactly where people use inline literals. If the destination is hoisted (`linkOptions(...)` outside render, or the React Compiler memoizing the literal), `deepEqual` short-circuits on `===` and costs nothing.

Since the nested condition clearly read wrong, c63f63d restructures the loop into two `if`s (reference check → `continue`; non-router key or `deepEqual` false → `return false`; −2 gzip), and adds a test that a new JSX child re-renders the Link and that a cyclic element prop does not throw (it would if anything deep-compared it).

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

packages/react-router/src/link.tsx
```diff
@@ -821,9 +830,39 @@ export const Link: LinkComponent<'a'> = React.forwardRef<Element, any>(
...
+): boolean {
+  let extraKeys = 0
+  for (const key in next) {
+    extraKeys++
+    if (prev[key] === next[key]) {
+      continue
+    }
+    if (
+      !ROUTER_OPTION_KEYS.has(key) ||
+      !deepEqual(prev[key], next[key], { ignoreUndefined: false })
```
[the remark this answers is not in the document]
schiller-manuel: IMO cyclic values inside activeProps dont make any sense. prove me wrong?

packages/react-router/src/link.tsx
Sheraff: but links will always have children, so this optimization only applies for the cases where children are pure strings. That feels a bit narrow, but maybe it's fine if memo is cheap enough (and for people using the compiler, this is probably nice)

packages/react-router/src/link.tsx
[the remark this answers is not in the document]
schiller-manuel: Not taking this one. The input it needs is a plain-object graph with back-references that is *rebuilt on every render* and passed inside `activeProps` of a custom link. `deepEqual` only recurses through plain objects and arrays, so class instances, DOM nodes, Maps, or React elements (their `_owner` is a Fiber) stop it, and a graph that keeps its identity (state, store, module scope) short-circuits on `===` before any recursion. `state` has had exactly the same constraint on main (`replaceEqualDeep` in `buildLocation`, `deepEqual` in `commitLocation`), so the router already treats cyclic plain objects as unsupported in these positions.

Both alternatives cost more than the bug they prevent. Comparing the two bags by reference drops the memoization for the most common pattern, `activeProps={{ className: 'font-bold' }}`: those Links would re-render on every parent render again. A one-level comparison keeps the hit but measured +46 gzip and adds a special case to the comparator (I tried it). So `activeProps`/`inactiveProps` stay compared by value, like the other router options.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

packages/react-router/src/link.tsx
schiller-manuel: Right, a hit needs `children` to be referentially equal, so out of the box that is strings and numbers, hoisted static elements, anything the React Compiler memoizes, and every Link whose parent did not re-render at all (the location subscription still bails per Link). That covers most nav, menu and breadcrumb links; content links with an icon child will miss.

What makes the miss acceptable is its price. The comparator is a `for…in` with `===` per key; only inline `params`/`search`/`activeProps` objects fall through to `deepEqual` on a few flat keys. Measured on a 6-key props object (`to`, inline `params`, `search`, `activeProps`, `preload`, `children`): ~140 ns per call, the same for a string-children hit and an element-children miss (`children` is the last key), versus a Link re-render in the microseconds. On the paired runner, the workloads whose props can never compare equal (updater functions, render children, inline `activeProps` functions) show the whole memo overhead as noise: param-updaters +2.5% [−5.2, 10.9], location-updaters −1.0%, active −2.0%, all inconclusive, while the hits are −54% to −60%. And with the Compiler the JSX children get a stable identity, so it becomes a hit there too.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

PR #8412, opened 2026-09-13
fix(start): don't log aborted server function requests as errors

oztune:
### Problem

`serverFnFetcher`'s `getResponse` logs **every** non-`Response` error before rethrowing:

```ts
} catch (error) {
  if (error instanceof Response) {
    response = error
  } else {
    console.log(error) // <- logs even expected cancellations
    throw error
  }
}
```

This includes the `AbortError` thrown when the caller **cancels** a server-function request, which is normal, expected control flow — not a failure. The most common trigger is TanStack Query aborting a query when its component unmounts: navigating away while a server-fn-backed query is in flight spams the console with

```
AbortError: signal is aborted without reason
```

even though nothing is wrong — the request was intentionally cancelled.

### Fix

Skip the log when the error is an abort (`DOMException` or any `Error` with `name === 'AbortError'`). The error is still **rethrown**, so callers (e.g. TanStack Query) handle the cancellation exactly as before, and genuine errors are still logged.

### Tests

Added `serverFnFetcher` unit tests covering both branches:
- an aborted request is **not** logged, but is still rethrown
- a genuine error **is** logged and rethrown

A changeset is included (`@tanstack/start-client-core`, `patch`).

## Summary by CodeRabbit

- **Bug Fixes**
  - Aborted server-function requests no longer appear as logged errors.
  - Genuine server-function errors continue to be logged and rethrown.
  - Aborted requests and other errors preserve their original error details.


packages/start-client-core/src/client-rpc/serverFnFetcher.ts
```diff
@@ -238,7 +252,11 @@ async function getResponse(fn: () => Promise<Response>) {
     if (error instanceof Response) {
       response = error
     } else {
-      console.log(error)
+      // A caller aborting the request is expected control flow, not a failure, so don't log it.
+      // We still rethrow so the caller can handle the cancellation.
+      if (!isAbortError(error)) {
```
Sheraff, line 257: Probably the only test we need is this 
```suggestion
      if (!(error instanceof Error) || error.name !== 'AbortError') {
```

PR #8422, opened 2026-09-14
perf(router-core): parse the location once in update and build rewrites without arrays

schiller-manuel:
## 🎯 Changes

- update(): compute the basepath/rewrite before the single updateLatestLocation
  call; the first update always saw basepathWasUnset, so every router
  construction parsed the location twice and re-published to a store created
  from the same location.
- invalidate(): fill the id set in one pass instead of spread/flat/filter/map.
- rewriteBasepath(basepath, caseSensitive?) takes positional arguments (it is
  internal) and the output rewrite uses cleanPath on a template string instead
  of joinPaths' array/filter/join.
- parseLocation(): inline the single-use fullPath local.

react-router.minimal 86019 -> 85989 gzip, react-router.full 89605 -> 89561 gzip.

### Verification on `main`

- router-core `test:unit` (3254 passed, including the new `rewriteBasepath` case-sensitivity and slash-normalization tests), `test:types`, `test:eslint`
- start-server-core `test:types` (it imports `executeRewriteInput`, unchanged here)

### Relationship to other PRs

Independent (applies to `main` byte-identically to the original commit). Touches `router.ts` regions (`update`, `invalidate`, `parseLocation`) that other PRs in this series do not.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved routing updates when the base path, rewrite configuration, or case-sensitivity setting changes.
  * Ensured locations are parsed consistently after routing configuration updates.
  * Improved custom URL rewriting, including string, URL, and mutation-based rewrites.
  * Preserved query parameters, hashes, external origins, and case-sensitive base paths during rewrites.
  * Improved route invalidation so affected routes and cached data refresh reliably.
  * Cleaned up repeated slashes in rewritten paths.
  * Updated server-side route handling when case-sensitivity settings change.
* **Chores**
  * Updated router core release metadata for a patch release.


packages/router-core/src/router.ts
```diff
@@ -1289,6 +1289,25 @@ export class RouterCore<
...
+    if (basepathChanged || rewriteChanged) {
+      this.basepath = nextBasepath
+
+      const trimmed = trimPath(nextBasepath)
+      const basepathRewrite =
+        trimmed && trimmed !== '/' ? rewriteBasepath(nextBasepath) : undefined
+      // The basepath is stripped first on input and re-added last on output.
+      this.rewrite =
+        basepathRewrite && nextRewriteOption
+          ? composeRewrites([basepathRewrite, nextRewriteOption])
```
Sheraff, line 1306: is `composeRewrites` *ever* used with anything other than exactly 2 rewrites? if not, maybe the function can be simplified (and we don't create an array for its argument)

Sheraff (changes requested):

packages/router-core/src/router.ts
schiller-manuel: users could use it to compose arbitrary chains. but we could specialize it indeed

Sheraff (approved):

Issue #4469, opened 2025-06-19
Absolute paths get included in the build artifacts (files)


antitoxic:
### Which project does this relate to?

Start

### Describe the bug

Cross-posting from discord: https://discord.com/channels/719702312431386674/1385153236604031026/1385159859091279872

This is generated by the default from tanstack start basic example where the only thing I've changed is removed tailwind and just use some basic global css imported like `import appCss from '../global.css?url';`

The `createServerFnStaticCache` includes a `publicUrl` which is absolute path.

Also `.tanstack-start/build/server/server-functions-manifest.json` has `extractedFilename` with absolute path

`.vercel/output/static/.vite/manifest.json` also  has the global css as absolute path. (same in .tanstack-start dir)

And `.vercel/output/static/assets/main-<hash>.js`  has some absolute path but it's minified so I can't really tell where. (same in .tanstack-start dir)

### Your Example Website or App

start-basic

### Steps to Reproduce the Bug or Issue

1. https://tanstack.com/start/latest/docs/framework/react/quick-start#impatient
2. Run `pnpm build` / `npm run build`

### Expected behavior

I don't have absolute paths in build artifacts.

### Screenshots or Videos

<img width="1382" alt="Image" src="https://github.com/user-attachments/assets/ff48bdd3-48bc-4812-b2e9-c30b5f7e4b3c" />

### Platform

- OS: macOS
- Browser: n/a
- Version: `^1.121.24`

gitpash:
I don't know it is related or now but we were having some assets processing issue recently, some assets get's corrupted (like wepb in our case), I tried all possible assets processing plugins variations - but nothing works...
after reading this I downgraded to @tanstack/router-plugin@1.102.6 and it goes away
also later I can see that setting autoCodeSplitting:false also helps us even on a newest version

schiller-manuel:
@gitpash this issue is about start , not router. please create a new issue including a complete example

gitpash:
> @gitpash this issue is about start , not router. please create a new issue including a complete example  

yes you are right! sorry for creating distraction here, my only intention was to give someone a clue(as myself get it here)

PR #8432, opened 2026-09-15
fix(start): omit build paths from production manifests

danieljvdm:
## 🎯 Changes

Production Start manifests embed absolute route source paths, so moving an unchanged project changes the emitted JavaScript. Remove those build-only paths after client chunk lookup, preserving the original route metadata for build and development use and leaving the general serializer unchanged.

Related to #4469; the absolute lookup paths were introduced in #4453.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Production manifests no longer include build-only route file paths.
  * Identical builds now produce the same manifest regardless of the checkout directory used during the build.
  * Route assets, including preloads, scripts, and stylesheets, continue to be included as expected.

* **Release**
  * Published a patch release for the Start core plugin.


danieljvdm:
Validated against main `8e164d28f9e9eb1a7301109230d7ece9817492a8` with the pinned Node 24.8.0 and pnpm 11.21.0 toolchain:

- The new regression calls the existing normalized-build, production-manifest builder, and serializer/deserializer at two checkout paths. It checks unchanged frozen source routes and exact emitted preloads, scripts, and CSS. Before the fix, those checks passed and the final serialized-output comparison failed on the absolute `filePath` values. The same test passes after the fix.
- The existing package checks passed: 541 tests across 34 files, all six TypeScript targets (5.6/5.7/5.8/5.9/6/7), and ESLint. These ran through the existing Nx graph, including the required dependency builds, using Vite+ as the command entry point.
- Existing serializer tests for shared asset identity and preserving explicitly supplied non-link fields remain unchanged and pass. This fix only removes paths from the builder's cloned production result.

The source and dependency inputs are unchanged since those checks. No performance improvement is claimed.

PR #8435, opened 2026-09-15
perf(react-router): avoid redundant link hydration updates

Sheraff:
## 🎯 Changes

Avoid the extra hydration-triggered render for React Links whose active matching does not compare URL hashes.

Every Link currently subscribes to a hydration snapshot that changes from `false` to `true`. That schedules another render and invalidates the Link's location selector, even when its output does not depend on hydration. An `@internal` overload of `useHydrated` lets these Links use equal server/client snapshots without a separate helper. The implementation defaults `enabled` to `true`; declaration emission strips the boolean overload, preserving the public no-argument API. Hash-sensitive Links retain the deferred active-state update they need.

Found while investigating the 1,000-link table in [e18e/framework-tracker#200](https://github.com/e18e/framework-tracker/issues/200).

### Verified work reduction

| Controlled hydration case | Before | After |
| --- | --- | --- |
| Ordinary Link render-prop calls during hydration | 2 | 1 |
| Matching hash-sensitive Link active-state sequence | `false → true` | `false → true` |

This avoids one hydration-snapshot-induced update per ordinary Link. Other parent/router updates can still cause renders. An improvement in e18e's FP/FCP, interaction latency, or SSR throughput has not been established.

The regression tests exercise distinct server/client fragment state, active/inactive props, DOM-node reuse, toggling hash matching after hydration, and a delayed Suspense hydration boundary. They check both recoverable errors and hydration diagnostics.

### Validation

The local checks below passed for commit `544927c03d`. The internal-overload follow-up (`88a3c208ea`) is left to CI, as requested.

Original validation used `NX_BASE=origin/main`, `CI=1`, and `NX_DAEMON=false`:

- `pnpm test:eslint --outputStyle=stream --skipRemoteCache` — passed.
- `pnpm test:types --outputStyle=stream --skipRemoteCache` — passed.
- `pnpm test:unit --outputStyle=stream --skipRemoteCache` — passed; React Router: 1,180 passed, one skipped.
- `pnpm nx run tanstack-react-start-e2e-basic:test:e2e--vite-ssr --outputStyle=stream --skipRemoteCache -- tests/navigation.spec.ts tests/client-only.spec.ts` — 11 passed.
- Formatting and `git diff --check` passed.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed unnecessary hydration-triggered rerenders for links that do not compare URL hashes.
  - Preserved correct active-state updates for links that include URL hash comparisons.
  - Improved link behavior when hydration occurs within delayed loading boundaries.
  - Prevented related hydration and console errors during link rendering.
  - Ensured ordinary links retain their server-rendered state during hydration while hash-sensitive links update correctly.


packages/react-router/src/ClientOnly.tsx
```diff
@@ -57,7 +57,17 @@ export function ClientOnly({ children, fallback = null }: ClientOnlyProps) {
  * @returns True if the JS has been hydrated already, false otherwise.
  */
 export function useHydrated(): boolean {
-  return React.useSyncExternalStore(subscribe, getSnapshot, getServerSnapshot)
+  return useHydratedWhen(true)
```
schiller-manuel: overload with @internal should be smaller

Issue #8241, opened 2026-09-04
[solid-router] errorComponent is SSR'd inside MatchInner but hydrated from CatchBoundary's Errored fallback — hydration key miss, error UI never interactive


TylerRick:
**Which project does this relate to?** Router

**Describe the bug**

In `@tanstack/solid-router`, a route whose loader throws gets its `errorComponent` instantiated at two different points in the component tree depending on the side. `MatchInner`'s `status === 'error'` branch (`packages/solid-router/src/Match.tsx`) renders the error component **inline** on the server, as a child of its own `<Solid.Match>`, but **throws** on the client, so `CatchBoundary`'s `<Solid.Errored>` catches it and renders the error component from its `fallback`, several levels higher:

```tsx
<Solid.Match when={currentMatch().status === 'error'}>
  {(_) => {
    if (isServer ?? router.isServer) {
      const RouteErrorComponent =
        (route.options.errorComponent ?? router.options.defaultErrorComponent) || ErrorComponent
      return <RouteErrorComponent error={currentMatch().error} info={{ componentStack: '' }} />
    }
    throw currentMatch().error
  }}
</Solid.Match>
```

Solid hydration is id-addressed, and the two sides mint the error component's ids under different namespaces — server: `Match → CatchBoundary → Errored → Show → RouteContent → … → MatchInner → Show → Switch → Match`; client: `Match → CatchBoundary → Errored`. The client's root element therefore names a key no server element carries. `@solidjs/web` logs `Hydration key miss` and builds the client copy detached; the server's copy stays in the document, unclaimed, with the markup the user sees and no event listeners. Any control inside an error component is dead after hydration. A one-element error component (a bare `<h1>`) looks fine, because nothing in it needed wiring — which is how this stays invisible until someone puts a button in one.

From the reproducer, the server id, the client's missed claim, and — for comparison — the id the same component gets when rendered from the route's normal `component`:

```text
server  …0004900010002 1011032370000   ← MatchInner renders <RouteErrorComponent> inline
client  …0004900010002 200100          ← <Errored fallback> instantiates it at the CatchBoundary
control …0004900010002 1032371102      ← MatchInner's success branch
```

The React router has the same server/client split and gets away with it because React reconciles by tree position and re-renders on a mismatch; Solid does neither.

**Complete minimal reproducer**

https://github.com/TylerRick/tanstack-solid-router-error-component-hydration-repro

**Steps to Reproduce the Bug**

[...]

PR #8438, opened 2026-09-15
fix(solid-router): hydrate route error components consistently

dvd233:
## What

- render route match errors from a route-specific boundary position on both the server and client
- keep render-time descendant errors handled by the existing `CatchBoundary`
- remount the route error UI when a subsequent loader failure supplies a new error
- add unit and Solid Start browser regressions for repeat failures, hydration warnings, and interactive controls

## Why

During SSR, `MatchInner` rendered a route `errorComponent` inline, while the client threw the same error into `CatchBoundary`. Solid assigns hydration IDs from the owner tree, so the client could not claim the server-rendered error UI and its controls remained unwired.

Rendering known route errors from one stable boundary position keeps the server and client owner trees aligned.

Fixes #8241

## Tests

- `cd packages/solid-router && pnpm run test:unit` (873 passed, 2 skipped; server mode: 7 passed)
- `cd packages/solid-router && pnpm run test:types` (TypeScript 5.6 through 7.0)
- `cd packages/solid-router && pnpm run test:eslint` (0 errors; 33 existing warnings)
- `cd packages/solid-router && pnpm run build`
- `cd e2e/solid-start/query-integration && pnpm run build`
- Playwright Chromium: `error-component-hydration.spec.ts` and `not-found-reload.spec.ts` (4 passed)

## Checklist

- [x] I have followed the steps in the contributing guide.
- [x] I have tested the code changes locally.
- [x] I fully understand the code in this pull request, including the AI-assisted portions I reviewed and rewrote.
- [x] This change affects published code, and I have added a changeset.


brenelz:
## Review

**Intent:** Fix #8241 — SSR rendered route `errorComponent` inside `MatchInner`, client threw into `CatchBoundary`/`Errored`, so hydration IDs didn’t match and controls stayed dead. `RouteCatchBoundary` renders known match errors at the same boundary position on both sides; render-time descendant errors still go through `CatchBoundary`.

**Verdict:** Approach looks right and matches the bug. No hard blockers from the diff; I’d want maintainer CI (or a green run on `solid-router-v2-pre`) before merge.

### Blockers
None in the code itself.

### Should confirm
1. **Thin CI on this base.** Checks here are basically Socket/CodeRabbit; author cites local unit/types/e2e. Branch is `solid-router-v2-pre`, not default — worth a maintainer re-run or explicit ACK that local green is enough.
2. **`getResetKey` is unused while `hasError` is true.** Loader-error UI lives outside `CatchBoundary`, so the existing reset-key effect won’t fire until the match leaves error. New errors remount via `createEffect(props.getError)` + `Object.is` — covered by the invalidate unit test. Confirm that’s intentional for “same error ref, `updatedAt` bump only” (rare).

### Nits
- **`MatchInner` still `throw`s on `status === 'error'`** — dead for the loader path once `RouteCatchBoundary` owns it; fine as a backstop for the `CatchBoundary` child path, or delete later for clarity.
- **SSR no longer passes `info={{ componentStack: '' }}`** (old inline path did). Only matters if something read `info` during SSR; default `ErrorComponent` doesn’t.
- **Loader `reset()`** is still remount-only (`setRetryKey`), not a route retry — same practical limit as before for sticky match errors; invalidate/navigation remains the real recovery. Fine, just don’t document it as “retry loader.”
- E2E falsy-error case (`?falsy=true` → message `"undefined"`) is a good edge pin; a bit odd as product UX but OK as a regression.

### What’s strong
- Clear root cause; shared `renderErrorComponent` helper.
- Unit test for second loader failure after invalidate + `onCatch` counts.
- Playwright checks hydration warning strings **and** that the button actually toggles `data-clicked`.
- Changeset for `@tanstack/solid-router`.

PR #8453, opened 2026-09-16
perf(benchmarks): compress bundle-size outputs once

Sheraff:
## Summary

- Measure each emitted file's raw/gzip/Brotli size once per scenario, then reuse the per-file measurements for initial-load totals.
- Preserve compression settings, metric schema, file ordering and multiplicity; keep reuse scoped to one measurement so rebuilt files are never stale.
- Add opt-in `--timings` diagnostics for package preparation, builds, manifest discovery, compression and source attribution, including failed phases. Document fresh execution versus Nx cached timing output.
- Register the helpers as Nx inputs and add compression parity, call-count, error-path, timing and runner tests.

## Performance

Same application/plugin artifacts and all 18 scenarios, separate warmups and three measured samples per implementation; package preparation excluded. Only compression behavior differs between comparison arms.

| Phase | Before (median) | After (median) |
| --- | ---: | ---: |
| Build + measurement | 12.432 s | **8.276 s** |
| Scenario builds | 4.065 s | 4.059 s |
| Gzip | 0.141 s | 0.073 s |
| Brotli | 8.198 s | 4.128 s |

**33.4% less measurement time.** Overall ranges were 12.418–12.452 s and 8.260–8.284 s. All 18 metric objects, benchmark-action rows and emitted output hashes matched exactly. An isolated five-pair experiment also showed approximately 50% less compression time. Bounded async compression did not improve on the synchronous implementation and is not included.

Measured on Linux / AMD Ryzen 7 255, Node 24.8.0 and pnpm 11.21.0. These are local-machine results.

## Validation already completed

- 30 benchmark-script unit tests passed through Nx.
- Full 18-scenario uncached bundle matrix passed; every raw/gzip/Brotli and initial-load metric matched the baseline.
- Formatting and `git diff --check` passed.

The timing output is diagnostic only; it does not add fields to metric artifacts. The route compiler optimization is a separate PR.

## Summary by CodeRabbit

* **New Features**
  * Added optional timing diagnostics to bundle-size benchmark runs.
  * Added a `--timings` option and corresponding environment-variable support.
  * Timing output now covers benchmark phases, scenarios, compression, and file measurement.
  * Bundle-size measurements now report current and baseline totals consistently.

* **Documentation**
  * Documented timing configuration, cache behavior, diagnostic output, measured phases, and failure handling.

* **Tests**
  * Expanded coverage for compression measurements, timing reporting, error handling, and command-line option behavior.


schiller-manuel:
do we need to update the bundlesize skill?

Sheraff:
> do we need to update the bundlesize skill?

I don't think so:
- this adds a `--timings` option, but that's for working on the script itself, i don't think an agent that is trying to get bundle measurements should care (or waste context on it)
- this reduces the compression work, but only in a way that should not change anything about the output

schiller-manuel (approved):

PR #8461, opened 2026-09-16
feat(solid-router): consume server-function response metadata

ryansolid:
## What

`RouterContextProvider` registers the router as the unnamed single-flight consumer of Solid's server-function transport on the client (`subscribeFlightData`), so the response helpers a mutation returns take effect without the caller wrapping the call. The transport hands `useMutation` the plain value; the router applies the metadata.

- **`redirect()`** — the carrier delivers the target resolved to an absolute url, so the soft/hard split is a real origin comparison. Same-origin navigates under the router with `replace` (the destination takes the submission's place in history, as HTTP gives a form post). Any other origin goes through `navigate({ href })`, so the document reload still passes the protocol allowlist and blockers.
- **Route data** — any declaration of a write (`redirect()`, `reload({ revalidate })`, `respond(value, { revalidate })`, the reserved `revalidate: '*'`) reloads through `router.invalidate()`. An empty declaration (`revalidate: []`) is honored as nothing to reload: a redirect still navigates, shared layouts are not rerun.
- **`X-Revalidate` keys** name entries in caches the router does not own; the router reads only the declaration's shape and leaves the keys to those caches' consumers (`@tanstack/solid-query` subscribes under its own source id).

## Why this order

Navigate, then invalidate. `navigate` commits the destination to history synchronously (the Solid adapter's load follows in a microtask); `invalidate` then marks the committed matches — the layouts the destination shares with the page the mutation ran on — and starts a load that already reads the destination as the latest location. The adapter's load adopts the loader generations that load started for the same match ids (core's donor-flight mechanism) rather than restarting them. Result: the destination commits with its shared layouts reloaded, and the abandoned page never reruns its own loaders. The other order does — the test caught it (`{ root: 2, index: 2, after: 1 }` vs the expected `{ root: 2, index: 1, after: 1 }`).

## Transport dependency

The router's consumer runs when the transport delivers to it. Against `@solidjs/web` 2.0.0-rc.8 that is only folded responses (a server collector contributed data), which covers same-origin redirects and keyed reloads in the Start/query template. A pending `@solidjs/web` change delivers every metadata-bearing mutation response to registered consumers, folded or not — cross-origin redirects, failed collectors, apps with no collector. The router code here covers both; the peer range (`>=2.0.0-0 <3.0.0`) picks the release up.

## Tests

`tests/flightData.test.tsx` drives the registered consumer the way the transport does: registration lifetime, same-origin redirect (location, `replace`, exact loader counts), cross-origin redirect (`navigate({ href })`, no invalidation), keyed `reload`, `respond` with keys, `'*'`, empty declaration (alone and with a redirect), metadata-free response, malformed carrier refused. Full `@tanstack/solid-router` suite green (882), types clean.

— *Claude via Cursor*


brenelz:
## Review

**Intent:** `RouterContextProvider` registers as the unnamed `subscribeFlightData` consumer so `redirect()` / `reload({ revalidate })` / `respond(..., { revalidate })` work without callers wrapping mutations.

**Behavior:** Same-origin redirect → soft `navigate` with `replace`, then `invalidate` if a write was declared. Cross-origin → `navigate({ href })` and **return** (no invalidate). Absent / named / `*` header → reload route data; `revalidate: []` → navigate only, no invalidate. Order is navigate-then-invalidate (test pins loader counts).

**Verdict:** Clean, well-documented, strong tests (registration, same/cross-origin, keyed/`*`/empty, metadata-free, malformed `javascript:` carrier refused). **No blockers** from the diff.

### Watch
- Relies on a **pending `@solidjs/web` transport** change for metadata without a folded collector slice; peer range is ready, but against rc.8 only folded responses hit the consumer (PR says so).
- Confirm the full Test job is green before merge (some checks were still unsettled when reviewed).
- `setupFlightDataConsumer` in the provider body + `onCleanup` is fine under Solid’s once-run components; don’t “React-ify” this later into a render loop.

### What’s strong
- Navigate-before-invalidate rationale and the loader-count regression that caught the reverse order.
- Cross-origin hard navigation still goes through `navigate({ href })` (protocol allowlist / blockers).
- Empty `revalidate: []` honored as “nothing to reload” while redirect still navigates.

brenelz:
1 test failed

>   1 failed
    [chromium] › tests/server-functions.spec.ts:165:1 › POST server functions return single-flight loader data

brenelz (approved):

PR #8472, opened 2026-09-19
fix(router-core): advance the pending boundary past settled matches

freshgiammi:
## 🎯 Changes

Fixes a regression from lane match loader rewrite, see #7805. 

`offerPending` in `packages/router-core/src/load-client.ts` picks a settled match as the pending boundary, especially with lazy-only layouts. A layout stays pending after it finishes loading until the slowest loader further down the branch resolves. Moving between sibling layouts also replaces the whole shell, even when the shell itself has nothing left to load.

This change makes the first presented pending match the painted boundary and skips settled matches with a successful status and no `_notFound`. Layouts with no loader render as soon as they settle, while the fallback stays on the match that is actually still loading.

Added a few test to gate future regressions:
- [`pending-boundary-settled-ancestor.test.ts`](https://github.com/TanStack/router/compare/main...freshgiammi:fix/pending-boundary-settled-ancestor?expand=1#diff-c5c9431940c22584eae2c5aa95cb6c1f9c5bce58722242b03a18212f315cf919) covers the boundary move for a sibling swap, an initial load, and a `pendingMinMs` case.
- [`pending-boundary-settled-ancestor.test.tsx`](https://github.com/TanStack/router/compare/main...freshgiammi:fix/pending-boundary-settled-ancestor?expand=1#diff-5f54527325efadd9a3de1d604eb5f4ca07b683d36b677aea1959873c72c6b31f) checks the shell renders while the leaf still loads.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pending-state handling during navigation.
  * Layout routes now render as soon as they settle, while fallback content remains visible for child routes that are still loading.
  * Pending boundaries now move correctly from settled layouts to pending descendants.
  * Fallback content remains visible for the configured minimum duration before transitioning.
  * Prevented settled routes from delaying updates when no asynchronous work remains.

* **Tests**
  * Added coverage for lazy layouts, nested loading states, boundary transitions, and pending-delay behavior.


Sheraff:
sorry what's the bug this is fixing?

freshgiammi:
We're facing an issue that was not present before the loader rewrite. The layout is this:
 
 ```
__root__
└── _auth (loader)
    ├── _sidebar
    │   └── /users ← current
    │
    └── _header (lazy only, no loader)
        └── /users/$userId            (route, with loader)
            └── /users/$userId/       (index, with loader) ← navigating here
```


Navigating `/users` → `/users/$userId/` moves across sibling layouts (`_sidebar` → `_header`).

Before the rewrite, the route's (`/users/$userId`) `pendingComponent` appeared while the loader resolved: a shell with skeleton breadcrumbs. Now `_header` stays pending for the entire leaf loader duration. Nothing below `/users/$userId` sets a `pendingComponent`, so it falls back to `defaultPendingComponent` and the shell/skeleton never shows.

The patch works so that in this scenario, once `_header` has settled, it stops blocking the boundary. The fallback moves to whatever is actually still loading, so `_header` renders and only the leaf spins.

I can set up a repro if you feel like that's needed: sorry I didn't file an issue, I pointed an agent at a private repo to chase this and it came back with a small patch, so a PR seemed more useful than an issue.
@Sheraff

Sheraff:
ok i understand now, that was an intentional change to avoid pendingComponent from moving too much during loading. I'll bring it up to the team see what they think.

Ideally if we *do* want the pending to change during a navigation, each one that is shown should at least be shown for `minPendingMs`, which i'm assuming would make this diff quite a bit larger?

freshgiammi:
> Ideally if we do want the pending to change during a navigation, each one that is shown should at least be shown for minPendingMs, which i'm assuming would make this diff quite a bit larger?

You're right, that was a miss on my part. I don't think the diff is that large though, if you only count logic changes and not tests.

I've added a few commits:
- [`7116510`](https://github.com/TanStack/router/pull/8472/commits/7116510f822e62013d39128b3d13e78ef41ce692) which applies the [Nx suggestion](https://github.com/TanStack/router/pull/8472#pullrequestreview-5260162359) above. I hadn't run the e2e suite locally, which caused this to slip through.
- [`6fdd448`](https://github.com/TanStack/router/pull/8472/commits/6fdd4480c54735203c380e5ee524c87aa815bcdc) holds the painted boundary through its `minPendingMs` window before moving it down the tree, plus a test covering it
- [`fbf3d11`](https://github.com/TanStack/router/pull/8472/commits/fbf3d11b4bbfc732aad2c545c5e3834ecc7413d0) additional coverage around the issue to make sure different loader combinations aren't problematic

---

**EDIT**

Two bot findings addressed:

- [`dc59eb9`](https://github.com/TanStack/router/pull/8472/commits/dc59eb9df4d3241998f04f7576703a1bced544a9): the hold test now runs on a controlled clock (`vi.useFakeTimers` plus `act`, same shape as `ancestor-loader-child-pending-min`) instead of a real 100 ms sleep. Converting it caught a second real-timer dependency: the final await navigation needed an advance of 400 ms because the leaf's own min gates commit.

I also tried  to cover the pending-descendant gate with a unit test for the nitpick, but the spy showed the lane jumps straight to commit, so there is nothing to observe there. The e2e suite should be enough coverage as it fails without it?

